Issue #66, opened 2025-12-22
[Feature Request] Agents Manager


wienans:
# Agent Manager

## Feature List

- [x] Running parallel opencode agents with the same prompt (completely leveraged MultiRunLauncher from Wev App)
  - [x] Base Banch selection
  - [x] Multi Model picker
  - [x] Agent Picker
  - [x] Add Files
- [x] Group multiple Agents created by MultiRunLauncher in the UI 
- [ ]  Review Changes (will likely leverage diff viewer from web app)
- [ ] easily Pull in changes from the worktree into the main worktree
- [x] delete AgentGroups (and worktrees) 

## UI

<img width="1741" height="997" alt="image" src="https://github.com/user-attachments/assets/02068de4-9ddf-4ddd-97d2-6b617e486d10" />

Select sessions within the group
<img width="1741" height="997" alt="image" src="https://github.com/user-attachments/assets/66bf5943-71e0-45d1-92bd-593db3633978" />

Reuse chat Component for every individual session
<img width="1741" height="997" alt="image" src="https://github.com/user-attachments/assets/dfb4355c-1e48-46dc-a5f9-41569df0c03f" />

## Initial Proposal

It would be nice to have a "Agents Manager" similar to the Cursor Agents View in the VS Code Extension

* Git Worktree support 
* Branch selection
* Running multiple different models with the same Prompt in parallel

Usage:

You can send of some Agents in Parallel on separate git worktrees implementing things in parallel, while you can work on your own feature directly with the Chat window.

### Extra context (optional)

Cursor View:
<img width="2480" height="456" alt="Image" src="https://github.com/user-attachments/assets/a1947bc3-7713-450b-9e4e-6f3299f8bb81" />

<img width="2193" height="849" alt="Image" src="https://github.com/user-attachments/assets/5c2e6b80-5082-41b9-a50d-21329951a803" />

<img width="1468" height="1346" alt="Image" src="https://github.com/user-attachments/assets/b80d6ba3-80fd-4a86-ad98-cc82d87e222c" />

Maybe similar to the one Kilo Code is offering (implemented as a seperate Window)

<img width="1598" height="866" alt="Image" src="https://github.com/user-attachments/assets/6f0276d5-82af-41d3-9ac8-5245112a0941" />

PR #87, opened 2025-12-31
feat(vscode) Agent Manager

wienans:
Agent Manager Mock up in VsCode for #66 

Mock Up (it sometimes took to much insparation from cursor, it should be the same chat interface we have in the normal chat):

Full Screen Mode (sidebars disabled)
image
image

With Sidebars enabled (going into background for Chat users)
image


wienans:
Agent Manager UI

For simplicity reuses the MultiRun Components for Branch and Model Selector and the general structure
image

Select sessions within the group
image

Reuse chat Component for every individual session
image

"Breaking" Change:
I changed the Session Title and Branch Names to:
groupSlug/provider/model/index

the AgentManager Groups Sessions based on that title and groupSlug

Can be accessed via title bar navigation (circuit board)

PR #96, opened 2026-01-02
Update the shortcuts in the desktop app for macOS

sakhnyuk:
## Summary

I’ve been using OpenChamber more frequently, and I’m generally pleased with my experience so far. However, it’s quite challenging to adopt new shortcuts in a completely different manner on macOS, particularly when using the Ctrl key. Therefore, I decided to address this issue.

Here are a few notes about shortcuts:

- Having `Ctrl+X` as the primary command palette seems to conflict with the native `Cut` action on Windows and Linux. A similar issue would arise for `Cmd+X`, so I opted for the `Cmd+K` (`Ctrl+K`) shortcut for the Command Palette.
- The `Cmd+N` shortcut will overlap with the native browser `New tab` action, so I’ve restricted my changes to the MacOS desktop app.

## Changes

- I’ve applied the metakey to all shortcuts to enable the Cmd key on macOS. I’ve only enabled this for the MacOS Desktop app.
- I’ve changed the Command Palette shortcut from `Ctrl+X` to `Ctrl+K` (`⌘+K` on macOS desktop).
- I’ve added helper functions `isMacOS()`, `hasModifier()`, and `getModifierLabel()`.

**Please treat it as a proposal. I am ready to adjust the PR.**

CleanShot 2026-01-02 at 16 56 47@2x
CleanShot 2026-01-02 at 16 52 00@2x

### DIFF
--- a/packages/ui/src/components/ui/HelpDialog.tsx
+++ b/packages/ui/src/components/ui/HelpDialog.tsx
@@ -10,7 +10,6 @@ import { useUIStore } from "@/stores/useUIStore";
 import {
   RiAddLine,
   RiArrowUpSLine,
-  RiArrowUpWideLine,
   RiBrainAi3Line,
   RiCloseCircleLine,
   RiCodeLine,


btriapitsyn:
Hey @sakhnyuk I'm totally fine with this change! Just a quick note: cmd+h hides the window and cmd+m minimizes the app in MacOS, so those shortcuts might conflict with the system commands, which could take precedence. It would be great if you could consider replacing those two shortcuts as well. For the web version, we should also use ctrl, so we need to make sure those are free there as well. How about trying cmd+; (or ctrl+; in the browser) for help dialog and cmd+u (or ctrl+u) for the model selector? Feel free to suggest your options as well. I apologize for the request, but having a fully functional suggestion would really help avoid any fixes later on!

sakhnyuk:
> cmd+h hides the window and cmd+m minimizes the app

@btriapitsyn 
Those shortcuts works well, tbh. It simply overrides the system behavior. 

<details>
<summary>Cmd+H and Cmd+M shortcuts GIF</summary>

CleanShot 2026-01-02 at 22 31 04
</details>

If there’s a need to change the bindings, for the `model switcher` I’d suggest using the right bottom corner of the keyboard to align with the UI placement. This includes the `/`, `,`, and `.` keys. The first two keys will replace the `theme switcher` or `settings` options.
But honestly any option would work, and we could even add shortcuts configuration later.

So, what do you think? should we unblock native system shortcuts and if yes, are you ok with`Cmd/Ctrl + +` and `Cmd/Ctrl + .` (or `/`, `,`)?

PR #107, opened 2026-01-06
fix(chat): prevent message send during IME composition

### LINKED ISSUE #104: [Bug] Japanese IME input submits message when pressing Enter to confirm conversion
## Description
When using Japanese (or other CJK) input methods, pressing Enter to confirm a character conversion immediately sends the message instead of just completing the IME conversion.
## Steps to Reproduce
1. Open OpenChamber chat
2. Switch to Japanese IME (or any CJK input method)
3. Type "nihongo" (にほんご in hiragana appears)
4. Press Enter to confirm the conversion to "日本語"
5. **Result:** Message is sent immediately with incomplete/partial text
6. **Expected:** Conversion should complete, message should NOT be sent
## Root Cause
In `packages/ui/src/components/chat/ChatInput.tsx`, the `handleKeyDown` function does not check for `isComposing`:
```typescript
if (e.key === 'Enter' && !e.shiftKey && !isMobile) {
    e.preventDefault();
    // sends message
}
```
## Proposed Fix
```diff
- if (e.key === 'Enter' && !e.shiftKey && !isMobile) {
+ if (e.key === 'Enter' && !e.shiftKey && !isMobile && !e.nativeEvent.isComposing) {
```
## References
- [MDN: KeyboardEvent.isComposing](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/isComposing)
---
*This issue was identified and analyzed using [opencode](https://github.com/anomalyco/opencode)* 🤖

### Where does it happen?

Desktop Web

### Version (if known)

1.4.3

_No response_

### Logs (optional)

```shell

```

madebyjun:
## Summary
Fixes an issue where pressing Enter to confirm IME (Input Method Editor) conversion immediately sends the message instead of completing the conversion.

Fixes #104

## Problem
When using Japanese, Chinese, Korean, or other CJK input methods, users press Enter to confirm character conversion. The chat input was intercepting this Enter keypress and sending the message with incomplete/partial text.

## Solution
This PR adds robust IME composition detection to `ChatInput`:

1. Add `isIMECompositionEvent()` helper that detects composition via:
   - `isComposing` (standard)
   - `keyCode === 229` fallback for some WebKit-based environments
     - Note: `keyCode` is deprecated, but `229` remains a practical fallback for IME detection in certain WebKit/Tauri cases.

2. Early-return from `handleKeyDown` while IME composition is active, intentionally avoiding interference with IME keystrokes.

3. Add an explicit IME guard to the Enter-to-send handler as an additional safety net.

## Changes
- packages/ui/src/components/chat/ChatInput.tsx
  - Add `isIMECompositionEvent()`
  - Skip key handling during IME composition
  - Guard Enter-to-send during composition

## Testing
Tested with Japanese IME on macOS:
- Web UI
- Desktop App (Tauri/WebKit)

Confirmed:
- Enter confirms conversion without sending
- Message sends only after composition completes

### DIFF
--- a/packages/ui/src/components/views/agent-manager/AgentManagerEmptyState.tsx
+++ b/packages/ui/src/components/views/agent-manager/AgentManagerEmptyState.tsx
@@ -191,7 +191,7 @@ export const AgentManagerEmptyState: React.FC<AgentManagerEmptyStateProps> = ({
     if (isIMECompositionEvent(e)) return;
 
     // Enter submits if valid, Shift+Enter adds newline
-    if (e.key === 'Enter' && !e.shiftKey) {
+    if (e.key === 'Enter' && !e.shiftKey && !isIMECompositionEvent(e)) {
       e.preventDefault();
       if (isValid && !isSubmittingOrCreating) {
         handleSubmit(e as unknown as React.FormEvent);


wienans:
@madebyjun please add the same for the AgentManagerEmptyState.tsx

packages/ui/src/components/chat/ChatInput.tsx
```diff
@@ -508,7 +521,7 @@ export const ChatInput: React.FC<ChatInputProps> = ({ onOpenSettings, scrollToBo
         }
 
         // Handle Enter/Ctrl+Enter based on queue mode
-        if (e.key === 'Enter' && !e.shiftKey && !isMobile) {
+        if (e.key === 'Enter' && !e.shiftKey && !isMobile && !isIMECompositionEvent(e)) {
```
wienans, line 524: If i am not tripping `!isIMECompositionEvent(e)` is always true, because in line 491 whe check `isIMECompositionEvent(e)` and return if true. Therefore it is always false and the negation always true and therefore not necessary and can be removed.

packages/ui/src/components/views/agent-manager/AgentManagerEmptyState.tsx
```diff
@@ -177,8 +187,11 @@ export const AgentManagerEmptyState: React.FC<AgentManagerEmptyStateProps> = ({
   };
 
   const handleKeyDown = (e: React.KeyboardEvent<HTMLTextAreaElement>) => {
+    // Early return during IME composition
+    if (isIMECompositionEvent(e)) return;
+
     // Enter submits if valid, Shift+Enter adds newline
-    if (e.key === 'Enter' && !e.shiftKey) {
+    if (e.key === 'Enter' && !e.shiftKey && !isIMECompositionEvent(e)) {
```
wienans, line 194: If i am not tripping `!isIMECompositionEvent(e)` is always true, because in line 491 whe check `isIMECompositionEvent(e)` and return if true. Therefore it is always false and the negation always true and therefore not necessary and can be removed.

PR #114, opened 2026-01-07
feat: add default identity selection with auto-apply

riderx:
The goal here is to make an identity default and get the one present in github automatically and put it default.

I mostly beleive that what people want and only some will use different identity 
## Summary
- Add checkbox to mark a git identity as default
- System identity (from git config) is the default fallback
- Only one identity can be marked as default at a time
- Automatically apply default identity to new sessions

## Test plan
- [ ] Check "Set as default identity" on a custom profile
- [ ] Verify star icon appears next to the default profile
- [ ] Switch default to another profile, verify only one has the star
- [ ] Create a new session and verify the default identity is applied
- [ ] Verify system identity is shown with star by default


riderx:
Thanks your implementation make sense !

PR #135, opened 2026-01-13
feat: Display default model in dropdowns and consolidate Git settings menu

nelsonPires5:
## Summary
- Makes `opencode/big-pickle` appear as the selected model in dropdowns when no default model is configured (display-only, no persistence)
- Consolidates Worktree settings into the Git menu section with proper visual hierarchy

### Default Model Display
- Added `getDisplayModel` helper to compute display model with fallback to `opencode/big-pickle`
- Falls back to first available provider/model if big-pickle is not found
- "Not selected" option remains available to explicitly clear stored settings
- Applies to both:
  - Session Defaults → Default model
  - Git → Commit Messages → Model for generation

### Git Menu Consolidation
- Removed separate Worktree menu item
- Expanded Git menu items: "Commit Messages", "Worktree"
- Git section now renders both settings with visual separator and section header:
  - Commit Messages (model config)
  - [separator]
  - Worktree (section header)
    - Branch prefix
    - Base branch  
    - Setup commands
    - Active worktrees

### Print
image

## Tests
- [X] Web app (Desktop and Mobile)
- [ ] Desktop app
- [ ] Vscode extension

Closes https://github.com/btriapitsyn/openchamber/issues/130

🤖 Generated with OpenCode

### DIFF
--- a/packages/desktop/src-tauri/src/commands/git.rs
+++ b/packages/desktop/src-tauri/src/commands/git.rs
@@ -2091,10 +2091,21 @@ Diff summary:
     );
  
     let settings = state.settings().load().await.unwrap_or(serde_json::Value::Null);
-    let model = settings.get("commitMessageModel")
+    let raw_model = settings.get("commitMessageModel")
         .and_then(|v| v.as_str())
-        .filter(|s| !s.trim().is_empty())
-        .unwrap_or("big-pickle");
+        .unwrap_or("");
+
+    let model_candidate = raw_model
+        .split('/')
+        .last()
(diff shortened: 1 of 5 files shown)


btriapitsyn:
I added restrictions to model selection for the Git commit message since we are using only opencode.ai/zen provider directly. Using the SDK will add the ability to choose other providers, but it will create a new session for each commit message generation with a lot of additional tokens spent on basic context like tools, MCP, and system prompts. The reason I'm using direct requests to Opencode Zen as a provider is because they have free model and we can avoid additional token burning.

PR #293, opened 2026-02-04
feat(mobile): split controls into separate Agent and Model buttons

Jovines:
## Summary

Split mobile controls bar into separate Agent and Model buttons for better UX.

## Motivation

Agent mode switching is frequent during usage. On mobile, users had to tap three times (open panel, select, close) to switch agents - too long a workflow.

## Changes

- Add `MobileAgentButton` with tap-to-cycle and long-press-to-open behavior
- Add `MobileModelButton` for direct model panel access
- Remove agent selection from `UnifiedControlsDrawer`
- Agent cycling remains accessible via Tab key shortcut

## Screenshots

image

## Verification

- bun run type-check passes
- bun run lint passes


btriapitsyn:
Nice idea, please resolve the branch conflict, so that it can be merged

PR #360, opened 2026-02-08
feat: improve session status handling and SSE batching

btriapitsyn:
### Summary
- Fixed session status parsing to accept both `sessionID` and `sessionId` across status event handlers.
- Restored/strengthened status reconciliation while sessions are active by triggering snapshot polls during busy/retry periods
- Improved assistant completion handling for short replies by finalizing on `status=completed|complete` and `finish=stop` (even when parts/timestamps are sparse)
- Reduced false busy flicker by refining streaming-part detection and adding a short guard after recently confirmed idle
- Improved restore/focus reliability with unconditional status polling on visibility/focus events
- Hardened snapshot sync with `no-store` fetch, queued in-flight immediate polls, and a follow-up settle poll
- Added global SSE coalescing/batching for session status events to reduce redundant updates and preserve latest state
### Notes
- No public API contract changes; this is internal status-sync and event-processing hardening
- Status indicator timing may be slightly less noisy due to coalescing and guarded fallback transitions
- Existing `triggerSessionStatusPoll` flow is now used more aggressively on reconnect/visibility/focus for faster correction

### DIFF
--- a/packages/ui/src/hooks/useEventStream.ts
+++ b/packages/ui/src/hooks/useEventStream.ts
@@ -784,17 +784,38 @@ export const useEventStream = () => {
         // Fallback: if we see assistant parts but session.status hasn't arrived yet, mark busy.
         if (roleInfo === 'assistant') {
           const partType = (messagePart as { type?: unknown }).type;
-          const isStreamingPart =
-            partType === 'step-start' ||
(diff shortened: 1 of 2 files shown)


gsxdsm:
Will test asap @shekohex @btriapitsyn

PR #379, opened 2026-02-10
fix: Gemini/Antigravity quota sources and update labels

gsxdsm:
## Summary
- Add multi-source Google quota support (Gemini, Antigravity) with per-model quotas aggregation
- Replace static 5h window handling with dynamic window selection per source
- Update window label formatting: 5h -> 5-Hour and daily -> Daily
- Introduce quota buckets fetch for per-model remaining/usage and per-model windows
## Why
- Enables correct quota reporting when multiple Google quota sources are configured
- Improves user-facing labels to be clearer and consistent with other windows
- Reduces reliance on model-level hard-coded window assumptions for Google quotas

## Testing
- [ ] Not run locally
- [ ] Verify UI displays per-model quotas for Gemini source with 5-Hour and Daily windows
- [ ] Verify UI displays per-model quotas for Antigravity source with appropriate window labels based on remaining time
- [ ] Validate token refresh flow uses correct client credentials for each source and that access tokens are refreshed when expired
- [ ] Ensure absence of model quota data does not crash the header and shows a sensible fallback

### DIFF
--- a/packages/vscode/src/quotaProviders.ts
+++ b/packages/vscode/src/quotaProviders.ts
@@ -91,6 +91,14 @@ const ANTIGRAVITY_ACCOUNTS_PATHS = [
   path.join(OPENCODE_DATA_DIR, 'antigravity-accounts.json'),
 ];
 
+// OAuth Secret value used to init client
+// Note: It's ok to save this in git because this is an installed application
+// as described here: https://developers.google.com/identity/protocols/oauth2#installed
+// "The process results in a client ID and, in some cases, a client secret,
+// which you embed in the source code of your application. (In this context,
+// the client secret is obviously not treated as a secret.)"
+// ref: https://github.com/opgginc/opencode-bar
+
 const ANTIGRAVITY_GOOGLE_CLIENT_ID =
   '1071006060591-tmhssin2h21lcre235vtolojh4g403ep.apps.googleusercontent.com';
 const ANTIGRAVITY_GOOGLE_CLIENT_SECRET = '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';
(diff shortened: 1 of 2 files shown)


packages/vscode/src/quotaProviders.ts
```diff
@@ -83,16 +91,21 @@ const ANTIGRAVITY_ACCOUNTS_PATHS = [
...
 ];
 
-const GOOGLE_CLIENT_ID =
+const ANTIGRAVITY_GOOGLE_CLIENT_ID =
   '1071006060591-tmhssin2h21lcre235vtolojh4g403ep.apps.googleusercontent.com';
-const GOOGLE_CLIENT_SECRET = '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';
+const ANTIGRAVITY_GOOGLE_CLIENT_SECRET = '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';
+const GEMINI_GOOGLE_CLIENT_ID =
+  '681255809395-oo8ft2oprdrnp9e3aqf6av3hmdib135j.apps.googleusercontent.com';
+const GEMINI_GOOGLE_CLIENT_SECRET = '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';
```
btriapitsyn, line 107: I would kindly ask you to add this [comment](https://github.com/google-gemini/gemini-cli/blob/2ae5e1ae20e401a018354ab867f894e09eef8ff3/packages/core/src/code_assist/oauth2.ts#L73-L78) here as well, to make it obvious for any future wanderers in this codebase :)

gsxdsm: Good call! Updated in both places it's used and provided a ref to opencodebar where the code and values are referenced

PR #528, opened 2026-02-26
feat: add openchamber nix flake

VDuchauffour:
This PR adds a nix flake for the project. All commands (`bun run type-check && bun run lint  && bun run build`) runs well.
I use the [opencode approach](https://github.com/anomalyco/opencode) to add this to the project.


VDuchauffour:
No problem, it makes sense 👍

PR #557, opened 2026-02-28
feat(chat): align activity timing UI with hover-only end timestamps

nelsonPires5:
## Summary
- add a shared timestamp formatter for activity/footer time display (`HH:mm:ss` for today, `yy-MM-dd HH:mm:ss` otherwise)
- update activity rows so duration remains visible while end timestamp appears only on hover
- keep final assistant footer timestamp always visible next to elapsed duration

## Screenshots
image
image

Resolves https://github.com/btriapitsyn/openchamber/issues/501


btriapitsyn (approved):

colinmollenhour:
Very nice, thanks @nelsonPires5 and @btriapitsyn !

PR #563, opened 2026-02-28
fix(chat): show in-flight tools immediately and keep collapsed activity live

nelsonPires5:
## Summary
- Show tool rows in `MessageBody` as soon as they become active (`pending`, `running`, `started`) instead of waiting for message completion.
- Keep finalized tools visible with the same existing behavior, while preserving standalone tool filtering and footer/copy behavior.

## Why
Users were not seeing live tool/activity progress until completion, which made streaming execution feel delayed and opaque.

### DIFF
--- a/packages/ui/src/components/chat/message/parts/ToolPart.tsx
+++ b/packages/ui/src/components/chat/message/parts/ToolPart.tsx
@@ -1455,18 +1455,14 @@ const ToolPart: React.FC<ToolPartProps> = ({
     const input = stateWithData.input;
     const time = stateWithData.time;
 
-    const [pinnedTaskTime, setPinnedTaskTime] = React.useState<{ start?: number; end?: number }>({});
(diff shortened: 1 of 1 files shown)


nelsonPires5:
> Does it show detailed/non-collapsed tool call results as the come in, then auto collapse when it is done? does this means it will show the bash output as it streams the output?

No, it solves a problem that happens today that the most recent tool call does not appear in the activity until ended. So now, the tool call appears when not finished yet, and do not appear the duration, so you know that is the tool call that's running. It doesn't streams the tool call output.

btriapitsyn:
@nelsonPires5 no, sorry, I'll review this. I just saw the commit from you recently here and thought you are still working on it 😅

btriapitsyn:
Nice work here. I pulled in `main`, resolved conflicts, and kept your intent intact. I also tightened the live experience by ensuring active non-`task` tools render immediately, and by stabilizing the duration timer so it no longer resets during streaming updates.

PR #564, opened 2026-03-01
fix(chat): align code block actions and restore horizontal scrolling across desktop and mobile

nelsonPires5:
## Summary
- move markdown code-block actions to Streamdown controls so actions sit in the header area and follow hover/focus behavior
- keep code block chrome constrained to chat width while making code content horizontally scroll inside the code body
- enable horizontal scrolling for virtualized code output rows
- restore horizontal code scrolling on mobile by overriding the generic mobile `.overflow-hidden` rule for Streamdown code bodies

### Before
image

### After

image
image

### DIFF
--- a/packages/ui/src/components/chat/MarkdownRenderer.tsx
+++ b/packages/ui/src/components/chat/MarkdownRenderer.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import { Streamdown } from 'streamdown';
-import { code } from '@streamdown/code';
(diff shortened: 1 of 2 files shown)


nelsonPires5:
> Hey @nelsonPires5, if you don't mind, I will also try to contribute to this PR a bit later with some cosmetic changes to how code blocks look, since they somehow got a visual regression lately. I think it will complement your change, so it doesn't need to be a separate PR from my side.

That’s a great idea! Feel free to add those changes

PR #591, opened 2026-03-03
feat: Agent loops

TilDoesCode:
This PR adds agents loops to Openchamber. Agent loops are a coordinated series of sessions automatically running after one another on a shared list of work packages. This is essentially an implementation of the popular Ralph Wiggum pattern, but with much better observability than most scripts can offer.
These loops can be easily created and controlled via a custom session UI that only applies to loop type sessions.

This PR also adds a devcontainer definition file to make working on OpenChamber easier, especially in Github Codespaces.
Bildschirmfoto 2026-03-03 um 21 12 20
Bildschirmfoto 2026-03-03 um 21 12 33
Bildschirmfoto 2026-03-03 um 21 12 42


TilDoesCode:
@nelsonPires5 If you havent implemented that yet, this could serve as a foundation for that. At some point you have to press play on the kanban board, at which point the loop implemented here would work its way across the tickets with the todo status. The workpackages file the loop works with already 80% of what youd need to store tickets on a board as well. If you then still expose the UI I made, or only show the board UI, is another story.

Issue #549, opened 2026-02-28
[Bug] The password-protected tunnel is locked.


nzlov:
### What's wrong? Maybe some steps to reproduce.

docker-compose.yaml
```
services:
  openchamber:
    image: openchamber
    container_name: openchamber
    ports:
      - "3000:3000"
    extra_hosts:
      - "host.docker.internal:host-gateway"
    volumes:
      - ./data/openchamber:/home/openchamber/.config/openchamber
      - ./data/opencode/share:/home/openchamber/.local/share/opencode
      - ./data/opencode/state:/home/openchamber/.local/state/opencode
      - ./data/opencode/config:/home/openchamber/.config/opencode
      - ./data/ssh:/home/openchamber/.ssh
      - ./workspaces:/home/openchamber/workspaces
    environment:
      HTTP_PROXY: http://100.64.0.15:9110
      HTTPS_PROXY: http://100.64.0.15:9110
      NO_PROXY: localhost,127.0.0.1,172.16.0.0/16,172.17.0.0/16,100.64.0.0/16,minimaxi.com
      UI_PASSWORD: "FhOMX1QatOj0m3TnGLBaAAJFP46gQSUOy9MOUO82TCZ2zFNfzRiMoyNgX5CnOm4e"  # Uncomment to set UI password
      CF_TUNNEL: password  # Enable Cloudflare Tunnel (true/qr/password)
      OH_MY_OPENCODE: true # enable oh-my-opencode
    # OPENCODE_HOST: http://172.17.0.1:4096 # Connect to external OpenCode server
    # OPENCODE_SKIP_START: true # skip start opencode
    restart: unless-stopped
```

use tunnel:

<img width="2945" height="1541" alt="Image" src="https://github.com/user-attachments/assets/abb71569-f404-4857-a2fb-b420e98c04eb" />

use local :

<img width="3036" height="1371" alt="Image" src="https://github.com/user-attachments/assets/f14aff6b-b508-4a1f-909c-cc92fac46d97" />

docker compose logs 
```
Attaching to oc
oc  | [entrypoint] SSH public key:
oc  | ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIEZ3r/2cfbhTMfm92m5W2x3p8uEZzWcBx5rvEq/rVRsz openchamber@d65d40c4ecfa
oc  | [entrypoint] UI password set, enabling authentication
oc  | [entrypoint] Cloudflare Tunnel enabled (password)
[...]

PR #601, opened 2026-03-05
feat: support custom localhost allowlist via environment variable

nzlov:
Allow tunnel auth localhost checks to include hosts from CUSTOM_LOCALHOST Parse CUSTOM_LOCALHOST with space/comma/semicolon separators and host normalization Cache parsed CUSTOM_LOCALHOST values to avoid repeated parsing at runtime

Fix: #549

### DIFF
--- a/packages/web/server/lib/opencode/tunnel-auth.js
+++ b/packages/web/server/lib/opencode/tunnel-auth.js
@@ -77,7 +77,18 @@ const normalizeHost = (candidate) => {
   if (!trimmed) {
     return null;
   }
-  return trimmed.replace(/:\d+$/, '');
+
+  const bracketedIpv6Match = trimmed.match(/^\[([^\]]+)\](?::(\d+))?$/);
+  if (bracketedIpv6Match) {
+    return bracketedIpv6Match[1];
+  }
+
+  const colonCount = trimmed.split(':').length - 1;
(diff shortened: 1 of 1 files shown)


packages/web/server/lib/opencode/tunnel-auth.js
```diff
@@ -80,11 +80,36 @@ const normalizeHost = (candidate) => {
...
+
+const getCustomLocalHostSet = () => {
+  const raw = typeof process.env.CUSTOM_LOCALHOST === 'string' ? process.env.CUSTOM_LOCALHOST : '';
+  if (raw === cachedCustomLocalHostRaw) {
+    return cachedCustomLocalHostSet;
+  }
+
+  const next = new Set();
+  for (const token of raw.split(/[\s,;]+/)) {
+    const normalized = normalizeHost(token);
```
nelsonPires5: This breaks unbracketed IPv6 literals in `CUSTOM_LOCALHOST`. `normalizeHost()` strips any trailing `:digits`, which is fine for `host:3000`, but it also mutates IPv6 addresses like `::1` into `:` and `2001:db8::1` into `2001:db8:` before they’re added to the allowlist. As a result, IPv6 entries in `CUSTOM_LOCALHOST` will never match the normalized request host.

PR #626, opened 2026-03-08
Feat: spell check toggle for desktop

### LINKED ISSUE #346: [Feature Request] Add an option to turn on spellcheck for the input box

I would be great if there was an option in the settings to turn on spell checking for the main input dialog (could also work for the git commit message dialog).

I find myself typing away rapidly and not really paying attention to spelling. I would appreciate spell checking, but I realize it might be distracting for most people as there will often be a lot of code or other jargon that will flag the spell checker. So I think an option in the chat settings to turn on spell checking (off by default) would be great.

The relevant code for the main input box is here: https://github.com/btriapitsyn/openchamber/blob/1bca4b460bea1e74530c0603311d7c99e9aef483/packages/ui/src/components/chat/ChatInput.tsx#L1823

### Extra context (optional)

_No response_

DocterZed:
Added a desktop spellcheck setting for chat input while preserving the current mobile behavior.

- `packages/ui/src/components/chat/ChatInput.tsx:158` now reads chatInputSpellcheckEnabled, and `packages/ui/src/components/chat/ChatInput.tsx:2685` uses spellCheck={isMobile || chatInputSpellcheckEnabled} so mobile stays on by default and desktop follows the new toggle.
- `packages/ui/src/components/sections/openchamber/OpenChamberVisualSettings.tsx:127` adds a new visible setting, `packages/ui/src/components/sections/openchamber/OpenChamberVisualSettings.tsx:201` persists it through shared settings, and `packages/ui/src/components/sections/openchamber/OpenChamberVisualSettings.tsx:1025` adds the checkbox row in Chat settings.
- `packages/ui/src/components/sections/openchamber/OpenChamberVisualSettings.tsx` The checkbox does not render on mobile (mobile still keeps spellcheck on by default, without showing the toggle).
- `packages/ui/src/components/sections/openchamber/OpenChamberPage.tsx:120` includes the new toggle in the Chat settings page.
- `packages/ui/src/stores/useUIStore.ts` add the new Zustand state, default, setter, and persisted local store field.
- `packages/ui/src/lib/desktop.ts:119`, `packages/ui/src/lib/api/types.ts:528`, `packages/ui/src/lib/persistence.ts:377`, `packages/ui/src/lib/persistence.ts:747`, and `packages/web/server/index.js:2097` wire the setting through the shared settings sync path so it survives reloads.
- `packages/ui/src/components/views/git/CommitInput.tsx:4` and `packages/ui/src/components/views/git/CommitInput.tsx:25` now reads chatInputSpellcheckEnabled from useUIStore, and spellCheck now follows `hasTouchInput || chatInputSpellcheckEnabled`, so touch/mobile keeps current behavior and desktop uses the same toggle as chat.

> ```
> bun run type-check
> bun run lint
> bun run build
> ```
All passed.

I tested the app on desktop (web) and mobile web. The setting correctly only shows on desktop, and the spellcheck toggle works for both the main chat input and the git commit input.

Closes #346

### DIFF
--- a/packages/ui/src/components/views/git/CommitInput.tsx
+++ b/packages/ui/src/components/views/git/CommitInput.tsx
@@ -9,6 +9,7 @@ interface CommitInputProps {
   placeholder?: string;
   disabled?: boolean;
   hasTouchInput?: boolean;
+  isMobile?: boolean;
 }
 
 const MIN_HEIGHT = 38; // Single line height
@@ -20,9 +21,10 @@ export const CommitInput: React.FC<CommitInputProps> = ({
   placeholder = 'Commit message',
   disabled = false,
   hasTouchInput = false,
+  isMobile = false,
 }) => {
   const textareaRef = React.useRef<HTMLTextAreaElement>(null);
-  const chatInputSpellcheckEnabled = useUIStore((state) => state.chatInputSpellcheckEnabled);
+  const inputSpellcheckEnabled = useUIStore((state) => state.inputSpellcheckEnabled);
 
   // Auto-resize based on content (layout phase to avoid mount flicker)
   React.useLayoutEffect(() => {
@@ -47,7 +49,7 @@ export const CommitInput: React.FC<CommitInputProps> = ({
       disabled={disabled}
       autoCorrect={hasTouchInput ? 'on' : 'off'}
       autoCapitalize={hasTouchInput ? 'sentences' : 'off'}
-      spellCheck={hasTouchInput || chatInputSpellcheckEnabled}
+      spellCheck={isMobile || inputSpellcheckEnabled}
       scrollbarClassName="hidden"
       className={cn(
         'rounded-lg bg-transparent resize-none overflow-y-hidden',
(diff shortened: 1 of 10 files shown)


packages/ui/src/components/views/git/CommitInput.tsx
```diff
@@ -45,7 +47,7 @@ export const CommitInput: React.FC<CommitInputProps> = ({
       disabled={disabled}
       autoCorrect={hasTouchInput ? 'on' : 'off'}
       autoCapitalize={hasTouchInput ? 'sentences' : 'off'}
-      spellCheck={hasTouchInput ? true : false}
+      spellCheck={hasTouchInput || chatInputSpellcheckEnabled}
```
nelsonPires5: On non-mobile touch-enabled devices (for example tablets in desktop-width layouts or touch laptops), this line can enable spellcheck in the commit input while the chat input stays disabled. CommitInput uses hasTouchInput, but ChatInput uses isMobile, so the two inputs no longer follow the same rule even though they share the same setting. I think this needs to be fixed so both inputs use the same condition and behave consistently.

DocterZed: Good catch. There is an inconsistency there. ChatInput.tsx uses:
```
autoCorrect={isMobile ? "on" : "off"}
autoCapitalize={isMobile ? "sentences" : "off"}
spellCheck={isMobile}
```
So the main chat input spell check etc currently only ever shows on small screen devices.

Whereas CommitInput.tsx uses:
```
autoCorrect={hasTouchInput ? 'on' : 'off'}
autoCapitalize={hasTouchInput ? 'sentences' : 'off'}
spellCheck={hasTouchInput ? true : false}
```
Which will apply on larger screen devices (as defined in device.ts: `prefersCoarsePointer || noHover || maxTouchPoints > 0`).

It is unclear to me if this is an intentional design choice to only have commit (but not chat) spell check on larger touch devices. So clarification on that would help.

For now I changed CommitInput to use isMobile to be consistent as that is what is also used to show or hide the spell check setting.

nelsonPires5: Nice

PR #695, opened 2026-03-17
feat(cli): add --foreground flag for systemd and process manager deployments

colinmollenhour:
## Summary

- Adds `--foreground` / `--no-daemon` to `openchamber serve`, which runs the server inline in the CLI process instead of spawning a detached daemon child
- Required for systemd `Type=simple` (and other process managers like Docker, supervisord) that track the direct child — the always-daemon behavior introduced in #640 broke this use case
- Documents `OPENCHAMBER_HOST` (bind address env var) in `--help`, which was implemented but never exposed to users

## Usage

```ini
# systemd service (Type=simple)
ExecStart=... openchamber serve --port 4094 --foreground
```

```bash
# or directly
openchamber serve --port 3000 --foreground
```

## How it works

When `--foreground` is set, the serve command skips the daemon child spawn entirely and instead dynamically imports `server/index.js` and calls `startWebUiServer()` inline with `attachSignals: true` / `exitOnShutdown: true`. The CLI process stays alive as the server, so the process manager tracks it correctly.

Signals (SIGTERM/SIGINT) are handled by the server's existing signal attachment logic.

### DIFF
--- a/packages/web/bin/cli.js
+++ b/packages/web/bin/cli.js
@@ -851,7 +851,7 @@ USAGE:
 COMMANDS:
   serve          Start the web server (daemon default)
   stop           Stop running instance(s)
-  restart        Stop and start the server (foreground instances stay attached)
+  restart        Stop and start the server
   status         Show server status
   tunnel         Tunnel lifecycle commands
   logs           Tail OpenChamber logs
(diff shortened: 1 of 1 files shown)


yulia-ivashko:
Thanks for the PR and for the clear systemd `Type=simple` use case — I agree `--foreground` is needed and the direction makes sense. I tested this branch locally, and I’m seeing a few blocking parity issues in the current implementation:

- Foreground instances are not discoverable by CLI lifecycle commands (`status` does not see them).

- Because of that, `stop` and `restart` do not work for foreground flows as expected.

- `--foreground --json` does not return stable machine-readable startup JSON (it behaves like normal text output).

- In foreground serve, `Ctrl+C` currently does nothing. This should trigger a graceful shutdown path (SIGINT/SIGTERM handling) and terminate cleanly.

Given these results, I don’t think this is merge-ready yet: the feature goal is correct, but CLI behavior is inconsistent across modes.
My expectation before merge is:

1. Fix interrupt handling so foreground responds to Ctrl+C and shuts down properly.
2. Restore lifecycle parity for foreground (`status` / `stop` / `restart`) or explicitly implement deterministic “unsupported” behavior with clear messaging.
3. Define deterministic --foreground --json semantics (either startup JSON + continue running, or explicit JSON error).

packages/web/bin/cli.js
```diff
@@ -3104,19 +3272,22 @@ const commands = {
...
+          port: restartPort,
           explicitPort: true,
           uiPassword: options.explicitUiPassword ? options.uiPassword : storedOptions.uiPassword,
+          foreground: isForeground,
           suppressStartupSummary: true,
-          quiet: true,
+          quiet: !isForeground,
           suppressUiPasswordWarning: true,
           suppressQuietOutput: true,
         });
```
colinmollenhour, line 3269: > **AI Code Review** · Flagged by: colin-review-opus, colin-review-gpt-codex, colin-review-glm

**Bug: `restart --json` with foreground instance blocks forever without emitting JSON output**

When `restart` operates on a foreground instance (either via `--foreground` flag or because the discovered instance has persisted `launchMode: 'foreground'`), `this.serve({ foreground: true })` blocks forever at `await new Promise(() => {})`. The `printJson()` call after the for-loop is never reached.

The `serve` command correctly rejects `--json + --foreground` with a `TunnelCliError`, but `restart` doesn't propagate `json: true` to the internal `serve()` call, so that guard is bypassed.

**Impact:** `openchamber restart --json` with a foreground instance produces zero output and hangs indefinitely — violating the CLI Parity and Safety Policy: _"The same functional outcome and safety gates MUST hold for all execution modes"_ and _"`--json` changes output shape only; it does not weaken validation."_

```
Fix packages/web/bin/cli.js:3276: restart --json with foreground instance blocks forever. Either propagate json: true to the serve() call so the existing guard fires and throws a TunnelCliError, or add an early check in restart that rejects --json when hasForegroundRestart is true (before entering the loop).
```

colinmollenhour: This was fixed by still supporting "restart --json" and emitting a JSON block for the restart command, just not for the "serve --foreground" command.

PR #757, opened 2026-03-24
fix: dispatch namespaced slash commands instead of sending literal text

### LINKED ISSUE #701: [Bug] Using nested command does not inject command template

I have my personal commands namespaced in a directory so my commands don't conflict with others - for example, "/colin/commit". These show up in the autocomplete, but when actually invoked, only the literal text is sent and the command template is not used at all. I tested this with the OpenCode WebUI and it does work there as expected.

### Where does it happen?

Desktop Web

### Version (if known)

1.8.7

Image

### Logs (optional)

```shell

```

colinmollenhour:
## Summary

- Fixes namespaced slash commands (e.g. `/colin/commit`) being sent as literal text instead of dispatching via the command template
- The `firstTokenLooksLikeAbsolutePath` heuristic in `messageStore.ts` was misidentifying namespaced commands as file paths because their names contain `/`
- Added a check against known commands from `useCommandsStore` before concluding a token is an absolute path

## Root Cause

The heuristic at `packages/ui/src/stores/messageStore.ts:1057-1067` checks whether the first token of user input looks like an absolute file path (to avoid treating `/usr/bin/python` as a command). Its logic: "if the token starts with `/` and the remainder contains `/`, it's a path."

A namespaced command like `/colin/commit` triggers this false positive because `colin/commit` contains `/`. This causes `commandPayload` to return `null`, and the dispatch falls through to `sendMessage()` — sending the literal text as a chat message.

## Fix

Before the heuristic returns `true`, check if the token (minus the leading `/`) matches a known command name via `useCommandsStore.getState().getCommandByName()`. If it matches, return `false` — it's a command, not a path.

This is a minimal 4-line change (1 import + 2 lines of logic + 1 comment line).

Closes #701


packages/ui/src/stores/messageStore.ts
```diff
@@ -1063,6 +1064,9 @@ export const useMessageStore = create<MessageStore>()(
                                 if (firstToken.length <= 1) return false;
                                 const tokenWithoutLeadingSlash = firstToken.slice(1);
                                 if (!tokenWithoutLeadingSlash.includes('/')) return false;
+                                // Namespaced commands (e.g. /colin/commit) contain '/' but
```
colinmollenhour:
```suggestion
                                // Namespaced commands (e.g. /my/commit) contain '/' but
```

btriapitsyn (approved):

PR #764, opened 2026-03-25
feat(web): add WebSocket transport for message event streaming with SSE fallback

geekifan:
## Summary

This PR adds WebSocket support for the UI message event stream while keeping SSE as a supported fallback.

OpenChamber currently relies on SSE for chat and session event streaming. This change adds a WebSocket transport between the browser UI and the OpenChamber web server, while the server still proxies the upstream OpenCode SSE stream. It also adds a user-facing setting to choose `Auto`, `WebSocket`, or `SSE`.

## What changed

- Add message event WebSocket endpoints on the web server:
  - `/api/global/event/ws`
  - `/api/event/ws`
- Keep upstream OpenCode integration on SSE and bridge SSE -> WebSocket on the OpenChamber server
- Preserve existing OpenChamber synthetic events on the WS path, including:
  - `openchamber:session-status`
  - `openchamber:session-activity`
  - `openchamber:notification`
  - `openchamber:heartbeat`
- Add a shared setting: `messageStreamTransport`
  - `auto`
  - `ws`
  - `sse`
- Add a Chat settings control for choosing the message stream transport
- Update the UI event client to:
  - prefer WebSocket in `auto` mode
  - fall back to SSE when WS fails before becoming ready
  - keep the existing event handling shape unchanged for consumers

## Notes

- `Auto` mode currently uses an optimistic WS-first strategy and temporarily falls back to SSE if WS fails before the stream becomes ready.


btriapitsyn:
> I noticed the WS logic (~200 lines) is inlined in index.js alongside the existing ~14K lines.

I would kindly ask you to update your code and resolve conflicts so that it will become reviewable.

geekifan:
@btriapitsyn I apologize for the delayed code fix; I was on vacation recently. I've rebased the code and decoupled the functional modules. It is ready for your review. Thank you :)

btriapitsyn (approved):

Issue #411, opened 2026-02-12
[Feature Request] Docker image and publication to ghrc.io


VDuchauffour:
### What should we add/change?

Hello,
Could be possible to add a dockerfile and a github action that build and push the docker image to the ghcr registry ? This would make way more easy to host the project on a remote server. A workflow triggered on every release would be perfect. I can open a PR if you want.
Thanks

TheFrenchGhosty:
I made one for those interested: https://github.com/TheFrenchGhosty/docker-openchamber-quay @VDuchauffour 

Image from the latest release (uses the npm package) and image that builds from source

x86_64 + ARM64

PR #801, opened 2026-03-31
Add github workflow for creating docker images. 

ForceConstant:
Addresses #411 

Note after accepting pull request, will need to update repo settings to make sure project docker repo is public.


.github/workflows/docker.yml
williamokano, line 1: I believe this would be better if added to the release.yml workflow.

Also, you could use the same approach as the npm release to extract the version in a more standardized way.

Like here: https://github.com/openchamber/openchamber/blob/main/.github/workflows/release.yml#L109-L111

PR #827, opened 2026-04-02
fix(files): refresh open file after external edits

kostazol:
Summary
This PR fixes stale file content in the Files view when a file is changed outside OpenChamber.

Previously, if a markdown or text file was already open in the UI and then edited through the filesystem, the open editor could keep showing old content. In practice, users often had to fully reload the page to see the latest file state.

What changed
Added external-change detection for the currently open file in the Files view.
Reload open file content when the file changes on disk and the editor has no unsaved local changes.
Updated the file-content cache to validate cached entries against file metadata instead of trusting path-only cache hits.
Extended statFile to include mtimeMs so shared UI logic can reliably detect external file updates.
Kept web and VS Code file-runtime contracts in sync.
Why this approach
The issue had two separate causes:

The open file view did not actively detect that the file had changed on disk.
The runtime file-content cache could return stale content for the same path even after an external edit.
Fixing only one of these was not enough. This PR addresses both, while keeping the behavior conservative:

we only refresh when the file actually changed
we do not overwrite unsaved local edits
User impact
After this change, if a file is open in the Files view and is modified externally:

the UI picks up the change automatically
closing and reopening the file shows fresh content
a full page reload is no longer required


jwcrystal (changes requested):
Thanks for this PR — the direction is correct. The cache validation layer is essential since the current `main` cache uses path-only hits, meaning reopening a file after external edits still returns stale content.

Two issues that should be addressed:

### 1. Double reload on external change
The poller in `FilesView.tsx` does both `setLoadedFilePath(null)` and `loadSelectedFile()` when detecting a change. This causes the manual call plus the existing effect (triggered by `loadedFilePath` becoming `null`) to fire twice. Pick one — either set `loadedFilePath` to `null` and let the existing effect handle it, or call `loadSelectedFile` directly without touching `loadedFilePath`.

### 2. Non-atomic read+stat (TOCTOU) in cache
`readFreshFile()` in `RuntimeAPIProvider.tsx` uses `Promise.all([readFile, statFile])`. If the file changes between the two, the cache may store old content with new metadata, making `statMatches` always pass on subsequent hits. Consider stat → read → stat, and only accept if metadata is consistent.

kostazol:
fixed with https://github.com/openchamber/openchamber/commit/3115d5eb0e04f7911dcb706630be799746cd0ec5

@jwcrystal  Thanks you!

PR #834, opened 2026-04-03
fix(chat): show thinking/justification before question card in sorted mode

nguyentamdat:
## Summary

- In sorted render mode during streaming, text/reasoning/justification parts were hidden until the message completed (`finish === 'stop'`). When the model called the `question` tool, the `QuestionCard` rendered immediately via SSE, but the preceding thinking/justification content was invisible.
- Adds `hasBlockingToolCall` detection for question tool parts, and bypasses the sorted-mode text/reasoning deferral when the model is paused waiting for user input.
- Also fixes justification rendering: only skips in flat loop when the activity group will render it, preventing hidden justification when no activity group is active.


btriapitsyn:
Hello @nguyentamdat, sorry but this feature needed more broad changes in relationships between text and tools than just fixing them for questions. I implemented fix that now makes text/tools/reasoning to be displayed right when the tool appears and not when the whole message is finished. This automatically satisfying your idea and making tool feel more live in Sorted mode.

PR #863, opened 2026-04-06
feat: use global OpenCode config for providers and agents

wpbiggs:
## Summary

Providers and agents now load from global OpenCode config (`~/.config/opencode/opencode.json`) instead of being cached per-project.

**Problem:** Users had to re-configure providers and agents for each project. Different ports also created isolated localStorage caches, making settings appear to change randomly.

**Root causes:**
1. `directoryScoped` state in `useConfigStore` cached providers/agents per directory, persisted to localStorage
2. Directory parameter passed to OpenCode SDK caused project-scoped config lookups
3. Browser localStorage is per-origin (different ports = different storage)

**Changes:**
- Removed `directoryScoped` caching from `useConfigStore` - providers/agents are now loaded once globally
- Removed directory parameter from `getProviders()` and `listAgents()` SDK calls
- Simplified `activateDirectory()` to not reload providers/agents per project
- Fixed server proxy to strip `content-encoding` header (Bun auto-decompresses responses, causing `ERR_CONTENT_DECODING_FAILED` when header was forwarded)

**Files changed:**
- `packages/ui/src/stores/useConfigStore.ts` - Removed per-directory caching (~500 lines simplified)
- `packages/ui/src/stores/useAgentsStore.ts` - Removed directory iteration in config refresh
- `packages/ui/src/lib/opencode/client.ts` - SDK calls now omit directory parameter
- `packages/web/server/index.js` - Added `content-encoding` to hop-by-hop headers

## Testing

1. Configure providers/agents in `~/.config/opencode/opencode.json`
2. Open OpenChamber on different ports
3. Verify providers/agents appear consistently across projects and ports


btriapitsyn:
Thanks a lot for this PR and for the clear write-up.
I pulled it locally and resolved conflicts to test the behavior end-to-end. After testing, I’m concerned that removing per-project provider/agent configuration could break several existing workflows and user use cases that rely on project-level separation.
So I’m not going to merge this in its current form right now. I want to think a bit more about the right implementation so we can address the original pain without losing project-specific behavior.
Really appreciate the contribution and effort here.

PR #873, opened 2026-04-08
fix: provide UI feedback when /compact command is issued

### LINKED ISSUE #840: [Bug] Command '/compact' make seesion get stuck

When I use /compact command to manually compact the context, there is no message sent. After restarting the openchamber and back to the session, it shows that the /compact command is successfully sent and executed by opencode agent. 
Sometimes I can get the notification of the task if I don't restart the openchamber that quick, but there is still nothing shows up.
This command works fine on my machine with both opencode tui and desktop app.

### Where does it happen?

Desktop (macOS)

### Version (if known)

1.9.3

_No response_

### Logs (optional)

```shell

```

jwcrystal:
Close #840 

## Summary

- Fixes #840 — the `/compact` command appeared to freeze the session with no visual feedback
- Adds `compactSession()` to `session-actions.ts` using the same optimistic message pattern as `optimisticSend()`, but adapted for the summarize API which has no `messageID` parameter

## Problem

When a user ran `/compact`, `ChatInput.tsx` called `sdk.session.summarize()` directly and returned. There was **no optimistic message**, **no busy indicator**, and **no immediate UI feedback**. The session looked stuck until SSE events eventually arrived — sometimes after a long wait, sometimes only visible after restarting.

## Approach

The fix mirrors `optimisticSend()` but accounts for a key difference: the summarize API doesn't accept a `messageID`, so the server generates its own message IDs for compaction messages.

**Strategy:**
1. Insert an optimistic user message (`/compact`) for instant visual feedback
2. Set session status to `busy` so the UI shows activity
3. Call `sdk.session.summarize()` — server runs the compaction agent
4. After the API resolves, **remove the optimistic placeholder** to avoid a duplicate (optimistic + real server message)
5. SSE events (`message.updated`, `message.part.updated`, `session.idle`) deliver the actual compaction result and clear the busy status

**Why not reuse `optimisticSend()` directly?** `optimisticSend` passes the client-generated `messageID` to the server via `promptAsync({ messageID })` so both share the same ID. The summarize API has no such parameter — the server always generates its own IDs. Keeping the optimistic message after the API resolves would cause a visible duplicate user message.

## Files changed

| File | Change |
|------|--------|
| `packages/ui/src/sync/session-actions.ts` | Add `compactSession()` with optimistic message, busy status, and post-resolve cleanup |
| `packages/ui/src/sync/index.ts` | Export `compactSession` |
| `packages/ui/src/components/chat/ChatInput.tsx` | Replace direct `sdk.session.summarize()` call with `sessionActions.compactSession()` |

## Test plan

- [ ] Run `/compact` in a session with enough context — user message appears immediately, busy indicator shows, compaction result appears after completion
- [ ] Run `/compact` when server is unreachable — optimistic message is removed, error is surfaced, session returns to idle
- [ ] Verify no duplicate user messages appear after compaction completes
- [ ] Verify session status returns to idle after compaction agent finishes (via `session.idle` SSE event)

### DIFF
--- a/packages/ui/src/components/chat/ChatInput.tsx
+++ b/packages/ui/src/components/chat/ChatInput.tsx
@@ -1462,13 +1462,11 @@ const ChatInputComponent: React.FC<ChatInputProps> = ({ onOpenSettings, scrollTo
                 return;
             }
             else if (commandName === 'compact' && currentSessionId) {
-                const { opencodeClient } = await import('@/lib/opencode/client');
-                const sdk = opencodeClient.getSdkClient();
                 const configState = useConfigStore.getState();
-                await sdk.session.summarize({
-                    sessionID: currentSessionId,
-                    modelID: configState.currentModelId || '',
+                await sessionActions.compactSession({
+                    sessionId: currentSessionId,
                     providerID: configState.currentProviderId || '',
+                    modelID: configState.currentModelId || '',
                 });
                 return;
             }
--- a/packages/ui/src/sync/index.ts
+++ b/packages/ui/src/sync/index.ts
@@ -146,6 +146,7 @@ export {
   shareSession,
   unshareSession,
   optimisticSend,
+  compactSession,
   abortCurrentOperation,
   respondToPermission,
   dismissPermission,
--- a/packages/ui/src/sync/session-actions.ts
+++ b/packages/ui/src/sync/session-actions.ts
@@ -336,6 +336,111 @@ export async function optimisticSend(input: {
 }
 
 // ---------------------------------------------------------------------------
+// Compact (summarize session)
+// ---------------------------------------------------------------------------
+
+/**
+ * Compacts a session by calling the server's summarize endpoint.
+ *
+ * Unlike optimisticSend (which passes the client messageID to the server so
+ * both share the same ID), the summarize API has no messageID parameter.
+ * The server generates its own IDs for compaction messages.
+ *
+ * Strategy:
+ * 1. Insert an optimistic user message ("/compact") for instant feedback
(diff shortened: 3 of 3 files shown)


btriapitsyn:
Hey @jwcrystal 
I would like you to re-test it without your changes, but with the latest main branch code, since I don't have such issues. There is immediate effect with /compact command in chat.

jwcrystal:
I’ll try it again. The stat is #840, and compact action has no reaction until restarting the app.

Briefly, performing compaction without any screen feedback, but the content is visible on the OPENCODE console.

jwcrystal:
> Hey @jwcrystal I would like you to re-test it without your changes, but with the latest main branch code, since I don't have such issues. There is immediate effect with /compact command in chat.


I checked later fixes like #817 and #889. My read is that they likely mitigate the symptom indirectly by improving SSE recovery and sync robustness, but they do not replace what this PR does.

This PR addresses the submit-time UX gap for `/compact`: the current flow still calls `sdk.session.summarize()` directly, without explicit optimistic local feedback before the first SSE update arrives.

So #817 / #889 may make the issue less visible in practice, but they are not a strict replacement for this PR. That also explains why the issue may be harder to reproduce now: later sync fixes can improve the overall behavior, even if this specific optimistic `/compact` flow is still not implemented.

The practical question, then, is whether the current behavior is already good enough after those sync fixes, or whether the explicit optimistic `/compact` feedback in #873 is still worth keeping for consistency and slower-event scenarios.

jwcrystal:
This situation never arises because the code undergoes significant changes during each iteration.
The PR was specifically for the openchamber and opencode versions at that time.

PR #889, opened 2026-04-10
fix(sync): remove stale-delta skip and add parts-gap recovery

jwcrystal:
## Summary

Fixes a regression introduced in PR #803 (`b32ed938`) where the event pipeline's stale-delta mechanism silently dropped valid `message.part.delta` events, causing assistant messages to render blank or incomplete.

### Root cause

PR #803 added event coalescing with a stale-delta optimization: when a `message.part.updated` coalesces (replaces) an earlier event in the queue, all `message.part.delta` events for the same `(messageID, partID)` were marked as stale and skipped during flush. However, the implementation did not distinguish between deltas that arrived **before** vs **after** the coalesced `part.updated` — it unconditionally skipped all matching deltas regardless of queue position.

This is incorrect because:
- `message.part.updated` is a **snapshot** (full replacement)
- `message.part.delta` is an **incremental append**
- A delta arriving _after_ a `part.updated` snapshot must still be applied on top of that snapshot

In practice, a single streaming session could trigger this within a 16ms flush window — no "massive concurrency" needed.

### Changes

**Layer 1 — Pipeline: Remove stale-delta mechanism** (`event-pipeline.ts`)
- Remove `staleDeltas` Set and `deltaKey()` function entirely
- Coalesce still replaces same-key events (correct dedup), but deltas are never skipped
- The reducer naturally handles ordering: snapshot first, then incremental deltas apply on top

**Layer 2 — Reducer diagnostics** (`event-reducer.ts`, `debug.ts`)
- Add gated diagnostic logging (zero cost when disabled)
- Log when `message.updated` is skipped (unchanged), when parts are missing, and when partIDs are not found
- Enable in browser console: `localStorage.setItem("openchamber:sync:debug", "1")`
- This debug flag is **intentionally kept** for future sync debugging — reviewers please note

**Layer 3 — Parts-gap recovery** (`sync-context.tsx`)
- Add `enqueuePartsRepair()` + `repairSessionParts()` — targeted SDK re-fetch for sessions with missing parts
- 5-second cooldown per `(directory, sessionID)` to avoid redundant fetches
- Trigger on reducer `return false` for `message.part.delta` / `message.part.updated` (parts missing or partID not found)
- Trigger on `message.updated` `return true` when role is `assistant` but `state.part[messageID]` is empty (blank bubble)
- Repair preserves existing parts: only patches entries where `!existing || existing.length < newParts.length`

**Layer 3.5 — Pipeline coalescing tests** (`event-pipeline.test.js`)
- Verify that `message.part.delta` events are delivered after a coalesced `message.part.updated`
- Verify that two `message.part.updated` events for the same part coalesce into one

### Debug flag

This PR adds `packages/ui/src/sync/debug.ts` — a lightweight diagnostic module gated behind `localStorage.setItem("openchamber:sync:debug", "1")`. It is purposefully included as a permanent observability tool for the SSE sync layer, not debug-only cruft. All checks are early-returns on the hot path with zero runtime cost when disabled.

### Test plan

- [x] `bun run lint` passes
- [x] `bun run build` passes
- [x] `bun test packages/ui/src/sync/__tests__/event-pipeline.test.js` — 6 pass
- [x] Manual: start a long assistant response under Slow 3G, verify content renders completely
- [x] Manual: switch tabs during streaming, verify no lost content on return
- [x] Manual: enable `localStorage.setItem("openchamber:sync:debug", "1")`, verify `[sync:pipeline]`, `[sync:reducer]`, `[sync:dispatch]` logs appear in console

### DIFF
--- a/packages/ui/src/sync/__tests__/event-pipeline.test.js
+++ b/packages/ui/src/sync/__tests__/event-pipeline.test.js
@@ -35,6 +35,22 @@ function createSdkWithSingleEvent(event, hold) {
   };
 }
 
+// Helper to create an SDK that yields multiple events in sequence, then holds.
+function createSdkWithEvents(events, hold) {
+  return {
+    global: {
+      event: async () => ({
+        stream: (async function* () {
+          for (const event of events) {
+            yield event;
+          }
+          await hold;
+        })(),
+      }),
+    },
+  };
+}
+
 describe('createEventPipeline', () => {
   it('falls back to payload.properties.directory when the SDK event omits top-level directory', async () => {
     installDomStubs();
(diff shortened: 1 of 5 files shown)


jwcrystal:
@btriapitsyn Please REVIEW this because it is annoying when the message disappears.

btriapitsyn:
I will, right when I'm back from the vacation. Thank you for contribution 🙂

btriapitsyn (approved):

Issue #810, opened 2026-04-01
[Bug] The subagent has already finished, but the UI still shows `waiting for subagent activity`.


InCerryGit:
### What's wrong? Maybe some steps to reproduce.

When the agent starts a subagent, the subagent has already finished running, but it still remains in a waiting state. After pressing Ctrl + F5, the problem is resolved.

### Where does it happen?

Mobile (Web/PWA)

### Version (if known)

1.9.2

### Screenshots / recordings (optional)

<img width="1364" height="720" alt="Image" src="https://github.com/user-attachments/assets/794e2918-7541-4f9c-9730-84f58771eb04" />

### Logs (optional)

```shell

```

jwcrystal:
Root cause identified: when a subagent (task tool) completes, the parent session's tool part relies solely on SSE `message.part.updated` events to reflect the final state. Two gaps existed:

1. **SSE reconnect**: Child sessions were resynced, but parent sessions weren't — tool parts stayed stale.
2. **Normal SSE flow**: If `message.part.updated` for the parent's task tool part was coalesced or delayed, the UI got stuck with no fallback to fetch the latest state.

Both gaps were missed because testing only verified the happy path where SSE events arrive in order. No test covered the case where SSE events for the parent session are lost or delayed after a child completes.

Fix in #869:
- On SSE reconnect: resync parent sessions alongside child sessions.
- On normal completion: when child polling stops, proactively fetch the parent session's latest parts for the target message, so the tool part reflects the final state even if SSE was coalesced.

jwcrystal:
@InCerryGit Could you test with the latest build from #869 when you get a chance? Specifically:

1. Start a task that spawns a subagent
2. Let the subagent complete normally — check if the parent session updates without manual refresh
3. If possible, test with an unstable network (e.g., throttle in devtools) to stress the SSE reconnect path

Thanks for the thorough reports 👍

PR #903, opened 2026-04-13
fix(task): prevent subagent silent failures in session resolution and polling lifecycle

jwcrystal:
## Summary

Fixes a remaining task/subagent handoff failure in the parent UI.

In some cases, a subagent appears to stop without returning usable output, and the parent ends up continuing the work itself. This happened when the parent either failed to bind to the child session in time or stopped polling before the child result was captured.

## Root cause

Two edge cases remained after #817:

1. **Fallback child-session binding was still too narrow**
   - `resolveFallbackTaskSessionId` only matched within a 3s window
   - under load, child sessions can appear later than that

2. **Polling/final-fetch timing could miss the child result**
   - especially when the child went busy → idle before the parent observed it as active

## Fix

- widen fallback child-session matching on retry (`3s -> 8s`)
- stop polling first, then run final fetch in a separate effect
- avoid races between final-fetch writes and in-flight polling writes
- retry safely if final fetch fails

## Note on out-of-scope behavior

This PR does **not** change how the UI distinguishes a child session that truly errored/aborted from one that completed normally. If a subagent genuinely stops mid-execution, the parent may still continue on its own and the UI may not explain why.

That is a separate observability issue, tracked in **#904**, and not required for the Mode 1 fix in this PR.

## Related

- #810
- #817
- #904

## Validation

- `bun run lint` ✅
- `bun run type-check` ✅

### DIFF
--- a/packages/ui/src/components/chat/message/parts/ToolPart.tsx
+++ b/packages/ui/src/components/chat/message/parts/ToolPart.tsx
@@ -1851,6 +1851,10 @@ const ToolPart: React.FC<ToolPartProps> = ({
         return parseTaskMetadataBlock(taskOutputString);
     }, [taskOutputString]);
 
+    // Track whether fallback session resolution has failed at least once.
+    // When true, resolveFallbackTaskSessionId widens its time window (3s → 8s).
+    const [taskFallbackRetried, setTaskFallbackRetried] = React.useState(false);
+
     const explicitTaskSessionId = React.useMemo<string | undefined>(() => {
         if (!isTaskTool) {
             return undefined;
@@ -1885,8 +1889,9 @@ const ToolPart: React.FC<ToolPartProps> = ({
                 isTaskFinalized: isFinalized,
                 sessions: storeState.session,
                 sessionStatusMap: storeState.session_status,
+                hasRetried: taskFallbackRetried,
             });
-        }, [explicitTaskSessionId, isTaskTool, currentSessionId, taskSessionResolutionStart, isFinalized]),
+        }, [explicitTaskSessionId, isTaskTool, currentSessionId, taskSessionResolutionStart, isFinalized, taskFallbackRetried]),
         currentDirectory,
     );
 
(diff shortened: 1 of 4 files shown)


jwcrystal:
> This issue is more common than it appears on the tracker, but manifests in two different ways:

**1. Child finished, but parent never captured the result** (addressed by this PR)
The symptom looks like "the subagent stopped without returning output, and the parent did the work itself." In reality, the child completed — the parent just failed to bind to the child session or stopped polling before capturing the result.

Most likely triggers:
- Subagent completes quickly before the parent starts observing
- Server under load — child sessions take >3s to appear in the store
- SSE events arrive out of order (child idle before child busy)

**2. Child actually didn't finish or was interrupted mid-execution** (NOT addressed here — separate gap)
The subagent truly stops without producing output — due to model early termination, server error, timeout, or SDK-level interruption.

This is harder to observe because:
- `session.error` and `session.idle` produce the same `session_status = { type: "idle" }` in the event reducer — the parent cannot distinguish a failed child from a completed one
- child session errors are explicitly skipped in notification dispatch (`if (session.parentID) return`)
- `ToolPart` treats all six terminal states (`completed`, `error`, `aborted`, `failed`, `timeout`, `cancelled`) identically for the polling lifecycle — no special handling for error paths
- only `status === 'error'` renders a styled error box; `aborted`, `failed`, `timeout`, `cancelled` silently stop without visible feedback

Both modes produce the same user-visible symptom: **subagent appears to stop, parent continues alone.** That's why this is hard to report as a distinct bug — users see "the model decided to do it itself."

This PR fixes mode 1. Mode 2 would require making `session.error` distinguishable from `session.idle` in the event reducer and adding visible error feedback for child sessions — out of scope but tracked.

See #810 for a related report where the UI stayed stuck on "waiting for subagent activity."

jwcrystal:
Dropped the earlier Mode 2 experiment from this PR.

Reason: making `session.error` observable by writing a local `{ type: "error" }` into `session_status` changed the SDK-backed store contract and made the review scope wider than necessary.

This PR now stays focused on Mode 1 only:
- child session binding retry (`3s -> 8s`)
- serialized final-fetch after polling stops

Known remaining gap: if a child session truly errors/aborts mid-execution, the parent may still continue on its own without a clear UI explanation. That is a separate observability concern tracked in #904, and not required for this handoff/polling fix.

btriapitsyn (approved):

PR #908, opened 2026-04-13
perf(sync): optimize multi-session event pipeline with per-directory queues and delta coalescing

jwcrystal:
## Summary

- Split the event pipeline's global queue into **per-directory queues** with independent flush timers, so a busy project's streaming deltas no longer delay events for other projects.
- **Coalesce consecutive `message.part.delta` events** for the same `(messageID, partID, field)` into a single accumulated delta per 16ms flush window.
- Adds 9 unit tests and a runnable synthetic benchmark for both changes.

## Why

Running multiple sessions concurrently — either across open projects, or a single project with parent + subagent tasks — causes UI latency. The bottleneck is the frontend event pipeline, not provider/model side: `event-pipeline.ts` funnels all directories through one SSE consumer with a single shared queue, so any busy directory blocks the rest. And `message.part.delta` is the hottest event type but was the only one not being coalesced, so every streamed token became a separate reducer call and `setState`.

Subagents make this worse because they are independent child sessions sharing the parent's directory (`parentID` linkage) — any delegating task doubles or triples the event rate on a single directory queue.

## Changes

- **`98d013a`** — `event-pipeline.ts` queue state keyed by directory; each directory owns its own flush timer. Public API unchanged; SSE source and network layer untouched.
- **`258acf0`** — `message.part.delta` added to the coalesce key set; on hit, the new delta string is appended to the already-queued event. Verified safe against `event-reducer.ts:241-258`, which treats deltas as pure string append with no per-event side effects. `staleDeltas` supersession still works.
- **`295738e`** — tests + benchmark.

## Benchmark

Run locally: `bun packages/ui/src/sync/__tests__/event-pipeline.bench.js`

| Scenario | In → Out | Delta reduction | Integrity |
|---|---:|---:|:---:|
| Single project, parent + 3 subagents, 500 tokens each | 2,012 → 13 | **99.8%** | ✓ |
| 5 projects × parent + 3 subagents, 200 tokens each | 4,060 → 70 | **99.5%** | ✓ |
| Stress: 10 projects × 5 sessions × 1,000 tokens | 50,150 → 200 | **99.9%** | ✓ |

"Integrity ✓" = concatenated delivered delta bytes exactly equal input total — no content dropped.

## Test Plan

- [x] `bun test packages/ui/src/sync/__tests__/event-pipeline.test.js` — 13/13 passing
- [x] `bun run --cwd packages/ui type-check` / `lint` — clean
- [ ] Open two projects, run a streaming task in one, verify the other project's UI updates arrive without lag
- [ ] Run a task that delegates to a subagent, verify parent and subagent streams are both smooth
- [ ] Visibility toggle past 15s heartbeat timeout, verify reconnect + flush on return


jwcrystal:
@btriapitsyn Perhaps this is more important. Please review this when you have some free time.

btriapitsyn (approved):

PR #913, opened 2026-04-15
feat: session worktree isolation

jwcrystal:
## Summary

- Session branches are now fixed to their attached worktree — switching branches in another session no longer causes drift
- Authoritative `SessionWorktreeAttachment` in `session-worktree-store` is the single source of truth for session directory and branch
- Header/GitView read attachment branch/status first, falling back to live git only when no attachment or degraded/legacy
- Web, desktop, and VS Code runtimes all expose `validateWorktreeDirectory` and `canonicalizeWorktreeState` (3-way parity)
- Mutation safety gates block branch checkout/rename/creation when worktree is missing, invalid, or has in-progress operations (merge/rebase/cherry-pick/revert)

### Key changes (10 commits, 25 files, +2603/-32)

| Phase | Commit | What |
|-------|--------|------|
| Task 1 | `3f4300ac` | Session/worktree contract types + canonicalizeWorktreeState API |
| Task 2 | `e34fa5ba` | Canonicalize metadata producers (worktreeManager fills canonical fields) |
| Task 3 | `bdbcd6f6` | Route sessions through authoritative worktree attachments |
| Task 4 | `3b49cdcf` | Clarify session worktree targets (ChatInput, buildSessionTargetOptions) |
| Task 5 | `de1579a6` | Show worktree-backed session state UI (badges, attention indicators) |
| Task 6 | `4e56ac85` | Enforce safe mutations for attached worktrees (blocking reasons) |
| Task 7 | `d43594e6` | Full verification + cleanup |
| Phase A+B | `ef1376bc` | Attachment first-priority for directory resolution; createSession/setCurrentSession recovery |
| Phase C | `d499bed6` | Header/GitView read attachment branch as primary source |
| Phase D | `e14b2786` | Web/VS Code runtime parity for validate + canonicalize |

### New files
- `session-worktree-contract.ts` — pure helpers (resolveSessionWorktreeState, formatSessionWorktreeBadge, getAttachmentBranchLabel, getMutationBlockingReasons, etc.)
- `session-worktree-store.ts` — Zustand authoritative store
- `session-worktree-contract.test.js` — 46 tests
- `session-ui-store.test.js` — 8 tests
- `session-worktree-store.test.js` — 3 tests

### Remaining (separate PRs)
- ~~**Phase E**: Dirty-tree safety — `getMutationBlockingReasons` should also accept git status to block checkout with uncommitted changes~~
- **P2**: Server canonicalization fixes — `legacy = isLinkedWorktree(...)` semantics, `headState: 'not-a-repo'` type mismatch, `bisect` detection

## Test plan
- [x] `bun run type-check` — all packages pass
- [x] `bun run lint` — 0 errors
- [x] `bun test session-worktree-contract.test.js` — 46 pass
- [x] Manual: Header branch label stays fixed when another session switches branches
- [x] Manual: Degraded/legacy sessions fallback to live git branch
- [x] Manual: Mutation blocking works for rebase/merge/cherry-pick states
- [ ] Full E2E across web, desktop, VS Code runtimes


jwcrystal:
## Design Decisions — Seeking Review

A few core architectural choices that could use a second pair of eyes:

### 1. Attachment Priority Chain

Session branch display priority:

```
authoritative attachment branch → live git branch → legacy metadata → catalog lookup
```

When the attachment is present and not degraded/legacy, we **only use the attachment branch**, ignoring live git entirely. This means if a user runs `git checkout` in the terminal, the session's branch label won't drift.

**Question**: Is this the intended behavior? Or should we show a warning when live git and attachment disagree?

### 2. Canonicalization Timing

- `createSession` → immediately canonicalize and write attachment
- `setCurrentSession` → if no attachment or legacy, async recovery once

**Question**: Should canonicalization also be triggered at other points, e.g. on SSE session updates?

### 3. Mutation Blocking Scope

Currently only blocks branch mutations for `missing / invalid / attention` states. Dirty-tree blocking (uncommitted changes) is planned for a separate PR (Phase E).

**Question**: Is this phased approach acceptable, or should dirty-tree safety be in this PR?

### 4. `legacy` Semantics in `SessionWorktreeAttachment`

`legacy: true` means "this session was created before worktree isolation, so it has no canonical metadata." Phase A recovery attempts to fill in canonicalization on `setCurrentSession`, flipping `legacy` to `false` on success.

**Question**: Is this definition and recovery strategy sound?

jwcrystal:
## Updated: Phase E (dirty-tree safety) now included

`getMutationBlockingReasons` now accepts an optional `gitStatus` param (`{ isClean, files? }`) and blocks branch mutations when the tree has uncommitted changes.

This closes the unused `{ reason: 'dirty' }` type variant — all `MutationBlockingReason` variants are now fully exercised.

### What changed
- **`session-worktree-contract.ts`**: `getMutationBlockingReasons(attachment, gitStatus?)` — adds dirty-tree check before attachment checks
- **`GitView.tsx`**: All 3 call sites now pass `status` as second argument
- **5 new tests**: dirty, clean, null attachment, combined dirty+missing, no file count

Commit: `f2d6a649`

**Resolved**: Dirty-tree blocking is now in this PR. All `MutationBlockingReason` variants are used.

btriapitsyn:
Thanks for the work, but I can’t approve the core model.

Right now this PR treats `session attachment branch` as authoritative.  
That is wrong for shared-directory sessions and can show a stale branch label.

Concrete source of truth already exists in code:
- `packages/ui/src/stores/useGitStore.ts` → `directories.get(directory)?.status?.current`
- populated by `fetchStatus()` via `git.getGitStatus(...)` (`useGitStore.ts`)
- consumed by `useGitBranchLabel(...)` (`useGitStore.ts`)
- refreshed on directory activation and git refresh hints in `packages/ui/src/components/views/GitView.tsx` (`ensureAll(...)`, `fetchStatus(...)`)

So the better model is:
- keep session attachment for binding/status (`worktreeRoot`, `cwd`, degraded/missing/repair)
- take branch label from live directory git status (`useGitStore`), not frozen attachment branch
- if attachment branch != live branch, show explicit drift warning instead of masking it

jwcrystal:
Agreed with you. I forgot that the code allows shared directories (legacy sessions, sessions without a worktree), which can lead to errors. After re-analyzing, the current PR has two problems:

### 1. Branch label priority should stay live-git-first

Main already does the right thing:
```ts
// main branch (correct)
const currentBranchLabel = gitBranchForDirectory || currentSessionWorktreeBranch || catalogWorktreeBranch;
```

Phase C reversed this to attachment-first, which is a regression. The spec also says *"branch is informational state, not the authority"* and *"refresh/display the worktree's actual branch state."*

### 2. Not all sessions use worktrees

This is a gap I missed. The current code assumes every session has a worktree attachment, but in reality:
- Legacy sessions share the main working directory
- New sessions created without worktree isolation also share a directory
- Only sessions explicitly created with a worktree have an attachment

For shared-directory sessions, live git is the only correct source — there's no worktree to "freeze" the branch to.

### Proposed fix

**Revert Phase C's branch priority change:**
- Remove `getAttachmentBranchLabel()` — go back to the original live-git-first chain
- Keep everything else: `authoritativeProjectRoot` in GitView, Phase A+B (directory resolution), Phase D (runtime parity), Phase E (mutation blocking)
- Attachment stays authoritative for `worktreeRoot`, `cwd`, degraded/missing/repair, and mutation blocking
- Attachment `branch` becomes fallback-only (used when live git returns null, e.g. missing/degraded worktree)

**Drift warning as follow-up:**
- When `attachment.branch !== liveGitBranch` and session has a non-degraded worktree attachment → show subtle indicator
- This is a UX enhancement, not a correctness fix — separate PR

Does this align with what you had in mind?

jwcrystal:
Thanks for the review. You're right — attachment branch should not be the branch label source.

**Updated in `b9d5eadf`:** Branch label now uses live git as the primary source (same as main), matching your suggestion:

```ts
const currentBranchLabel = gitBranchForDirectory || currentSessionWorktreeBranch || catalogWorktreeBranch;
```

**What changed:**
- Removed `getAttachmentBranchLabel()` and its 7 tests
- Header reverted to the original live-git-first priority chain

**What's preserved (unchanged):**
- Attachment is still authoritative for: `worktreeRoot`, `cwd`, degraded/missing/repair status
- `authoritativeProjectRoot` in GitView (re-integrate uses correct repo root)
- Mutation blocking (missing/invalid/attention/dirty)
- Session create canonicalization + legacy recovery
- Runtime parity across web/desktop/VS Code

The worktree isolation goal (session doesn't drift) is achieved through directory isolation — each session gets its own worktree with its own checked-out branch. Branch label display doesn't need to be frozen because the underlying directory is already isolated.

btriapitsyn (approved):

jwcrystal:
Thank you for the straightforward code. 
My fault lies in the redundancy.😂

PR #916, opened 2026-04-15
fix(sync): deduplicate overlapping delta after coalesced part.updated

jwcrystal:
## Summary

When `message.part.updated` coalesces in the event pipeline and a `message.part.delta` for the same part arrives in the same 16ms flush window, the reducer appends the delta verbatim to the already-complete field value, producing duplicated text in tool output and assistant messages.

This is an intermittent regression from #889, which removed the stale-delta skip to fix dropped/incomplete streaming content. That fix was correct, but left a gap: when a coalesced `part.updated` already contains the complete text, an overlapping delta from the same flush window gets appended on top, duplicating the trailing portion.

## Root Cause

Event sequence that triggers the bug:

```
T0: message.part.updated(part-A, text="Hello ")     ← enters queue
T1: message.part.delta(part-A, delta="world")        ← enters queue
T2: message.part.updated(part-A, text="Hello world") ← coalesces with T0
```

After flush, the reducer receives:
1. `part.updated` → sets `text` to `"Hello world"`
2. `part.delta` → appends `"world"` → `text` becomes `"Hello worldworld"` ← **duplicate**

The existing test `event-pipeline.test.js:212-291` already proves deltas are delivered after coalesced updates (intended behavior from #889). The gap is in the reducer's delta handler, which always appends without checking for overlap.

## Fix

Add **targeted overlap reconciliation** scoped to the suspicious `updated → delta` window:

- When `message.part.updated` replaces an existing part with overlapping string content (`text` or `output`), mark the next delta for those fields as dedupe-eligible
- On the next delta, use suffix-overlap detection instead of blind append
- The dedupe marker clears automatically after one delta — normal streaming is untouched

This preserves the #889 guarantee (deltas are never dropped by the pipeline) while preventing the append-side duplication.

## Test Coverage

| Test | Protects |
|------|----------|
| Full overlap: delta already present in updated value → no duplicate | Bug fix |
| Partial overlap: only non-overlapping suffix appended | Variant |
| No overlap: unchanged append behavior | No regression |
| Legitimate repeated output (`ha` + `ha` → `haha`) | No false dedupe |

## Impact Assessment

- **Frequency**: Intermittent. Requires `updated → delta → updated` timing within the same 16ms flush window. More likely during long tool output, subagent tasks, or high-frequency streaming (~60 deltas/sec).
- **User symptom**: Trailing text duplication in tool output or assistant messages. Can make JSON output unparseable.
- **Without fix**: No good live workaround for users; refresh/reconnect eventually self-heals but user may have already copied wrong content.
- **With fix**: Minimal scope — only activates in the suspicious `updated → delta` window. Normal streaming behavior unchanged.

## Files Changed

- `packages/ui/src/sync/event-reducer.ts` — overlap-aware delta merge logic
- `packages/ui/src/sync/__tests__/event-reducer.test.js` — 4 regression tests (new file)

## Test Plan

- [x] `bun test packages/ui/src/sync/__tests__/` — 10 pass, 0 fail
- [x] `bun run type-check:ui` — clean
- [x] Manual testing: bash tool output, file read, subagent task, general streaming — no duplicates, no missing content

### DIFF
--- a/packages/ui/src/components/session/sidebar/SessionGroupSection.tsx
+++ b/packages/ui/src/components/session/sidebar/SessionGroupSection.tsx
@@ -184,57 +184,30 @@ export function SessionGroupSection(props: Props): React.ReactNode {
     return { folder, nodes };
   });
 
-  const allFoldersForGroup = React.useMemo(() => {
-    const folderMapById = new Map(allFoldersForGroupBase.map((entry) => [entry.folder.id, entry]));
-    const childFolderIdsByParentId = new Map<string, string[]>();
-    for (const { folder } of allFoldersForGroupBase) {
-      if (!folder.parentId) continue;
-      const existing = childFolderIdsByParentId.get(folder.parentId);
-      if (existing) {
-        existing.push(folder.id);
-      } else {
-        childFolderIdsByParentId.set(folder.parentId, [folder.id]);
-      }
+  const folderMapById = new Map(allFoldersForGroupBase.map((entry) => [entry.folder.id, entry]));
+  const shouldKeepFolder = (folderId: string): boolean => {
+    const entry = folderMapById.get(folderId);
+    if (!entry) return false;
+
+    // For archived buckets, hide folders with no sessions
+    // This prevents showing empty "project root" folders
+    if (group.isArchivedBucket && entry.nodes.length === 0) {
+      // Check if any sub-folders have content
+      const hasContentInChildren = allFoldersForGroupBase
+        .filter(({ folder }) => folder.parentId === folderId)
+        .some((childEntry) => shouldKeepFolder(childEntry.folder.id));
+      return hasContentInChildren;
     }
 
-    const keepByFolderId = new Map<string, boolean>();
-    const shouldKeepFolder = (folderId: string): boolean => {
-      const cached = keepByFolderId.get(folderId);
-      if (cached !== undefined) return cached;
+    if (!hasSessionSearchQuery) return true;
+    const folderMatches = entry.folder.name.toLowerCase().includes(normalizedSessionSearchQuery);
+    if (folderMatches || entry.nodes.length > 0) return true;
+    return allFoldersForGroupBase
+      .filter(({ folder }) => folder.parentId === folderId)
+      .some(({ folder }) => shouldKeepFolder(folder.id));
+  };
 
-      const entry = folderMapById.get(folderId);
-      if (!entry) {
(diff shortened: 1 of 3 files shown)


jwcrystal:
### Additional context on user impact

This bug is **intermittent but has no good live workaround for users**:

- It triggers when `updated → delta → updated` lands within the same 16ms flush window — more likely during tool-heavy sessions with high-frequency streaming (~60 deltas/sec)
- Users see trailing text duplication in tool output or assistant messages, which looks like "the tool ran twice" or "the model repeated itself"
- If the output is structured (JSON), the duplicated suffix can break parsing
- **Refresh/reconnect eventually self-heals** (canonical parts overwrite local state), but the user may have already copied wrong content or made decisions based on the duplicated output before that happens
- There is no way for users to prevent it during live streaming

The fix scope is intentionally narrow — it only activates in the suspicious `updated → delta` window and clears after one delta, leaving normal streaming behavior completely untouched.

jwcrystal:
### Request for manual testing

@btriapitsyn 
I'd appreciate if you could help verify this fix in a running dev environment. The bug is intermittent (timing-dependent), so it won't reproduce every time, but here's how to trigger it:

**Setup:**
```bash
bun run dev:web:full
```

**Enable sync debug in browser console:**
```javascript
localStorage.setItem("openchamber:sync:debug", "1")
```

**Test cases:**

| # | Action | What to check |
|---|--------|---------------|
| 1 | Ask the model to run a bash command with long output (e.g. `find . -type f -name "*.ts" \| head -80`) | Tool output should appear once, no duplicated trailing text |
| 2 | Ask the model to read a large file (e.g. `packages/ui/src/sync/event-reducer.ts`) | File content should be complete and not repeated |
| 3 | Ask the model to spawn a subagent task with multiple tool calls | Task summary should be clean, no repeated segments |
| 4 | Normal streaming chat (no tools) | Text should stream smoothly with no missing content |

**Expected results:**
- ✅ Cases 1-3: Clean output, no duplicated text
- ✅ Case 4: Complete streaming, no missing/cut-off content
- ✅ Browser console: no unexpected sync debug warnings

**What this PR changes:** Only the delta merge path in `event-reducer.ts`. When a `message.part.updated` replaces an existing part with overlapping text, the next delta for that field uses suffix-overlap detection instead of blind append. Normal streaming is completely untouched.

I'm happy to adjust the approach based on review feedback. Leaving the merge decision to the reviewer.

btriapitsyn (approved):

PR #923, opened 2026-04-16
feat(vscode): support multi-root workspaces

geekifan:
## Summary

- add multi-root workspace support to the VS Code extension runtime
- align the VS Code session sidebar with web-style project/worktree grouping
- allow selecting the target workspace for new chats in VS Code
- keep single-root VS Code behavior unchanged
- polish VS Code compact-layout draft navigation and hide unsupported project actions

## What changed

- treat VS Code `workspaceFolders` as projects instead of assuming a single `workspaceFolder`
- pass `workspaceFolders` and `activeWorkspaceFolder` into the webview config
- resolve files, config, git, and session defaults against the active workspace root instead of hardcoding the first root
- render multiple workspace sections in the session sidebar, with sessions grouped under the correct workspace root
- show the workspace selector in the VS Code chat composer when multiple roots are open
- keep single-root mode visually and behaviorally aligned with current `main`
- hide unsupported `Rename` and `Close Project` actions in VS Code project menus
- fix compact VS Code layout so opening a new draft from the sidebar navigates to chat
- preserve back navigation from chat to sessions while still allowing `New session` to reopen the draft view

## Why

VS Code multi-root workspaces were previously treated like a single-root environment. That caused sessions, directory scoping, and chat targeting to drift toward the first workspace folder and made the VS Code UX fall behind the web runtime.

This change brings VS Code much closer to the existing web mental model while preserving the current single-root experience.

## Validation

Ran VS Code-only checks and packaging during development:

- `bun run vscode:type-check`
- `bun run vscode:build`
- `bun run vscode:package`

## Notes

- this keeps the current single OpenCode manager model; it does not introduce one backend instance per workspace root
- single-root workspaces intentionally degrade to the existing behavior

### DIFF
--- a/packages/ui/src/components/chat/ChatInput.tsx
+++ b/packages/ui/src/components/chat/ChatInput.tsx
@@ -2824,6 +2824,7 @@ const ChatInputComponent: React.FC<ChatInputProps> = ({ onOpenSettings, scrollTo
 
     const footerGapClass = 'gap-x-1.5 gap-y-0';
     const isVSCode = isVSCodeRuntime();
+    const showWorkspaceSelector = isVSCode && projects.length > 1;
     const showDraftTargetSelectors = newSessionDraftOpen && !isVSCode;
 
     const selectedDraftProject = React.useMemo(() => {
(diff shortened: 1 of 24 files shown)


geekifan:
Update: there are some bugs in this PR. I will investigate and continue working until this draft PR is ready for review.

PR #950, opened 2026-04-18
feat: show file change summary bar 

jwcrystal:
## Summary

Adds a global file-change preview bar above the chat input, letting users quickly see which files were modified and open inline diffs.

**Dual-mode design:**
- **Git mode** — reads workspace git diff status (`useGitStore`), auto-clears on commit/restore. Label: _"N files changed in workspace"_
- **Non-Git mode** — reads tool-part metadata from the latest completed assistant turn, auto-clears on new message send + manual dismiss. Label: _"AI updated N files in the last reply"_

**Also includes:** `FileChangeSummary` — a multi-file diff aggregation card rendered inside `ToolPart` for inline diff preview with expand/collapse and git revert per file.

## Bar Lifecycle (when does it appear / disappear)

**No Accept/Reject buttons** — those actions belong in the Git Status sidebar. The bar is informational only: "here's what changed."

### Git mode
| Event | Behavior |
|-------|----------|
| Files modified in workspace | Bar appears |
| User commits or restores files (git status → clean) | Bar **auto-disappears** |
| User clicks dismiss (✕) | Bar hides until file set changes |

### Non-Git mode
| Event | Behavior |
|-------|----------|
| AI turn completes with file edits | Bar appears (hidden during streaming to prevent flicker) |
| User sends next message | Bar **auto-clears** |
| User clicks dismiss (✕) | Bar hides for current signature |
| AI edits same files again (different content) | Bar **re-appears** (signature includes additions/deletions) |

**Dismiss = "acknowledged"** — there is no confirm/accept action. Dismissing means the user has seen the changes. The bar will re-appear if the underlying state changes (new edits, new AI turn).

HapiGo_2026-04-19_01 38 19

## Changes

| File | Purpose |
|------|---------|
| `PendingChangesBar.tsx` | Main bar component (collapsed/expanded, dual-mode, dismiss) |
| `FileChangeSummary.tsx` | Multi-file diff card for ToolPart |
| `fileChangeHelpers.ts` | Shared `parsePatchStats` / `parseCount` utilities |
| `ChatContainer.tsx` | Renders bar above chat input (main + draft branches) |
| `session-ui-store.ts` | Per-session dismiss state (`Map<string, string>`) |

## Key design decisions

- **Streaming gate**: Non-git mode hidden during active streaming via `useStreamingStore` — prevents flicker
- **Per-session dismiss**: Uses `Map<sessionId, signature>` to prevent cross-session state leaks
[...]

Issue #980, opened 2026-04-22
[Bug] File change bar cannot be dismissed, redundant with right sidebar git status


jwcrystal:
## Bug

Commit `d1553ba` ("fix(chat): unify changes row with tasks") removed the dismiss mechanism from `PendingChangesBar`, causing two regressions:

### What was removed
- `computeSignature()` function
- `dismissedSignature` read from `useSessionUIStore`
- `handleDismiss` click handler and ✕ button
- `isDismissed` visibility guard

### What remains in store (dead code)
`session-ui-store.ts` still has `pendingChangesBarDismissed: Map<string, string>` and `dismissPendingChangesBar()` action. The `sendMessage` flow still writes to this map, but nothing reads it.

### Impact

1. **Bar cannot be dismissed** — In non-git mode, once AI edits files, the bar stays visible until the user sends a new message. No way to acknowledge and close it mid-conversation.

2. **Blocks todo progress** — The bar renders via `StatusRow`'s `leftAccessory` prop, which is mutually exclusive with todo content (`hasContent = hasAssistantContent || hasTodoContent || hasLeftAccessory`). When the bar is visible, active todo progress is hidden behind the file change display.

3. **Dead state in store** — `pendingChangesBarDismissed` map is written on every message send but never read, wasting state updates.

### Fix options
- **A**: Restore dismiss button + signature logic + `isDismissed` guard (original design)
- **B**: If permanent visibility is intentional, clean up dead code in `session-ui-store.ts` (remove `pendingChangesBarDismissed` and `dismissPendingChangesBar`)

### Repro
1. Open a non-git project
2. Let AI edit files → bar appears
3. No ✕ button to dismiss → bar stays until next message sent
4. Meanwhile, if there are active todos, they are hidden

jwcrystal:
## Context from original PR design intent (#950)

The `PendingChangesBar` was designed as an **informational notification** — "here's what changed" — so users can see what the AI modified, acknowledge it, and dismiss it. It is **not** meant to be a live mirror of git working tree changes.

### Original design (PR #950 description)

> **No Accept/Reject buttons** — those actions belong in the Git Status sidebar. The bar is informational only: "here's what changed."
>
> **Dismiss = "acknowledged"** — there is no confirm/accept action. Dismissing means the user has seen the changes. The bar will re-appear if the underlying state changes (new edits, new AI turn).

Key lifecycle from the original spec:

| Event | Behavior |
|-------|----------|
[...]

PR #981, opened 2026-04-22
fix(chat): restore desktop editor file-open in PendingChangesBar

### LINKED ISSUE #979: [Bug] Changed files cannot be opened in desktop editor
## Bug

Commit `d1553ba` ("fix(chat): unify changes row with tasks") rewrote `handleOpenFile` in `PendingChangesBar.tsx`, removing the `runtime?.editor` branch that handles VS Code / Electron desktop environments.

### Before (worked)
```ts
const editor = runtime?.editor;
if (editor && !isGitFile(file) && file.patch) {
    void editor.openDiff('', absolutePath, undefined, { patch: file.patch });
} else if (editor) {
    void editor.openFile(absolutePath);
} else {
    useUIStore.getState().openContextDiff(currentDirectory, absolutePath);
}
```

### After (broken)
```ts
const store = useUIStore.getState();
if (!store.isMobile) {
    store.openContextDiff(currentDirectory, targetPath);
    return;
}
store.navigateToDiff(targetPath);
store.setRightSidebarOpen(false);
```

### Impact
- **VS Code / Electron**: clicking a changed file does nothing — `editor.openDiff` and `editor.openFile` are never called
- **Non-git patch mode**: `file.patch` data is no longer passed to the diff viewer
- **Path handling**: `openContextDiff` may receive a relative path when it expects absolute

### Fix
Restore the 3-branch logic: `runtime.editor.openDiff(patch)` → `runtime.editor.openFile()` → `openContextDiff()` fallback.

### Repro
1. Open a project with file edits in desktop (VS Code / Electron)
2. Click on a file in the changes bar
3. Nothing opens

jwcrystal:
## Summary

Fixes #979

Commit `d1553ba` (within PR #950) removed the `runtime?.editor` branch from `handleOpenFile` during UI refactoring, breaking file-click in VS Code and Electron desktop runtimes.

### Two fixes in this PR

**Commit 1 — Restore `runtime.editor` branch (VS Code fix)**

Restores the 3-branch logic matching `ToolPart.tsx`:

| Priority | Condition | Action | Path type |
|----------|-----------|--------|-----------|
| 1 | `editor` + non-git + has `patch` | `editor.openDiff('', absolutePath, undefined, { patch })` | Absolute |
| 2 | `editor` exists | `editor.openFile(absolutePath)` | Absolute |
| 3 | No editor (web/desktop) | `openContextDiff` or `openContextFile` fallback | Relative |

**Commit 2 — Route non-git files to file view (Desktop fix)**

`DiffView` requires a git repository to display diffs. In desktop (Electron/Tauri), `runtime.editor` is unavailable, so non-git files fell through to `openContextDiff` which shows "Not a git repository".

Route non-git files to `openContextFile` instead, which works without git. Git-tracked files continue to use `openContextDiff`.

### Behavior matrix

| Runtime | Git file click | Non-git file click |
|---------|---------------|-------------------|
| **VS Code** | `editor.openFile(absolutePath)` | `editor.openDiff(patch)` or `editor.openFile()` |
| **Desktop (Electron/Tauri)** | `openContextDiff` (diff panel) | `openContextFile` (file content) |
| **Web** | `openContextDiff` (diff panel) | `openContextFile` (file content) |
| **Mobile** | `navigateToDiff` | `navigateToDiff` |

## Test plan

- [x] `bun run lint` passes
- [ ] VS Code: click changed file → opens in editor via `runtime.editor.openFile`
- [ ] Electron desktop: click git-tracked file → opens diff panel
- [ ] Electron desktop: click non-git file → opens file content view (not "Not a git repository" error)
- [ ] Web: click git-tracked file → opens context diff panel
- [ ] Web: click non-git file → opens file content view
- [ ] Mobile: click changed file → navigates to diff view

### DIFF
--- a/packages/ui/src/components/chat/ChangedFilesList.tsx
+++ b/packages/ui/src/components/chat/ChangedFilesList.tsx
@@ -1,0 +1,63 @@
+import React from 'react';
+import { FileTypeIcon } from '@/components/icons/FileTypeIcon';
+import { type ChangedFileEntry, getDisplayPath, getFileStats } from './changedFiles';
+
+interface ChangedFilesListProps {
+    files: ChangedFileEntry[];
+    currentDirectory: string;
+    onOpenFile: (file: ChangedFileEntry) => void;
+}
+
+export const ChangedFilesList: React.FC<ChangedFilesListProps> = ({ files, currentDirectory, onOpenFile }) => {
+    return (
+        <>
+            <div className="flex items-center gap-1.5 px-2 py-1 typography-ui-label font-medium text-muted-foreground">
+                <span>Changed files</span>
+                <span className="typography-meta tabular-nums">{files.length}</span>
+            </div>
+
+            <div className="max-h-[260px] overflow-y-auto">
+                {files.map((file, index) => {
+                    const { fileName, dirPart } = getDisplayPath(file, currentDirectory);
+                    const stats = getFileStats(file);
+
+                    return (
+                        <button
+                            key={`${file.path}:${index}`}
+                            type="button"
+                            className="relative flex w-full cursor-pointer items-center gap-2 rounded-lg px-2 py-1 typography-ui-label outline-hidden select-none text-left hover:bg-interactive-hover"
+                            title={`Open ${file.path}`}
+                            onClick={() => onOpenFile(file)}
+                        >
+                            <FileTypeIcon filePath={file.path} className="h-3.5 w-3.5 flex-shrink-0" />
+                            <span className="min-w-0 flex-1 flex items-baseline overflow-hidden" title={file.path}>
+                                {dirPart ? (
+                                    <>
+                                        <span
+                                            className="min-w-0 truncate text-muted-foreground"
+                                            style={{ direction: 'rtl', textAlign: 'left' }}
+                                        >
+                                            {dirPart}
+                                        </span>
+                                        <span className="flex-shrink-0">
+                                            <span className="text-muted-foreground">/</span>
+                                            <span className="text-foreground">{fileName}</span>
+                                        </span>
+                                    </>
+                                ) : (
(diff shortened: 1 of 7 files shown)


btriapitsyn:
Pushed a few follow-ups on top of the two original commits:

- PendingChangesBar now seeds the git store itself (ensureStatus on mount + subscribes to onGitRefreshHint). Previously the bar relied on RightSidebarTabs/DiffView/GitView being mounted to populate the store. VS Code doesn't mount RightSidebarTabs, so the bar never rendered there. Now works across all runtimes.
- Bar is git-only. Non-git changes moved to a new per-turn dropdown (TurnChangedFilesDropdown) in the footer row of each completed assistant turn — history is preserved per turn instead of a single global "last reply" state that reset on every new turn.

btriapitsyn (approved):

jwcrystal:
> Pushed a few follow-ups on top of the two original commits:
> 
> * PendingChangesBar now seeds the git store itself (ensureStatus on mount + subscribes to onGitRefreshHint). Previously the bar relied on RightSidebarTabs/DiffView/GitView being mounted to populate the store. VS Code doesn't mount RightSidebarTabs, so the bar never rendered there. Now works across all runtimes.
> * Bar is git-only. Non-git changes moved to a new per-turn dropdown (TurnChangedFilesDropdown) in the footer row of each completed assistant turn — history is preserved per turn instead of a single global "last reply" state that reset on every new turn.

@btriapitsyn you could review Issue #980, similar to your response. Need your point to decide how to proceed in further.

btriapitsyn:
@jwcrystal answered there.
Overall, I see this as a complete feature which is not relying on OpenCode data and we can be sure that the feature will work with other harnesses in future. This PR closes all the leftovers regarding this feature. I'm happy with how it turned out eventually :) thank you for contribution!

PR #988, opened 2026-04-23
feat: add 'Open files in preview mode' setting

daveotero:
## Summary

Adds a new user preference that allows files to open in preview mode by default, rather than edit mode.

## Why

Most work with modern agentic harnesses involves reviewing plans, markdown files, and generated output rather than manually editing source code. This setting gives users the option to default to preview mode, reducing friction for review-heavy workflows where the edit view is rarely needed.

## Changes

- **Backend** (settings-helpers.js): Adds `defaultFileViewerPreview` to the settings sanitizer
- **Config Store** (useConfigStore.ts): Adds state field and setter for the new preference
- **Desktop Settings Type** (desktop.ts): Adds the new field to `DesktopSettings`
- **Settings UI** (DefaultsSettings.tsx): Adds checkbox in Session Defaults section labeled **"Open files in preview mode"**
- **Files View** (FilesView.tsx): Reads the setting and initializes file viewer mode accordingly

## Behavior

- When enabled, files open in preview/view mode by default
- When disabled (default), files open in edit mode (existing behavior)
- Markdown/HTML/JSON files still respect their per-file-type localStorage persistence when available
- Setting persists across sessions via `/api/config/settings`

## Testing

- TypeScript type-check passes
- ESLint passes
- Dev server tested locally with existing OpenCode backend

### DIFF
--- a/packages/ui/src/components/sections/openchamber/DefaultsSettings.tsx
+++ b/packages/ui/src/components/sections/openchamber/DefaultsSettings.tsx
@@ -31,6 +31,7 @@ export const DefaultsSettings: React.FC = () => {
   const setSettingsDefaultVariant = useConfigStore((state) => state.setSettingsDefaultVariant);
   const setSettingsDefaultAgent = useConfigStore((state) => state.setSettingsDefaultAgent);
   const setSettingsDefaultFileViewerPreview = useConfigStore((state) => state.setSettingsDefaultFileViewerPreview);
+  const settingsDefaultFileViewerPreview = useConfigStore((state) => state.settingsDefaultFileViewerPreview);
   const showDeletionDialog = useUIStore((state) => state.showDeletionDialog);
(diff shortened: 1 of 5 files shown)


daveotero:
@greptile-apps All three issues addressed in commit 4d3178a:

1. **HTML localStorage override (P1):** The file-change effect now reads HTML_VIEWER_MODE_KEY from localStorage first and falls back to the global setting only when nothing is stored. The standalone mount-only localStorage effect was removed to avoid duplicate logic.

2. **Redundant local state (P2):** defaultFileViewerPreview is now read directly from useConfigStore (same pattern as showDeletionDialog). The local useState, separate fetch, and population logic were removed.

3. **Duplicated toggle logic (P2):** Extracted into a single handleToggleFileViewerPreview callback shared by both onClick and onKeyDown.

btriapitsyn (approved):

PR #995, opened 2026-04-23
fix: make todo list update dynamically when task status changes

Yabuku-xD:
## Summary

Fixes the todo/task list UI so it updates dynamically when the AI marks tasks as completed, in-progress, or pending. Previously the list would get "stuck" showing stale status.

## Root Cause

`areRenderRelevantPartsEqual` in `renderCompare.ts` compared tool parts by `state`, `metadata`, `status`, `time`, and `tool` fields — but **not** the `output` field. When the AI streamed updated todo JSON (e.g., marking items complete), the `output` string changed but the comparator returned `true`, so `React.memo(ToolPart)` skipped re-render and the UI stayed frozen.

## Fix

Add `output` reference comparison for tool parts in `areRenderRelevantPartsEqual`:

```js
const leftOutput = (leftPart as { output?: unknown }).output;
const rightOutput = (rightPart as { output?: unknown }).output;
if (leftOutput !== rightOutput) {
  return false;
}
```

This ensures any streaming tool output update triggers a re-render.

## Testing

- [x] `bun run lint` passes
- [x] Todo list now updates when task status changes from pending → in-progress → completed


Yabuku-xD:
Closing to recreate with cleaned-up commit history.

PR #993, opened 2026-04-23
perf: lazy-load heavy dependencies (MarkdownRenderer + CodeMirror languages)

Yabuku-xD:
## Summary

Defers loading of heavy parsing and language dependencies until they are actually needed. Reduces initial bundle by ~400KB+ without changing any consumer code.

### MarkdownRenderer dynamic import
- **Before**: `marked`, `react-markdown`, `beautiful-mermaid`, `react-syntax-highlighter`, `prismjs`, `katex` were all bundled in the initial chunk (~1500 lines, ~200-400KB)
- **After**: Moved to `MarkdownRendererImpl.tsx`, loaded on-demand via `React.lazy` when markdown content is first rendered
- All 11 existing imports (`import { MarkdownRenderer } from '@/components/chat/MarkdownRenderer'`) work unchanged — zero consumer code modified
- Suspense fallback renders an empty div (no visual flash) while the chunk loads

### CodeMirror language lazy loading
- **Before**: `languageByExtension.ts` statically imported 16+ language packages (`@codemirror/lang-go`, `lang-rust`, `lang-sql`, `lang-xml`, `lang-yaml`, `codemirror-lang-elixir`, and 6 legacy modes)
- **After**: Only the 6 most common languages are statically imported:
  - JavaScript/TypeScript, JSON, CSS, HTML, Markdown, Python, Shell
- Less common languages (Go, Rust, SQL, XML, YAML, C/C++, Ruby, Elixir, Erlang, TOML, etc.) now return `null` from `languageByExtension()`, causing callers to fall back to `loadLanguageByExtension()` which dynamically loads from `@codemirror/language-data`
- The fallback path already existed in `FilesView.tsx` — it was just never triggered because all languages were statically available

## Impact

| Dependency | Before | After |
|------------|--------|-------|
| Markdown stack | In initial bundle | Lazy chunk |
| CodeMirror languages (16+) | All static | 6 static, rest dynamic |
| Estimated initial bundle | ~1.5MB parsed | ~1.1MB parsed |

## Testing

- [x] `bun run type-check` passes (all packages)
- [x] `bun run lint` passes (all packages)
- [x] Markdown rendering works (chat messages, files, tool output)
- [x] CodeMirror syntax highlighting works for common languages
- [x] Dynamic language loading works for less common languages (tested via FilesView fallback)

## Notes

- ChatMessage MessageBody extraction was evaluated but deferred — the component has 1154 lines of deeply intertwined state/hooks/refs that require a dedicated refactor PR
[...]

PR #998, opened 2026-04-23
feat: add global AGENTS.md support

Yabuku-xD:
## Summary
Add support for global (non-project) AGENTS.md files that provide context instructions across all projects.

### Files read
- `~/.config/opencode/AGENTS.md` (opencode config directory)
- `~/AGENTS.md` (user home directory)

These complement project-level AGENTS.md files read by the OpenCode backend.

### Changes
- `packages/web/server/lib/global-agents-md.js`: async reader with 5-second caching
- `packages/web/server/lib/opencode/bootstrap-runtime.js`: endpoint wiring after auth middleware
- `packages/ui/src/hooks/useGlobalAgentsMd.ts`: React hook for UI consumption

### Endpoint
`GET /api/global-agents-md` returns:
```json
{
  "active": [{ "path": "...", "content": "..." }],
  "count": 1,
  "paths": ["~/.config/opencode/AGENTS.md", "~/AGENTS.md"]
}
```

## Dependency
Builds on top of **#993** — should be merged after the lazy-load PR.

## Test Plan
- [ ] `bun run type-check`
- [ ] `bun run lint`
- [ ] `bun run build`
- [ ] Verify endpoint returns correct data when files exist
- [ ] Verify endpoint handles missing files gracefully


btriapitsyn:
I don't understand the intended consumer here. This PR adds a server endpoint and a UI hook for global AGENTS.md files, but I don't see where this data is actually used: not in prompt/context assembly and not in any UI surface. What behavior is this meant to enable?

Yabuku-xD:
> I don't understand the intended consumer here. This PR adds a server endpoint and a UI hook for global AGENTS.md files, but I don't see where this data is actually used: not in prompt/context assembly and not in any UI surface. What behavior is this meant to enable?

You're right — the consumer is missing. The intended use is a Settings section that shows the user which global AGENTS.md files are active (like ~/.config/opencode/AGENTS.md and ~/AGENTS.md). This gives users visibility into cross-project instructions that OpenCode already consumes.

I should have included the UI component in this PR. Would you like me to add a "Global Instructions" settings section that uses the hook, or close this and come back with the full implementation?

btriapitsyn:
> I should have included the UI component in this PR. Would you like me to add a "Global Instructions" settings section that uses the hook, or close this and come back with the full implementation?

I think this could be a useful feature, so I would kindly ask you to make it complete within this PR. Having a way of managing context on global level would be useful. Right now there is an easy way of editing any per-project AGENTS.md as any other files in project, but not the global one. This could be a good addition to the setting. Please also merge main into this PR with conflict resolution, since I added a localization to the project and will ask you to maintain it as well for Global Context settings section, so that we will have it available in different languages like other sections.

btriapitsyn:
But again, a lot of unrelated changes in this PR, please clean it up as well down to this feature only :)

Yabuku-xD:
I'll be creating a new clean PR

PR #994, opened 2026-04-23
feat: add global AGENTS.md support

Yabuku-xD:
## Summary

Adds support for global `AGENTS.md` files that apply across all projects, not just project-specific ones. Reads from two locations:
- `~/.config/opencode/AGENTS.md` (opencode config directory)
- `~/AGENTS.md` (user home directory)

These complement project-level `AGENTS.md` files handled by the OpenCode backend.

## Changes

- `packages/web/server/lib/global-agents-md.js`: Utility module with caching (5s TTL) that reads global AGENTS.md files
- `packages/web/server/lib/opencode/core-routes.js`: New `GET /api/global-agents-md` endpoint
- `packages/ui/src/hooks/useGlobalAgentsMd.ts`: React hook for UI consumption

## API Response

```json
{
  "active": [
    { "path": "/Users/yabuku/.config/opencode/AGENTS.md", "content": "..." },
    { "path": "/Users/yabuku/AGENTS.md", "content": "..." }
  ],
  "count": 2,
  "paths": ["/Users/yabuku/.config/opencode/AGENTS.md", "/Users/yabuku/AGENTS.md"]
}
```

Only files that exist and have non-empty content are included in `active`. The `paths` field shows all checked locations.

## Usage

Place your global instructions in either location:
```bash
echo "Always use TypeScript strict mode" > ~/.config/opencode/AGENTS.md
# or
echo "Follow conventional commits" > ~/AGENTS.md
```

The UI can use `useGlobalAgentsMd()` to show an indicator when global agents are active.

## Testing

- [x] `bun run type-check` passes
- [x] `bun run lint` passes
- [x] Endpoint returns correct data when files exist
- [x] Endpoint returns empty active array when files don't exist
- [x] Caching works (5s TTL)

PR #999, opened 2026-04-23
fix: make todo list update dynamically when task status changes

Yabuku-xD:
## Summary
Fix todo list UI getting stuck when AI marks tasks as completed / in-progress / pending.

### Root cause
`areRenderRelevantPartsEqual` did not compare the tool part `output` field. When the AI updated a todo list (e.g., marking items completed), the `ToolPart` memo comparator returned `true` and skipped re-render. The UI showed stale task status.

### Fix
Add `output` reference comparison for tool parts in `areRenderRelevantPartsEqual` so streaming tool output updates trigger re-renders correctly.

### Files changed
- `packages/ui/src/components/chat/message/renderCompare.ts` (+7 lines)

## Dependency
Builds on top of **#994** — should be merged after the global AGENTS.md PR.

## Test Plan
- [ ] `bun run type-check`
- [ ] `bun run lint`
- [ ] Start a session with todo list generation
- [ ] Verify task status updates dynamically as AI marks items done

### DIFF
--- a/packages/ui/src/components/chat/message/renderCompare.ts
+++ b/packages/ui/src/components/chat/message/renderCompare.ts
@@ -80,6 +80,13 @@ export const areRenderRelevantPartsEqual = (left: Part[], right: Part[]): boolea
       if (leftTool !== rightTool) {
         return false;
       }
+      // Compare tool output so streaming updates (e.g., todo list progress)
+      // trigger re-renders when the output payload changes.
+      const leftOutput = (leftPart as { output?: unknown }).output;
+      const rightOutput = (rightPart as { output?: unknown }).output;
+      if (leftOutput !== rightOutput) {
+        return false;
+      }
       continue;
     }
 


btriapitsyn:
Please rebase/cherry-pick this down to the actual todo fix only. Right now #999 still includes the full #997 perf stack plus the closed #994 global-AGENTS work, so this PR is not reviewable as an isolated bugfix.

Yabuku-xD:
> Please rebase/cherry-pick this down to the actual todo fix only. Right now #999 still includes the full #997 perf stack plus the closed #994 global-AGENTS work, so this PR is not reviewable as an isolated bugfix.

Done — rebased down to the actual todo fix only. PR now contains a single commit with the isolated change to renderCompare.ts.

btriapitsyn (approved):

leonezhu:
noop

PR #1032, opened 2026-04-26
Improve assistant message action placement for split responses

vhqtvn:
## Summary

This PR improves assistant message actions when a response is split across multiple text and tool segments.

- show text-scoped actions for intermediate assistant text blocks instead of only exposing them on the final assistant message
- place the standalone action row directly after the text block it belongs to, rather than after later tool output
- keep the final assistant message footer merged into a single row for answer actions and turn metadata
- add lightweight visual grouping for standalone per-text action rows so the association with the text above is clearer
- preserve message image export behavior by hiding action groups while keeping metadata visible

image

image

## Testing

- `bun run type-check`
- `bun run lint`

## Build note

- `bun run build` on the clean latest-`upstream/main` worktree currently fails in `packages/vscode` with an unresolved import for `@fontsource/ibm-plex-sans/latin-600.css` from `packages/ui/src/styles/fonts.ts`
- this appears unrelated to the `MessageBody.tsx` change in this PR


btriapitsyn:
Sorry, but in my opinion it just adds additional noise without any value. This intermediate text is just a noise ment to entertain user while the work is in progress. Only the final summary of what was changed are valuable for user to operate with. I appreciate your interest in project and all the contributions, but this is something that might be just not really needed this time.

vhqtvn:
For my specific usecase: My workflow always ask the agent to return a final "Close out" so i can copy to another master agent, but sometime it returned the close out, then run a command, then output "All finished". Copying only include the "All finish" final message.

If you want i'll add toggle-able to settings.

btriapitsyn:
@vhqtvn Yes, adding such option in setting with default to off would be an acceptable compromize.

vhqtvn:
I added the settings.

image

btriapitsyn (approved):

PR #1059, opened 2026-04-28
fix: allow Enter/Ctrl+Enter to submit custom answer in QuestionCard

daveotero:
## Summary

The "Other..." custom answer textarea in the QuestionCard component had no keyboard handler — pressing Enter would just insert a newline instead of submitting, unlike the normal chat input where Enter/Ctrl+Enter sends the message.

## Changes

- Added `handleKeyDown` callback on `QuestionCard.tsx` that intercepts Enter (no Shift) and Ctrl+Enter on the custom answer textarea
- Matches the submit button's behavior: submits when all questions are answered (`requiredSatisfied`), otherwise navigates to the next unanswered question
- Shift+Enter continues to insert a newline (default textarea behavior)

## Testing

Verified manually via the Vite dev server — selecting "Other...", typing a custom answer, and pressing Enter now submits the answer.

### DIFF
--- a/packages/ui/src/components/chat/QuestionCard.tsx
+++ b/packages/ui/src/components/chat/QuestionCard.tsx
@@ -175,6 +175,20 @@ export const QuestionCard: React.FC<QuestionCardProps> = ({ question }) => {
     }
   }, [buildAnswersPayload, question.id, question.sessionID, requiredSatisfied, respondToQuestion]);
 
+  const handleKeyDown = React.useCallback(
+    (e: React.KeyboardEvent<HTMLTextAreaElement>) => {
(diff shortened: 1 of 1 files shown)


btriapitsyn:
Looks mostly fine, but please add the same IME guard we use in `ChatInput`:

```ts
import { isIMECompositionEvent } from '@/lib/ime';

if (isIMECompositionEvent(e)) return;

btriapitsyn (approved):

daveotero:
Pushed a follow-up to add the same `isIMECompositionEvent` guard used in `ChatInput`.

I also matched the mobile Enter behavior so plain Enter still behaves like a newline on mobile, with `Ctrl+Enter` still submitting.

Issue #1033, opened 2026-04-26
[Bug] GitHub MCP works for opencode TUI but not openchamber


Jazz23:
### What's wrong? Maybe some steps to reproduce.

E.g.
```json
    "github": {
      "type": "remote",
      "url": "https://api.githubcopilot.com/mcp/",
      "headers": {
        "Authorization": "Bearer {env:GITHUB_TOKEN}"
      },
      "enabled": true
    }
```
Works fine with opencode TUI but not with openchamber. Error code 400. Hard coding the credentials also does not fix the issue.

### Where does it happen?

Mobile (Web/PWA)

### Version (if known)

1.9.8

### Logs (optional)

```shell

```

PR #1072, opened 2026-04-29
fix(opencodeConfig): retain headers for remote MCP configurations

ricautomation:
Resolves issue #1033 where remote GitHub MCP setups failed to work in OpenChamber because the `headers` property (which contains authorization credentials) was being dropped during configuration parsing in `packages/vscode/src/opencodeConfig.ts`.


btriapitsyn (approved):

btriapitsyn:
@ricautomation is it still in progress? Let me know when you're done working on PR

PR #1082, opened 2026-04-30
fix: cross-verify update API claims against npm registry

corrm:
## Problem

The update API (`api.openchamber.dev/v1/update/check`) can return `updateAvailable: true` with a version ahead of what's actually published on npm. The server trusted the API unconditionally, creating a **perpetual update loop**: clicking "update" runs `npm install -g @openchamber/web@latest` (installs same version since npm doesn't have the phantom version), server restarts, API returns the same claim, update button reappears.

## Solution

Cross-verify API update claims against the npm registry before reporting an update as available. If the API says a newer version exists but npm doesn't have it yet (`npm's latest ≤ current version`), ignore the API's claim and report no update available.

## Changes

- `packages/web/server/lib/package-manager.js` — in `checkForUpdates()`, after the API returns `available: true`, fetch npm's latest version and confirm it's actually newer than current before trusting the API
- `packages/web/server/lib/package-manager.test.js` — 8 new tests covering all scenarios (both agree, API claims phantom, npm fallback, both unreachable, etc.)

## Verification

- 120/120 tests pass across 19 test files (1 pre-existing failure for `sse-routes.test.js` using `bun:test` in vitest runner)
- Type-check passes cleanly
- No regressions


btriapitsyn (approved):
Reviewed and pushed a small Electron regression fix for the About dialog version path. Update API verification logic and tests look good now.

PR #1083, opened 2026-04-30
fix: preserve per-session scroll position on session switch

jwcrystal:
## Summary
- Scroll position is now saved per-session (pixel-level `scrollTop/scrollHeight/clientHeight`) and restored via ratio mapping when switching back to a session that was not at the bottom
- Adds `restoreSavedScrollPosition` to the timeline controller as a separate path from `resumeToBottomInstant`, preserving its existing "always go to bottom" semantics for reselection and turn navigation
- Suppresses intermediate scroll events during session transition with an explicit `restoreInProgressForRef` flag to prevent saved state from being overwritten by DOM content reshaping

## Test plan
- [x] Open 3+ sessions, scroll up in each to different positions, switch between them — each should restore independently
- [x] Session left at bottom should stay at bottom on return
- [x] Session actively streaming should continue following bottom (not restore to an old position)
- [x] Click already-active session in sidebar (reselection) should still scroll to bottom
- [x] Rapidly switch between sessions — no flicker or position jumping

---
**Chore:** Removed debug `console.log` statements from `SessionAuthGate.tsx` that were printing auth response details in all environments including production. Catch-block error logging preserved.

### DIFF
--- a/packages/ui/src/components/auth/SessionAuthGate.tsx
+++ b/packages/ui/src/components/auth/SessionAuthGate.tsx
@@ -25,7 +25,6 @@ const STATUS_CHECK_ENDPOINT = '/auth/session';
 const TRUST_DEVICE_STORAGE_KEY = 'openchamber.uiAuth.trustDevice';
 
 const fetchSessionStatus = async (): Promise<Response> => {
-  console.log('[Frontend Auth] Checking session status...');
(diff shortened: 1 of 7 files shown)


jwcrystal:
@greptile-apps check again, thank you

btriapitsyn (approved):

PR #1095, opened 2026-05-03
test(sync): add edge-case coverage for reconnect recovery with parts

jwcrystal:
## Summary

The implementation fix landed in `3f300700` (main). This PR is now **test-only** — it adds test coverage for edge cases not covered by the author's existing test.

### What changed
- Added `Part` import and `createPart()` helper for realistic test data
- Existing tests updated with `part` state (required since `3f300700`'s implementation now checks `state.part`)
- Added 3 new tests:
  - **Completed assistant with no parts entry** (`part: {}`) — same scenario as author's test but with explicit naming
  - **Completed assistant with empty parts array** (`part: { id: [] }`) — not covered by author's test
  - **Completed assistant with populated parts** (negative case) — verifies the check correctly excludes sessions with real parts

### Why
`3f300700` added one test for the "parts missing" case. These additional tests cover the empty-array edge case and the negative case (populated parts should NOT trigger reconnect), providing a more complete safety net around `getReconnectCandidateSessionIds`.

### Test plan
- [x] `bun test packages/ui/src/sync/reconnect-recovery.test.ts` — 6/6 pass
- [x] `bun run lint` — clean
- [x] `reconnect-recovery.ts` matches `origin/main` exactly (no implementation changes)


btriapitsyn:
Thanks for digging into this and writing up the reconnect scenario clearly.
I looked through the current `main` state, and this case is already covered by `3f300700`:
```ts
lastAssistantComplete && state.part && (state.part[lastMessage.id]?.length ?? 0) === 0
```
That handles the real runtime case where the UI sleeps/disconnects, then reconnects with the assistant message marked completed but with missing or empty parts in the sync store. In the actual app state, `state.part` is initialized as `{}`, so a missing `state.part[messageId]` is treated as empty and the session is selected for reconnect recovery.
So I think this PR is now functionally superseded by the existing fix.
The extra tests here are useful, especially the populated-parts negative case, but I do not think we need the PR’s code change itself. If we want to keep anything from this, I’d prefer a small follow-up test-only PR/commit rather than merging overlapping logic.

jwcrystal:
Yeah, the commit 3f30070 fixed this issue. I’ll keep it as a test-only feature.

jwcrystal:
Commit 3f300700 — confirmed the implementation fix covers the same case. I've scoped this PR down to test-only.

One minor thought on the message loop in `getReconnectCandidateSessionIds`: the current version computes the completion check twice (once as `lastAssistantComplete`, once inverted in the `if`), and the `else if` hangs outside the inner block. A slightly flatter alternative:

```ts
for (const [sessionId, messages] of Object.entries(state.message ?? {})) {
  const lastMessage = messages[messages.length - 1]
  if (!lastMessage || lastMessage.role !== "assistant") continue

  const incomplete = typeof (lastMessage as { time?: { completed?: number } }).time?.completed !== "number"
  const messageId = lastMessage.id
  const partsEmpty = !state.part?.[messageId] || state.part[messageId].length === 0

  if (incomplete || partsEmpty) {
    ids.add(sessionId)
  }
}
```

Same semantics — early continue skips non-assistant, then two named booleans make the intent explicit. Happy to send a original follow-up if you think it's worth cleaning up, or no worries if the current version is intentional.

tomzx:
@openchamber-bot review

btriapitsyn:

PR #1104, opened 2026-05-04
feat(ui): enable TimelineDialog with full-text search across all message roles in one session

jwcrystal:
## Summary

- Enable the existing but unused `TimelineDialog` component by wiring it into `ChatContainer`
- Add keyboard shortcut `Cmd+T` / `Ctrl+T` (`mod+t`) to open the timeline dialog
- Enhance search to cover **all messages** (user + assistant) with **full-text matching** instead of user-only 80-char previews
- Add role icons (user / assistant) to search results for easy identification
- Add search snippet display with surrounding context for matched text

## Why

Long conversations are hard to navigate — users had no way to search within a session and jump to a specific message. The TimelineDialog component existed but was dead code (never imported or rendered).

## Changes

| File | Change |
|------|--------|
| `shortcuts.ts` | Register `open_timeline_dialog` shortcut (`mod+t`) |
| `useKeyboardShortcuts.ts` | Handle `mod+t` keydown → `setTimelineDialogOpen(true)` |
| `TimelineDialog.tsx` | Full-text search all roles, role icons, search snippets |
| `ChatContainer.tsx` | Import + render `<TimelineDialog>` with scroll controller |
| `*.settings.ts` (6 locales) | Shortcut label translations |

## Test Plan

- [ ] Press `Cmd+T` / `Ctrl+T` — TimelineDialog opens
- [ ] Search a keyword — matches both user and assistant messages
- [ ] Click a result — chat scrolls to that message
- [ ] Escape closes the dialog
- [ ] Settings → Keyboard Shortcuts shows the new "Open conversation timeline" entry
- [ ] Revert and Fork actions still work on user messages
- [ ] `bun run type-check` passes
- [ ] `bun run lint` passes (0 errors)

### DIFF
--- a/packages/ui/src/components/chat/ChatContainer.tsx
+++ b/packages/ui/src/components/chat/ChatContainer.tsx
@@ -14,6 +14,7 @@ import ScrollToBottomButton from './components/ScrollToBottomButton';
 import { ScrollShadow } from '@/components/ui/ScrollShadow';
 import { useChatScrollManager, type AnimationHandlers, type ContentChangeReason } from '@/hooks/useChatScrollManager';
 import { useChatTimelineController } from './hooks/useChatTimelineController';
+import { TimelineDialog } from './TimelineDialog';
 import { useChatTurnNavigation } from './hooks/useChatTurnNavigation';
 import { useDeviceInfo } from '@/lib/device';
 import { Button } from '@/components/ui/button';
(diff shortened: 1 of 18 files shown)


btriapitsyn:
I pushed a follow-up fix for the timeline dialog.

Summary:
- Kept the timeline user-message only.
- Preserved full-text search, but scoped it to all text parts of user messages.
- Removed assistant/system/tool messages from the timeline because the available actions (`Revert` and `Fork`) are only valid for user messages. Running those actions on assistant messages would produce incorrect behavior and does not match the existing session action contracts.
- Added `/timeline` as a slash command in addition to the `mod+t` shortcut.
- Added keyboard navigation in the dialog: search is focused on open, arrow keys move through results, and Enter navigates to the selected message.
- Cleaned up the render loop and removed unused snippet metadata.

btriapitsyn (approved):

PR #1106, opened 2026-05-04
Feat/add ide opened file to chat context

davidsazg:
## What

Adds the ability to attach the currently open file (or an active text selection) from the VS Code editor directly to the chat context, without leaving the chat panel.

Two new UI elements appear above the chat input when using the VS Code extension:

- **File suggestion chip** — shown when a file is open in the editor and hasn't been attached yet. Click the `+` button to add the full file to the context.
- **Selection pin chip** — shown when text is selected in the editor. Click the pin button to attach that specific selection (with its line range, e.g. `index.ts:12-34`) to the context.

Both chips are dismissed automatically once the attachment is added, and can be removed individually via the `×` button.

> **Note:** This feature is VS Code extension only. It has no effect on the web or desktop runtimes, which continue to work as expected.

## Why

This mirrors the workflow from GitHub Copilot Chat in VS Code, where the active editor file and selections can be added to the chat context with a single click. Users working in VS Code shouldn't need to manually copy-paste code snippets or file paths into the chat — the relevant context should be reachable from the input area directly.

## Changes

- **`ChatViewProvider.ts`** — listens to `onDidChangeActiveTextEditor` and `onDidChangeTextEditorSelection`; broadcasts file path, name, relative path, size, and current selection to the webview via `postMessage`.
- **`webview/main.tsx`** — handles the `activeEditorFile` command and updates the input store.
- **`input-store.ts`** — adds `activeEditorFile` state and two new actions: `addVSCodeFileAttachment` and `addVSCodeSelectionAttachment`.
- **`FileAttachment.tsx`** — adds `AttachedVSCodeFileChips` (renders attached VS Code files/selections) and `ActiveEditorFileSuggestion` (renders the suggestion UI).
- **`ChatInput.tsx`** — renders both new components above the textarea.
- **`sessionTypes.ts`** — extends `AttachedFile` with `source: 'vscode'`, `vscodePath`, and `vscodeSource` fields.
- **i18n** — new keys added to all 6 supported locales.

## Testing

| Environment | Result |
|---|---|
| VS Code on Windows | ✅ Works as expected |
| VS Code on Ubuntu (Linux) | ✅ Works as expected |
| Web runtime | ✅ Feature does not apply, no regressions |

## Validation

- bun run type-check ✅
- bun run lint ✅
- bun run build ✅

## Example GIF
openchamber-feat-add-opened-file-to-chat


davidsazg:
I've just addressed the feedback from Greptile. I added a debounce to the selection events, fixed that race condition in the duplicate check, and handled the floating promises. I also made the source filter more explicit and documented the file:// URL contract. Ready for another look!

btriapitsyn:
I pushed a follow-up fix after testing the VS Code editor context chips locally.

Changes included:
- Fixed duplicate selection attachments by guarding pending VS Code selection reads.
- Normalized VS Code `file://` attachment URLs so Windows paths are handled correctly.
- Made selection attachment matching path-aware to avoid conflicts between files with the same name/range.
- Fixed the layout overlap between the new VS Code chips and the `@file` mention highlight layer.
- Reduced unnecessary active-editor updates by skipping unchanged payloads.
- Prevented the chip from disappearing briefly while switching editor tabs, avoiding the vertical layout shift/flicker.

btriapitsyn (approved):

btriapitsyn:
Thanks again for the contribution, this is a really useful workflow improvement for the VS Code extension ❤️

PR #1126, opened 2026-05-07
fix(ui): keep out-of-order assistant replies on the correct turn

### LINKED ISSUE #1113: [Bug] Rendering issue when using the web interface at MSTSC (or other remote solution)

When using Openchamber web at a remote PC, the response will be placing above the prompt or below the prompt randomly. As investigated by AI, it's causing by racing issue.

>>

The issue is any network latency or timing variation that causes events to arrive out of expected order:

The bug is in Openchamber's code, not your remote solution:

projectTurnRecord.ts doesn't handle out-of-order events gracefully
If assistant response arrives before user message is processed → duplicate rendering
Any network jitter can trigger it

### Where does it happen?

Desktop Web

### Version (if known)

1.10.0

Image

### Logs (optional)

```shell

```

zerone0x:
## Summary
- keep assistant messages with unresolved parent IDs deferred instead of attaching them to the current turn
- resolve deferred assistant chains once their parent user/assistant message arrives
- add regression coverage for out-of-order user/assistant delivery and deferred assistant chains

## Why
When network jitter delivers assistant events before their matching user message is projected, the previous fallback path attaches that reply to the current turn. That makes the response render above/below the wrong prompt, which matches #1113.

## Testing
- `npx -y tsx /tmp/openchamber-turns-check.ts`
- `git diff --check -- packages/ui/src/components/chat/lib/turns/projectTurnRecords.ts packages/ui/src/components/chat/lib/turns/projectTurnRecords.test.ts`

Fixes #1113

### DIFF
--- a/packages/ui/src/components/chat/lib/turns/projectTurnRecords.test.ts
+++ b/packages/ui/src/components/chat/lib/turns/projectTurnRecords.test.ts
@@ -1,0 +1,56 @@
+import { describe, expect, test } from 'bun:test'
+import type { Message, Part } from '@opencode-ai/sdk/v2'
+import { projectTurnRecords } from './projectTurnRecords'
+import type { ChatMessageEntry } from './types'
+
+function createMessageEntry({
+  id,
+  role,
+  parentID,
+  createdAt,
+}: {
+  id: string
+  role: 'user' | 'assistant'
+  parentID?: string
+  createdAt: number
+}): ChatMessageEntry {
+  return {
+    info: {
+      id,
+      role,
+      ...(parentID ? { parentID } : {}),
+      time: { created: createdAt },
+    } as Message,
+    parts: [] as Part[],
+  }
+}
+
+describe('projectTurnRecords', () => {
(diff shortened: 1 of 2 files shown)


btriapitsyn:
Thanks for taking a look at this and for identifying the right area of the code.

After reviewing the issue more deeply and comparing the behavior with OpenCode's current implementation, I don't think this PR is the right direction to merge as-is. The pending-assistant approach improves the final grouping, but it still builds on the old `currentTurn` fallback model, which is the root of the rendering race.

I pushed a separate fix in `43ef7b58` that addresses this more directly:

- turn projection is now user-rooted, matching OpenCode's timeline model;
- assistant messages are only rendered under the user message referenced by `parentID`;
- assistant messages whose parent user turn has not arrived yet are not rendered as standalone/ungrouped entries;
- direct prompt/command sends now always pass a stable OpenCode-compatible `messageID`, making retry/latency scenarios more idempotent;
- regression tests cover the out-of-order and transient missing-parent cases.

Because of that, I'm going to close this PR in favor of the new commit. Thanks again for the contribution.

PR #1138, opened 2026-05-08
fix(electron): preserve desktop hosts when saving SSH instances

isanchez404:
## Summary
- Preserve manually configured desktop hosts when SSH instances are saved.
- Only remove desktop host entries that correspond to deleted SSH instances.

## Testing
- `node --check packages/electron/ssh-manager.mjs`
- `node --check packages/electron/main.mjs`
- `node --check packages/electron/preload.mjs`
- Manual Node scenario for SSH host filtering
- `bun run type-check` — all packages pass
- `bun run lint` — all packages pass

### DIFF
--- a/packages/electron/ssh-manager.mjs
+++ b/packages/electron/ssh-manager.mjs
@@ -579,16 +579,20 @@ export class ElectronSshManager {
 
   async setInstances(config) {
     const root = readJsonRoot(this.settingsFilePath);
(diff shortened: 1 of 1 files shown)


isanchez404:
Addressed the validation gap: `bun install`, `bun run type-check`, and `bun run lint` have all been run and pass clean.

btriapitsyn (approved):

Issue #1135, opened 2026-05-07
[Bug] flickering scroll in mobile, modals showing at bottom.


procodingtools:
### What's wrong? Maybe some steps to reproduce.

When ai is generating, the scroll working well on mobile. But when finished generating the scroll flicker. Also, dialogs showing at bottom.

### Where does it happen?

Mobile (Web/PWA)

### Version (if known)

1.10.2

### Screenshots / recordings (optional)

<img width="1080" height="2400" alt="Image" src="https://github.com/user-attachments/assets/5e845dce-c34f-41ca-8a3e-e638152a109c" />

https://github.com/user-attachments/assets/9ba74f41-d994-46ba-bf14-01c7320cbed6

### Logs (optional)

```shell

```

fengT-T:
If the dialogue is long enough, scrolling and flickering will occur.

youfch:
PC Edge It also appeared

yulong-ge:
This issue does not exist in version 1.10.1; it is newly introduced in version 1.10.2

procodingtools:
> This issue does not exist in version 1.10.1; it is newly introduced in version 1.10.2

yes for flickering but the dialog issue exists.

damienmiheev:
Also same problem on mac client

tomzx:
Can you confirm the issue is still present as of 1.12.3?

youfch:
> 你能确认这个问题在1.12.3版本时依然存在吗？

does not exist

makeittech:
Closing as fixed based on reporter/contributor confirmation that the flicker was gone as of 1.12.3, plus subsequent scroll/auto-follow and mobile modal work (#1289 / #1371 and later chat scroll hardening). Please open a new issue if flicker or bottom-clipped modals return on a current mobile PWA build.

PR #1162, opened 2026-05-08
fix(chat): prevent materialization resume from snapping unpinned view

Long-louis:
## Summary
- add a session-scoped one-shot guard around `ensureSessionRenderable` retries to avoid repeated resume attempts during hydration
- skip `resumeToLatestInstant()` when the chat view is unpinned, so user-driven scroll position is preserved
- keep the earlier `sessionIsWorking` hardening (`computeSessionIsWorking`) that avoids stale historical incomplete messages being treated as live work without live status

## Validation
- bun run type-check
- bun run lint
- bun test packages/ui/src/components/chat/lib/computeSessionIsWorking.test.ts

## Issue
- Related to #1135


yulong-ge:
Closing per reporter request; continuing investigation and will reopen with a more complete fix.

PR #1175, opened 2026-05-08
Fix chat list sorting by user activity

### LINKED ISSUE #1174: [Feature Request] Chat Sorting Improvement

chats switch positions between eachother when there are multiple of them open and multiple agent sessions that are active. suggesting to follow industry standard of the chat session list being ordered by last message sent by the user instead of the last message sent by the agent

### Extra context (optional)

_No response_

zoubenr:
## Summary
- Sort chat/session sidebar entries by the latest user message instead of assistant/session update activity.
- Track user-message activity from optimistic sends, message events, loaded history, and cold-start hydration.
- Keep pinned sessions first and add focused tests for ordering and activity reconciliation.

Fixes #1174

## Verification
- bun test packages/ui/src/sync/session-user-activity-store.test.ts packages/ui/src/components/session/sidebar/utils.test.ts
- bun run type-check
- bun run lint

### DIFF
--- a/packages/ui/src/components/session/SessionSidebar.tsx
+++ b/packages/ui/src/components/session/SessionSidebar.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import type { Session } from '@opencode-ai/sdk/v2';
+import type { Message } from '@opencode-ai/sdk/v2/client';
 import { RiLayoutLeftLine } from '@remixicon/react';
 import { toast } from '@/components/ui';
 import { Tooltip, TooltipContent, TooltipTrigger } from '@/components/ui/tooltip';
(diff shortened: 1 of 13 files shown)


zoubenr:
Updated with a simpler activity-ordering fix:

- Sorts sessions by pinned status, then latest user-message timestamp, then creation fallback.
- Rolls child/subagent user activity up to the root/apex chat for sidebar ordering.
- Adds per-item hydration error handling so one failed message fetch does not stop the queue.
- Keeps Active Now entries while user activity is unresolved, including unresolved child activity sources.
- Renames the misleading created-time helper and adds focused coverage for these cases.

Verified:
- bun test packages/ui/src/sync/session-user-activity-store.test.ts packages/ui/src/components/session/sidebar/utils.test.ts packages/ui/src/components/session/sidebar/activitySections.test.ts
- bun run type-check
- bun run lint

PR #1195, opened 2026-05-10
fix(sessions): isolate folder save temp files

isanchez404:
## Summary
- Use a unique temporary file per session-folder save before atomic rename.
- Add route-level coverage that concurrent saves do not share the same temp path.

## Validation
- `vet "ensure concurrent session folder saves use unique temp files" --agentic --agent-harness claude`
- `bun run --cwd packages/web test server/lib/session-folders/routes.test.js`
- `bun run type-check`
- `bun run lint`
- `git diff --check`

### DIFF
--- a/packages/web/server/lib/session-folders/routes.js
+++ b/packages/web/server/lib/session-folders/routes.js
@@ -45,7 +45,7 @@ export const registerSessionFoldersRoutes = (app, dependencies) => {
     }
     try {
       await ensureDir();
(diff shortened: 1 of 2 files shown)


isanchez404:
Fixed the Greptile P2 outside-diff feedback in 1b616f1a: a failed rename now attempts to unlink the unique temp file before returning the write error.

btriapitsyn (approved):

PR #1208, opened 2026-05-10
fix(files): match Windows prefixes case-insensitively

isanchez404:
## Summary
- Compare Windows drive paths case-insensitively when removing open files by prefix.
- Add coverage for closing an uppercase stored path with a lowercase prefix.

## Validation
- `vet "make files-view tab prefix removal case-insensitive for Windows drive paths" --agentic --agent-harness claude`
- `bun test packages/ui/src/stores/useFilesViewTabsStore.windows.test.ts`
- `bun run type-check`
- `bun run lint`
- `git diff --check`

### DIFF
--- a/packages/ui/src/stores/useFilesViewTabsStore.ts
+++ b/packages/ui/src/stores/useFilesViewTabsStore.ts
@@ -238,13 +238,18 @@ export const useFilesViewTabsStore = create<FilesViewTabsStore>()(
               return state;
             }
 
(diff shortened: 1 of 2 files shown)


isanchez404:
Fixed the Greptile P2 feedback in d3528ab5: `removeOpenPath` now compares Windows drive paths case-insensitively, matching `removeOpenPathsByPrefix`, with a focused regression test for single-path removal.

isanchez404:
Fixed the Greptile P2 outside-diff feedback in d3528ab5: removeOpenPath now uses Windows comparable-path matching and has a regression test.

btriapitsyn (approved):

PR #1209, opened 2026-05-10
fix(status): classify multiedit as editing

isanchez404:
## Summary
- Treat running `multiedit` tool parts as editing activity, matching the existing status phrase for that tool.

## Validation
- `vet "classify multiedit assistant tool activity as editing" --agentic --agent-harness claude`
- `bun run type-check`
- `bun run lint`
- `git diff --check`


isanchez404:
Fixed the Greptile P1 feedback in 936b2914: `multiedit` remains classified as editing, but its status text now uses the existing plural `editing files` phrase instead of falling back to the generic singular editing copy.

isanchez404:
Fixed the Greptile P1 outside-diff feedback in 936b2914: multiedit still reports editing activity but now uses the existing editing files phrase.

btriapitsyn (approved):

PR #1229, opened 2026-05-11
fix(ui): persist selection-store state across reloads

xomikronx:
Noticed that agent and model selections weren't sticking after a page refresh—they’d always revert back to the defaults. It looks like when these were moved into the new selection-store, the persist middleware didn't make the jump.

This PR wraps useSelectionStore in persist() using the existing getSafeStorage helper. I also added a merge handler to make sure the Maps rehydrate correctly from local storage, otherwise they'd just end up as empty objects.

Tested locally and my agent/model choices now survive a hard refresh as expected.

### DIFF
--- a/packages/ui/src/sync/selection-store.ts
+++ b/packages/ui/src/sync/selection-store.ts
@@ -5,13 +5,23 @@
 
 import { create } from "zustand"
 import { persist, createJSONStorage } from "zustand/middleware"
-import { getSafeStorage } from "../utils/safeStorage"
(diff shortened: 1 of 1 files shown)


xomikronx:
Added version: 1 and a migrate stub for future schema bumps.

Added a 150-session LRU eviction cap during the partialize step to prevent localStorage exhaustion. I also updated the set methods to delete the key first, ensuring the Map insertion order correctly reflects recency so slice() drops the oldest sessions.

btriapitsyn (approved):

PR #1260, opened 2026-05-13
fix: merge skills from both OpenCode and local .agents/skills sources

maximillions-dev:
## Problem

When OpenCode is running, `fetchOpenCodeDiscoveredSkills()` can return an empty array `[]` (no skills set up in `.opencode/skills`). The route handler used `||` between the two sources:

```js
const skills = (await fetchOpenCodeDiscoveredSkills(directory)) || discoverSkills(directory);
```

Since `[]` is truthy, `discoverSkills()` was never evaluated — making `.agents/skills` skills invisible while OpenCode was running.

## Fix

Replaced both occurrences (main skills listing + browse sources endpoint) with a `mergeDiscoveredSkills` helper that:

1. Fetches **both** sources in parallel via `Promise.all`
2. Merges results by `name` with dedup
3. Local skills take precedence

## Test plan

- Existing tests pass (41 pass, 4 pre-existing failures unrelated to skills)
- Verified `discoverSkills()` returns `.agents/skills` skills correctly
- Merge is a no-op when OpenCode is not running (returns `null` → `openCodeSkills` is `null` → `|| []` kicks in → only local skills returned)

### DIFF
--- a/packages/web/server/lib/opencode/skill-routes.js
+++ b/packages/web/server/lib/opencode/skill-routes.js
@@ -164,6 +164,24 @@ export const registerSkillRoutes = (app, dependencies) => {
     }
   };
 
+  const mergeDiscoveredSkills = async (directory) => {
+    const [openCodeSkills, localSkills] = await Promise.all([
+      fetchOpenCodeDiscoveredSkills(directory),
+      discoverSkills(directory),
+    ]);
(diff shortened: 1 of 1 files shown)


tomzx:
@openchamber-bot review

PR #1263, opened 2026-05-14
fix(mobile): collapse session status bar by default and on dismiss

mrosnerr:
## Summary

On mobile, the session status bar stays expanded in situations where it should collapse:

1. **Starts expanded by default** — even when a session is already active, the full session list is visible on load
2. **Stays open after selecting a session** — tapping a session switches to it but leaves the list open, wasting vertical space
3. **No tap-outside-to-dismiss** — tapping the chat area does not collapse the bar

The collapsed view still shows the current session title, running/unread indicators, context usage, and a "New" button — so discoverability is preserved.

## Changes

- `useUIStore.ts`: Default `isMobileSessionStatusBarCollapsed` to `true`
- `MobileSessionStatusBar.tsx`: Call `setIsMobileSessionStatusBarCollapsed(true)` in `handleSessionClick`
- `ChatContainer.tsx`: Collapse the bar when tapping the chat viewport on mobile

## Validation

- Tested in Chrome DevTools mobile emulation (iPhone 14 Pro, 390×844, touch enabled)
- Verified bar starts collapsed on load
- Verified bar collapses after session selection
- Verified tapping chat area dismisses the expanded bar
- `bun run lint` — passes

### DIFF
--- a/packages/ui/src/components/chat/ChatContainer.tsx
+++ b/packages/ui/src/components/chat/ChatContainer.tsx
@@ -191,7 +191,11 @@ const ChatViewport = React.memo(({
         }
 
         scrollRef.current?.focus({ preventScroll: true });
-    }, [scrollRef]);
+
+        if (isMobile && !useUIStore.getState().isMobileSessionStatusBarCollapsed) {
+            useUIStore.getState().setIsMobileSessionStatusBarCollapsed(true);
+        }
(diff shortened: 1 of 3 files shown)


mrosnerr:
Closing — this has gone stale with merge conflicts and no review activity. Happy to revisit if there's interest; the fix is small enough to re-create. Thanks!

PR #1279, opened 2026-05-16
feat(ui): revert indicator with undo/redo, message list, and attachment restore

youfch:
## Overview

Add a bidirectional undo/redo system with a revert indicator in the chat StatusRow. Users can see how many messages are reverted, restore individual messages or all at once, fork from any reverted message, and navigate through the undo/redo stack.

---

## Screenshots

PixPin_2026-05-16_12-38-37

PixPin_2026-05-16_12-37-55

PixPin_2026-05-16_12-38-15

---

### New components

**RevertIndicator.tsx** (+248)

- Compact trigger button in StatusRow with reverted message count badge
- Portal popover with per-session ID to prevent multi-session collisions
- Per-button loading spinners during restore/revert operations
- Panel-locked state prevents concurrent operations
- Click-outside detection to close popover
- Dual-source revert state: UI store (responsive) then sync store fallback (survives refresh)
- revertExplicitlyClearedRef suppresses stale sync fallback after sending a new message

**RevertPopover.tsx** (+155)

- Scrollable list of reverted messages with inline preview text
- Per-message "Revert" and "Fork" action buttons with tooltips
- "Restore All" footer button with loading spinner
- Styled to match existing StatusRow todo popover

### UI integration

**StatusRow.tsx** (+21)

- New showRevertIndicator prop, rendered in left section

**ChatInput.tsx** (+1)

- Passes showRevertIndicator to StatusRow

**ChatMessage.tsx** (+2, -2)

- Switches to session-ui-store for revertToMessage and forkFromMessage

### State management

**session-ui-store.ts** (+215)

- revertHistory Map with redo stack and reverted message previews
- activeRevertMessageIDs for per-session revert point indicators
- pushRevertRedo and popRevertRedo for deterministic forward navigation
- ensureRevertSnapshot for lazy rebuild from sync store after page refresh
- handleSlashUndo reads target parts before optimistic deletion (fixes [No text] toast)
- handleSlashRedo uses redo stack; fullUnrevert bypasses stack
- deleteSession, archiveSession, handleSend auto-clear revert history

**session-actions.ts** (+33)

- Snapshots file parts from target message and restores attachments to input
- Clears previous attachments when target message has none
- API failure rollback: restores previous input text and attachments

**input-store.ts** (+21)

- addRestoredAttachment for server-file attachments with inline image preview

### i18n

**locales (7 files, +12 each)**

- chat.revertIndicator.redo and redoAria
- chat.revertPopover.title, noTextContent, restoreAll, revert, fork
- chat.revert.toast.undo, redo, restored using formatMessage
- All 7 locales: en, es, ko, pl, pt-BR, uk, zh-CN

---

## Validation

- `bun run type-check`
- `bun run lint`
- `bun run build`


youfch:
Most of the coding was completed by DeepSeekv4Pro and GLM5.1, and I reviewed and tested it. I have been using this feature for over 20 hours now.

youfch (commented):
**Issue 1** (missing sessionId dependency): Fixed in 4fcda04 - added sessionId to the effect dependency array. **Issue 2** (direct sonner imports): These follow the existing pattern in session-ui-store.ts - zustand store actions cannot use React hooks. Pre-existing forkFromMessage toasts use the same pattern. **Issue 3** (fork toast hardcoded English): These are pre-existing code, not introduced by this PR. I will address them in a separate PR.

btriapitsyn:
Thanks for putting this together and for the initial idea here. The revert dock is a useful UX improvement, and the PR helped surface the right direction.

I ended up refactoring the implementation quite a bit because the original version kept a separate in-memory `revertHistory` / redo stack and tried to keep that in sync with server/session state. That made the UI fragile: the dock could disappear, redo could depend on stale local state, and some selectors could trigger React update loops.

The new version follows the upstream OpenCode model more closely: the dock derives directly from the authoritative `session.revert.messageID` plus the existing message/part sync state. Reverted messages stay in the local store, and the visible timeline is filtered from the session revert marker. This keeps undo/redo/restore behavior simpler and consistent across refreshes.

I also kept the useful row actions from the original PR: restoring and forking individual reverted messages.

btriapitsyn (approved):

PR #1283, opened 2026-05-16
fix(ui): load agents from OpenCode /agent endpoint

SaM13997:
## Summary
- use the OpenCode `/agent` endpoint as the primary UI agent source so plugin-provided agents like oh-my-openagent remain visible in OpenChamber
- keep the existing SDK `app.agents()` call as a fallback when the proxied endpoint is unavailable
- route bootstrap agent loading through the shared client and add a regression test covering the `/agent` path

## Testing
- `bun run --cwd packages/ui type-check`
- `bun test packages/ui/src/lib/opencode/client.test.ts`

## Context
- addresses the agent discovery gap reported around OMO/OpenAgent integration in OpenChamber UI
- related to #1002

### DIFF
--- a/packages/ui/src/lib/opencode/client.test.ts
+++ b/packages/ui/src/lib/opencode/client.test.ts
@@ -1,0 +1,57 @@
+import { beforeEach, describe, expect, mock, test } from 'bun:test'
+
+import { opencodeClient } from './client'
+
+type MutableClientState = {
+  client: {
+    app: {
+      agents: (options?: { directory?: string }) => Promise<{ data?: Array<{ name: string }> }>
+    }
+  }
+  currentDirectory?: string
+}
+
+const clientState = opencodeClient as unknown as MutableClientState
+const originalClient = clientState.client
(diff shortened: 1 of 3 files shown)


btriapitsyn:
I’m not convinced this change fixes the stated issue.

From OpenCode’s side, the JS SDK `app.agents()` method appears to be a typed wrapper around the same `GET /agent` endpoint. Since OpenChamber creates the SDK client with `baseUrl: /api`, `sdk.app.agents()` should already call `/api/agent`.

This PR replaces the SDK call with a manual `fetch('/api/agent')`, but that seems to hit the same endpoint while losing SDK behavior and breaking our existing retry semantics during bootstrap.

Before changing this path, can you please confirm with evidence that `sdk.app.agents()` and direct `GET /api/agent` return different results for the OMO/OpenAgent case?

If they return the same data, this PR should not be needed. If they differ, we need to understand why the SDK wrapper behaves differently instead of bypassing it with a manual fetch.

PR #1290, opened 2026-05-16
optimized instructions to minimize redundancy between vscode and the host

calebomondi:


btriapitsyn:
Thanks for the PR, but I don’t think this is ready to review/merge in its current shape.

The intended feature seems to be adding VS Code commands for Analyze / Test & Debug / Secure selected code. That sounds reasonable, but the implementation is much broader than necessary and the PR has no description, so it is hard to understand the actual scope or intent.

We already have a very similar pattern in the VS Code extension with `openchamber.explain` and `openchamber.improveCode`: read the active editor/selection, build a prompt, reveal the chat view, and call `createNewSessionWithPrompt(...)`. This feature should likely extend that existing flow instead of introducing a new shared package, duplicated prompt files, new session editor stubs, and cross-package relative imports.

Specific concerns:

- The PR description is empty, so reviewers have to reverse-engineer the feature from the diff.
- `packages/shared/src/taskPrompts.ts` is duplicated by `packages/ui/src/stores/taskPrompts.shared.ts`, which defeats the stated goal of reducing redundancy.
- `useTaskPrefixStore.ts` still imports the local duplicate instead of the shared source.
- `getSystemPrompt(task, toolName)` accepts `toolName`, but the prompts do not use `{toolName}`, so the parameter is currently a no-op.
- `SessionEditorPanelProvider` gets new integration/interaction session stubs that are not implemented and appear unrelated to the main feature.
- The new stubs introduce unused locals that will likely fail type-checking.
- `packages/vscode/src/extension.ts` imports shared code via `../../shared/src/...` instead of using a proper workspace/package import.
- Overall, the diff mixes a simple VS Code command feature with unfinished architecture changes.

Could you please simplify this PR?

A much smaller version would be easier to review:

- Add the three VS Code commands near the existing `explain` / `improveCode` commands.
- Reuse the existing selected-code prompt flow.
- Avoid touching `packages/shared`, UI stores, or `SessionEditorPanelProvider` unless there is a clear need.
- If `prefaceText` is required, please explain why in the PR description and implement it through the existing session creation path with a focused diff.
- Add a proper PR description explaining the user-facing behavior and why these architecture changes are needed.

Right now I’d prefer not to merge this as-is because the scope is unclear and the implementation is unnecessarily broad for the feature being added.

PR #1292, opened 2026-05-16
fix: merge local + upstream skills so Claude Code skills appear in Skills tab, not Commands

cmdaltctr:
## Problem

Claude Code skills (`~/.claude/skills/*/SKILL.md`) were showing up in the **Commands** tab instead of the **Skills** tab in Settings.

## Root cause

Two interacting bugs:

### Bug 1: `||` prevents local filesystem scanning

In `skill-routes.js:198`:
```js
const skills = (await fetchOpenCodeDiscoveredSkills(directory)) || discoverSkills(directory);
```

The `||` operator treats empty arrays `[]` as **truthy** in JavaScript. Since the upstream OpenCode `/skill` API returns `[]` (no skills), the `discoverSkills()` function — which scans `~/.claude/skills/`, `~/.agents/skills/`, `.opencode/skills/` etc. — **never ran**.

This same pattern was replicated across all 6 skill endpoints.

### Bug 2: Upstream contamination

The OpenCode SDK `command.list()` API returns `.claude/skills/` entries as commands. The `CommandsSidebar` already had a dedup filter to strip these out (matching command names against skill names), but since the Skills store was empty due to Bug 1, the dedup had no names to filter against.

## Changes

| File | Change |
|------|--------|
| `packages/web/server/lib/opencode/skill-routes.js` | Added `mergeSkillLists()` and `findDiscoveredSkill()` helpers. Replaced all 6 `||` fallback patterns with proper source merging. |
| `packages/web/server/lib/opencode/skills.js` | Added `fs.existsSync` guard before using `matchedDiscovered.path` in `getSkillSources()` to handle upstream built-in skills with synthetic paths like `"<built-in>"`. |

## Verification

Tested with the server running locally:

- `GET /api/config/skills` now returns **47 skills**: 41 from `~/.claude/skills/` (`source=claude`) + 6 from `~/.config/opencode/skills/` (`source=opencode`)
- Individual skill lookups work correctly
- Non-existent skills return `exists: false` gracefully
- The existing CommandsSidebar dedup filter now functions correctly since the Skills store is properly populated

---

_Discovered and fixed while investigating the "skills appearing in commands tab" bug report._

### DIFF
--- a/packages/web/server/lib/opencode/skill-routes.js
+++ b/packages/web/server/lib/opencode/skill-routes.js
@@ -112,6 +112,35 @@ export const registerSkillRoutes = (app, dependencies) => {
     return { scope: SKILL_SCOPE.USER, source };
   };
 
+  /**
+   * Merge upstream (OpenCode server) skills with local (filesystem-scanned) skills.
+   * Deduplicates by name — local skills take priority because they have more
+   * complete scope/source/path metadata.
+   */
+  const mergeSkillLists = (upstreamSkills, localSkills) => {
+    const merged = new Map();
+    for (const skill of localSkills) {
+      merged.set(skill.name, skill);
+    }
+    for (const skill of upstreamSkills) {
+      if (!merged.has(skill.name)) {
+        merged.set(skill.name, skill);
+      }
+    }
+    return Array.from(merged.values());
+  };
+
+  /**
+   * Search for a specific skill across both upstream and local discovery.
+   * Used by individual skill file endpoints.
+   */
+  const findDiscoveredSkill = async (directory, skillName) => {
+    const upstreamSkills = await fetchOpenCodeDiscoveredSkills(directory) ?? [];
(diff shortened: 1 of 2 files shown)


btriapitsyn:
Thanks again for the investigation and PR. You correctly identified the root issue around our skills list being empty and skill-backed commands leaking into Commands.

I ended up taking a slightly different approach in `0f49bc5b`: OpenChamber now uses the OpenCode SDK skills endpoint as the source of truth for the available skills list, including OpenCode’s own discovery, precedence, and deduplication behavior. We still use filesystem access, but only for reading/editing the exact `SKILL.md` path returned by OpenCode.

Because that replaces the local filesystem merge approach from this PR, I’m going to close this one. The PR was still very helpful in confirming the bug and pointing at the right area. Thank you!

cmdaltctr:
Well done, thank you very much for the clarification.

PR #1296, opened 2026-05-17
fix installed skills discovery and improve editor UX

jkker:
- Fix installed skills discovery by using OpenCode's SDK-backed skills list together with OpenChamber's local scan.
- Preserve built-in OpenCode skills such as `customize-opencode` and resolve their SDK-provided content without treating `<built-in>` as a filesystem path.
- Keep global skills from `~/.agents/skills` visible when OpenCode returns a partial list.
- Improve the installed skills details experience with a larger CodeMirror-backed markdown editor, preview mode, supporting-file editing, and read-only handling for built-in skills.
- Improve mobile skills navigation so selecting a skill opens a detail history entry, the top-left Back button returns to the skills list, and browser Back returns from detail to list without closing settings.

## Root Cause

OpenChamber previously preferred OpenCode's live skill discovery whenever it was reachable. The response can include virtual built-in skills with `location: "<built-in>"`, and it can omit local global skills that OpenChamber can still discover from disk. The installed skills UI could therefore either lose global skills or fail to resolve built-in skill content correctly.

The mobile skills settings flow also treated split-page detail navigation as generic settings navigation. That made the top-left Back button jump out to the settings section list while a separate top-right list button returned to the skills list, which was inconsistent with browser history expectations.

## Validation

lint, build, test, agent-browser e2e, human validation

### built-in skill

image

### global skill (`~/.agents/skills/`) + preview mode

image

### mobile

image

## Notes

> development was mostly done by codex - i've fully tested everything and been using it for a day. 
> it's now ready for review. thanks @btriapitsyn and everyone for the awesome work

### DIFF
--- a/packages/ui/src/components/sections/skills/SkillsPage.tsx
+++ b/packages/ui/src/components/sections/skills/SkillsPage.tsx
@@ -270,9 +270,7 @@ const SkillsInstalledPage: React.FC = () => {
   const handleSkillMarkdownChange = React.useCallback((nextMarkdown: string) => {
     setSkillMarkdown(nextMarkdown);
     const parsed = parseSkillMarkdown(nextMarkdown);
-    if (parsed.description !== null) {
-      setDescription(parsed.description);
-    }
+    setDescription(parsed.description ?? '');
     setInstructions(parsed.instructions);
   }, []);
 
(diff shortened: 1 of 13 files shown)


jkker:
Follow-up status: the Greptile stale-description feedback was addressed in `d5a5a69` and both inline Greptile threads are resolved. I also re-ran local validation on the PR branch: `bun run type-check` and `bun run lint` both pass. Ready for review when you have time.

btriapitsyn (approved):

PR #1299, opened 2026-05-18
Scroll restore on session revisit + optimistic activation spinner

vhqtvn:
## Summary

Two coupled changes to the session-switch UX:

### 1. Scroll restore fires when messages render, not when a snapshot flag flips

`ChatContainer` had two correctness bugs on `upstream/main`:

- The `lastScrolledSessionRef` guard was never invalidated, so once a session had been restored once in a tab session, returning to it was treated as `alreadyRestored = true` and skipped. The user lost their scroll position on revisits.
- The restore effect was gated on `hasRenderableSessionSnapshot`, which can stay `false` for the rest of the visit if any assistant message has unmaterialized parts. Combined with the guard bug, this produced inconsistent restores.

Fix:

- Reset `lastScrolledSessionRef` when `currentSessionId` changes, so every visit triggers a fresh restore.
- Gate the restore on `sessionMessages.length > 0` instead of the snapshot flag. By the time messages render, the scroll container is attached — so the `pendingInitialRestoreRef` queue (which deferred restore when the container wasn't mounted yet) is unreachable in normal flow. The queue, its replay effect, and the cross-session cleanup are removed. The no-container code path now no-ops.

### 2. Optimistic activation spinner

Large sessions take a real amount of time to render. Without feedback, the click looks frozen. The fix is a visual-only overlay that paints on the next frame so the click feels instant while React does the heavy work.

- `SessionSidebar` holds `pendingTransitionSessionId`, written via `flushSync` inside `markSessionActivating` so the spinner is in the DOM on the very next paint.
- `dispatchSessionTransition` defers the actual `setCurrentSession` by two `requestAnimationFrame`s — first frame paints the spinner, second frame dispatches the switch. A token cancels any older in-flight transition if the user clicks again.
- `ChatContainer` emits `openchamber:session-ready` once the new session's snapshot is renderable; the sidebar listener matches it to the pending id and clears the overlay.
- A 4 s `setTimeout` failsafe drops the overlay if the ready event never fires (target deleted mid-transition, sync stalled, etc.).
- `SessionNodeItem` accepts a separate `optimisticActiveSessionId` prop. When it matches the row's id (and the row isn't already active), the row renders a small spinner before the title and gets a `bg-primary/18 ring-1 ring-primary/45` outline.

### What's different from the parent branch

The parent branch derived `effectiveActiveSessionId = pending ?? current` and passed it as `currentSessionId` to the whole tree, so the sidebar's `isActive` flipped to the pending id immediately. That widened the failure mode: any other code path calling
`setCurrentSession` mid-transition (route handler, sync recovery, command palette) would land its update while the sidebar still treated the pending id as truth. The 4 s failsafe would then clear the overlay and snap back to the real id — the bounce-back the
reviewer reported.

This PR keeps the optimistic id as a **visual-only overlay**: `currentSessionId` passed to the tree stays the real store value. `isActive` follows the truth immediately; the spinner is rendered from a separate `optimisticActiveSessionId` prop. Any out-of-band write wins on the next paint.

## Test plan

- [x] Open session A, scroll up to a known anchor, switch to session B, switch back to A. A scrolls to the same anchor on the first paint after switch (no scroll-to-bottom-then-restore flicker, no skipped restore).
- [x] Open a session with unmaterialized assistant parts. Restore still fires once messages render.
- [x] Click a large session. Spinner + ring appears on the target row on the next frame (click feels instant). Spinner clears once the chat content renders.
- [x] Click A → B → C rapidly. Only C shows the spinner; B's transition is cancelled via the token bump. Final state lands on C.
- [x] **Bounce-back regression check:** click a session in the sidebar, then immediately switch via the command palette (or any other `setCurrentSession` caller) before the chat finishes hydrating. The sidebar's `isActive` row should follow the command-palette target instantly. After the 4 s failsafe (or sooner if `session-ready` fires for either id), the spinner clears. No bounce-back to the originally-clicked row.
- [x] Hash-deep-link still wins over restore.


vhqtvn:
@btriapitsyn any concerns about this PR? Maybe you can mock a large session (with many tools calls outputs, reasoning), you'll see the pain point when activating that session

PR #1314, opened 2026-05-18
fix(ui): update viewport height without visualViewport

isanchez404:
## Summary
- Keep mobile layout height updates working when `window.visualViewport` is unavailable by falling back to `window.resize`.
- Preserve existing `visualViewport` resize/scroll behavior and cleanup captured listeners safely.
- Add targeted coverage for viewport state calculation with and without `visualViewport`.

## Validation
- `vet "fix visual viewport height updates when visualViewport is unavailable" --agentic --agent-harness claude`
- `bun test packages/ui/src/hooks/useVisualViewport.test.ts`
- `bun run type-check`
- `bun run lint`
- `git diff --check`

### DIFF
--- a/packages/ui/src/hooks/useVisualViewport.test.ts
+++ b/packages/ui/src/hooks/useVisualViewport.test.ts
@@ -1,0 +1,35 @@
+import { describe, expect, test } from 'bun:test';
+
+import { getVisualViewportState } from './useVisualViewport';
+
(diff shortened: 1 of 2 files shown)


isanchez404:
Addressed Greptile P2 by initializing useVisualViewport state with getVisualViewportState and removing the stale getInitialHeight path in aad2d158.

btriapitsyn (approved):

PR #1316, opened 2026-05-18
fix: resolve symlinks in project directory paths

### LINKED ISSUE #1315: [Bug] Sessions not visible when project path contains symlinks

When a project directory is accessed via a symlink, sessions created in that directory are not visible in OpenChamber. The project appears in the sidebar but shows no sessions.

OpenCode resolves symlinks when storing sessions (using the canonical path), but OpenChamber passes the unresolved symlink path to `session.list({directory})`. The server-side directory filter is a string match, so the paths don't match.

Example:
- Symlink: `/home/user/workspace` → `/workspace/user`
- OpenChamber sends: `/home/user/workspace/project`
- OpenCode DB has: `/workspace/user/project`
- Result: no match, no sessions shown

Ideally, OpenChamber should resolve symlinks before querying the OpenCode API so users don't need to worry about whether they loaded the symlinked directory vs the canonical one.

### Where does it happen?

Desktop (macOS)

### Version (if known)

_No response_

_No response_

### Logs (optional)

```shell

```

jeremysamuel13:
OpenCode stores sessions using the canonical (realpath) directory, but OpenChamber passed the unresolved symlink path in several places. The string-match directory filter would fail when a project was accessed via a symlink, making sessions invisible.

**Changes:**
- Add `safeRealpathSync` to settings normalization — project paths and `lastDirectory` are canonicalized at persistence time
- Add Express middleware before the API proxy to resolve symlinks in `?directory=` query params on in-flight requests
- Resolve symlinks in `/api/fs/list` so the directory browser returns canonical paths, allowing the "already added" check to work correctly
- Reconcile the in-memory projects store when the server responds with normalized paths, preventing temporary duplicates

Fixes #1315

----

*note for transparency: used opus 4.6 + opencode to assist with this PR*

### DIFF
--- a/packages/ui/src/components/session/DirectoryTree.tsx
+++ b/packages/ui/src/components/session/DirectoryTree.tsx
@@ -42,6 +42,18 @@ interface DirectoryTreeProps {
   disabledPaths?: Iterable<string>;
 }
 
+const areStringSetsEqual = (left: Set<string>, right: Set<string>) => {
+  if (left.size !== right.size) {
+    return false;
+  }
+  for (const value of left) {
+    if (!right.has(value)) {
+      return false;
+    }
+  }
+  return true;
+};
+
 export const DirectoryTree: React.FC<DirectoryTreeProps> = ({
   currentPath,
   onSelectPath,
@@ -244,7 +256,10 @@ export const DirectoryTree: React.FC<DirectoryTreeProps> = ({
           const normalizedPath = path.replace(/\\/g, '/');
           return (stripTrailingSlashes(normalizedPath) as string) ?? normalizedPath;
         });
-      setPinnedPaths(new Set(normalized));
+      setPinnedPaths((prev) => {
+        const next = new Set(normalized);
+        return areStringSetsEqual(prev, next) ? prev : next;
+      });
     };
 
     const loadFromLocalStorage = () => {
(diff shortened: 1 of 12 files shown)


btriapitsyn:
Thanks for taking this on. The core issue you were addressing is real, and resolving symlinked project paths is the right direction.

I pushed a follow-up commit to tighten the implementation. The main problem was that the original fix added `fs.realpathSync()` to the OpenCode proxy path, which means every directory-scoped proxied API request could block the Node event loop on filesystem IO. That is too risky for a request hot path, especially for something that should be a compatibility fallback rather than the primary source of truth.

The follow-up changes:
- replace the proxy `realpathSync()` call with cached async canonicalization;
- dedupe concurrent `realpath` lookups for the same path;
- keep strict validation behavior where validation errors should still surface;
- keep proxy canonicalization best-effort so failed `realpath` does not break requests;
- route UI settings reconciliation through the existing `settings-synced` / `synchronizeFromSettings` path instead of direct store mutation;
- avoid a settings autosave loop caused by re-applying unchanged pinned directories;
- add targeted tests for the cache and proxy behavior.

For future changes like this, please be careful not to put synchronous filesystem work or other blocking operations on high-traffic request paths. The fix was directionally useful, but the first implementation was too mechanical and did not account for runtime cost.

btriapitsyn (approved):

PR #1358, opened 2026-05-21
Add custom font controls

lederniermagicien:
## Problem

Appearance settings only offered a fixed list of bundled font presets for the Interface Font and Code Font controls. Users with local fonts installed — for example MonoLisa, Berkeley Mono, or a system corporate font — had no way to use them without changing code or CSS.

The custom font control also needs to avoid two UX/perf traps:

1. **Do not put an always-visible text field beside the preset select.** It makes the row noisy, widens the control group, and breaks the compact settings rhythm.
2. **Do not apply font-family changes on every keystroke.** Changing `--font-sans`, `--font-mono`, and `document.body.style.fontFamily` while typing can force broad style recalculation and make the input feel sluggish.

## Solution

Add a `Custom…` entry to both font selects. Selecting it reveals a compact text field directly below the select, using the same 24px settings-control height as adjacent selects.

The entered font family is treated as a local installed font: it is prepended to the selected preset stack, so if the font is not installed the UI naturally falls back to the chosen preset.

### How it works

- Interface and code font preferences now each have two persisted values:
  - the existing preset id (`uiFont`, `monoFont`)
  - an optional custom family (`customUiFont`, `customMonoFont`)
- `getUiFontStack()` and `getMonoFontStack()` build the effective CSS stack:

```ts
"Custom Font", <selected preset fallback stack>
```

- The custom input keeps a local draft while the user types. The draft is committed:
  - after a 250ms debounce, so typing does not repaint the app on every keypress
  - immediately on blur, so leaving the field never loses the latest value
- Choosing a preset clears the custom override. Reset returns to the default preset and clears both draft and persisted custom values.

## Safety details

Custom font names are sanitized before they reach CSS:

- control characters are replaced
- input is capped at 120 chars
- surrounding quotes are normalized
- family names are quoted and escaped
- generic CSS families such as `ui-monospace`, `system-ui`, and `serif` stay unquoted

This keeps the feature flexible enough for real installed font names while avoiding malformed CSS font-family strings.

## Changes

**Font stack logic**
- Add custom font normalization/sanitization helpers in `fontOptions.ts`
- Add `getUiFontStack()` / `getMonoFontStack()` so app-level font effects do not duplicate fallback logic
- Add focused tests for custom stack construction, escaping, generic families, blank fallback, and typing-friendly sanitization

**Settings UI**
- Add `Custom…` to Interface Font and Code Font select menus
- Reveal a compact custom-family input only when `Custom…` is selected
- Align the input with existing settings controls instead of using the larger default text input styling
- Debounce custom font application while typing

**Persistence/application**
- Persist `customUiFont` and `customMonoFont` in `useUIStore`
- Include custom font values in `useFontPreferences()`
- Apply effective stacks in `useAppFontEffects()`

**Localization**
- Add labels/placeholders for all existing settings locales.

## Verification

- [x] `bun test packages/ui/src/lib/fontOptions.test.ts`
- [x] `bun run type-check`
- [x] `bun run lint`
- [x] `bun run build`


lederniermagicien:
@greptileai Re-review this PR after fc942b9922910acfa6bfee88000d0dc01e8f8ac2. I addressed the custom font comments: persisted custom font fields are sanitized during store merge, C1 controls are stripped, and focused drafts are not overwritten by debounced sanitized store commits.

lederniermagicien:
Hi @btriapitsyn , would there any changes be required for this PR to be approved? Happy to amend as needed 👍

tomzx:
@openchamber-bot review

makeittech:
Thanks for this, @lederniermagicien. 

The sanitization/CSS-escaping/persistence logic all looks solid (confirmed by both automated reviews too). Two things to fix before this can merge:

**Merge conflicts:** main's Appearance settings page was refactored to the SettingsSection/SettingsTwoColumn/SettingsStackedField primitives after this branch was cut, so OpenChamberVisualSettings.tsx now conflicts structurally (the custom-font select + input need to move into the new SettingsStackedField control area instead of the old manual layout). useUIStore.ts has a one-line import conflict. Could you rebase/merge main and resolve those?

**Missing locale coverage:** fr.settings.ts, ja.settings.ts, and zh-TW.settings.ts are missing the 5 new keys (settings.openchamber.visual.option.font.custom and the 4 custom-font placeholder/aria keys) that the other 7 locale files got. All 10 locale files need them, fully translated (not English fallback).

**Small polish while you're in there:** the custom font <Input> elements are missing maxLength={120} to match the store's cap — pasting a long name currently overflows visibly until it's truncated on blur/debounce.

btriapitsyn:

PR #1359, opened 2026-05-21
feat: Redesign git changes to split stage/unstaged files.

ShogunPanda:
This PR changes the Git panel staging workflow across UI, web runtime, and VS Code runtime.

Primary areas to review:

- Staged and unstaged file presentation in the Git panel
- `+` / `-` stage and unstage controls
- Bulk stage/unstage API support
- Optimistic staged/unstaged UI updates
- Serialized Git index mutation queue
- Delayed Git reconciliation
- Cross-runtime behavior for web/Electron and VS Code
- Test coverage for the above

## Disclaimer

Assisted-By: OpenAI:GPT-5.5 <openai/gpt-5.5>


ShogunPanda:
I've addressed all issues reported earlier.

btriapitsyn:
Thanks for the work on this, the staging flow is a nice direction.

I pushed a follow-up that keeps your core approach but smooths out a few pieces:

- unified staged and unstaged changes into one `ChangesPanel`
- kept the grouped `Staged` / `Changes` layout while removing the draggable split
- preserved tree/flat view behavior and directory-level stage/unstage actions
- deduplicated `Revert all` paths so partially-staged files are only reverted once
- moved the tree-building logic into a small helper
- moved stashes access into the Git header menu
- added safer `git add` fallback handling for optimistic queued stage operations
- applies stashes with `--index` first so staged/unstaged state is restored when possible

btriapitsyn (approved):

PR #1362, opened 2026-05-21
fix: switch model when selecting an agent

Adrian-Eckardt:
## Bug

When selecting an agent from the chat dropdown, the model selector does not reliably update to the model configured on that agent. It can keep the previously selected model, or a stale per-session agent/model choice can be re-applied shortly after the switch.

## Root Cause

There were two paths preventing the selected agent's configured model from winning:

1. In `useConfigStore.setAgent()`, an early return kept the current model whenever it was valid. Since the current model is usually valid, the agent model fallback was effectively unreachable.
2. In `ModelControls`, the agent-switch effect restored a persisted per-session agent/model selection before considering the agent file's configured model. That could override the model from the selected agent.

## Fix

Explicit agent selection now prefers the model configured on the selected agent:

- `useConfigStore.ts`: applies the selected agent's configured model before saved/default fallback logic.
- `ModelControls.tsx`: applies the selected agent's configured model before restoring persisted per-session agent/model choices.

If the selected agent has no valid configured model, existing fallback behavior remains: saved selection, settings default, then current valid model.

## Testing

- `bun run lint` — passes
- `bun run build` — passes
- Deployed locally and verified the updated build starts successfully

### DIFF
--- a/packages/ui/src/components/chat/ModelControls.tsx
+++ b/packages/ui/src/components/chat/ModelControls.tsx
@@ -618,6 +618,7 @@ export const ModelControls: React.FC<ModelControlsProps> = ({
     ];
 
     const prevAgentNameRef = React.useRef<string | undefined>(undefined);
+    const explicitAgentSwitchRef = React.useRef<string | null>(null);
     const latestLoadedUserChoiceRestoreRef = React.useRef<string | null>(null);
 
     const currentSessionDirectory = currentSessionId ? getDirectoryForSession(currentSessionId) : undefined;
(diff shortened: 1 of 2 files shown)


Adrian-Eckardt:
Updated the fix after reproducing the remaining issue: stale per-session agent-model selections were re-applied after the agent switch and could override the agent file's configured model. The PR now makes explicit agent selection prefer the agent's configured model in both the config store and the ModelControls agent-switch effect. Validation still passes: type-check, lint, and build.

Adrian-Eckardt:
Rebased this PR on latest upstream main after updating local OpenChamber to 1.11.6. Validation still passes locally: `bun run type-check`, `bun run lint`, and `bun run build`.

btriapitsyn (approved):

PR #1370, opened 2026-05-22
fix(ui): keep PWA dialogs visible on Android

lilyzhaun:
## Summary

Dialogs in the installed PWA (and any standalone display-mode browser tab) are pushed below the visible viewport on Android. In the Skills catalog **Install** dialog this is bad enough that the **Install / Cancel** buttons can land off-screen and the dialog becomes unusable.

Root cause is in `packages/ui/src/styles/mobile.css` under

```css
@media (display-mode: standalone) and (max-width: 768px) {
  .pwa-dialog-content { ... }
}
```

Two problems for non-iOS devices:

1. The rule unconditionally adds `top: calc(50% + (safe-area-top * 0.22))` and `--tw-translate-y: calc(-50% + (safe-area-top * 0.6))`. These were written to clear the iOS status bar / notch, but the popup is already centered by the portal wrapper (`fixed inset-0 flex items-center justify-center` in `dialog.tsx`). On Android the safe-area-top is non-trivial and the extra offsets just shove the dialog downward.
2. `max-height` is calculated against `100vh`. On Android Chrome the collapsible URL bar means `100vh` exceeds the actually visible viewport while the bar is shown, so the dialog's footer (action buttons) is pushed off-screen. This is what causes the **Install** button to be unreachable in the Skills install dialog.

## Fix

- Switch the shared rule to `max-height: calc(100dvh - safe-area-top - safe-area-bottom-visual - 20px)`. `dvh` tracks the dynamic visible viewport, so the dialog footer stays on-screen on Android even when the URL bar is visible. Older browsers without `dvh` still get the previous behavior via the unprefixed declarations earlier in the file.
- Move the `top` / `--tw-translate-y` overrides into `@supports (-webkit-touch-callout: none)`, so they only apply on iOS where they were originally intended.

The iOS code path is byte-for-byte preserved, just scoped.

## How it was tested

- `bun run type-check` and `bun run lint` in `packages/ui` (CONTRIBUTING.md required checks). Both clean.
- `bun run build` in `packages/web`. The emitted `dist/assets/index-*.css` shows:
  - The shared `.pwa-dialog-content` rule now uses `max-height: calc(100dvh - ...)` with no `top` / `--tw-translate-y`.
  - Inside `@supports (-webkit-touch-callout: none)`, `.pwa-dialog-content` keeps the original `top: calc(50% + (--oc-safe-area-top * 0.22))` and `--tw-translate-y: calc(-50% + (--oc-safe-area-top * 0.6))`.
- Manual smoke test against a self-hosted instance from Android Chrome PWA:
  - **Settings → About → OpenChamber version** dialog: centers vertically and the close / update buttons stay in the viewport.
  - **Skills catalog → Install** dialog: footer **Install / Cancel** buttons are now visible without scrolling and remain reachable when the URL bar collapses / expands.
- No visual regressions observed on desktop Chrome / Firefox at narrow breakpoints (`<=768px` non-PWA does not match `display-mode: standalone`, so this block is a no-op there).

## Risk / scope

- Pure CSS change in a single `@media` block.
- Only affects PWA / standalone display-mode on viewports up to 768px wide.
- Falls back to the previous behavior on browsers that don't support `dvh`.
- iOS PWA behavior intentionally unchanged.

## Files changed

- `packages/ui/src/styles/mobile.css` (+20 / -3)

### DIFF
--- a/packages/ui/src/styles/mobile.css
+++ b/packages/ui/src/styles/mobile.css
@@ -385,9 +385,15 @@
 @media (display-mode: standalone) and (max-width: 768px) {
   .pwa-dialog-content {
     --pwa-dialog-padding: clamp(0.75rem, 3vw, 1.25rem);
-    /* Use dvh so Android Chrome's collapsible URL bar doesn't push
-       the dialog footer off-screen. Fallback to vh on older browsers
-       via the earlier declaration. */
+    /* Browsers that don't understand dvh ignore the second declaration and
+       keep the vh-based fallback. Newer browsers use 100dvh so Android
+       Chrome's collapsible URL bar doesn't push the dialog footer off-screen. */
+    max-height: calc(
+      100vh
+      - var(--oc-safe-area-top, 0px)
+      - var(--oc-safe-area-bottom-visual, 0px)
+      - 20px
+    );
     max-height: calc(
       100dvh
       - var(--oc-safe-area-top, 0px)
--- a/packages/ui/src/styles/mobile.css
+++ b/packages/ui/src/styles/mobile.css
@@ -385,11 +385,17 @@
 @media (display-mode: standalone) and (max-width: 768px) {
   .pwa-dialog-content {
     --pwa-dialog-padding: clamp(0.75rem, 3vw, 1.25rem);
-    top: calc(50% + (var(--oc-safe-area-top, 0px) * 0.22));
-    max-height: calc(100vh - var(--oc-safe-area-top, 0px) - var(--oc-safe-area-bottom-visual, 0px) - 20px);
+    /* Use dvh so Android Chrome's collapsible URL bar doesn't push
+       the dialog footer off-screen. Fallback to vh on older browsers
+       via the earlier declaration. */
+    max-height: calc(
+      100dvh
+      - var(--oc-safe-area-top, 0px)
+      - var(--oc-safe-area-bottom-visual, 0px)
+      - 20px
+    );
     padding: var(--pwa-dialog-padding);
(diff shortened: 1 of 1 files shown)


lilyzhaun:
You're absolutely right — apologies for that, the comment described a pattern I never actually shipped. The previous version of `mobile.css` had a single `max-height: calc(100vh - ...)` line, and my first patch replaced it with `100dvh` in place. There was no earlier vh declaration left in the file, so on any engine that doesn't parse `dvh` the entire `max-height` would be dropped from the dialog inside this media block — which, as you note, is strictly worse than the pre-patch behavior.

Fixed in 7c34d475: I restored the `100vh` line as a first declaration and kept `100dvh` as the override on the following line, which is the standard progressive-enhancement pattern (older engines silently drop the second invalid declaration and keep the first; modern engines apply the override). I also reworded the inline comment so the two-line pattern is clearly intentional and won't be tidied away as accidental duplication in a future cleanup.

Verified in the rebuilt CSS — the shared rule now emits both declarations in order:

```css
.pwa-dialog-content{
  ...
  max-height:calc(100vh - var(--oc-safe-area-top,0px) - var(--oc-safe-area-bottom-visual,0px) - 20px);
  max-height:calc(100dvh - var(--oc-safe-area-top,0px) - var(--oc-safe-area-bottom-visual,0px) - 20px);
  ...
}
```

`type-check` and `lint` in `packages/ui` are still clean. Thanks for the careful read.

btriapitsyn (approved):

PR #1375, opened 2026-05-22
feat: plugin settings

Quat3rnion:
## Summary

Adds a Plugins page under Settings → OpenCode for managing the opencode `plugin` array end-to-end, plus npm registry awareness for update + invalid-spec detection.

**CRUD**
- View, add, edit, remove `plugin` config entries across user (`~/.config/opencode/opencode.json`) and project (`<project>/.opencode/opencode.json`) scopes.
- Supports all opencode spec shapes: bare npm name, scoped npm (`@scope/name`), versioned (`name@1.2.3`), dist-tagged (`name@beta`), and local paths.
- Optional `options` JSON per entry; tuple form (`[spec, options]`) auto-collapses to string when options are empty.
- Manages auto-loaded plugin files in `~/.config/opencode/plugins/` and `<project>/.opencode/plugins/` (list, create, edit, delete) with strict filename validation (`^[a-z0-9][a-z0-9-_.]*\.(js|ts|mjs|cjs)$`, rejects `..`/`/`).
- All mutations call `refreshOpenCodeAfterConfigChange`, mirroring the MCP/skills restart flow.

**npm registry integration**
- New `GET /api/config/plugins/registry?specs=...&refresh=` endpoint backed by `npm-registry.js` (in-flight dedup, 5s timeout, 1h cache; 404s cached, network failures NOT cached per repo's distinguish-failure rule).
- `plugin-spec.js` parses spec strings and detects exact semver; `hasUpdate` is only flagged when current is exact semver and differs from `latest` (range/dist-tag/bare specs stay quiet).
- Sidebar shows per-row update badge (`↑ x.y.z`), group header update counts, and a kebab "Update to latest" action wired through the existing PATCH path.
- PluginsPage shows a banner for update-available / missing-version / missing-package / malformed / missing-path / unreadable-path / offline-registry states.
- Sidebar header gets a refresh button for forced cache bypass.

**UI / infra**
- New section module `packages/ui/src/components/sections/plugins/` with `PluginsSidebar`, `PluginsPage`, `AddPluginDialog`, `RegistryBadge`, `RegistryBanner`.
- New Zustand slice in `usePluginsStore.ts` covers entries, files, registry info, draft, cache TTL, in-flight dedup, and per-spec narrow subscriptions.
- Settings metadata gains a `plugins` slug (group `opencode`, kind `split`); SettingsView wires nav icon, page title, sidebar, content render, and store load on slug activation.
- Whitelists `/api/config/plugins` in the JSON body-parser middleware so POST/PATCH bodies parse.
- 64 new i18n keys across all 7 locales (English source; non-English copy English verbatim — matches existing deferred-translation pattern).

## Why

opencode supports plugins via the `plugin` array in `opencode.json`, but until now there was no UI surface for managing them — users had to hand-edit JSON, remember exact spec syntax, and discover updates manually. This PR:

- Removes the JSON-editing friction with a typed UI that mirrors the MCP/skills patterns already familiar to users.
- Surfaces silent failure modes that previously went unnoticed: misspelled package names, pinned versions that were unpublished, local plugin files that moved or got deleted.
- Makes outdated plugins discoverable at a glance and updatable with one click, instead of requiring users to check npm manually for each entry.
- Lays the groundwork for future plugin-aware features (deprecation warnings, range-aware updates, bulk update) without changing the public API.

## Testing

**Automated (110/110 pass, 0 fail):**
- `bun test packages/web/server/lib/opencode/plugins.test.js` — 13 cases (data layer round-trip, tuple↔string collapse, scope routing, file CRUD, filename traversal, id encode/decode, env-isolated tmp dirs).
- `bun test packages/web/server/lib/opencode/plugin-routes.test.js` — 28 cases (full CRUD HTTP contract + 16 registry route cases: npm-ok, missing-version, missing-package, malformed, network, path-ok/missing/unreadable, dedup, refresh bypass, 100-spec cap, exact-semver guard).
- `bun test packages/web/server/lib/opencode/plugin-spec.test.js` — 31 cases (parseNpmSpec across all spec shapes, isExactSemver edge cases, parsePathSpec home/cwd/abs).
- `bun test packages/web/server/lib/opencode/npm-registry.test.js` — 14 cases (200/404/500/network/timeout, cache hit/miss, force-refresh, in-flight dedup, 404 cached but network not, scoped name URL encoding, User-Agent header). All fetches mocked.
- `bun test packages/ui/src/stores/usePluginsStore.test.ts` — 24 cases (CRUD, cache TTL, URL chunking >1500 chars, force refresh, mutation→registry refresh, updateToLatest happy/no-update paths, deletion clears selectedId).

**Static gates:**
- `bun run type-check` exits 0 across all 5 packages; strict TypeScript, zero `: any` / `as any` in any new file.
- `bun run lint` exits 0 across all 5 packages.
- Zero hardcoded Tailwind colors in any new UI file (theme tokens only: `text-[var(--status-success)]` etc.); zero direct `@remixicon/react` imports (Icon sprite component only); narrow per-spec store subscriptions throughout (no broad `registryInfo` map subscriptions).

**Live API smoke against `bun dev` on :3001:**
- `GET /api/config/plugins/registry?specs=opencode-wakatime@1.3.0` → `kind: npm-ok`, `latestVersion: 1.3.8`, `hasUpdate: true`.
- `GET /api/config/plugins/registry?specs=nonexistent-pkg-xyz-99999@1.0.0` → `npm-missing-package`.
- `GET /api/config/plugins/registry?specs=opencode-wakatime@99.99.99` → `npm-missing-version` with `latestVersion: 1.3.8`.
- `GET /api/config/plugins/registry?specs=%2Ftmp%2Fnonexistent.js` → `path-missing`.
- `GET /api/config/plugins/registry?specs=%40%40bad` → `npm-malformed`.
- Dedup: `specs=lodash@1,lodash@2,lodash@3` → 3 results from 1 registry fetch.
- Cache timing: cold 247 ms → warm 8 ms (30× speedup) → `&refresh=true` 176 ms (cache bypass).

**Live CRUD against real `~/.config/opencode/opencode.json`:**
- POST string spec → file shows string form. PATCH adding options → file flips to tuple form. PATCH `{options:{}}` → file collapses back to string. DELETE → entry removed and `plugin` key pruned when array empty. Verified at every step with `jq '.plugin'`. Final state: 9 plugins (baseline preserved).
- File CRUD: POST `qa-test.ts` writes to `~/.config/opencode/plugins/`; PUT updates content; DELETE unlinks. Filename traversal `../escape.js` rejected with 400.

**Browser QA (Playwright against built UI on :3001):**
- Settings → Plugins renders 13 user-scope items (9 config entries + 4 dir files) grouped under "User · Config entries" + "User · Files".
- Update badges visible: `opencode-wakatime@1.3.0 ↑ 1.3.8`.
- Group header shows `"User · Config entries 2 updates available"` (plural rendered correctly via split `_one`/`_other` keys — i18n runtime does not parse ICU plural format, so dedicated singular/plural keys are required).
- Selecting an updatable entry opens PluginsPage with the registry banner showing `"Update available 1.3.0 → 1.3.8"` and an `"Update to 1.3.8"` action button.
- Add dialog: three tabs (From npm / From local path / New file); filename validator rejects `../escape.js`, `Foo.JS`, `foo.txt` with inline error and disables Submit; accepts `foo.js`.
- Long-spec rendering verified at the DOM level: spec text truncates with ellipsis (`flex-1 min-w-0 truncate`), badge stays anchored at the right of the row via `shrink-0`, badge right edge equals row right edge for both 42 px text badges (`↑ x.y.z`) and 12 px icon-only badges (warning states).
- Zero plugin-related console errors (the two pre-existing `/api/fs/read` 404s for missing project config files are unrelated and present on `main`).
- Theme toggle (dark/light) leaves badges and banners legible in both modes.


btriapitsyn:
Added a follow-up commit: `1ace8fce fix(settings/plugins): address review follow-ups`.

What changed:
- Fixed the file editor dirty-state regression by tracking the original loaded plugin file content instead of referencing the undefined `readFileCache`.
- Treated Windows absolute and UNC paths as local plugin paths, so they are not queried against npm; added targeted route/spec tests.
- Removed duplicate textarea resize handles on the plugins page.
- Replaced the Add plugin dialog's custom tab markup with the shared `SortableTabsStrip`.
- Cleaned up plugin copy (for example `Installed plugin`, `User config`, `Project plugin file`) and added proper translations for all `settings.plugins.*` strings across es, ko, pl, pt-BR, uk, and zh-CN.

btriapitsyn (approved):

PR #1377, opened 2026-05-22
fix(ui): use dvh units for PWA dialog max-height on mobile

### LINKED ISSUE #1371: [Bug]  Modals and confirmations are not fully visible

When using Chrome on a mobile device running Android 15 (in PWA mode), modal windows do not display fully. This affects update notifications, file renaming and deletion windows, action confirmations, and similar elements.

Image

### Where does it happen?

Mobile (Web/PWA)

### Version (if known)

1.11.4

_No response_

### Logs (optional)

```shell

```

tamatyan99:
Fixes #1371

## Problem
On Android 15 PWA (Chrome), modal dialogs were not fully visible because the CSS used `100vh` for max-height calculation, which includes the dynamic toolbar/keyboard area on modern mobile browsers.

## Solution
1. **mobile.css**: Changed `100vh` → `100dvh` in the `.pwa-dialog-content` max-height calculation. This uses the dynamic viewport height that excludes dynamic toolbars and keyboard areas.

2. **dialog.tsx**: Added explicit mobile-safe max-height `max-h-[calc(100dvh-2rem)]` to the default `DialogContent` component, with `sm:max-h-[80vh]` for desktop. This ensures dialogs are properly constrained even outside of PWA mode.

## Testing
- [x] `bun run lint` passes


tamatyan99:
@greptileapp Good catch on the CSS specificity issue.

Reverted `dialog.tsx` back to `max-h-full` — the `mobile.css` fix (`100vh` → `100dvh`) is the actual root cause and sufficient on its own. This avoids the specificity conflict between the Tailwind utility and the PWA CSS rule, ensuring the safe-area-aware `mobile.css` rule wins cleanly in standalone PWA mode.

PR #1385, opened 2026-05-23
fix(chat): restore scroll on revisit and keep auto-follow glued on updates

vhqtvn:
## Summary

Correctness fixes for two unrelated-symptom-but-shared-area chat scroll bugs: scroll-restore on session revisit, and auto-follow staying glued to the bottom while a message updates. No new UI, no new state in stores, no behavior change for first-visit sessions or for a freshly released auto-follow.

### 1. Guard ref invalidates on session change

`ChatContainer` keeps `lastScrolledSessionRef` so the restore effect skips already-restored sessions. The ref was never invalidated, so once a session had been restored in a tab session, returning to it was treated as `alreadyRestored = true` and the restore was skipped. The user lost their scroll position on revisits.

Reset the ref when `currentSessionId` changes, so every visit triggers a fresh restore.

### 2. Gate on rendered messages, not on snapshot flag

The restore effect was gated on `hasRenderableSessionSnapshot`. That flag can stay `false` for the rest of the visit if any assistant message has unmaterialized parts — so the restore would never fire on those sessions.

Switch the gate to `sessionMessages.length > 0`. By the time messages have rendered, the chat scroll container is attached in the normal flow, so the previously-needed `pendingInitialRestoreRef` queue (which deferred restore until the container mounted) is unreachable and is removed.

### 3. Retry on container un-mount race

With the queue gone, the no-container path needs to be safe under Suspense re-mount / concurrent-transition races: messages can render while the container is briefly detached, and the rAF callback might see `scrollRef.current === null`.

`restoreSnapshot`'s return value now distinguishes "container missing, retry" (`false`) from "container ready, scroll applied" (`true`, covering both the scroll-to-bottom path and the snapshot-restore path). The hook also exposes `scrollContainerEl`, which the effect depends on; the guard ref is only set after a successful restore. If the container re-mounts later, the effect re-runs and the restore applies cleanly.

### 4. Decouple the follow loop from `sessionIsWorking`

Every kick path in `useChatAutoFollow` (`tickFollow`, `startFollowLoop`, the `ResizeObserver` reaction, `notifyContentChange`, and the re-pin branch in `handleScrollEvent`) used to gate on `sessionWorkingRef.current === true` *in addition to* `state === 'following'`. The moment `time.completed` landed on the trailing assistant message, every kick path turned into a no-op — even though late reflows kept arriving for hundreds of ms (async markdown chunk load, code highlighter hydration, image height adjustment, reasoning block auto-collapse on stream end). A pinned user watched the chat freeze a few dozen px above the bottom.

Gate only on `state === 'following'` now. `tickFollow` self-terminates when content actually settles (`SETTLE_EPSILON` for `SETTLE_FRAMES`), so removing the streaming gate doesn't busy-loop.

### 5. Settle burst on stream completion instead of a hard stop

The `sessionIsWorking`-tracking effect used to call `stopFollowLoop()` the moment streaming ended — tearing down the loop right before the post-completion reflow burst hit. Switch to `startSettleBurst()`, which snaps scrollTop to the bottom each frame for `SETTLE_BURST_DURATION_MS` (the exact window where the reflows happen). `tickFollow` continues in parallel and cleans up when content stabilizes.

### 6. Suppress layout-clamp false-releases

`handleScrollEvent` released auto-follow on any `currentTop < previousTop` while pinned. But when `scrollHeight` shrinks (reasoning block collapses on stream end, tool placeholder is replaced by a shorter output, image height drops after load), the browser auto-clamps `scrollTop` down — a scroll event fires with `currentTop < previousTop`, but the user did not scroll. The most user-visible flake: *"I'm pinned, the chat shrinks, suddenly I'm released and stuck above the bottom for 1.2 s"* (the `REPIN_GRACE_AFTER_RELEASE_MS` grace window).

Track `scrollHeight` per scroll event (and from the `ResizeObserver`, so there's a baseline before the first scroll). When `scrollHeight` has strictly decreased since the last observation, treat the scroll-up as a clamp and skip the release branch.

## Files

- `packages/ui/src/components/chat/ChatContainer.tsx` — `+34 / −4` (invalidate guard on session change, gate on `sessionMessages.length`, defensive `scrollContainerEl` check, only mark restored on success)
- `packages/ui/src/hooks/useChatAutoFollow.ts` — `+61 / −37` (remove `pendingInitialRestoreRef` queue, refine `restoreSnapshot` return semantics, expose `scrollContainerEl`, decouple kicks from `sessionWorking`, settle-burst on completion, layout-shrink release suppression)

### DIFF
--- a/packages/ui/src/components/chat/ChatContainer.tsx
+++ b/packages/ui/src/components/chat/ChatContainer.tsx
@@ -775,6 +775,68 @@ export const ChatContainer: React.FC<ChatContainerProps> = ({ autoOpenDraft = tr
         // triggering a spurious re-restore that overwrites user scroll.
         let rafId: number | null = null;
         let cancelled = false;
+        let settleObserver: ResizeObserver | null = null;
+        let settleTimer: number | null = null;
+
+        const teardownSettle = () => {
+            if (settleObserver) {
+                settleObserver.disconnect();
+                settleObserver = null;
+            }
+            if (settleTimer !== null && typeof window !== 'undefined') {
+                window.clearTimeout(settleTimer);
+                settleTimer = null;
+            }
+        };
+
+        const setupSettleObserver = (container: HTMLElement) => {
+            // The relaxed restore gate (`sessionMessages.length > 0`) lets
+            // the initial restore fire before tool parts (mermaid SVG, lazy
+            // markdown chunks, syntax highlighter, image dimensions) have
+            // finished materializing. The first `restoreSnapshot()` is
+            // therefore calibrated against an intermediate layout, so the
+            // user lands at a scrollTop whose visible region will drift
+            // away from the saved position as parts settle. Watch
+            // scrollHeight for a brief window and re-apply the saved
+            // snapshot whenever it changes — but only while the user has
+            // not scrolled away from where we placed them.
+            if (cancelled) return;
+            if (typeof ResizeObserver === 'undefined' || typeof window === 'undefined') return;
+
+            const SETTLE_TOLERANCE_PX = 4;
+            const SETTLE_MAX_MS = 1500;
+            let lastAppliedTop = container.scrollTop;
+
+            settleObserver = new ResizeObserver(() => {
+                if (cancelled || !settleObserver) return;
+                const currentTop = container.scrollTop;
+                if (Math.abs(currentTop - lastAppliedTop) > SETTLE_TOLERANCE_PX) {
+                    // User scrolled away from our placement — they have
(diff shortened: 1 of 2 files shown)


vhqtvn:
@btriapitsyn I closed the previous PR and split out only the scroll-correctness slice per your last bullet ("make scroll restoration synchronous and non-flickery before adding more behavior"). New PR: split/scroll-restore. 2 files, +197/−40. No SessionSidebar changes, no spinner, no project-to-session reconciliation, no backend, no pagination work. Just the restore-on-revisit + auto-follow-on-update correctness fixes.

PR #1402, opened 2026-05-24
feat: add native macOS dynamic app icon

btriapitsyn:
## Summary
- Adds a native macOS dynamic app icon asset for the Electron app using Icon Composer output.
- Packages the compiled Assets.car into the macOS app bundle and sets CFBundleIconName.
- Adds a helper script to regenerate macOS icon assets when the source .icon changes.

## Why
- Runtime Dock icon overrides only work while the app is running.
- macOS needs a real bundle-level icon asset so Finder, Launchpad, Spotlight, and Dock can render the correct icon based on system appearance.

## Testing
- Ran bun run type-check.
- Ran bun run lint.
- Verified the Electron afterPack hook copies Assets.car into the app resources.


btriapitsyn:
Addressed the follow-up script hardening notes:

- Removed the unnecessary async from the afterPack hook.
- Added an explicit guard that fails clearly if actool does not emit Assets.car before copying.

Pushed in 6815f402.

PR #1418, opened 2026-05-25
Fix iOS PWA keyboard viewport alignment

DevNewbie1826:
## Summary
- Expose visual viewport origin fields (`offsetTop`, `pageTop`) and include them in viewport state equality checks.
- Derive the mobile root style from the current visual viewport rectangle so iOS keyboard panning keeps the chat composer aligned with the visible area.
- Preserve desktop/non-mobile behavior by returning no mobile inline style outside the mobile path.

## Testing
- `bun test packages/ui/src/hooks/useVisualViewport.test.ts`
- `bun run type-check`
- `bun run lint`
- `bun run build`

## Manual verification
- Tested on an iOS standalone PWA installed from the home screen.
- With the keyboard open, the composer remains attached to the keyboard-visible area.
- No large blank region remains above the keyboard.

Screenshot evidence will be attached in the PR discussion.

### DIFF
--- a/packages/ui/src/components/layout/MainLayout.tsx
+++ b/packages/ui/src/components/layout/MainLayout.tsx
@@ -20,7 +20,7 @@ import { useUIStore } from '@/stores/useUIStore';
 import { useSessionUIStore } from '@/sync/session-ui-store';
 import { useUpdateStore } from '@/stores/useUpdateStore';
 import { useDeviceInfo } from '@/lib/device';
-import { useVisualViewport } from '@/hooks/useVisualViewport';
+import { useVisualViewport, deriveMobileRootStyle } from '@/hooks/useVisualViewport';
 import { cn } from '@/lib/utils';
 import { lazyWithChunkRecovery } from '@/lib/chunkLoadRecovery';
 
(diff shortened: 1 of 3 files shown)


DevNewbie1826:
Manual iOS PWA verification:

Before:
- In the iOS standalone PWA, focusing the chat input could move the composer/toolbar up near the status bar.
- A large blank area appeared above the keyboard.

After:
- Tested with the app installed to the iOS home screen and launched as a standalone PWA.
- With the keyboard open, the composer stays attached to the keyboard-visible area.
- No large blank region remains above the keyboard.
- The message viewport remains visible and scrollable.

Screenshots:

**Before — composer jumps upward / blank region above keyboard**

Before: composer jumps upward and blank region appears above keyboard

**After — composer remains above the keyboard-visible area**

After: composer remains above the keyboard-visible area

btriapitsyn:
Thanks for digging into the iOS keyboard issue and for the detailed tests/manual verification.

I ended up solving this in a different direction: instead of manually tracking `visualViewport.offsetTop/pageTop` and applying dynamic root positioning, I simplified the mobile layout to rely on `interactive-widget=resizes-content` plus a stable `100dvh` shell. That avoided the iOS pan/resize fighting we were seeing with visualViewport-driven transforms/fixed positioning.

Because this PR reintroduces the more complex visualViewport positioning path and now conflicts with the current implementation, I don’t think we should merge it as-is.

Closing this out, but the investigation was useful and helped validate that the keyboard handling needed a simpler approach. Thank you!

PR #1427, opened 2026-05-26
OpenChamber CLI

tomzx:
# What

- Add `occtl` CLI  for programmatic control of a running OpenChamber server.
- Register `occtl` as a package bin entry.

The CLI provides six command groups with full CRUD and status operations:

- **project** (aliases: `proj`): list, add, remove, info, configure, set-active.
- **session** (alias: `sess`): list, info, status, attention.
- **task** (alias: `tasks`): list, add, remove, run, enable, disable, status.
- **worktree** (alias: `wt`): list, create, remove, preview.
- **config** (alias: `cfg`): get, set, themes, reload.
- **server** (alias: `srv`): health, info, status.

Supports `--json` for machine-readable output, `--quiet` to suppress informational messages, `--plain` / `NO_COLOR` for colorless output, and configurable `--port` / `--host` with `OPENCHAMBER_PORT` env fallback.

Includes a custom arg parser, HTTP client wrapper with connection-error handling, and an inline table renderer for TTY output.

# Why

OpenChamber needs a scriptable interface so automation (agents!), CI pipelines, and other tools can manage projects, sessions, tasks, worktrees, and config without the interactive web UI.

# How to test

1. Start the dev server: `bun run dev`
2. Run the CLI against it: `node packages/web/bin/occtl.js server health`
3. List projects: `node packages/web/bin/occtl.js project list`
4. Verify JSON output: `node packages/web/bin/occtl.js --json project list`
5. Run the test suite: `bun test packages/web/bin/occtl.test.js`
6. Check help output: `node packages/web/bin/occtl.js --help`

---
Generated with [create-pr-description](https://github.com/tomzx/dot-claude/blob/a11ca12932847dadd5aeaa5255b1a7fb3606d547/skills/create-pr-description/SKILL.md) (`a11ca12`)


tomzx:
Note that this is more of a prototype than something I'd like to see merged in.
The idea is to enable users to interact with openchamber from within openchamber itself, where we could ideally update settings that would be reflected without having to go into the settings ourselves.

One use case I'm particularly interested in is to run scheduled tasks that create a new worktree on each execution.
Instead of implementing this feature in OpenChamber I could then use regular cron + the `occtl` CLI to create a worktree with a given name, submit a session with a specific prompt and have the whole thing orchestrated for me by OpenChamber.

colinmollenhour:
@tomzx You might be interested in this: https://github.com/colinmollenhour/occtl - same name, great minds think alike. 😉 

I'd strongly suggest you check out https://worktrunk.dev, it's awesome!

It appears that OpenChamber now discovers and lists sessions from worktrees that weren't created by OpenChamber which is super awesome. It does seem to take a minute though so if the detection mechanism was faster it seems you don't realyl need another CLI to manage worktrees. That is, for CLI just use OpenCode and git in their native ways and OpenChamber should discover and update quickly to reflect what you're doing, ideally. IMHO.

colinmollenhour:
Linking this PR to the instant worktree detection PR: https://github.com/openchamber/openchamber/pull/1480

tomzx:
@openchamber-bot review

tomzx:
@openchamber-bot review

PR #1429, opened 2026-05-26
fix: session rename exits immediately due to focus race

youfch:
## Summary

Fixes a bug where clicking "Rename" in the session dropdown menu enters rename mode briefly then immediately exits, making sessions unrenamable in the web UI. The issue affected auto-named sessions (new sessions with AI-generated titles) but was intermittent — some sessions worked while others didn't.

The root cause was a focus race between the rename input's `onBlur` handler and Base UI's dropdown popup cleanup (FocusScope focus restoration, CSS transition teardown). When the dropdown closes after clicking "Rename", Base UI's internal effects could steal focus from the newly-mounted rename input, triggering `onBlur` → `handleSaveEdit()` → immediate exit.

## Changes

`packages/ui/src/components/session/sidebar/SessionNodeItem.tsx` (+34/-12)

1. **Replaced `onBlur` with document `mousedown` listener** — saves only on click-outside, immune to focus management timing
2. **Deferred dropdown-initiated rename to `onOpenChangeComplete`** — fires after CSS transition + all cleanup effects complete
3. **Replaced dead `onCloseAutoFocus` with `finalFocus`** — `onCloseAutoFocus` was discarded in the wrapper (`void`), never forwarded to Base UI; `finalFocus` is properly passed via `{...props}`

## Why

The `DropdownMenuContent` wrapper extracts `onCloseAutoFocus` but then runs `void onCloseAutoFocus` — a no-op that suppresses lint warnings but never forwards the prop to `BaseMenu.Popup`. The `finalFocus` prop IS forwarded, and the `editingIdRef` (synced every render) avoids stale closure timing issues.

The `onOpenChangeComplete` approach guarantees the dropdown's close transition (150ms CSS + FocusScope cleanup) is fully done before the rename form mounts, eliminating the focus race entirely.

The document `mousedown` listener replaces the fragile `onBlur` handler — it's a simpler, more reliable mechanism that's immune to any focus management side effects.

## Testing

- Open any session dropdown → click "Rename" → rename form appears and stays open
- Type a new title → press Enter → title saves and edit mode exits
- Click ✓ button → title saves and edit mode exits
- Click ✗ button or press Escape → edit mode exits without saving
- Click outside the rename form → title saves and edit mode exits
- Double-click on a session title → rename form appears
- Auto-named sessions (new sessions with AI-generated titles) can now be renamed
- Previously renamed sessions still rename correctly
- Multiple sessions in sidebar — rename one, verify others unaffected

```
  Type check: 5/5 packages pass
  Lint:       5/5 packages pass
```

### DIFF
--- a/packages/ui/src/components/session/sidebar/SessionNodeItem.tsx
+++ b/packages/ui/src/components/session/sidebar/SessionNodeItem.tsx
@@ -284,7 +284,6 @@ function SessionNodeItemComponent(props: Props): React.ReactNode {
         : (showQuickArchiveAction ? 'group-hover:pr-12 group-focus-within:pr-12' : 'group-hover:pr-5 group-focus-within:pr-5'));
   const alwaysActionPaddingClass = showQuickArchiveAction ? 'pr-13' : 'pr-7';
   const suppressNextSelectRef = React.useRef(false);
-  const editCancelledRef = React.useRef(false);
   const [isTouchPressed, setIsTouchPressed] = React.useState(false);
   const editingIdRef = React.useRef(editingId);
   editingIdRef.current = editingId;
@@ -489,7 +488,6 @@ function SessionNodeItemComponent(props: Props): React.ReactNode {
               onKeyDown={(event) => {
                 if (event.key === 'Escape') {
                   event.stopPropagation();
-                  editCancelledRef.current = true;
                   handleCancelEdit();
                   return;
                 }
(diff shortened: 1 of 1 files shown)


youfch:
Both issues addressed:

**1. editCancelledRef is dead code** — Fixed in 77f2583f. The ref declaration and its single write in the Escape handler have been removed.

**2. finalFocus returning true vs null** — Per Base UI Popup docs: finalFocus accepts boolean, RefObject, or function returning boolean | void | HTMLElement | null. Returning true means "Move focus based on the default behavior". This is the documented API and is correct for the non-rename path.

youfch:
Fix the bug causing renaming failures on the Web UI since version 1.11.5. After clicking rename, the edit session briefly opens then exits immediately.

btriapitsyn (approved):

PR #1435, opened 2026-05-26
perf: bounded caches and prevent client/server memory leak fixes

ricautomation:
Fix server memory leaks that required openchamber stop/start to recover, plus add client-side performance optimizations (BoundedCache for PR store, debounced localStorage, memory leak tests).


tomzx:
@openchamber-bot review

PR #1445, opened 2026-05-27
fix: change agent cycle shortcut from Tab to Ctrl+Tab

nerdosaurus:
## Problem

The agent cycle shortcut is bound to `Tab`, which conflicts with several other Tab behaviors:
- Command autocomplete (navigating suggestions)
- Skill autocomplete
- Snippet autocomplete
- File mention autocomplete
- Natural tab-to-indent in the chat input

## Fix

Changed the default combo for `cycle_agent` from `tab` to `ctrl+tab` in `packages/ui/src/lib/shortcuts.ts`. Also updated the shortcut hint displayed in the model picker to use the computed shortcut (`formatShortcutForDisplay`) instead of a hardcoded `'Tab'` string, so it respects user customization.

The shortcut is still customizable — users can restore the old behavior in their `opencode.json`:

```json
{
  "shortcuts": {
    "cycle_agent": "tab"
  }
}
```

## Changes
- `packages/ui/src/lib/shortcuts.ts` — defaultCombo: `'tab'` → `'ctrl+tab'`
- `packages/ui/src/components/chat/ModelControls.tsx` — hardcoded `'Tab'` → `formatShortcutForDisplay(cycleAgentShortcut)`

### DIFF
--- a/packages/ui/src/components/chat/ModelControls.tsx
+++ b/packages/ui/src/components/chat/ModelControls.tsx
@@ -37,7 +37,7 @@ import { useIsTextTruncated } from '@/hooks/useIsTextTruncated';
 import { formatEffortLabel, getCycledPrimaryAgentName, isPrimaryMode, type MobileControlsPanel } from './mobileControlsUtils';
 import { getCurrentIntlLocale, useI18n } from '@/lib/i18n';
 import { useOpenCodeReadiness } from '@/hooks/useOpenCodeReadiness';
-import { eventMatchesShortcut, getEffectiveShortcutCombo, normalizeCombo } from '@/lib/shortcuts';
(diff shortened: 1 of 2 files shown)


nerdosaurus:
Rebased onto `main` (v1.12.3). No conflicts.

yulia-ivashko:
Thank you for this, and sorry it stayed open so long.

Closing because the ground it stands on is gone. The centralized shortcut system landed on 26 August: `packages/ui/src/lib/shortcuts.ts`, the file this PR edits, was removed and replaced by the `packages/ui/src/lib/shortcuts/` module (schema, config, bindings, registry, dispatcher), and the default layout was rebuilt around a `mod+k` leader.

For the record, your underlying point still holds: `cycle_agent` still defaults to `tab`. What changed is that it is now `customizable: true`, so it can be rebound in Settings without a code change. We are closing the patch, not dismissing the complaint.

Thanks for the time you put into it.

PR #1446, opened 2026-05-27
fix: load sub-agent (delegation) sessions on bootstrap for sidebar visibility

nerdosaurus:
## Problem

Sub-agent sessions created via the `task` tool (agent delegation) did not appear in the session sidebar on startup. The bootstrap session list used `roots: true`, which only returns sessions without a `parentID`. Sub-agent sessions have a `parentID` pointing to their parent session, so they were excluded.

Users could only see sub-agent sessions if the async global session store happened to load them in time, or if SSE events arrived during the current session. After restart, previous sub-agent sessions were invisible.

## Fix

Two-phase session load on bootstrap:

1. **First query**: fetch root sessions with `roots: true` (for accurate `sessionTotal` counter)
2. **Second query**: fetch all sessions without `roots`, extract child sessions (those with `parentID` and not already in the root list)
3. **Merge**: root sessions + child sessions into the store

The second query is best-effort — if it fails, the store falls back to roots only (same behavior as before).

## Changes

- `packages/ui/src/sync/sync-context.tsx` — `loadSessions` now performs a second `session.list` call to fetch child sessions and merges them into the store

### DIFF
--- a/packages/electron/main.mjs
+++ b/packages/electron/main.mjs
@@ -4004,6 +4004,7 @@ const COMMANDS_SAFE_FOR_REMOTE = new Set([
   'desktop_get_app_version',
   'desktop_get_lan_address',
   'desktop_capture_page_rect',
+  'desktop_tray_update',
 ]);
 
 ipcMain.handle('openchamber:invoke', async (event, command, args) => {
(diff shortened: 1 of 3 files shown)


nerdosaurus:
Rebased onto `main` (v1.12.3). Includes the two-phase child session bootstrap fix + the 15s periodic child discovery poll in the sync watchdog for cross-OpenCode-instance detection.

PR #1453, opened 2026-05-27
fix: atomic file writes to prevent concurrent read/write truncation

nerdosaurus:
Concurrent reads and writes to the same file via `fs.writeFile` (which uses `O_TRUNC`) create a race window where the file is briefly empty.

## Fix

1. **Write side** (`/api/fs/write`): replaced direct `writeFile` with write-to-temp-then-rename. `fs.rename()` is atomic on POSIX.

2. **Read side** (`/api/fs/read`): when `readFile` returns empty but `stat` reported non-zero size, retry up to 3 times with 50ms backoff.

3. **FilesView** (`FilesView.tsx`): refuse to save empty `draftContent` when the original `fileContent` was non-empty.

The write-to-temp-then-rename pattern already exists in the codebase (`settings-runtime.js`, `project-config.js`, `github/auth.js`) for the same reason.

### DIFF
--- a/packages/web/server/lib/fs/routes.js
+++ b/packages/web/server/lib/fs/routes.js
@@ -679,6 +679,9 @@ export const registerFsRoutes = (app, dependencies) => {
           content = await fsPromises.readFile(canonicalPath, 'utf8');
           if (content.length > 0) break;
         }
+        if (content.length === 0) {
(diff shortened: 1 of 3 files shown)


nerdosaurus:
Rebased onto `main` (v1.12.3). Includes the write-to-temp-then-rename atomic write fix + read retry + FilesView empty-draft guard against tab-switch zeroing.

btriapitsyn (approved):

PR #1471, opened 2026-05-30
fix(opencode): accept non-2xx status codes in probe-url to support re…

mdbetancourt:
…directs


btriapitsyn:
Resolved the merge conflict by merging current main into the PR branch and keeping both the new probe-url status coverage and the existing preview proxy credential coverage.
Validation run:
- `bun test packages/web/server/lib/opencode/core-routes.test.js`
- `bun run type-check`
- `bun run lint`
- `git diff --check`

btriapitsyn (approved):

PR #1510, opened 2026-06-03
feat(settings): add YOLO mode for notification suppression

raz123:
## Summary

Adds a **YOLO mode** toggle (Settings → Permissions) that suppresses ALL desktop/push notifications at the server source. When enabled, the notification runtime silently returns before dispatching any trigger path (completion, error, question, permission), letting power users enter a distraction-free flow state.

The toggle state is persisted in OpenChamber settings on disk so it survives server restarts and page reloads.

## Why

Power users running long OpenChamber sessions want a single global kill switch for notifications without having to dig into OS-level notification settings or toggle per-session auto-accept. YOLO mode fills this gap.

## Key Changes

- **`runtime.js`** — New `yoloSuppressed` flag + `setYoloSuppression()`/`getYoloSuppression()` pair. Early return at the top of `maybeSendPushForTrigger` skips ALL notification paths when set.
- **`routes.js`** — `GET/POST /api/notifications/yolo-suppress` endpoints. POST persists the flag to OpenChamber settings via `writeSettingsToDisk`. An init block reads the persisted value on server registration so suppression survives restarts.
- **`index.js` & `bootstrap-runtime.js`** — Wires the two new functions through the dependency injection chain from server → bootstrap → notification routes.
- **`useYoloStore.ts`** — Zustand store. `refresh()` reads state from the server endpoint (source of truth). `setEnabled()` arms server-side suppression *before* calling `opencodeClient.setYolo()` to eliminate a race window.
- **`PermissionsSettings.tsx`** — UI toggle with Checkbox, keyboard support, confirmation prompt on first enable, and active warning indicator. Beta badge shown.
- **i18n** — Messages added for all 10 locales (en, es, ko, pl, pt-BR, uk, zh-CN, zh-TW).

## Notable Details

- **No toast on toggle** — Toggle operates silently.
- **No revert on `setYolo` failure** — If the OpenCode SDK round-trip fails, suppression stays armed (already persisted server-side). Suppressing too aggressively is safer than leaking a notification.
- **`stopPropagation` on Checkbox** — Prevents double-toggle when both the div `onClick` and checkbox `onChange` fire.
- **Source of truth is the server** — UI reads `GET /api/notifications/yolo-suppress`, not `opencodeClient.getYoloStatus()`, because the OpenCode SDK may silently drop unknown config fields.

## Testing

- Manual smoke test on macOS Electron: toggle on → all notifications suppressed, toggle off → notifications resume.
- Server restart preserves toggle state (read from persisted settings on init).
- Page reload preserves toggle state (fetch from server on mount).
- No double-toggle or double-API-call on checkbox click.
- Race condition verified: suppression flag is armed before `setYolo()` round-trip begins.


raz123:
@greptileai All fixes are in. Please do a re-review.

raz123:
@greptileai All fixes are in. Please do a re-review.

PR #1523, opened 2026-06-03
fix: restore SSH auto-connect when clicking remote device in host switcher

panzeyu2013:
## Summary

Fixes a regression where clicking a remote SSH device in the Electron host switcher triggered a full page navigation (`window.location.assign`) instead of an in-place runtime switch.

**File:** `packages/ui/src/components/desktop/DesktopHostSwitcher.tsx` (+29, -26)

## Root cause

The early `isElectronShell()` guard sat at the top of `handleSwitch`, intercepting **all** host clicks in the Electron shell (including SSH hosts) before they could reach the SSH connection logic below. SSH hosts in Electron therefore couldn't flow into the SSH-aware code path.

## Fix

1. **Removed** the early `isElectronShell()` block from the top of `handleSwitch`, allowing SSH hosts to reach the SSH connection logic.
2. **Inside the SSH path** (both the already-ready fast path and the fresh-connect success path), the switching method now checks `isElectronShell()`:
   - Electron shell → uses `switchRuntimeEndpoint()` (in-place, no page reload)
   - Desktop shell (browser) → uses `window.location.assign()` (full navigation, as before)
3. **Added** cleanup of the SSH switch modal state in the connect-success path (was previously leaked).
4. **Moved** the Electron probing + switching logic to the bottom as a fallback for non-SSH remote hosts, keeping that behavior intact.

### DIFF
--- a/packages/ui/src/components/desktop/DesktopHostSwitcher.tsx
+++ b/packages/ui/src/components/desktop/DesktopHostSwitcher.tsx
@@ -488,28 +488,6 @@ export function DesktopHostSwitcherDialog({
     const apiOrigin = host.id === LOCAL_HOST_ID ? localOrigin : (normalizeHostUrl(getDesktopHostApiUrl(host)) || '');
     if (!origin) return;
 
-    if (isElectronShell()) {
-      if (!apiOrigin) return;
-      setSwitchingHostId(host.id);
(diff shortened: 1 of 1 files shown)


panzeyu2013:
already fixed

PR #1533, opened 2026-06-04
fix(docker): stabilize workspace dependency install

Youzini-afk:
## Summary

- Pin the Bun image tag used by the Docker build so dependency installation is reproducible.
- Copy the Electron workspace package manifest before install so workspace dependency resolution has the manifest it expects.

## Validation

- bun install --frozen-lockfile
- git diff --check upstream/main..HEAD


jakintosh:
Had to make these changes locally to so that the docker image would build. Should probably get merged asap.

btriapitsyn (approved):

PR #1538, opened 2026-06-04
fix: lighten session list payloads

### LINKED ISSUE #1536: [Bug] iOS WebKit crash from huge /experimental/session payload with revert fields
## Summary

OpenChamber mobile UI can crash iOS WebKit when the session list endpoint returns very large session objects. In my case, the large field was `revert` metadata included in `/api/experimental/session` list responses.

## Actual issue

Opening an OpenChamber instance on iPhone caused Safari to refresh several times and then show:

```text
A problem repeatedly occurred on "https://agent.mydomain.dev/"
```

The same instance worked on desktop browsers and Android. Another OpenChamber instance with fewer or smaller session payloads worked on the same iPhone.

The crash happened in Safari, other iOS browsers, and PWA mode, which all use iOS WebKit.

## Investigation findings

On mobile startup, OpenChamber requested large session lists:

```text
/api/experimental/session?archived=false&limit=500
/api/experimental/session?archived=true&limit=500
```

On the affected machine, a smaller active-session request with `limit=250` returned:

```text
182,852,461 bytes
```

Inspecting that payload showed two sessions where almost the entire size came from the `revert` field:

```text
ses_1ded1df05ffesjOjI5YG0PLpA7  some session title 1     revert about 87.10 MB
ses_1deed19e9ffeA06ZFNOWque6tR  me session title 2    revert about 87.08 MB
```

Those two sessions alone made the session-list response large enough to plausibly crash iOS WebKit while downloading, parsing, storing, or rendering the JSON.

## Mitigation tested

I tested a local OpenChamber proxy patch that removed `revert` from `/api/experimental/session` list responses while preserving the session count.

After stripping `revert`, the same style of `limit=500` responses were roughly:

```text
active sessions:   about 335 KB
archived sessions: about 335 KB
```

The response still contained 500 session rows, but no `revert` fields. The iOS crash stopped with this mitigation.

I then tested deleting the two sessions with huge `revert` fields, reinstalled OpenChamber to restore upstream code, and tested again without the local patch. Current upstream responses are now small:

```text
active sessions limit=500:   about 423 KB
archived sessions limit=500: about 425 KB
```

The iOS crash no longer occurs after removing those oversized sessions.

## Expected behavior

Session-list endpoints should return lightweight session summaries only.

For list views, the UI usually needs fields like:

- `id`
- `title`
- `directory`
- `project`
- `parentID`
- `agent`
- `model`
- `time`
- `cost`
- `tokens`
- maybe small summary/status fields

It should not return large per-session payloads such as:

- `revert` or rollback metadata
- message history
- full diffs
- large tool payloads
- file snapshots or other detail-only data

Those fields should be loaded on demand when a user opens a specific session or invokes a feature that needs them, such as viewing details or reverting changes.

## Why this matters

The current behavior makes the list endpoint size depend on worst-case session internals. A single large session can make the mobile startup payload enormous. Desktop browsers may tolerate this, but iOS WebKit has stricter memory/process limits and can crash the page entirely.

## Suggested fix

Make `/api/experimental/session` return summary objects by default and omit large fields such as `revert`.

Potential API approaches:

- Omit large fields from list responses unconditionally.
- Add an explicit `include=` or `fields=` mechanism for detail fields.
- Add a dedicated endpoint for revert metadata, for example session detail or session revert data loaded only when needed.
- Add regression coverage that list responses do not include unbounded fields.

The local proof-of-fix was simply stripping `revert` from list responses, which reduced hundreds of MB to hundreds of KB and stopped the iOS crash.

### Where does it happen?

Mobile (Web/PWA)

### Version (if known)

1.12.1

_No response_

### Logs (optional)

```shell

```

karamanliev:
## Summary
- sanitize `GET /api/experimental/session` and `GET /api/session` list responses so they drop heavy detail-only fields
- apply the same list sanitizer in shared UI global session loading so VS Code and other runtimes do not retain heavy list payloads
- preserve `revert.messageID` and `revert.partID` in list records, and hydrate full session detail when opening a session

## What changed
- added server-side session list sanitizers in the OpenCode proxy for exact list endpoints only
- sanitized Windows merged `/api/session` list output and shared global session list records
- removed heavy list-only fields such as `metadata`, `permission`, `summary.diffs`, `revert.snapshot`, and `revert.diff`
- kept aggregate summary counts and lightweight revert markers needed for reverted-session UI
- updated the client-side defensive sanitizers and session-open hydration path so reverted sessions still render correctly after reload

## Validation
- `npx bun test packages/web/server/opencode-proxy.test.js`
- `npx bun test packages/web/server/lib/opencode/proxy.test.js`
- `npx bun test packages/ui/src/sync/sanitize.test.ts packages/ui/src/stores/globalSessions.test.ts packages/ui/src/sync/use-sync.test.ts`
- `npx bun test packages/vscode/src/bridge-proxy-runtime.test.js`
- `npx bun run type-check`
- `npx bun run lint`
- `npx bun run vscode:type-check`
- `npx bun run --cwd packages/vscode lint`

Closes #1536


btriapitsyn:
Fixed the review findings in `4d03b6aa`:

- Preserved `session.metadata` in both server-side and client-side session list sanitizers so OpenChamber extension metadata, including handoff/review links, survives global session reloads.
- Kept trimming the heavy detail fields that caused the payload issue: `revert.snapshot`, `revert.diff`, `summary.diffs`, and list `permission`.
- Adjusted summary sanitization to remove only list-heavy diff entries instead of dropping the whole summary when aggregate count fields are absent.
- Added/updated regression coverage for metadata preservation, summary sanitization, global sessions, and proxy list responses.
- Fixed the VS Code webview `Response` binary body typing issue so full type-check passes.

Validation run:

- `bun test packages/ui/src/sync/sanitize.test.ts packages/ui/src/stores/globalSessions.test.ts packages/ui/src/sync/use-sync.test.ts`
- `bun test packages/web/server/opencode-proxy.test.js packages/web/server/lib/opencode/proxy.test.js`
- `bun run type-check`
- `bun run lint`
- `git diff --check`

btriapitsyn:
Resolved the merge conflict with latest `main` and pushed merge commit `3e28a9a7`.

Conflict was in `packages/vscode/webview/main.tsx`; kept the current `main` binary `Response` body handling.

Re-ran validation after resolving conflicts:

- `bun test packages/ui/src/sync/sanitize.test.ts packages/ui/src/stores/globalSessions.test.ts packages/ui/src/sync/use-sync.test.ts`
- `bun test packages/web/server/opencode-proxy.test.js packages/web/server/lib/opencode/proxy.test.js`
- `bun run type-check`
- `bun run lint`
- `git diff --check`

btriapitsyn (approved):

PR #1539, opened 2026-06-05
feat: add PDF file preview support

Dav1dch:
Clicking a PDF file in the file tree now opens it in a new browser tab with the correct `application/pdf` content type.

## Changes

- **`packages/web/server/lib/fs/routes.js`**: Added `.pdf` → `application/pdf` to the MIME type map in `/api/fs/raw` endpoint. Previously PDFs were served as `application/octet-stream`, causing browsers to download them instead of displaying.

- **`packages/ui/src/lib/toolHelpers.ts`**: Added `isPdfFile()` helper function to detect PDF files by their `.pdf` extension.

- **`packages/ui/src/components/views/FilesView.tsx`**: 
  - PDFs are detected and excluded from text/copy/edit operations
  - On click, refreshes the URL auth token and opens the PDF in a new tab via `window.open()` using the authenticated `/api/fs/raw` URL
  - Desktop (Electron) path preserved for future native handling

### DIFF
--- a/packages/ui/src/components/views/FilesView.tsx
+++ b/packages/ui/src/components/views/FilesView.tsx
@@ -35,7 +35,7 @@ import { useDebouncedValue } from '@/hooks/useDebouncedValue';
 import { useFileSearchStore } from '@/stores/useFileSearchStore';
 import { useDeviceInfo } from '@/lib/device';
 import { cn, getModifierLabel, getRevealLabelKey, hasModifier } from '@/lib/utils';
-import { getLanguageFromExtension, getImageMimeType, isImageFile } from '@/lib/toolHelpers';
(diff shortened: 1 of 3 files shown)


btriapitsyn:
Thanks for the PR! We ended up solving this a bit differently in `main`.

Instead of opening PDFs in a new browser tab, OpenChamber now previews PDFs inline inside the file viewer using the existing authenticated `/api/fs/raw` path. This keeps the user in the app, avoids popup-blocker issues from async `window.open()`, and preserves outside-workspace grant handling.

The useful server-side part of this PR, serving `.pdf` as `application/pdf`, was included as part of that implementation.

Implemented in: f9ce2ae8

PR #1548, opened 2026-06-06
fix: prevent orphaned OpenCode processes on server exit

nerdosaurus:
## Problem

When the OpenChamber dev server or Express server is killed (Ctrl+C, SIGTERM, nodemon restart), the auto-started `opencode serve` child process is orphaned. These accumulate on every restart:

- Each `bun run dev` spawns a new `opencode serve` instance
- A single dev session can leave 3-4 orphans
- They consume memory and port resources indefinitely

## Root cause

The graceful shutdown code already existed (`gracefulShutdown` stops OpenCode child, releases port), but it was never wired to process signals. It was only callable via the API or CLI.

A previous attempt in `dev-web-hmr.mjs` used a raw `pkill -f "opencode serve"` on shutdown, but that was both lost in the rebase and too broad (it would also kill the user's external server on port 4096).

## Fix

1. **Express server** (`index.js`): Register `process.on('SIGTERM'/'SIGINT'/'SIGHUP')` handlers that call `gracefulShutdown({ exitProcess: true })`, which properly stops the OpenCode child process and closes the HTTP server.

2. **Dev script** (`dev-web-hmr.mjs`): Restore the `pkill` fallback scoped to `--hostname 127.0.0.1` (auto-started instances only), so the external 4096 server at `0.0.0.0` is never touched.

### DIFF
--- a/packages/web/server/index.js
+++ b/packages/web/server/index.js
@@ -1624,6 +1624,16 @@ async function main(options = {}) {
   }, 60_000);
   relayReconcileTimer.unref?.();
 
+  // Wire process signals to graceful shutdown so the auto-started OpenCode
+  // child process is killed when the server exits (Ctrl+C, SIGTERM, etc.).
+  // The existing attachProcessHandlers in server-startup-runtime.js covers
+  // SIGTERM/SIGINT/SIGQUIT when attachSignals is enabled (CLI mode), but not
+  // SIGHUP. Register SIGHUP here unconditionally since no other handler covers it.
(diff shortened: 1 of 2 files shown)


nerdosaurus:
Fixed all issues raised by review:

1. Moved signal handlers before the `return` statement (was dead code)
2. Removed redundant SIGTERM/SIGINT (existing `attachProcessHandlers` covers them); kept only SIGHUP which was not covered
3. Removed `typeof process === "undefined"` guard
4. Removed redundant `await import("child_process")` in dev script (`spawnSync` already imported)

PR #1555, opened 2026-06-06
fix: prevent cascade rollback from restoring deleted session descendants

panzeyu2013:
The OpenCode server cascade-deletes all child sessions when a parent is removed (`session.remove` recurses into children). The client was sending individual `DELETE` requests for every descendant alongside the parent, creating a race where descendants returned 404 (already deleted by the server cascade). The 404 triggered the rollback path in `deleteSessionAction`, which restored the deleted sessions back into the global store — making them appear as "failed to delete" in the UI while actually gone on the server.

### Changes

- **`useSessionActions.ts`** — When hard-deleting a session that has descendants, only send the delete request for the root session. The server cascade handles the rest. Previously, every descendant was sent as a separate request, each hitting 404 after the parent's cascade removed them.

- **`session-actions.ts`** — In both `deleteSession` and `deleteSessionInDirectory`, treat HTTP 404 as success in the catch block. This acts as a safety net for any remaining code paths that may still send individual descendant deletes (e.g., the sidebar bulk action bar when a parent and child are both checkbox-selected).

### Verification

- `bun run lint` — passes

### DIFF
--- a/packages/ui/src/components/session/sidebar/hooks/useSessionActions.ts
+++ b/packages/ui/src/components/session/sidebar/hooks/useSessionActions.ts
@@ -214,16 +214,18 @@ export const useSessionActions = (args: Args) => {
 
       const ids = [session.id, ...descendantIds];
       if (shouldHardDelete) {
-        const { deletedIds, failedIds } = await args.deleteSessions(ids);
-        if (deletedIds.length > 0) {
-          toast.success(deletedIds.length === 1
-            ? t('sessions.sidebar.bulkActions.deletedSingle', { count: deletedIds.length })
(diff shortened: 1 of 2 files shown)


panzeyu2013:
Thanks for the review. Regarding the suggested fix to proactively remove descendant sessions from local stores:

I believe this is unnecessary because the SSE event for cascade deletion will arrive almost immediately after the delete API returns — the server processes the cascade synchronously. Any delay is in the range of hundreds of milliseconds, so the "ghost" descendant sessions will be imperceptible to users. Adding an explicit store removal introduces a race condition risk (double-removal when SSE arrives) and expands the change surface beyond the minimal fix. This can be revisited if users actually report visible issues.

btriapitsyn (approved):

PR #1561, opened 2026-06-07
fix: collapse user messages to a one-line plain-text preview

robertoberto:
Fixes collapsed user messages that still render at full height when they
contain block-level markdown content (e.g., `<pre>`, `<ul>`, `<blockquote>`).

The current CSS line-clamp approach silently fails on block descendants.
This builds an explicit one-line plain-text preview from the raw message
content, shows a compact "+N lines" indicator, and keeps the expanded view
unchanged.

Adds i18n strings for collapse/expand labels across 7 locales.

- [ ] 16 unit tests for `userTextPreview`
- [ ] Manual check with user messages containing code blocks, lists


btriapitsyn:
Thanks for the PR and for digging into the `line-clamp` issue with markdown/block elements.

I'm going to close this because we ended up taking a narrower fix for this bug: we kept the existing UX of showing the first two lines of the message, and adjusted the collapsed markdown rendering so code blocks/margins don't break the clamp. That keeps expanded markdown, `collapsibleUserMessages`, skill/agent links, and DOM-based truncation behavior working as before.

This PR has a good idea with the reliable plain-text preview, but it changes the UX model: it shows a one-line summary with `+N lines` instead of the actual first two rendered lines of the message. It also doesn't cover long single-line messages that wrap into multiple visual lines, and it adds a separate markdown-stripping helper, which feels like too much scope for this specific bugfix.

So I'm closing this not because the direction is bad, but because for this particular fix we chose the smaller change with lower regression risk. Thanks again, especially for identifying the root cause around block-level markdown descendants.

robertoberto:
> Thanks for the PR and for digging into the `line-clamp` issue with markdown/block elements.
> 
> I'm going to close this because we ended up taking a narrower fix for this bug: we kept the existing UX of showing the first two lines of the message, and adjusted the collapsed markdown rendering so code blocks/margins don't break the clamp. That keeps expanded markdown, `collapsibleUserMessages`, skill/agent links, and DOM-based truncation behavior working as before.
> 
> This PR has a good idea with the reliable plain-text preview, but it changes the UX model: it shows a one-line summary with `+N lines` instead of the actual first two rendered lines of the message. It also doesn't cover long single-line messages that wrap into multiple visual lines, and it adds a separate markdown-stripping helper, which feels like too much scope for this specific bugfix.
> 
> So I'm closing this not because the direction is bad, but because for this particular fix we chose the smaller change with lower regression risk. Thanks again, especially for identifying the root cause around block-level markdown descendants.

glad it's now resolved, that huge block was annoying me. smaller fix was the right call.

Issue #1519, opened 2026-06-03
[Feature Request] Adding a 'Question List' allows you to quickly view all questions in the session


firesahc:
### What should we add/change?

Add a question list, which can be a sidebar or a standalone window. Display all questions in the question list, and you can jump to the corresponding position in the session by clicking on a question in the list.

### Why is this useful?

When having multiple conversations in a session, the AI's responses can be very long, which makes it very difficult to review the previous messages, and we often need to scroll a long distance on the page. The more conversations there are in a session, the more obvious the problem becomes.

firesahc:
1. Each message is displayed in the list
2. Display a numbered index and a preview of the question
3. a collapsible panel within the chat view is ok，or a sidebar timeline format similar to DeepSeek or Codex

firesahc:
I have noticed that the current software already has the function of navigating to questions via the keyboard. However, I still think that a question list is necessary because it allows for a clear view of the questions. The key point is to first determine the target before making the jump, which is more efficient than filtering the target during the jump process.

firesahc:
Can some fixed replies in the question list be hidden? For example, /compact commands.

A lot of canned replies take up space in the issue list, making the information density of the list really low.

<img width="455" height="413" alt="Image" src="https://github.com/user-attachments/assets/c68e7e63-0cc3-477f-8612-39cf0ef7f0ca" />

PR #1583, opened 2026-06-10
feat: add jump to last user message button in chat

phantomObserver:
## Description
This PR implements a smooth-scrolling **Jump to last user message** button in the chat view. It provides users with a fast, intuitive way to return to their latest prompt from anywhere down the conversation history, streamlining navigation in long threads.

Somewhat addresses #1519

---

### 1. New UI Component (`JumpToPreviousMessageButton.tsx`)
* Added a modern floating action button positioned dynamically to the right of the conversation column.
* Wired to layout states (`isRightSidebarOpen`, `wideChatLayoutEnabled`) from `useUIStore`.
* Uses a robust CSS positioning formula to slide cleanly out of the text area into the margins when space permits, while pushing safely inward when resized:
  `right: max(minPadding, calc((100% - colWidth) / 2 - 1.5rem))`
* Centered tooltip (`side="bottom" align="center"`) beneath the button to match standard UI patterns.

### 2. Precise Scroll Shadow Offset (`MessageList.tsx`)
* Added `getMessageScrollTopOffset` to dynamically query `--scroll-shadow-size` and message padding.
* Replaced the hardcoded `50px` offset so jumped-to messages land perfectly flush below the top gradient shadow.

### 3. Programmatic Jump Guard (`ChatContainer.tsx`)
* Introduced `beginJumpScrollGuard` to suppress passive history preloads during programmatic smooth scrolls.
* Listens for native `scrollend` (with a `3s` timeout safety fallback) before restoring scroll listeners and checking if older segments need to be loaded. This guarantees smooth animations are never interrupted by viewport jump compensations.
* Disables button clicks during active scroll+fetch transitions.

### 4. Scrollbar Visibility Feedback (`OverlayScrollbar.tsx`)
* Added a `forceVisible` prop to the custom scrollbar component to keep the track visible during programmatically driven scrolls.

### 5. Multi-Language Translations
* Localized `'chat.jumpToPreviousMessage.aria'` ("Jump to last user message") across all 8 supported languages (`en`, `es`, `ko`, `pl`, `pt-BR`, `uk`, `zh-CN`, `zh-TW`).

---

## How to Test & Verify

1. **Basic Jump:** Scroll down in a long session. Click the jump button; verify a smooth transition that lands flush below the top shadow.
2. **Layout Adaptability:** Resize the window and toggle the right sidebar. Verify the button glides smoothly and stays visible without off-screen clipping.
3. **Older segment preloading:** Scroll near the top boundary; click jump. Verify smooth scroll completes fully *before* older history is fetched and mounted.
4. **Build & Standards:** Verified `bun run type-check` and `bun run lint` pass successfully.

### DIFF
--- a/packages/ui/src/components/chat/ChatContainer.tsx
+++ b/packages/ui/src/components/chat/ChatContainer.tsx
@@ -694,7 +694,8 @@ export const ChatContainer: React.FC<ChatContainerProps> = ({ autoOpenDraft = tr
         jumpScrollActiveRef.current = false;
         setIsJumpScrollActive(false);
         guard?.clear();
-    }, []);
+        finishJumpNavigation();
+    }, [finishJumpNavigation]);
 
     const beginJumpScrollGuard = React.useCallback(() => {
         clearJumpScrollGuard();
@@ -736,11 +737,13 @@ export const ChatContainer: React.FC<ChatContainerProps> = ({ autoOpenDraft = tr
         };
 
         container?.addEventListener('scrollend', finish, { once: true });
+        // Backup safety timeout for browsers lacking native 'scrollend' support 
+        // (e.g. older Safari/Chrome/Firefox or when scroll is interrupted by user).
         timeoutId = window.setTimeout(finish, 3000);
         jumpScrollGuardRef.current = { clear };
     }, [clearJumpScrollGuard, finishJumpNavigation, scrollRef, timelineController]);
 
-    React.useEffect(() => clearJumpScrollGuard, [clearJumpScrollGuard]);
+    React.useEffect(() => clearJumpScrollGuard, [clearJumpScrollGuard, currentSessionId]);
 
     const isTurnUserMessageAboveViewport = React.useCallback((turnId: string | null) => {
         if (!turnId) {
--- a/packages/ui/src/components/chat/ChatContainer.tsx
+++ b/packages/ui/src/components/chat/ChatContainer.tsx
@@ -688,17 +688,19 @@ export const ChatContainer: React.FC<ChatContainerProps> = ({ autoOpenDraft = tr
         setIsJumpNavigationBusy(false);
     }, []);
 
-    const clearJumpScrollGuard = React.useCallback(() => {
+    const clearJumpScrollGuard = React.useCallback((resetBusy = true) => {
         const guard = jumpScrollGuardRef.current;
         jumpScrollGuardRef.current = null;
         jumpScrollActiveRef.current = false;
         setIsJumpScrollActive(false);
         guard?.clear();
-        finishJumpNavigation();
+        if (resetBusy) {
+            finishJumpNavigation();
+        }
     }, [finishJumpNavigation]);
 
     const beginJumpScrollGuard = React.useCallback(() => {
-        clearJumpScrollGuard();
+        clearJumpScrollGuard(false);
         jumpScrollActiveRef.current = true;
         setIsJumpScrollActive(true);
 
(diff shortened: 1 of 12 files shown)


phantomObserver:
### PR Update & Refinements

Based on review feedback and further polish, the following updates and fixes have been added to this PR:

1. **Robust Session Switch Safety & Cleanup:**
   * **Fixes Stuck State:** Corrected a bug where `clearJumpScrollGuard` failed to reset the busy/navigation flag. It now cleanly calls `finishJumpNavigation()`.
   * **Prevents Cross-Session Leaks:** Bound the cleanup effect in `ChatContainer.tsx` to `currentSessionId`. Switching sessions during an active jump scroll now immediately tears down the active guard, preventing safety timeouts from executing stale `timelineController.handleHistoryScroll()` handlers on the newly loaded session.

2. **Tooltip Alignment & Wording:**
   * Centered the tooltip beneath the button (`side="bottom" align="center"`) to remain consistent with standard desktop app tooltip layouts.
   * Updated the tooltip text from "Jump to user message" to **"Jump to last user message"** as requested, updating localized dictionaries across all 8 supported languages.

3. **Code Quality & Documentation:**
   * Added inline comments clarifying the `scrollend` event's browser compatibility, noting the `3s` timeout safety fallback for older environments.
   * Eliminated redundant container variable shadowing in the `beginJumpScrollGuard` closure.

phantomObserver:
### PR Update: Disabled State & Tooltip Padding Polish

The following additional refinements and bug fixes have been applied:

1. **Fixed Active Scroll Disabled State:**
   * **The Issue:** Introducing the busy state cleanup inside `clearJumpScrollGuard` to prevent session-switch leaks inadvertently caused the button to immediately re-enable at the start of a scroll. This happened because `beginJumpScrollGuard` calls `clearJumpScrollGuard` to reset any stale active scroll guards prior to starting.
   * **The Fix:** Updated `clearJumpScrollGuard` to accept a `resetBusy = true` boolean parameter. When initiating a scroll, `clearJumpScrollGuard(false)` is invoked—cancelling prior listeners but securely preserving the active busy state. Standard cleanup hooks (like unmounting or session-switching) default to `true` to fully reset all navigation states. The button now correctly disables and prevents double-clicks for the entire duration of the smooth scroll and post-arrival history load.

2. **Tooltip Layout Polish:**
   * Reduced the gap between the button and its centered bottom tooltip by half (changing `sideOffset={8}` to `sideOffset={4}`) for a tighter, more cohesive visual appearance.

phantomObserver:
### PR Update: Performance Blocker Fixed & Code Cleanup

The following performance and architectural improvements have been added to this PR:

1. **Performance Blocker Fixed (ChatViewport Memoization Restored):**
   * **The Issue:** `handleChatHistoryScroll` previously referenced the entire `timelineController` object returned by `useChatTimelineController()`. Because that custom hook returns a fresh object literal on every single render of `ChatContainer`, `handleChatHistoryScroll` was re-instantiated on every single token chunk event during streaming (60/sec). This defeated `ChatViewport`'s `React.memo` comparator (`prev.handleHistoryScroll === next.handleHistoryScroll`), forcing unnecessary and expensive viewport re-renders.
   * **The Fix:** Extructured `const handleHistoryScroll = timelineController.handleHistoryScroll` at the component scope level and referenced only this stable, single-reference callback in the dependency arrays for both `beginJumpScrollGuard` and `handleChatHistoryScroll`. This satisfies ESLint's static property access rules while restoring `ChatViewport`'s memoization to 100% efficiency, stopping all streaming-related re-render fanouts.

2. **Removed Redundant Code in Navigation Error Paths:**
   * Cleaned up explicit `finishJumpNavigation()` calls inside `handleJumpToPreviousMessage`'s `try/catch` block. Since `clearJumpScrollGuard()` (using default parameters) already resets the busy navigation flags internally, the secondary calls were redundant dead code.

3. **Standardized History Load Threshold:**
   * Aligned the `HISTORY_SCROLL_THRESHOLD` inside the jump guard's completion callback from `100` to the canonical `200` to maintain perfect consistency with `useChatTimelineController.ts:66` for triggering older history loads.

phantomObserver:
Shutting down this pull request as it has been further improved into a more comprehensive navigation overhaul, will create a new pull request with specific details. I'm finding it much easier to navigate with the changes I've made.

PR #1591, opened 2026-06-10
feat: improve mobile UX

btriapitsyn:
## Summary
This improves the mobile experience across the main chat shell, usage surfaces, MCP access, and OpenChamber settings so the app feels usable on touch-first screens instead of only being a compressed desktop layout.

## What changed
- Added a mobile MCP overlay so MCP tools can be opened and managed from the mobile UI without relying on desktop-only dropdown behavior.
- Improved mobile session panel touch handling so tapping the status/session area opens the right panel reliably on phones and tablets.
- Cleaned up mobile usage provider metadata by removing duplicate rows, hiding unset providers, and showing provider logos consistently.
- Added eager loading for provider logos used in mobile usage views to avoid delayed or missing icons when the panel opens.
- Refined the mobile update and about flows in OpenChamber settings so release/update information is easier to read on small screens.
- Adjusted related layout, header, VS Code layout, command palette, and settings text/localization details needed for the mobile polish.

## Why
Mobile users need the same core controls that desktop users have, but several flows were either hard to reach, visually noisy, or depended on desktop interaction patterns. These changes make the mobile UI more direct, reduce confusing empty/duplicate usage data, and keep key settings and update information readable on small screens.

## Testing
- bun run type-check
- bun run lint


btriapitsyn:
@openchamber-bot help

PR #1614, opened 2026-06-11
fix: stabilize question custom textarea resizing

### LINKED ISSUE #1613: [Bug] Question custom-answer textarea flickers on every keystroke when chat context is long
## Description

When answering a `question` tool prompt with the custom answer option, the textarea starts flickering on every keystroke once the input grows beyond roughly two lines. The issue is much more noticeable when the current session has a long chat history / large context.

This makes typing custom answers uncomfortable because the input area visibly flashes or jumps after each character.

## Steps To Reproduce

1. Open a session with a long chat history / large rendered context.
2. Trigger a `question` tool prompt.
3. Select the custom answer option.
4. Type enough text to exceed two lines.
5. Continue typing one character at a time.

## Actual Behavior

The custom answer textarea flickers or visibly reflows on every keystroke after the content exceeds around two lines.

## Expected Behavior

Typing in the custom answer textarea should be stable and should not visibly flicker or jump, regardless of chat history length.

## Suspected Cause

The custom answer textarea appears to resize synchronously on every `onChange` by setting height to `auto`, reading `scrollHeight`, then writing the height back. Once the content exceeds two lines, `scrollHeight` changes frequently, causing forced synchronous layout/reflow.

Because the question card is rendered inside the main chat scroll container, a long chat history makes this reflow more expensive and visible.

A likely fix would be to isolate the textarea into a smaller memoized component and avoid layout thrashing during every input event, for example by moving sizing logic into a layout effect and only updating height when the computed target height actually changes.

## Environment

- App: OpenChamber / OpenCode UI
- Repro condition: long chat context, `question` tool custom answer input

### Where does it happen?

Desktop (macOS)

### Version (if known)

1.12.4

https://github.com/user-attachments/assets/8ad65e9b-70e6-487b-805f-bf6810e1fd27

### Logs (optional)

```shell

```

bigcoder84:
## Summary

Fixes #1613 

This MR fixes the custom-answer textarea in `QuestionCard` so it no longer flickers on every keystroke in long chat sessions, and so long custom answers remain readable while typing.

## Background

When an agent uses the `question` tool, users can choose predefined options or provide a custom answer. In long chat sessions, typing a custom answer became visually unstable once the answer grew beyond a couple of lines.

This was especially noticeable when the chat history was large, because each keystroke could trigger expensive layout work inside the main chat scroll container.

## Problem

The previous custom-answer textarea implementation resized itself directly inside the render ref and `onChange` handler:

- Set textarea height to `auto`
- Read `scrollHeight`
- Write the new height back
- Update parent `QuestionCard` state on every keystroke

This caused two issues:

- The textarea could flicker/reflow on every keystroke in long sessions.
- The textarea was capped at a small height and used hidden overflow, so longer answers could hide earlier text and feel difficult to edit.

## Changes

This MR updates the custom-answer textarea behavior by:

- Extracting the custom answer input into a memoized `CustomAnswerTextarea` component.
- Keeping per-keystroke input state local to the textarea component.
- Storing custom answer text in a ref at the parent level, while only updating parent state when the answer changes between empty and non-empty.
- Moving textarea height measurement into a layout effect.
- Adding `getQuestionCustomTextareaHeight` to clamp textarea height consistently.
- Increasing the max auto-resize height from 4 lines to 10 lines.
- Allowing vertical scrolling once the content exceeds the max textarea height.
- Adding regression tests for the textarea sizing behavior.

## Behavior After Fix

After this change:

- Typing in the custom-answer textarea no longer flickers in long chat sessions.
- The textarea grows automatically from 2 lines up to 10 lines.
- Very long answers remain accessible through internal textarea scrolling.
- The parent question card avoids unnecessary full re-renders on every keystroke.
- Submit validation still works: empty custom answers remain incomplete, non-empty custom answers can be submitted normally.

## Before Recording

https://github.com/user-attachments/assets/07a4aaa7-39c7-4930-bc19-4bcc0f145d02

## After Recording

https://github.com/user-attachments/assets/2e38ac91-01cb-4d18-a51e-353460bebbfc

## Testing

```bash
bun test packages/ui/src/components/chat/__tests__/questionTextareaSizing.test.ts
bun run lint
bun run type-check

### DIFF
--- a/packages/ui/src/components/chat/QuestionCard.tsx
+++ b/packages/ui/src/components/chat/QuestionCard.tsx
@@ -14,7 +14,7 @@ import { useSessions } from '@/sync/sync-context';
 import * as sessionActions from '@/sync/session-actions';
 import { useI18n } from '@/lib/i18n';
 import { serializeQuestionAsJson, serializeQuestionAsMarkdown } from './questionSerializers';
-import { getQuestionCustomTextareaHeight } from './questionTextareaSizing';
+import { QUESTION_CUSTOM_TEXTAREA_MIN_HEIGHT, getQuestionCustomTextareaHeight } from './questionTextareaSizing';
 
 interface QuestionCardProps {
   question: QuestionRequest;
@@ -40,7 +40,7 @@ const CustomAnswerTextarea = React.memo(function CustomAnswerTextarea({
 }: CustomAnswerTextareaProps) {
   const textareaRef = React.useRef<HTMLTextAreaElement | null>(null);
   const [localValue, setLocalValue] = React.useState(value);
-  const [height, setHeight] = React.useState(40);
+  const [height, setHeight] = React.useState(QUESTION_CUSTOM_TEXTAREA_MIN_HEIGHT);
   const [isScrollable, setIsScrollable] = React.useState(false);
 
   React.useEffect(() => {
@@ -231,6 +231,8 @@ export const QuestionCard: React.FC<QuestionCardProps> = ({ question }) => {
   const handleSelectCustom = React.useCallback(() => {
     setCustomMode((prev) => ({ ...prev, [activeIndex]: true }));
     setSelectedOptions((prev) => ({ ...prev, [activeIndex]: [] }));
+    const hasValue = (customTextRef.current[activeIndex] ?? '').trim().length > 0;
+    setCustomTextFilled((prev) => (prev[activeIndex] === hasValue ? prev : { ...prev, [activeIndex]: hasValue }));
   }, [activeIndex]);
 
   const handleCustomValueChange = React.useCallback((value: string) => {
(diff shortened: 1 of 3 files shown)


btriapitsyn:
Addressed the review findings and bot nits.

Changes made:

- Fixed the custom-answer state bug when switching from `Other` to a predefined option and back to `Other`: the previously typed custom text is preserved and `customTextFilled` is restored from the saved ref value, so submit validation matches what the user sees.
- Replaced the initial textarea height magic number with a shared `QUESTION_CUSTOM_TEXTAREA_MIN_HEIGHT` constant from `questionTextareaSizing.ts`.
- Added a small targeted assertion for the exported initial textarea height.
- Confirmed there were no merge conflicts with current `main`.

Validation run locally:

- `bun test packages/ui/src/components/chat/__tests__/questionTextareaSizing.test.ts` — 3/3 passed
- `bun run type-check`
- `bun run lint`
- `git diff --check`

PR is mergeable; GitHub checks are queued after the push.

btriapitsyn (approved):

PR #1627, opened 2026-06-12
fix: refresh worktree list on dropdown open and window focus

jwcrystal:
## Problem

Worktree list in the UI only refreshes when `projects` array changes (sidebar mount/switch) or after UI-initiated create/remove. External `git worktree add` in the terminal was never detected — no polling, no focus listener, no refresh-on-dropdown-open.

## Solution

Lightweight client-side only approach with three triggers:

1. **Window focus** → `refreshAllProjectWorktrees()` — when the user returns from terminal after `git worktree add`
2. **Dropdown open** → `refreshProjectWorktrees(project)` — when the user opens the branch selector in ChatInput
3. **Exported cache invalidation** — `invalidateWorktreeList()` now public for external use

Implemented with:
- 5s throttle on`refreshAllProjectWorktrees` to prevent burst on rapid focus events
- Single read-modify-write cycle to avoid race conditions across parallel per-project fetches
- Empty worktree entries are cleaned from the store map
- Inflight dedup prevents duplicate requests

## Scope / relation to server-side watching

See [PR #1480](https://github.com/openchamber/openchamber/pull/1480) which takes a heavier server-side approach (`fs.watch` on git dirs + event stream).

This PR intentionally takes a lightweight client-side approach.

The motivation was to improve worktree discovery at common interaction points without introducing additional server-side state, lifecycle management, or file-watching infrastructure.

This is not intended as a replacement for a future watcher/event-based solution. Client-side refreshes and server-side watching can coexist and address different tradeoffs.

If maintainers decide that worktree state should be synchronized automatically under all circumstances, a watcher-based approach may still be appropriate. This PR is intended as an incremental improvement that can remain useful as a fallback even in that model.

## Changes

3 files, +115 lines:
- `packages/ui/src/lib/worktrees/worktreeManager.ts` — export `invalidateWorktreeList`, add `refreshProjectWorktrees` and `refreshAllProjectWorktrees`
- `packages/ui/src/components/chat/ChatInput.tsx` — refresh on branch selector dropdown open
- `packages/ui/src/components/session/SessionSidebar.tsx` — refresh on window focus

## Verification

- `bun run lint` — passed
- `bun run type-check` — no new errors
- Manual code review of all paths, race conditions, throttling, and lifecycle cleanup


colinmollenhour:
> the common workflow is: run git command in terminal → return to OpenChamber → window.focus fires in <1s

I don't really agree with this assumption.. For example, OpenChamber has its own terminal, so what if you use the OpenChamber terminal to run the git command? There is never a blur/focus. Also what if a different user (likely an agent or same user from another device) runs the git command? The point in my opinion should be that the user never, under any circumstance, has to manually refresh something to see new worktrees appear - so client-side triggers alone are not sufficient. Don't get me wrong, they are far better than nothing, but just speaking ideally..

Regarding the `fs.watch` approach, Node already makes this mostly platform agnostic, but it is true that it has some limitations.

1. I wouldn't worry about obscure platforms like AIX and things like NFS.. who runs OpenChamber on projects via NFS? Unless this is stated as a target, I think these don't matter and all that matters is native filesystems on Mac, Windows, WSL and Linux.
2. Is the Windows issue actually an issue? I'd have to inspect the way git works and how that meshes with the Windows limitations but I think it might be a non-issue.. IDEs very commonly use file watching for stuff like this.

jwcrystal:
> > the common workflow is: run git command in terminal → return to OpenChamber → window.focus fires in <1s
> 
> 
> 
> I don't really agree with this assumption.. For example, OpenChamber has its own terminal, so what if you use the OpenChamber terminal to run the git command? There is never a blur/focus. Also what if a different user (likely an agent or same user from another device) runs the git command? The point in my opinion should be that the user never, under any circumstance, has to manually refresh something to see new worktrees appear - so client-side triggers alone are not sufficient. Don't get me wrong, they are far better than nothing, but just speaking ideally..
> 
> 
> 
> Regarding the `fs.watch` approach, Node already makes this mostly platform agnostic, but it is true that it has some limitations.
> 
> 
> 
> 1. I wouldn't worry about obscure platforms like AIX and things like NFS.. who runs OpenChamber on projects via NFS? Unless this is stated as a target, I think these don't matter and all that matters is native filesystems on Mac, Windows, WSL and Linux.
> 
> 2. Is the Windows issue actually an issue? I'd have to inspect the way git works and how that meshes with the Windows limitations but I think it might be a non-issue.. IDEs very commonly use file watching for stuff like this.
> 
> 

Good point.

I agree that client-side triggers alone are not a complete solution if the goal is for worktree changes to appear automatically under all circumstances.

My intent with this PR was much narrower: provide a lightweight UX improvement that covers some common refresh points with minimal complexity, rather than argue against a future watcher/event-based approach.

I’ve updated the PR description to clarify that this should be viewed as complementary rather than competing with a server-side watcher. Even if a watcher is added later, I think refresh-on-focus and refresh-on-open are still useful fallback/recovery paths when events are unavailable, delayed, or missed.

Thanks for calling out the assumption in the original description.

btriapitsyn:

PR #1631, opened 2026-06-12
feat: remote Open in Desktop App, HTML preview, image fix, download hardening

bashrusakh:
## Summary

Three features + multiple security/QoL fixes for FilesView:

### Remote "Open in Desktop App"
- **New IPC handler** `desktop_download_and_open` — downloads file from remote server via `runtimeFetch`, saves to `os.tmpdir()/openchamber-downloads/`, opens with `shell.openPath`
- **22-type extension blocklist** with multi-segment bypass protection (.bat, .exe, .sh, .ps1, .pkg, .dmg, .deb, .rpm, etc.)
- **100MB size limit** with Content-Length + post-arrayBuffer checks
- **crypto randomBytes** suffix for unique temp filenames (no collisions)
- **dialog.showMessageBox** user confirmation before opening
- **One-at-a-time dialog guard** against spam
- **Periodic temp cleanup** — on startup + hourly `setInterval`
- **Fallback chain** in `handleOpenInApp`: local app → local path → remote download → parent directory

### HTML Preview
- **New `/api/fs/serve/:path(*)`** endpoint for path-based file serving (replaces broken `srcDoc` iframe)
- `FILE_MIME_MAP` — `Object.freeze` constant with 28 MIME types, shared between `/api/fs/raw` and `/api/fs/serve`
- **100MB serve limit**, `X-Content-Type-Options: nosniff`, `allowOutsideWorkspace` explicitly rejected
- **URL slash guard** for robust path construction
- **"Open in Browser"** button for HTML files in preview mode (with auth URL + remote-aware fallback)
- Default `htmlViewMode` → `'preview'`

### PNG Preview Fix
- Replaced `authenticatedAsset` URL with `runtimeFetch` + blob + `URL.createObjectURL` — avoids `oc_url_token` race condition
- Proper blob URL cleanup: pre-branch revocation, cancelled-aware IIFE, unmount cleanup, early-return revocation
- `allowOutsideWorkspace` parameter in `runtimeFetch` query

### Download Fix
- `downloadFile` uses `runtimeFetch` + blob instead of unauth `<a>` navigation (eliminates "permission required" dialog)
- `setTimeout` revokeObjectURL (100ms) to prevent download race
- All 3 download callers have `.catch()` with `console.error` + toast

### Code Review
- 3 rounds of OCR review: **37 issues found, 35 fixed, 2 accepted risks**
- All packages pass `type-check` and `lint`

### Files Changed
| File | Changes |
|------|---------|
| `packages/electron/main.mjs` | IPC handler, COMMANDS_SAFE_FOR_REMOTE, cleanup |
| `packages/ui/src/lib/desktop.ts` | `openRemoteFileInApp()` |
| `packages/ui/src/components/views/FilesView.tsx` | HTML preview, image fix, remote open, download |
| `packages/ui/src/components/layout/SidebarFilesTree.tsx` | download .catch() |
| `packages/web/server/lib/fs/routes.js` | /api/fs/serve, FILE_MIME_MAP, size limits, nosniff |
| `packages/web/server/lib/ui-auth/ui-auth.js` | URL auth whitelist |
| `packages/web/src/api/files.ts` | downloadFile runtimeFetch+blob |
| `packages/ui/src/lib/i18n/messages/*.ts` (9 files) | i18n keys |

### DIFF
--- a/packages/electron/main.mjs
+++ b/packages/electron/main.mjs
@@ -4081,6 +4081,7 @@ const BLOCKED_EXTENSIONS = new Set([
   '.command', '.applescript', '.scpt',
   '.jar', '.jnlp',
   '.reg', '.inf',
+  '.hta', '.js',
   '.pkg', '.dmg', '.deb', '.rpm',
 ]);
 
--- a/packages/electron/main.mjs
+++ b/packages/electron/main.mjs
@@ -7,6 +7,7 @@ import fsp from 'node:fs/promises';
 import os from 'node:os';
 import path from 'node:path';
 import { execFile, spawn, spawnSync } from 'node:child_process';
+import crypto from 'node:crypto';
 import { fileURLToPath, pathToFileURL } from 'node:url';
 import { promisify } from 'node:util';
 import updaterPkg from 'electron-updater';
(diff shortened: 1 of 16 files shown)


bashrusakh (commented):
**Review summary response:**

**1. Referrer-Policy — false positive.** The `Referrer-Policy: no-referrer` header is preserved in `/api/fs/raw` at lines 880-882 (resolved during rebase conflict resolution — kept both `X-Content-Type-Options: nosniff` AND `Referrer-Policy`).

**2. App selection — by design.** The user explicitly chose `shell.openPath` (default OS app) for remote downloads. Windows `shell.openPath` has no app-verb support; macOS could use `spawn('open', ['-a', appName])` but this was explicitly scoped to default handler. Dialog text already reflects this: "Open with default ... application?"

**3. .hta + .js — fixed.** Added to BLOCKED_EXTENSIONS in 75fb635.

btriapitsyn:
Thanks for putting this together. The PR surfaced several useful issues and ideas around FilesView previews/downloads, especially authenticated downloads, HTML preview serving, and image preview auth races.

We are not going to merge this PR as-is because it combines several unrelated changes with a security-sensitive Electron flow: downloading bytes from a remote runtime and opening them through the local OS default application. We do not want to introduce a remote-runtime-to-local-OS-open bridge, even with extension blocklists and confirmation dialogs. The PR also had a few correctness issues during review, including a committed conflict marker and an auth/path handling issue around the new browser-consumed route.

Instead, we landed the safe subset directly on `main` in:

https://github.com/openchamber/openchamber/commit/f36b8b72227fb9a97ba5b4ee6eb9f379ada72d3d

That commit includes:
- authenticated downloads via `runtimeFetch` + blob;
- HTML preview through a workspace-scoped `/api/fs/serve` endpoint;
- URL-token auth allowlist coverage for that endpoint;
- a fix for mounted `/api` auth path handling;
- desktop image preview via blob URLs with cleanup;
- a stale selected-file fix for FilesView tabs/previews.

We intentionally did not include the remote “Open in Desktop App” flow or any Electron IPC for opening remote files locally.

Thanks again for the direction here. The useful parts helped shape the smaller, safer fix.

Issue #1387, opened 2026-05-23
[Feature Request] Show cost along with context consumption


abrouaux:
### What should we add/change?

I pay on demand. It's easy to forget a session gets too long, the only way to know how much it's costing is to open the context summary, which takes up valuable space.

charpeni:
This is the same thing I was looking for. One specific placement that would be useful: show the current estimated session cost next to the existing context percentage in the top bar, e.g. `13.8% · $0.42`.

If there is no associated cost metadata, or the computed cost is `0`, I would skip rendering it to avoid noisy UI.

That would make cost visible during normal chat without needing to open the context summary.

yulia-ivashko:
Shipped in v1.23.0 (#2776). The work-status panel beside the transcript on desktop now shows the session cost right next to the context percentage, with a breakdown of what the session and its subagents spent underneath. Thanks for the specific placement — that is the shape it took. It is desktop-only for now; mobile and the VS Code panel do not show the panel.

PR #1635, opened 2026-06-13
feat: show session cost in desktop header

charpeni:
## Summary

Show actual session cost beside desktop header context usage:

Screenshot 2026-06-12 at 9 59 23 PM

This gives users the same cost visibility they already get in the context panel, without making them open the panel just to check spend.

The header reuses the existing actual assistant-message cost metadata used by the context panel. It only shows cost when real positive cost data is present, so free or subscription-based sessions remain uncluttered and we avoid showing estimates.

Related to: https://github.com/openchamber/openchamber/issues/1387.

## Testing

- `bun test "packages/ui/src/stores/utils/costUtils.test.ts"`
- `bun run type-check`
- `bun run lint`


btriapitsyn:

Issue #636, opened 2026-03-10
[Bug] Thinking variant resets to default within the same session


eengad:
Selecting a thinking variant (e.g. `xhigh`) reverts to `default` during the same session without any manual change.

**Version:** OpenChamber v1.8.5 (latest)

**Steps to reproduce:**
1. Select a model (e.g. openai/gpt-5.4)
2. Change thinking variant to `xhigh`
3. Send a message
4. Variant resets to `default`

**Expected:** Variant stays on `xhigh` until manually changed.

**Note:** This may be a regression of #492.

PR #1639, opened 2026-06-13
fix: preserve settings default thinking variant when switching agents

weixiang1862:
# What

When a user sets a default thinking variant (e.g. 'high') in settings and switches between plan and build agents in a session, the variant was reset to 'default' (undefined) instead of respecting the settings default.

# Why
Root cause: two code paths failed to fall back to settingsDefaultVariant:

1. ModelControls variant sync effect: when no per-session+agent+model variant was saved, the effect set currentVariant to undefined instead of falling back to settingsDefaultVariant.

2. setAgent in useConfigStore: when the target agent had a configured model, the variant was always passed as undefined to applyResolvedModelSelection, ignoring both the saved per-session variant and the settings default.

Fix both paths to resolve variants in priority order: saved variant > settingsDefaultVariant > undefined.

# May releated issue
[[Bug] Thinking variant resets to default within the same session](https://github.com/openchamber/openchamber/issues/636)


btriapitsyn:
Fixed the review findings and conflicts in the PR branch.

What changed:
- Merged latest `origin/main` into the PR branch; no manual conflict resolution was needed.
- Centralized `setAgent` variant resolution: saved per-session variant → agent configured variant → settings default variant → `undefined`.
- Applied the same fallback to the saved session-agent-model path.
- Added targeted `useConfigStore` tests for configured-agent default variant, saved/agent/settings priority, and saved session-agent-model fallback.

Validation run:
- `bun test packages/ui/src/stores/useConfigStore.test.ts`
- `git diff --check`

btriapitsyn:
Follow-up after checking historical sessions restore logic:

- Verified `ModelControls` already waits for an existing session to become renderable via `sync.ensureSessionRenderable(...)` before falling back to default agent/model selection.
- Verified it reads the latest loaded user message and restores `agent`, `providerID`, `modelID`, and `variant` from history when no per-session override exists.
- Fixed the missing last step: the historical `variant` is now validated against the model variants and applied immediately through `applyModelSelectionWithVariant(...)`, instead of only being written to the selection store. This prevents settings default variant from temporarily or permanently winning when the historical session model is already selected.

Additional validation run:
- `bun test packages/ui/src/stores/useConfigStore.test.ts`
- `git diff --check`

btriapitsyn (approved):

PR #1648, opened 2026-06-14
feat: add provider model context limits

### LINKED ISSUE #1624: [Feature Request] Surface per-model context window caps in OpenChamber
I'm trying to intentionally limit the context window used by large-context models from the OpenChamber UI.

This is already possible by editing OpenCode configuration directly, for example by overriding custom provider model metadata with a lower `limit.context`. However, OpenChamber currently does not appear to expose this in the UI.

Some models expose very large context windows, such as ~1M tokens, but I often do not want OpenChamber/OpenCode to use that much context. I would like to cap models like Opus 4.8 or GPT 5.5 around 200k tokens instead.

It would be great if OpenChamber allowed configuring a per-model effective context limit, separate from the provider/model's advertised maximum context window.

Use cases:

- Keep sessions cheaper and more predictable.
- Avoid very large prompts when I do not need the full model context.
- Trigger compaction earlier at a user-defined threshold.
- Use the same model differently across workflows, e.g. 200k for normal work and full context only when explicitly needed.
- Avoid manually editing `opencode.json` for a common provider/model setting.

Ideally this could be configured in the model/provider settings UI, for example:

- Provider: Anthropic
- Model: Opus 4.8
- Advertised context: 1M
- Effective context cap: 200k

OpenChamber should then write the corresponding OpenCode config, reload OpenCode if needed, and use the effective cap for context usage display and compaction decisions.

charpeni:
## Summary

Closes #1624.

Adds user-scoped per-model context limits in Provider settings. The UI now shows each model's advertised context window by default and only opens a custom limit input when the user chooses `Set custom limit` or `Edit`.

Setting a custom limit writes an `OpenCode` model context override, which makes `OpenCode` compact earlier so sessions avoid exceeding that configured limit.

## Changes

- Added `GET /api/provider/:providerId/model-context-limits`.
- Added `PATCH /api/provider/:providerId/model-context-limit`.
- Persists custom caps to the user `OpenCode` config as `provider.<providerId>.models.<modelId>.limit.context`.
- Enforces that custom caps cannot exceed the advertised model context limit.
- Resolves advertised max from `models.dev` metadata first, with live `OpenCode` metadata as fallback.
- Added VS Code bridge/webview parity for the new provider limit routes.
- Updated Provider settings UX:
  - Default state shows advertised/model max and `Using model max`.
  - Custom state shows `{tokens} custom`.
  - Inline editor appears only when adding/editing a limit.
  - `Use model max` clears a custom override.
- Added locale keys and Settings search coverage for model context limits.
- Added focused provider config tests for reading, writing, clearing, and validating custom limits.

## Validation

- `bun run type-check`
- `bun run lint`
- `bunx vitest run packages/web/server/lib/opencode/providers.test.js`
- `bun run build:web`

## Manual QA Notes

Manual QA should verify:
- Provider model rows do not show inputs by default.
- `Set custom limit` opens the inline editor.
- Saving `200000`, `200k`, `1m`, and `1.5m` works when within max.
- Values above the advertised max are rejected.
- `Use model max` clears the override.
- Reloading preserves saved custom limits.


btriapitsyn:

PR #1668, opened 2026-06-15
Fix session unarchive flow

### LINKED ISSUE #1463: [Bug] There is no option to cancel archiving

As shown in the image below, there is no option to unarchive this session. How can I fix this?

Image

> I accidentally archived my session, but I can't find an option to unarchive it. How can I undo this?

### Where does it happen?

Mobile (Web/PWA)

### Version (if known)

1.10.4

_No response_

### Logs (optional)

```shell

```

charpeni:
## Summary

Fixes #1463

Archived sessions could get stuck in the archived list with no reliable way to restore them. OpenCode session update API accepts archive timestamps but does not currently clear `time_archived` when receiving `{ time: { archived: null } }`, so the UI could send an unarchive request while the server still classified the session as archived.

This PR adds a reliable unarchive flow from the archived-session sidebar context menu. While unarchive is pending, the archived row stays visible with a spinner and disabled menu item. Success feedback only appears after the server confirms the session is no longer archived.

## What Changed

- Added `unarchiveSession` / `unarchiveSessions` sync actions.
- Added global session store support for moving sessions from archived back to active.
- Added an archived-session context-menu action for unarchive.
- Added row-level pending UI for unarchive operations.
- Added localized unarchive strings for all supported locales.
- Updated `opencodeClient.updateSession()` to preserve `{ time: { archived: null } }` in the SDK payload.
- Added server compatibility handling for exact `{ time: { archived: null } }` session patches.
- Hardened session-list proxy filtering so `archived=true` / `archived=false` responses match actual `time.archived` state and preserve pagination.
- Added JSON body parsing for `/api/session` patch routes.

## Notes

The server compatibility route clears `session.time_archived` directly in `opencode.db`, then fetches the session through OpenCode for the normal response shape.

This follows existing precedent in the repo: `syncSandboxesToOpenCodeDb()` already writes directly to `opencode.db` for project sandbox syncing.

## Testing

Passed:

- `bun test packages/web/server/lib/opencode/proxy.test.js`
- `bun test packages/web/server/lib/opencode/core-routes.test.js`
- `bun test packages/ui/src/lib/opencode/client.test.ts`
- `bun test packages/ui/src/stores/useGlobalSessionsStore.test.ts`
- `bun test packages/ui/src/sync/session-actions.test.ts`
- `bun test packages/ui/src/lib/i18n/messages.test.ts`
- `bun run type-check`
- `bun run lint`
- `git diff --check`

Manual QA:

- Verified archived sessions appear only in archived results after archive.
- Verified unarchived sessions return to active results after unarchive.
- Verified archived list pagination returns only archived rows.
- Verified pending spinner remains visible until server-confirmed unarchive.


tomzx:
@openchamber-bot review

tomzx:
@openchamber-bot review

PR #1673, opened 2026-06-16
fix: handle non-ISO-8859-1 characters in fetch headers and Content-Disposition

FanFan4204:
## Problem

Opening a project with a Chinese path like `D:\文件` crashes in the browser:

```
TypeError: Failed to construct 'Headers': String contains non ISO-8859-1 code point
```

## Why

`new Headers()` only accepts ISO-8859-1. The custom `x-opencode-directory` header passes raw filesystem paths — any character above U+00FF breaks it. Separately, `Content-Disposition` doesn't follow RFC 5987 for non-ASCII filenames, so downloads of Chinese-named files produce garbled names.

### `packages/ui/src/lib/runtime-fetch.ts`
- `isLatin1Safe()` and `sanitizeHeadersForBrowser()`: URL-encode non-Latin-1 header values before `new Headers()`
- `mergeHeaders()` calls sanitize for all `HeadersInit` types (object, array, `Headers`)

### `packages/vscode/webview/main.tsx`
- `headersToRecord()` applies same sanitization
- `getRequestDirectoryHint()` uses `decodeURIComponent()` when reading the header

### `packages/web/server/lib/opencode/project-directory-runtime.js`
- `safeDecodeURIComponent()` helper
- Decode path in `resolveProjectDirectory()` and `resolveOptionalProjectDirectory()` before use

### `packages/web/server/lib/fs/routes.js`
- `Content-Disposition` uses `filename*=UTF-8''` encoding (RFC 5987) for non-ASCII filenames

## Tests

All existing tests pass. Type-check and lint clean across 4 packages. Manually tested file browsing and downloads with Chinese filenames and paths.

### DIFF
--- a/packages/ui/src/lib/runtime-fetch.test.ts
+++ b/packages/ui/src/lib/runtime-fetch.test.ts
@@ -394,6 +394,15 @@ describe('runtimeFetch header sanitization', () => {
     expect(result).toBeFalsy();
   });
 
+  test('sanitizeHeadersForBrowser always encodes directory hints with marker', () => {
+    const path = 'C:\\work\\foo%20bar';
+    const result = sanitizeHeadersForBrowser({ 'x-opencode-directory': path });
+    expect(result).toBeTruthy();
+    const encoded = Object.fromEntries(result!);
+    expect(encoded['x-opencode-directory']).toBe(encodeURIComponent(path));
+    expect(encoded['x-opencode-directory-encoding']).toBe('uri');
+  });
+
   test('sanitizeHeadersForBrowser returns undefined for empty/undefined input', () => {
     expect(sanitizeHeadersForBrowser(undefined)).toBeFalsy();
     expect(sanitizeHeadersForBrowser({})).toBeFalsy();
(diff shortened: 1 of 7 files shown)


woohahahaaa:
Hi! Ran into this exact issue today on **OpenChamber v1.13.2 / OpenCode v1.17.9** (Windows, both Desktop Electron and CLI `opencode web`) — a project at `D:\projects\测试项目` makes the entire **files** tab fail to load.

This PR looks like exactly the right fix — single defense point in `mergeHeaders()` covering all call sites, plus the server-side decode and the bonus Content-Disposition RFC 5987 fix for non-ASCII filenames. Tests + lint + type-check are clean per the description.

Is there anything blocking this from being reviewed/merged? Happy to help test the RC build if it helps move things forward. Thanks for the work on this!

btriapitsyn:
Пофіксив знайдений regression із percent-decoding у `x-opencode-directory`.

Що змінилось:
- `x-opencode-directory` тепер завжди percent-encodeʼиться клієнтським transport layer.
- Додано marker header `x-opencode-directory-encoding: uri`, щоб сервер декодував тільки значення, які точно закодував наш transport.
- Direct/raw API clients із шляхами типу `foo%20bar` або `foo%25bar` більше не будуть випадково декодитись у інший шлях.
- VS Code webview path також використовує marker-aware decode.
- Додано regression tests для encoded Unicode path і raw percent paths.

Перевірка пройдена:
- `bun test packages/ui/src/lib/runtime-fetch.test.ts`
- `bun test packages/web/server/lib/opencode/project-directory-runtime.test.js`
- `bun test packages/web/server/lib/fs/routes.test.js`
- `bun run type-check`
- `bun run lint`
- `git diff --check`

btriapitsyn (approved):

PR #1674, opened 2026-06-16
perf(right-sidebar): gate live effects, memoize lookups, always-mount tabs

bashrusakh:
### P0 — Correctness / leak fixes
- [x] RightSidebar: drop dead `useEffect`, collapse `width`/`minWidth`/`maxWidth` triple
- [x] useUIStore: clamp `setRightSidebarWidth`, simplify `setRightSidebarOpen`
- [x] RightSidebarTabs / `useRightSidebarGitSync`: gated poll (right tab + main tab + online + visible)
- [x] GitView: commit-files fetch `cancelled` + `Promise.all`
- [x] GitView + `useGitmojiList`: extract gitmoji fetch/cache with module-level inflight + subscribers
- [x] ProjectNotesTodoPanel: debounce cancel on blur, chained `persistProjectData` per project

### P1 — Render fanout
- [x] RightSidebarTabs: always-mounted tabs with `hidden` attribute, redirect effect, type-guarded `onSelect`
- [x] GitView: 13 selectors → one `useShallow` block
- [x] GitView: `isGitViewActive` gates 7 live effects
- [x] GitView: LRU for `gitViewSnapshots` (cap 20), remove dead `unique.set` dedup
- [x] SidebarFilesTree: precomputed `statusByPath`/`badgeByDir` maps, O(1) per-node render
- [x] SidebarFilesTree: `React.memo(FileRow)` + local context-menu state
- [x] SidebarFilesTree: `isCancelled` predicate for `loadDirectory`
- [x] SidebarFilesTree: `fileTreeCacheByRoot` LRU cache (cap 8 roots)

### Review fixes (post-review)
- [x] **Blocker:** `inFlightDirsRef` leak — removed `isCancelled` guard from `finally` (99a6f70)
- [x] **Non-blocker 1:** `useRightSidebarGitSync` poll — already gated on `rightTab === 'git'`, no change needed
- [x] **Non-blocker 2:** duplicate `RIGHT_SIDEBAR_MAX_WIDTH` — exported from `useUIStore`, imported in `RightSidebar` (7bf171e)

---

## Summary

Performance fixes for the right sidebar (git/files/context tabs). The left sidebar already shipped a similar pass — this PR is the right-side counterpart.

### Correctness / leak fixes (P0)

- **RightSidebar**: drop dead `useEffect` that re-nulled refs the resize handler already nulled; collapse redundant `width`/`minWidth`/`maxWidth` triple into `width` + the existing `--oc-right-sidebar-width` variable.
- **useUIStore**: clamp `setRightSidebarWidth` to `[MIN, MAX]`; simplify `setRightSidebarOpen` (22 lines → 12).
- **RightSidebarTabs / `useRightSidebarGitSync`**: now takes right tab + main tab; only polls when the right git tab is the visible consumer AND the browser is online + visible. Replaces a poll that ran for the lifetime of any open sidebar.
- **GitView**: commit-files fetch refactored to `cancelled` + `Promise.all` (was a per-hash loop that could `setState` after unmount); `getRemoteUrl` and `refreshRemotes` gated on `cancelled` / `mountedRef`; new module-scoped `mountedRef` guards `setIsSettingIdentity` from firing after unmount.
- **GitView + new `useGitmojiList` hook**: extract gitmoji fetch/cache into a hook with module-level inflight promise + subscribers Set; stale-while-revalidate from localStorage; `ensureLoaded()` for call-site-initiated hydration; `cancelled` flag on `setIsLoading` to avoid the React setState-after-unmount race.
- **ProjectNotesTodoPanel**: 400 ms notes debounce now cancels on blur (was double-saving); `persistProjectData` chained per project through a module-level `Map<projectId, Promise>` so a fast todo toggle racing the debounced save no longer hits the server in parallel; resize auto-adjust guards against same-value pings.

### Render fanout (P1)

- **RightSidebarTabs**: all three tab content components are now always mounted with the `hidden` attribute. State and cache survive tab switches. When `activeMainTab === 'git'` (or `'context'`) the matching right tab is filtered out of the tab strip and a redirect effect snaps any persisted-but-now-hidden right tab to `'files'`. `onSelect` is now a type-guarded handler instead of `as RightTab`.
- **GitView**: 13 separate `useGitStore` action selectors collapsed into one `useShallow` block (one re-evaluation per store change instead of 13).
- **GitView**: new `isGitViewActive` flag (true when this instance is the visible consumer) gates the 7 live effects — load identities, fetch remote URL, refresh remotes, `ensureAll`, `sessionEvents.onGitRefreshHint`, worktree bootstrap poll, default-identity auto-apply. Hidden GitView instances no longer run these.
- **GitView**: `gitViewSnapshots` module-level Map is now backed by an LRU wrapper (cap 20) so per-directory draft snapshots cannot leak across hundreds of project switches. Removed the dead `unique.set` dedup in `changeEntries` — `GitStatus.files` is already unique by path.
- **SidebarFilesTree**: `statusByPath` `Map<path, FileStatus>` and `badgeByDir` `Map<dirPath, { modified, added }>` are precomputed once per `gitStatus` change. Tree render is O(1) per node instead of O(N) per node via the previous per-row find/scan. `badgeByDir` walks each file's path segments and increments counters for every ancestor dir, so total cost is O(N + total_dirs_in_files) per `gitStatus` change.
- **SidebarFilesTree**: `FileRow` wrapped in `React.memo` with a custom comparator. Context-menu open state moved INTO `FileRow` as local state — opening a menu in one row no longer re-renders siblings.
- **SidebarFilesTree**: `loadDirectory` accepts an `isCancelled` predicate; the batch-load effect for `expandedPaths` passes a stable predicate so per-dir fetches stop touching state once the effect tears down.
- **SidebarFilesTree**: module-level `fileTreeCacheByRoot` Map (LRU, cap 8 roots) hydrates `childrenByDir` / `loadErrorsByDir` / `loadedDirsRef` on mount or root change. Mirror effects write state back to the cache. Survives close-and-reopen of the right sidebar; populated entries are dropped on unmount only when they had no data.

## Result

- 6 files modified, 1 new (`useGitmojiList.ts`), 682 insertions, 283 deletions.
- `bun run type-check` and `bun run lint` clean for staged files.
- 37 related tests pass; full UI test suite preserves the existing baseline (537 pass / 58 fail / 1 error — pre-existing, unrelated to this PR).
- Manual review pass completed: 2 High + 2 Medium + 1 OCR Medium — all addressed before push.

## Architecture notes

- P1.1's redirect effect snaps `rightSidebarTab` to `'files'` whenever `activeMainTab === 'git'`, so the right and main `GitView` instances are mutually exclusive — `isGitViewActive` cannot be true for both.
- The 7 gated effects plus the `useRightSidebarGitSync` poll cover all cases where git state should advance: visible consumer fetches; the poll keeps the store warm when only the right git tab is visible.
- The aborted `loadDirectory` predicate is sufficient because `inFlightDirsRef` and `loadedDirsRef` dedup at the call site before any network IO is initiated.

## Out of scope (deferred to follow-up)

- P2.1-P2.2 virtualization of `SidebarFilesTree` (large refactor; current state already addresses the main bottlenecks).
- P2.3 lazy plan titles (requires `openchamberConfig` schema changes).
- P2.5 extracting the remaining sub-views from `GitView` (large refactor, not perf-critical).

### DIFF
--- a/CHAT_PERF_PLAN.md
+++ b/CHAT_PERF_PLAN.md
@@ -1,0 +1,540 @@
+# Chat Session Performance Plan
+
+## Goal
+
+Reduce render cost and history-load latency when switching between chat sessions
+in `packages/ui/src/components/chat/`, especially on Windows desktop with
+projects that have a long-running chat history (50–500 messages per session,
+tool-heavy workloads where each tool call returns 10–200 KB of output).
+
+The plan is **strictly behavior-preserving**: every change keeps the visible
+behavior 1:1 with the current build. Differences are only in the cost of doing
+the work, not in the UX, layout, limits, persistence, or order of operations.
+
+The two scenarios this targets:
+
+- Switching between sessions in the sidebar (especially in/out of a session
+  with a long history) — the main window currently blocks the main thread
+  for 200–500 ms on first paint while `JSON.parse` + `materializeSessionSnapshots`
+  + `projectTurnRecords` run synchronously over the full initial page.
+- Returning to a session you visited a few minutes ago — the in-memory LRU
+  (`SESSION_MESSAGE_RECORDS_CACHE_MAX = 40` desktop) helps, but the second
+  visit to a session outside that window still pays the full network + parse
+  cost.
+
+A secondary goal is to avoid hammering the OpenCode server with large
+single-shot fetches; we want smaller, more frequent, deduped requests.
+
+## Upstream context
+
+Before implementing, I scanned open/closed PRs for overlap and lessons:
+
+- **#1651 — "perf: migrate chat rendering to virtua"** (`b920fd6f`, MERGED).
+  Already moved chat history to `virtua` virtualization with `bufferSize: 900`
+  and `MESSAGE_LIST_VIRTUALIZE_THRESHOLD = 5`, plus deferred Prism highlight,
+  mermaid/file annotation, table listeners, stable plugin lists. This plan
+  builds on top of that.
+- **#1650 — "perf: instant startup via cache hydration + decoupled readiness"**
+  (`c62f0d1c`, MERGED). Instant startup + decoupled readiness — orthogonal.
+- **#1503 — "fix(chat): do not force scroll in thinking block during streaming"**
+  (OPEN). Scroll behaviour during streaming — not in our hot path.
+- **#1584 — "feat: improve long user message collapse, scrolling, and navigation"**
+  (DRAFT). Touches long user messages; no direct conflict with this plan.
+- **#1621 — "feat: comprehensive scrolling and navigation improvements"**
+  (DRAFT). Broad scroll/navigation PR; no direct conflict.
+- **#1282 — "Optimize large-session sidebar trees, switching UX, and chat
+  scroll restore stability"** (CLOSED by maintainer). Introduced session-
+  switching instability, selection/chat desync, optimistic transitions.
+  Lesson: keep this plan focused on render-cost reduction and stable memoization;
+  do NOT add optimistic transitions, backend search endpoints, scroll-
+  restoration loops, or pagination on top of an unstable base.
+- **#1448 — "feat: implement unified multi-server sidebar"** (DRAFT). Rewrites
+  large parts of `SessionSidebar`. Independent of this plan.
+
+There is currently NO open PR targeting the same render-cost or history-load
+problems for the chat panel. We can proceed without coordination.
+
+## Current Architecture Notes
+
+### Data flow on session switch
+
+1. `ChatContainer` subscribes to `currentSessionId` from `useSessionUIStore`.
+2. `useSessionMessageRecords(currentSessionId, directory)` reads from the
+   directory-scoped sync store via `React.useSyncExternalStore`. Snapshot
+   cache: `SESSION_MESSAGE_RECORDS_CACHE_MAX = 40` desktop, `4` vscode/mobile.
+3. `useChatTimelineController.turnWindowModel` is rebuilt via
+   `buildTurnWindowModel(messages)` on first open of a session; cached per
+   sessionId in module-level `turnModelCache` (max 30 desktop, 4 constrained).
+4. `MessageList.baseDisplayMessages` reverses + dedupes + normalizes each
+   message via `getNormalizedMessageForDisplay` (per-message `WeakMap` cache).
+5. `useTurnRecords` projects turn records via `projectTurnRecords` (O(N) over
+   messages + per-turn work for summary/activity/diffStats/changedFiles).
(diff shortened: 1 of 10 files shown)


bashrusakh:
Re: the `inFlightDirsRef` leak — applied the fix, removed the `isCancelled` guard from `finally` so `inFlightDirsRef` always cleans up. The inflight flag is purely resource-tracking; cancelling should not prevent its cleanup.

Two alternatives considered and rejected:
- **Remove `isCancelled` from `loadDirectory` entirely and have the batch-load effect manage cleanup:** this would move ownership of `inFlightDirsRef` cleanup out of `loadDirectory`, but `loadDirectory` is also called from `refreshDirectory` and `toggleExpandedPath` — every caller would need to remember the cleanup, which is more fragile than keeping it self-contained.
- **AbortController instead of predicate:** would require threading `AbortSignal` through `listDirectory`/`listLocalDirectory`, changing the API contract for all consumers. Correct engineering-wise but disproportionate blast radius for this bugfix. The predicate guards in `try`/`catch` already prevent stale state writes — only the `finally` cleanup was wrong.

bashrusakh:
Re: non-blockers —

1. **`useRightSidebarGitSync` poll gating:** the `shouldPoll` condition already includes `rightTab === 'git'` (line 53 of `RightSidebarTabs.tsx`). The hook does not poll when the active right tab is `files` or `context`. No code change needed — description matches code.

2. **Duplicate `RIGHT_SIDEBAR_MAX_WIDTH`:** fixed in `7bf171e7`. Both constants (`RIGHT_SIDEBAR_MIN_WIDTH`, `RIGHT_SIDEBAR_MAX_WIDTH`) now exported from `useUIStore.ts` and imported in `RightSidebar.tsx`. Single source of truth, no divergence risk.

btriapitsyn:
Fixed the remaining review finding in `9a3eccde`.

What changed:
- `GitView` no longer infers activity from global `activeMainTab` / right-sidebar state.
- Each `GitView` instance now receives explicit `isActive` from its render owner.
- Main Git view, right-sidebar Git tab, and mobile right drawer now gate live effects by actual instance visibility, so hidden always-mounted `GitView` instances do not run polling/fetch/effect work.

Also rechecked prior bot findings:
- `inFlightDirsRef` cleanup fix remains in place.
- right sidebar min/max width constants remain deduped.
- right-sidebar git polling remains gated on visible right Git tab + non-main Git tab.

Local validation passed:
- `bun run type-check`
- `bun run lint`
- `git diff --check`

Pushed directly to the PR branch and left the local repo on `perf-right-sidebar` for artifact testing.

btriapitsyn (approved):

PR #1675, opened 2026-06-16
fix(chat): stabilize session switch to prevent scroll jumps and empty viewports

herjarsa:
## Summary

Fix multiple race conditions in the chat session-switching flow that caused:
- Blank/black viewport flashes on session switch
- Scroll jumping to an intermediate point or staying at the top
- Sessions stuck in an unrenderable state (no messages ever loaded)
- Stale turn/message hash from the previous session stranding the view

## Root causes

1. **ChatViewport `key={currentSessionId}`** forced a full DOM remount on every session switch. The remount gap raced with message delivery, producing transient empty/blank states.
2. **Scroll restore by ratio** ran while the Virtualizer had not yet flushed its measurements. With `scrollHeight === 0`, the ratio math produced `targetTop = 0`, jumping the view to the top.
3. **`effectiveSessionDirectory !== syncDirectory` guard** silently skipped `ensureSessionRenderable`, leaving the session in an unrenderable state when the directory had not yet reconciled after the atomic store update.
4. **`setTimeout(0)` for previous-session anchor save** interleaved with the new session's `restoreSnapshot` effect, racing against the same viewport-store entry.
5. **Stale `window.location.hash`** from a previous session released auto-follow and then failed to scroll to a non-existent turn, stranding the view at an intermediate point.

## Changes

- `packages/ui/src/components/chat/ChatContainer.tsx`
  - Remove `key={currentSessionId}` from `ChatViewport`. The Virtualizer still remounts on `sessionKey` change internally, but the scroll-container DOM is now preserved across session switches.
  - Replace the silent `return` in the `ensureSessionRenderable` effect with a short-interval retry (80ms, capped 2s) that polls `useDirectoryStore.getState().currentDirectory` so the message fetch is no longer skipped.
- `packages/ui/src/hooks/useChatAutoFollow.ts`
  - Add a `scrollHeight` guard to `restoreSnapshot`: when the container's current height is too small to map the saved ratio, defer the restore and let the `ResizeObserver` replay it once the Virtualizer has flushed.
- `packages/ui/src/sync/session-ui-store.ts`
  - Replace `setTimeout(0)` with `queueMicrotask` for the previous-session anchor save. The microtask runs after the current sync call stack but before the next paint / `restoreSnapshot` effect, eliminating the race. Bail if the user has already switched again.
- `packages/ui/src/components/chat/hooks/useChatTurnNavigation.ts`
  - Clear any stale `window.location.hash` on session change so a hash pointing to a turn that does not exist in the new session no longer releases auto-follow and strands the view.

## Verification

- `bun run type-check` clean across all packages
- `bun run lint` clean (one pre-existing warning in `SettingsView.tsx` unrelated to this change)
- Manual reproduction of the original failure (rapid session switching, sessions with many turns, sessions with virtualization enabled) now stays fluid without flashes or scroll jumps.

### DIFF
--- a/packages/ui/src/components/chat/ChatContainer.tsx
+++ b/packages/ui/src/components/chat/ChatContainer.tsx
@@ -34,6 +34,7 @@ import {
 
 // New sync system imports
 import { useSessionUIStore } from '@/sync/session-ui-store';
+import { useDirectoryStore } from '@/stores/useDirectoryStore';
 import { useStreamingStore } from '@/sync/streaming';
 import {
     useSessionMessageCount,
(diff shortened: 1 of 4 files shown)


btriapitsyn:
Thanks a lot for digging into this — the root-cause analysis here is genuinely solid, and every one of the five issues you listed (blank-frame on switch, ratio-restore before the Virtualizer flushes, the silently-skipped renderable fetch, the save/restore race, and the stale hash stranding the view) is real and worth fixing.

We're going to close this PR, though, because a few of the implementation choices introduce new regressions or risk on a code path we just finished stabilizing, and we'd rather address the underlying problems with a different approach:

- The hash clear on session change also wipes valid cross-session deep-links (jump-to-turn / permalinks): the hash is parsed-and-cleared before applyHash() can use it, so a target that *does* exist in the new session never gets applied. We want to clear only when the target can't resolve in the new session.
- The renderable-fetch retry reads hasRenderableSessionSnapshot from a stale closure inside the interval (it never updates), and the real divergence is effectiveSessionDirectory vs the live useDirectoryStore — we'd prefer fixing that source-of-truth and subscribing rather than polling on an interval.
- The restoreSnapshot defer can strand a session that legitimately became shorter (current scrollHeight stays below savedHeight*0.5 indefinitely) and re-fires via the ResizeObserver. Needs a bound/fallback.
- Removing key={currentSessionId} persists the scroll container + ChatViewport effects/refs across sessions — a broad reset-semantics change we don't want to take without proving no cross-session state bleeds; the inner Virtualizer already remounts on its own key.

The queueMicrotask + 'bail if switched again' change for the anchor save is great and we'll likely take that part directly.

We'll track the fixes on our side. Really appreciate the thorough write-up and the repro notes — they made the issues easy to confirm.

PR #1688, opened 2026-06-17
fix(mobile): restore missing settings pages in new layout

lilyzhaun:
## Problem

The new mobile layout (`MobileApp.tsx`) uses a hardcoded `MOBILE_SETTINGS_PAGES` whitelist that only includes 12 settings pages. The old mobile layout (which renders the full `SettingsView` without a whitelist filter) exposes all 17+ pages. This means users on the new mobile layout cannot access:

- **snippets** — code snippet management
- **projects** — project configuration
- **remote-instances** — remote server connections
- **agents** — agent configuration
- **commands** — custom command configuration
- **plugins** — plugin management
- **skills.installed** — installed skills
- **skills.catalog** — skills catalog
- **tunnel** — remote tunnel configuration

## Fix

Add all missing page slugs to `MOBILE_SETTINGS_PAGES`. Each page already has its own `isAvailable` guard in `settings/metadata.ts`, so the whitelist only needs to include them — runtime availability (e.g. excluding `tunnel` and `remote-instances` in VS Code) is handled downstream by `SettingsView`.

`shortcuts` is intentionally excluded — `SettingsView` already filters it out for mobile devices (line 345).

## Changed files

- `packages/ui/src/apps/MobileApp.tsx`: Added 9 missing slugs to `MOBILE_SETTINGS_PAGES`


btriapitsyn:
What a weird PR :)
So, you will manage from a web instance in a mobile view, which is just a UI for web server, not a desktop app or any sort of bundled runtime (with that mobile UI) "remote instances", "tunnels"?
They are hidden since needs to be adopted for mobile, to have a better mobile~friendly UX.
Closing this PR, and will re-introduce missing pages while adopting them for mobile in the same time.

PR #1689, opened 2026-06-17
fix(markdown): restore paragraph block spacing in chat messages

foundryseven:
When the assistant writes more than one paragraph in a single response, the paragraphs are squashed together on screen — the second one starts on the line right after the first, with no gap. It looks like one long paragraph, even though the text underneath is clearly two separate ones.

The giveaway is to copy the message out: paste it into another app and the blank lines between paragraphs are right there. So the source data is fine, the rendering is what's broken.

## Root cause

A missing CSS rule. The design system defines a token called `--markdown-paragraph-spacing` for exactly this purpose, but no rule ever applied it. `.markdown-content` sets font and line-height but nothing about `<p>` margins, so adjacent paragraph blocks collapse to a zero-gap stack. The streaming path renders into the same container, so the bug is visible the whole time a response is coming in, not just at the end.

## Fix

- Wire the existing `--markdown-paragraph-spacing` token to `.markdown-content p` with a bottom margin.
- Bump the token value from `0.35rem` to `0.75rem` so the gap is unambiguous, not just barely-visible.
- No JS, no React, no markdown parser changes. The data path was already correct.

Callers like `MessageBody` and `UserTextPart` already nullify the first and last child's margin through the `[&_.markdown-content>*]` pattern, so adding the rule here doesn't double-up spacing at the top or bottom of a message. Reasoning and tool variants have their own margin overrides that win by specificity, so this rule only affects assistant-text paragraphs.

## Files changed

- `packages/ui/src/index.css` — added `.markdown-content p` margin rule
- `packages/ui/src/styles/design-system.css` — bumped `--markdown-paragraph-spacing` from `0.35rem` to `0.75rem`
- `CHANGELOG.md` — added entry under `[Unreleased]`

3 files, +13/-1.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+- Chat: adjacent paragraphs in assistant messages now render with a visible gap instead of collapsing into a single visual line, matching the spacing users see when copying the text out.
+
 ## [1.13.1] - 2026-06-17
 
 - Chat: inline math delimiters no longer incorrectly treat currency amounts like `$50` as LaTeX math expressions — only `$$...$$` display math and `\(...\)` inline math are recognized.
--- a/packages/ui/src/index.css
+++ b/packages/ui/src/index.css
@@ -932,6 +932,16 @@ html:not(.dark) .chat-scroll {
   line-height: inherit;
 }
 
+/* Restore paragraph block spacing inside markdown content.
+   --markdown-paragraph-spacing is defined in design-system.css but
+   was never wired up, so adjacent <p> blocks collapsed to a single
+   visual line. Callers (MessageBody, UserTextPart) already nullify
+   the first/last child margins via the [&_.markdown-content>*] pattern,
+   so this rule is safe to apply unconditionally. */
+.markdown-content p {
+  margin: 0 0 var(--markdown-paragraph-spacing) 0;
+}
+
 /* Tool markdown should render at code size to match tool card typography. */
 .markdown-content.markdown-tool {
   font-size: var(--text-code) !important;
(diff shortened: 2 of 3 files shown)


btriapitsyn:
Thanks for catching this — the bug is real: the --markdown-paragraph-spacing token was defined but never wired up, so adjacent paragraphs collapsed into one visual line. Good find, and the copy-out repro made it easy to confirm.

We've landed the fix directly. One adjustment from your version: the markdown renderer wraps each block in a display:contents [data-md-block] element, so the message-level [&_.markdown-content>*:last-child]:mb-0 nullifiers actually target the wrapper (where margin is ignored), not the paragraph — meaning the last paragraph's bottom margin wouldn't have been cancelled, adding trailing space under every message. We drop that trailing margin on the last paragraph of the last block directly. We also kept tool-card and reasoning markdown compact (they don't carry their own p-margin overrides, so the unscoped rule would have loosened those too).

Wiring the token + the 0.35rem→0.75rem bump are exactly as you proposed. Really appreciate the clear write-up.

foundryseven:
> Thanks for catching this — the bug is real: the --markdown-paragraph-spacing token was defined but never wired up, so adjacent paragraphs collapsed into one visual line. Good find, and the copy-out repro made it easy to confirm.
> 
> We've landed the fix directly. One adjustment from your version: the markdown renderer wraps each block in a display:contents [data-md-block] element, so the message-level [&_.markdown-content>*:last-child]:mb-0 nullifiers actually target the wrapper (where margin is ignored), not the paragraph — meaning the last paragraph's bottom margin wouldn't have been cancelled, adding trailing space under every message. We drop that trailing margin on the last paragraph of the last block directly. We also kept tool-card and reasoning markdown compact (they don't carry their own p-margin overrides, so the unscoped rule would have loosened those too).
> 
> Wiring the token + the 0.35rem→0.75rem bump are exactly as you proposed. Really appreciate the clear write-up.

My pleasure! Love the project, keep it up.

PR #1694, opened 2026-06-17
fix(files): prevent HTML preview auth token cycling

bashrusakh:
## Summary

The HTML file preview in the right sidebar cycles between showing the page and displaying "authentication required" every ~50 seconds. Image and PDF previews have the exact same issue.

## Root cause

1. URL auth tokens (`oc_url_token`) have a 60-second server TTL with a 10-second client-side skew window, making them effectively valid for ~50 seconds
2. The token was fetched **once** at mount time and never refreshed periodically
3. When the token expired inside the skew window, `getRuntimeUrlAuthTokenSync()` returned `""` and fired a fire-and-forget async refresh — but nothing triggered a re-render when the new token arrived
4. The iframe/img `src` kept the URL without `oc_url_token`, the server returned 401, and the preview showed "authentication required"
5. Eventually an unrelated file-poll re-render (every 2s) would pick up the new token, restoring the preview — until the next expiry, creating the cycling pattern

## Fix

- Added a 45-second `setInterval` in all three asset auth effects (HTML, image, PDF)
- Each tick calls `clearRuntimeUrlAuthToken()` then `refreshRuntimeUrlAuthToken()` to force a fresh token fetch
- A nonce counter is incremented on each successful refresh, used as `key` on the iframe/img to force React to remount with the new authenticated URL
- The image effect previously swallowed errors silently (`.catch(() => {})`) — now reports them properly
- All three effects clear stale `fileError` on successful refresh so transient network errors don't permanently block the preview
- Intervals are properly cleaned up on unmount and file change

## Validation

- `diff` reviewed — only `packages/ui/src/components/views/FilesView.tsx` changed
- Pre-existing type-check errors are unrelated (missing `virtua`, `shiki`, `cron-parser` module declarations)
- No new dependencies or architectural changes

### DIFF
--- a/packages/ui/src/components/views/FilesView.tsx
+++ b/packages/ui/src/components/views/FilesView.tsx
@@ -855,6 +855,8 @@ export const FilesView: React.FC<FilesViewProps> = ({ mode = 'full' }) => {
   const [pdfAssetAuthReadyKey, setPdfAssetAuthReadyKey] = React.useState('');
 
   const [htmlPreviewNonce, setHtmlPreviewNonce] = React.useState(0);
+  const [imagePreviewNonce, setImagePreviewNonce] = React.useState(0);
+  const [pdfPreviewNonce, setPdfPreviewNonce] = React.useState(0);
 
   const [loadedFilePath, setLoadedFilePath] = React.useState<string | null>(null);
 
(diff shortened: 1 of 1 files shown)


btriapitsyn:
Thanks — the diagnosis here is spot on: the oc_url_token has a ~50s effective lifetime, it was only fetched once at mount, and nothing forced a re-render when it expired, so the iframe/img src kept a stale token and 401'd until an unrelated poll re-render picked up a fresh one. Good catch, and the image-effect error-swallowing fix was a nice touch.

We're going to close this and land a centralized version instead, for a couple of reasons:

- oc_url_token is a shared module-level token (FilesView, ContextPanel previews, and the SSE pipeline all use it). Doing clearRuntimeUrlAuthToken() on a 45s interval from FilesView empties the token globally for the duration of the async refetch, so any other consumer that builds a URL in that window gets a tokenless URL and 401s. We'd rather not mutate shared auth state from a feature view.
- The clear-then-refetch leaves a brief empty-token window even for the happy path. A proactive fetch-then-swap (mint the new token, then atomically replace the old one) keeps the previous token valid until the new one lands — no window at all.
- The nonce remount fires unconditionally every 45s, which reloads the iframe even when the token didn't change — that resets in-iframe state for interactive HTML previews. We'll remount only when the token value actually changes.

The fix we're landing: a consumer-gated proactive refresh inside runtime-auth (a single timer that mints a fresh token just before the skew window, only while there's an active url-token consumer), plus a small subscribe hook so previews remount only on a real token change. That fixes the root cause for every consumer at once. Really appreciate the thorough write-up and repro — made this easy to confirm.

PR #1696, opened 2026-06-18
ci: add Build Windows Electron workflow

bashrusakh:
Adds a reusable workflow to build and package OpenChamber for Windows x64 (unsigned), triggered via `workflow_dispatch`.

- Builds on `windows-2022` with Bun + Node.js 22
- Bundles web assets, Electron main process, rebuilds native modules
- Runs `package.mjs --win --x64 --publish=never`
- Uploads `.exe`, `.blockmap`, and `latest.yml` as artifacts

Allows maintainers to produce test builds on demand without a local Windows environment.

### DIFF
--- a/.github/workflows/build-windows-electron.yml
+++ b/.github/workflows/build-windows-electron.yml
@@ -1,0 +1,75 @@
+name: Build Windows Electron
+
+on:
+  workflow_dispatch:
+    inputs:
+      ref:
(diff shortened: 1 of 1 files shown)


bashrusakh:
Bot finding addressed in 24be3e9f: all four actions are now SHA-pinned with `# vX.Y.Z` comments, matching the convention in the other workflows.

PR #1700, opened 2026-06-18
fix(settings): persist per-model visibility and sibling selector state

attilaszasz:
The server-side settings sanitizer only allowlisted favoriteModels and recentModels, so hiddenModels, collapsedModelProviders, recentAgents, and recentEfforts were stripped on every write to settings.json — per-model visibility and collapsed-provider state silently reset on every container redeploy or settings reload.

Add the four missing fields to sanitizeSettingsUpdate:
- hiddenModels: sanitizeModelRefs(..., 1024) — same shape as favoriteModels; 1024 covers dense multi-provider setups while bounding persistence/memory.
- collapsedModelProviders: normalizeStringArray with Array.isArray gate (matches usageDropdownProviders).
- recentAgents: normalizeStringArray (Array<string> per ui-store).
- recentEfforts: new sanitizeRecentEfforts validating Record<string, string[]>
  (shape confirmed in ui-store + addRecentEffort action); trims/dedupes keys and variants, caps at 128 keys x 5 variants/key (5 matches client slice).

No ui-store version bump or migration: zustand's default merge spreads persisted state over defaults, so missing fields fall back to [] / {} until the next toggle. favoriteModels and recentModels are untouched.

Tests: 8 new cases in settings-helpers.test.js using the real sanitizeModelRefs / normalizeStringArray — round-trips, empty-[] parity with favoriteModels, garbage rejection, and a full-payload regression test.


btriapitsyn:
Fixed end-to-end persistence for model selector settings.

What changed:
- Merged the PR branch with the latest `origin/main`; there were no manual conflicts.
- Added `hiddenModels`, `collapsedModelProviders`, `recentAgents`, and `recentEfforts` to the UI `DesktopSettings` contract.
- `sanitizeWebSettings()` now preserves these fields from the server settings response.
- `applyDesktopUiPreferences()` now restores these fields into `useUIStore` during startup/settings sync.
- `modelPrefsAutoSave` now saves the full model selector state: favorites, hidden models, collapsed providers, recent models, recent agents, and recent efforts.
- Added targeted UI tests for server settings → `useUIStore` restore and autosave payload.

Validation passed:
- `bun test packages/web/server/lib/opencode/settings-helpers.test.js`
- `bun test packages/ui/src/lib/persistence.test.ts`
- `bun run type-check`
- `bun run lint`
- `git diff --check`

btriapitsyn (approved):

PR #1703, opened 2026-06-18
Preserve archived session cache entries

### LINKED ISSUE #1577: [Bug] Can not continue or create new session ，after send message，the ai answer just show and disappeared

Openchamber version: 1.12.3（1.12.4 the same problem,1.13.0 also）
OpenCode version: 1.16.2 （1.17.3 the same problem）

### Where does it happen?

VS Code extension

### Version (if known)

1.12.3

after send message,the ui just like below，and models info from ai just disappeared:
**it seems after installed windows desktop.exe that cause the issue**
https://github.com/user-attachments/assets/34e4d99c-f4b1-4ebf-a3fb-695de749223c

Image
### Logs (optional)

```shell

```

Federicorao:
Fixes #1577.

## Summary
- Preserve archived session cache entries during event reducer cleanup.
- Add regression coverage for archived sessions so AI responses remain available after archive transitions.

## Verification
- `bun test packages/ui/src/sync/__tests__/event-reducer.test.ts`

`bun run type-check:ui` was started but stopped after running too long in the local handover environment.

### DIFF
--- a/packages/ui/src/sync/__tests__/event-reducer.test.ts
+++ b/packages/ui/src/sync/__tests__/event-reducer.test.ts
@@ -199,6 +199,50 @@ describe("applyDirectoryEvent", () => {
     expect((draft.session_status.ses_1 as Extract<SessionStatus, { type: "retry" }>).attempt).toBe(2)
   })
 
+  test("archives sessions without dropping materialized messages", () => {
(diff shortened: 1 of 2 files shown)


tomzx:
@openchamber-bot review

PR #1707, opened 2026-06-18
fix(agents): apply agent's pinned model + variant on explicit switch

### LINKED ISSUE #1662: [Bug] Web UI: switching agents doesn't update provider/model display
### What's wrong?

When switching agents in the OpenChamber web UI, the provider and model dropdowns are not updated to reflect the new agent's configured model. The opencode server correctly routes requests using the per-agent model from `opencode.json`, so actual API calls work fine — but the UI displays stale provider/model info.

### Steps to reproduce

1. Configure multiple agents with different models in `opencode.json`
2. Open the OpenChamber web app
3. Switch from one agent to another via the agent selector
4. Observe that the provider/model dropdown still shows the previous agent's values

### Expected behavior

When an agent is selected, the provider and model displayed in the UI should match the `model` field configured for that agent in `opencode.json`.

### Root cause (from analysis)

The web app uses a Zustand store with separate fields: `currentAgentName`, `currentProviderId`, and `currentModelId`. The agent switch handler updates `currentAgentName` but never propagates the new agent's configured `model` to `currentProviderId`/`currentModelId`. The opencode TUI handles this correctly via `item.model ?? prev?.model` logic, but the web UI lacks this linkage.

### Suggested fix

When the agent store's active agent changes, look up the agent's `model` from the fetched OpenCode config and update `currentProviderId`/`currentModelId` accordingly.

### Where does it happen?

Desktop app (Electron) / Web UI

### Where does it happen?

Desktop Web

### Version (if known)

1.13.0

_No response_

### Logs (optional)

```shell

```

bashrusakh:
Fix #1662


btriapitsyn:
Updated the branch to current `main` to clear the stale Node 20 / `better-sqlite3` install failure and re-reviewed the agent-switch behavior:

- the explicit agent switch still applies the agent’s pinned provider/model/variant;
- current config-store and runtime changes merge cleanly;
- verified 18 focused config-store tests, UI type-check, and lint.

btriapitsyn:
Looks like the current main has no such issue. Thank you for your effort

PR #1710, opened 2026-06-18
fix(shortcuts): remove Tab as default agent cycle shortcut

### LINKED ISSUE #1440: Tab completion of slash commands during auto-complete differs from OpenCode

When you tab complete a slash command from the auto-complete selector, it also changes the agent. I'm not sure if this is intentional behavior as it differs from OpenCode. If it's unintentional then I'm happy to submit a PR for it. 

### Where does it happen?

Desktop Web

### Version (if known)

1.11.7

_No response_

### Logs (optional)

```shell

```

bashrusakh:
Fix #1440

### DIFF
--- a/packages/ui/src/lib/shortcuts.ts
+++ b/packages/ui/src/lib/shortcuts.ts
@@ -300,7 +300,7 @@ const SHORTCUT_ACTIONS: ReadonlyArray<ShortcutAction> = [
   },
   {
     id: 'cycle_agent',
-    defaultCombo: 'tab',
+    defaultCombo: '',
     label: 'Cycle agent',
     description: 'Cycle agent while the model selector is open',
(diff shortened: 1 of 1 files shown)


yulia-ivashko:
Thank you for this, and sorry for the long wait.

Closing for the same reason as #1445, which proposed the same change: the centralized shortcut system landed on 26 August and removed `packages/ui/src/lib/shortcuts.ts`, the file this PR edits. Shortcuts now live in `packages/ui/src/lib/shortcuts/` and the default layout was reorganized around a `mod+k` leader.

Your observation stands on its own — `cycle_agent` still defaults to `tab` — but it is now `customizable: true` and rebindable from Settings.

Appreciate the time you spent on it.

PR #1715, opened 2026-06-18
feat(agents): expose thinking variant configuration in agent settings

bashrusakh:
Fix #1425

PR #1718, opened 2026-06-18
fix(agents): send null to clear temperature/topP overrides on update

bashrusakh:
Fix for temperature/topP null clearing bug found during PR #1715 review. When clearing temperature or topP on an existing agent, the UI sent undefined which JSON.stringify drops, so the server never received the clear command. Now sends null to properly remove the override.


btriapitsyn (approved):

btriapitsyn:
Resolved the merge conflict with current main and pushed the fix to the PR branch.

What changed:
- Kept the current agent thinking variant settings from main.
- Preserved the temperature/topP null-clearing fix so clearing those fields sends null instead of dropping the keys.
- PR is now mergeable again.

PR #1722, opened 2026-06-18
fix: stabilize Electron dev HMR startup

phantomObserver:
## Summary

This PR fixes several Electron dev-mode startup issues that were breaking or degrading the HMR workflow:

- fixes Windows dev server port handoff so the HMR proxy targets the actual backend port
- fixes Electron local-origin detection so the HMR UI is treated as a trusted local renderer for desktop IPC
- prevents a second Vite reload during startup by making `shiki` resolvable from the web workspace and pre-optimizing it up front

## Changes

- added `packages/web/scripts/dev-server.mjs` to launch the web dev server with cross-platform `OPENCHAMBER_PORT` handling
- updated `packages/web/package.json` dev scripts to use the Node wrapper instead of Unix-only shell expansion
- updated `packages/electron/main.mjs` so dev HMR uses the local UI origin as the trusted local renderer origin while keeping the local API URL separate
- updated `packages/web/vite.config.ts` to pre-optimize `shiki`
- added `shiki` to `packages/web/package.json` so Vite can resolve it from the web workspace during dependency optimization
- refreshed `bun.lock`

## Why

Before this change:

- Windows dev startup could launch the backend on `3000` while Vite proxied to `3901` or `3903`
- Electron HMR pages loaded from `http://127.0.0.1:5173` were incorrectly rejected by the desktop IPC local-origin gate
- the first use of the markdown Shiki worker caused Vite to discover `shiki` late and force a full reload after initial render

After this change:

- the backend and HMR proxy agree on the same port across platforms
- Electron dev HMR pages can use desktop IPC like the packaged local UI
- `shiki` is optimized during startup instead of triggering a follow-up reload later

## Validation

- ran `node --check packages/electron/main.mjs`
- ran `node --check packages/web/scripts/dev-server.mjs`
- ran `bun run type-check` in `packages/web`
- ran `bun install`
- manually verified `bun run electron:dev` now:
  - starts the backend on the expected HMR API port
  - no longer rejects desktop IPC from the HMR UI origin
  - no longer hits the delayed `shiki` optimize-and-reload cycle

## User Impact

- Electron dev mode is more reliable on Windows
- HMR startup behaves consistently with packaged local UI trust rules
- developers avoid an unnecessary reload shortly after the app first renders

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -99,7 +99,7 @@
     },
     "packages/electron": {
       "name": "@openchamber/electron",
-      "version": "1.13.0",
+      "version": "1.13.2",
       "dependencies": {
         "@openchamber/web": "workspace:*",
         "electron-context-menu": "^4.1.2",
@@ -114,7 +114,7 @@
     },
     "packages/ui": {
       "name": "@openchamber/ui",
-      "version": "1.13.0",
+      "version": "1.13.2",
       "dependencies": {
         "@base-ui/react": "^1.4.0",
         "@codemirror/autocomplete": "^6.20.0",
@@ -214,7 +214,7 @@
     },
     "packages/vscode": {
       "name": "openchamber",
-      "version": "1.13.0",
+      "version": "1.13.2",
       "dependencies": {
         "@openchamber/ui": "workspace:*",
         "@opencode-ai/sdk": "^1.17.7",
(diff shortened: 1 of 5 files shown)


phantomObserver:
I double-checked the current branch directly in `HEAD`, and `shiki` is already under `devDependencies` in `packages/web/package.json`. It is not present under `dependencies`.

The finding appears to be a false positive against the current branch.

btriapitsyn:

PR #1724, opened 2026-06-19
feat(chat): Comments and review in VS Code, like the OpenChamber desktop app

felipegenef:
# Comments and review in VS Code, like the OpenChamber desktop app

Comment on code without leaving it. Select lines or click the `+` in the gutter, write, and the comment stays anchored there until you send.

01-comment-opens-tab

## What you get

- Comment from the gutter `+` or right-click → **OpenChamber → Add Comment**
- Threads stay anchored while you keep reading
- Each comment is its own chip in the composer and its own card in the conversation
- Remove from either side and both stay in sync
- No OpenChamber tab open? One opens

### Review a diff

Comment on either side of a Source Control diff. The comment is filed against the real file, not the diff's temporary document.

02-review-diff-comment

### Remove from either side

Removing the thread in the editor drops its chip above the chat, and removing the chip drops the thread. Neither side can be left holding a comment the other has dropped.

05-remove-syncs

### Send

Sending clears the threads that rode along with the message.

06-send-clears-threads

Comments written across different files go out together, each as its own card.
03-send-cards

04-cards

### At narrow widths

Each pending comment is its own chip, so the row wraps and the file name truncates rather than pushing the composer out of shape.

07-chips-narrow

## Intent

Comments about code get written while reading code. The extension had nowhere to do that, and several comments collapsed into one wall of fenced text with no way to edit or drop just one.

A comment is now a first-class object: anchored in the editor, its own chip in the composer, its own card in the conversation.

## Non-goals

- Existing context sources (terminal captures, dev-server logs, preview annotations, PR comments and checks) keep their current capture and rendering behavior. This PR only adds VS Code editor comments to the shared context-card pipeline
- Those context sources keep their counter chips in the composer
- No changes to the in-app file and diff viewers, which already render inline comments
- No approvals, resolution state, or replies

## Affected surfaces

| Surface | What changes |
|---|---|
| `packages/vscode` | Comment controller and threads, `Add Comment` command, gutter range, thread menus, webview sync |
| `packages/ui` | Composer submit assembly, context chips, comment formatting, queue entries, card rendering |
| Web / Desktop / Mobile | Inherit the shared composer change. No runtime branch added |
| Persisted data | Queue entries gain optional `contextParts`; drafts accept an optional caller-provided `id`. Both absent on existing data |

## How it works

The composer's draft store stays authoritative. The extension mints the draft id and disposes its thread when the webview reports that draft gone, so the two sides cannot drift.

The webview reports its whole draft list rather than events, so a dropped notification cannot strand a thread. Sending empties the list, which clears the threads through the same path.

Every chat surface — the sidebar and each session tab — runs its own draft store, so each snapshot is tagged with the surface that produced it and only decides that surface's own threads. To any other surface the draft simply never existed, and its silence is not evidence of removal.

A comment that opens a session tab is held until that panel's webview sends its first message. Posting into a panel that has not booted drops the message outright.

## Repository guidance

| Guidance | How the change complies |
|---|---|
| `AGENTS.md` — thin entrypoints, logic in owning modules | Thread lifecycle stays in `InlineCommentThreads.ts`; pure logic stays in `inlineCommentSelection.ts`, free of the `vscode` import so it is testable, following `activePanelRouting.ts`. Context rendering uses the shared composer/message context modules rather than a VS Code-specific parallel renderer |
| `AGENTS.md` — update owning docs when invariants change | `composer/DOCUMENTATION.md` records the card-vs-text split and the queue-time capture contract |
| `AGENTS.md` — no failure that looks like success | A comment that cannot be delivered disposes its thread instead of sitting in the editor looking attached; delivery into a booting webview is held until it proves it is listening |
| `openchamber-change-discipline` — preserve tested behavior | The 25 existing `buildOutgoingMessage` tests pass unmodified; new behavior is additive. `addDraft`'s new `id` is optional and refused on collision |
| `sync-state-invariants` — capture send config at queue time | Comment cards are built at queue time and stored on the queue entry, beside `sendConfig` |
| `sync-state-invariants` — never infer disappearance from a first snapshot | `reconcileThreadFate` disposes a thread only after the composer has confirmed holding its draft at least once, with tests for the in-flight case |
| `sync-state-invariants` — roll back optimistic state on failure | A failed send restores every consumed draft, including those that became cards |
| `ui-api-decoupling` — no leaked bridge requests | Snapshot sync uses a fire-and-forget `postBridgeNotification`; the removal handler resolves the owning draft key within the current runtime before mutating |
| `locale-ui-patterns` | Command titles via `%command.*.title%` with en/fr entries; runtime strings in both l10n bundles; new keys added to every locale, and the unused `chat.chatInput.reviewComments` key removed from all of them, keeping the parity test green |
| `theme-system` | Existing tokens only, no new colors. The editor thread is native VS Code chrome |

## Validation

| Check | Result |
|---|---|
| `bun run type-check` (mobile, ui, web, electron, root) | Pass, exit 0 each |
| `bun run --cwd packages/vscode type-check` (extension + webview) | Blocked locally by duplicate CodeMirror versions in unchanged `packages/ui/src/lib/codemirror/languageByExtension.ts`; focused tests and packaging pass |
| `bun run lint` (mobile, ui, web, electron, root) + vscode | Pass |
| `bun run build` | Pass, exit 0 |
| `bun run --cwd packages/vscode package` | `openchamber-1.21.1.vsix`, 475 files, 8.82 MB |
| `bun run dead-code` | No entry for any file in this diff |
| `bun test` — `buildOutgoingMessage` | 28 pass |
| `bun test` — `contextParts` | 15 pass |
| `bun test` — `useInlineCommentDraftStore.callerId` | 7 pass |
| `bun test` — `inlineCommentSelection` | 40 pass |
| `bun test` — `messageQueueStore` / `i18n/messages` | 11 / 2 pass (the file also carries upstream's 3 in-flight send tests, merged in the rebase) |
| `bun test` — `packages/vscode`, all test files | 113 pass, 14 files |
| `bun test` — full `packages/ui`, per file | 213 of 214 files pass, 1810 tests |
| nls/l10n en-fr parity, every `%token%` resolves, every menu names a real command | Pass |

Manually exercised the packaged `.vsix` in VSCodium 1.108 on Linux: gutter `+` and right-click entry points, single and multi-line selections, comments on both sides of a Source Control diff, several comments across several files, commenting with no OpenChamber tab open, removal reaching a tab that is not the active one, and a fresh window before the extension had otherwise activated.

**One pre-existing failure, unrelated to this PR.** It fails on `main` as well:

- `mainLayoutMobileSidebarMount.test.ts` asserts against the source text of `SessionSidebar.tsx`, whose most recent commit is `7bb9b89` on `main`. That refactor changed the string the test expects. Not touched here.

**Not verified:** rendering, focus, and IME are untested in `packages/ui` by design (`composer/DOCUMENTATION.md`). Not run on Electron, mobile, Windows, macOS, or Cursor. No performance measurement taken.

## Visual evidence

The recordings show the editor-thread flow end to end: the gutter `+`, a thread anchored on its lines, a comment on a Source Control diff, removal from either side staying in sync, sending clearing the threads, and the comments arriving as cards.

All seven assets show the native editor-thread flow this PR ships. The original 2026-06-19 uploads were replaced, not kept: they were `059f695f`, `6f88094c`, `8c42f3ac` and `c0258647`, none of which appear above. `01`-`04` were re-recorded on 2026-07-31 and `05`-`07` after the surface-tagging fix. `05` shows removal syncing in both directions.

Not filmed: the right-click menu path. The gutter `+` is shown instead, and both entry points run the same command through one shared workspace predicate.

## Risks and failure behavior

- **Failed send:** consumed drafts are restored, including those that became cards
- **Nothing listening:** the thread is disposed and the user told, rather than left looking attached
- **Accepted but never delivered:** a comment the composer does not confirm holding within 30s is retracted from every surface's pending hold, its thread dropped, and the user told. The retraction runs first, so a webview booting after the deadline cannot still file and send a comment that was already reported as discarded. It used to sit there reading "Not sent yet" forever, unsendable and unrewritable
- **Booting webview:** comments are held until that panel's webview sends its first message, then all delivered in order. The hold is a list, so a second comment written during boot does not displace the first. Covers both a panel opened by the comment and an existing panel still restoring
- **Removal across surfaces:** each webview owns its own draft store, so removal is broadcast to every session panel and the sidebar. Only the one holding the draft acts; it is a no-op elsewhere
- **Removal during delivery:** a comment removed before its draft has landed is dropped from the pending hold and remembered by the webview, so a payload still in flight does not arrive as a chip the user already dropped
- **Multiple session tabs:** a comment attaches to the chat you last used, which is revealed on delivery. Each webview holds its own draft store, so every snapshot is tagged with the surface that produced it and only that surface can decide its own threads. Another tab reporting an empty list is not evidence that a comment is gone
- **Directory not resolved:** the webview waits up to 10s, then logs instead of dropping silently. This was previously a silent loss on a cold sidebar
- **Older queue entries:** `contextParts` is optional; entries without it send exactly as before
- **Colliding draft ids:** refused, so an external owner cannot retarget another draft
- **Activation:** `onStartupFinished` was added so the gutter `+` exists before OpenChamber is opened. This activates the extension on every window
- **Files outside the workspace:** refused by both entry points through one shared predicate, since the comment is filed against a workspace-relative path. The gutter `+` does not appear and the command explains why
- **Security:** payloads carry only the workspace-relative path, line range, comment text, and the snippet the user selected


felipegenef:
Hi @btriapitsyn ,

Checking in on this one. I've rebased it onto current main locally and resolved the conflicts coming from the draft store re-keying in #2360, but I haven't pushed yet. Before I do, I'd like to know whether this is a direction you still want in OpenChamber.

It's been open since mid June and I've rebased it a few times since, so I'd rather confirm the approach is wanted before pushing another round. If it is, I'm glad to make any changes you'd like, or to split it into smaller PRs if that's easier to review. If it isn't a fit, that's completely fine, just say so and I'll close it.

Thanks for the work you put into the project.

btriapitsyn:

felipegenef:
@btriapitsyn No worries at all, and thanks for the heads-up. I’m still interested in landing this.

I’ve rebased the branch onto the latest main and force-pushed the updated history. I resolved the conflicts using the current ContextPart architecture, reran the focused UI and VS Code test suites, and manually tested the rebuilt VSIX.

Everything is working and the PR is conflict-free again. It’s ready for review whenever you have the bandwidth.

mattv8:
I'll help review and QC this PR. I'll let you know once I've had a chance to review this- you'll hear back from me in the next 24 hrs or so.

mattv8:
Apologies for taking longer than the 24 h I promised. Two command paths still need fixing before I can sign off on this.

1. **Inline comments disable local slash commands.** In `packages/ui/src/components/chat/ChatInput.tsx`, any `additionalParts` entry prevents local command parsing. An inline comment creates one, so `/compact`, `/undo`, `/redo`, `/timeline`, `/btw`, and local magic prompts fall through and send their command text to the model. The command palette still offers them in this state. Parse local commands before consuming comment drafts. State-changing commands should run without sending their text; prompt-producing commands should retain the attached comment. Add coverage for a local command with one inline-comment part.

2. **Pinned knowledge changes server-command behavior.** `packages/ui/src/sync/session-ui-store.ts` now bypasses `session.command` whenever any additional part exists. Pinned project knowledge injects a synthetic part for ordinary sends, so commands and skills can take the prompt route even when the user attached no comment. Preserve `session.command` when the only extra part is system-injected knowledge. Use the prompt route only when user-attached structured context requires it. Add a test for a command with pinned knowledge and no user context.

felipegenef:
@mattv8 thanks for taking the time to review this and for catching both cases.

Local commands are now handled before composer context is consumed. State-only commands run locally and leave comments and files attached, while /btw and the magic prompt commands send that context with the generated prompt. Raw local commands also no longer end up in the generic message queue.

Pinned session knowledge is now tagged separately, so it no longer pushes recognized commands or skills onto the prompt route. User-attached and other prepared context still uses the prompt route.

I added regression coverage for both paths. The focused suite passes with 130 tests, and I verified /timeline, /summary, and inline comment cards manually in VSCode.

mattv8:
Thanks, this fixes the two routing cases from the earlier review. Two small items remain:

1. `planLocalSlashCommand` claims session-dependent actions even when there is no session. From a new-session draft, manually submitting `/compact`, `/undo`, `/redo`, or `/timeline` reaches `ChatInput.tsx:1181` and returns without sending or explaining anything. Keep unavailable actions on the normal send path, or give them explicit behavior, and add regression coverage for the new-session case.

2. `packages/ui/src/components/chat/composer/submit/slashCommands.ts:225` fails `oxlint` with `anti-slop(no-known-value-widening)`. Remove the anonymous return type or replace it with an accepted typed pattern, then rerun the changed-file `oxlint` check.

felipegenef:
@mattv8 thanks for the quick follow-up. Fixed both items.

Local action handling now requires an existing session. Manually entering /compact, /undo, /redo, or /timeline in a new-session draft keeps the command on the normal send path instead of silently swallowing it. I added regression coverage for all four commands.

I also removed the explicit return type that triggered anti-slop(no-known-value-widening). The focused oxlint check for slashCommands.ts and its test is now clean.

mattv8:
Looks good to me.

@btriapitsyn Should be safe to merge.

btriapitsyn:
Thanks for sticking with this one. I merged current `main` into the branch (ec46926f0) and resolved the conflicts myself rather than sending it back, since most of them came from a change on our side.

What changed and why:

- `main` now keeps the message queue on the server, and queueing already captures the composer's context with the message (`buildComposerContext`, delivered by the server). The branch's own queue capture (`contextParts` on `QueuedMessage`, the sending-id claim in `ChatInput`, the auto-send forwarding) duplicated that, so I dropped it in favour of main's version. Your queued-comment scenario is covered by the same path as every other context chip now.
- Kept the local slash command planning and re-applied it on main's `ChatInput`: `/undo`, `/redo`, `/compact`, `/timeline` leave attached comments in place, `/btw` and the magic prompts send them, and a local command runs instead of being queued as text. The command-route change in `session-ui-store` is kept as is; I checked OpenCode's `CommandInput` and it really does accept file parts only.
- `contextParts` keeps the `opencodeComment` metadata mirror in both directions (that is real OpenCode Desktop interop, I verified against its `comment-note.ts`), but I removed the two text-format regex fallbacks. OpenCode persists part metadata verbatim, so there is nothing for them to catch, and we avoid parsing prose. The queued-message schema now carries the mirror too so it survives the server round trip.
- Comment labels from VS Code are now the plain relative path, like in-app comments. The chip and card titles add the lines themselves, so `src/app.ts:3-5` was showing up as "app.ts:3-5 lines 3-5".
- Removed `onStartupFinished`. `activate()` calls `openCodeManager.start()`, so that would have spawned a managed OpenCode in every VS Code window whether or not OpenChamber was used. The right-click "Add Comment" activates the extension, and the gutter `+` is there from that point on.
- Dropped the `autoEditDraftId` channel and the chip changes built on it (a thread reply always arrives with its text, so nothing used it), the `prt-` part ids in `client.ts` (OpenCode assigns ids to id-less parts), and a few type assertions in the bridge handlers where the union already narrows. Added a README row and a module entry in the VS Code docs.

Validation on the merged head: `type-check` for ui and vscode, the full ui suite (381 files) and the vscode suite (33 files), `bun run --cwd packages/vscode build`, `dead-code` with no new entries. I have not exercised the packaged extension by hand after the merge; your recordings cover the editor flow and nothing in that layer changed beyond the label and activation event.

PR #1727, opened 2026-06-19
fix(cli): externalize Windows startup PowerShell to .ps1 wrapper (schtasks /TR 261-char limit)

### LINKED ISSUE #1725: [Bug] `openchamber startup enable` fails on Windows — schtasks `/TR` exceeds 261-character limit
#### Summary
On Windows, `openchamber startup enable` always fails because the value passed
to `schtasks.exe /Create /TR ...` is far longer than the documented 261-character
limit. The `/TR` payload contains an inlined PowerShell script (env-file parser
+ node launch) plus full absolute paths, which on any typical install totals
well over 500 characters.

#### Environment
- OS: Windows (any locale; on zh-CN the error is additionally printed as mojibake
  because schtasks emits GBK-encoded text)
- OpenChamber: current main
- Node path example: `C:\Program Files\nodejs\node.exe`
- CLI entry: `C:\Users\<user>\node_modules\@openchamber\web\bin\cli.js`

#### Repro

```powershell
PS> openchamber startup enable
Actual result
■  schtasks.exe /Create /TN dev.openchamber.web /SC ONLOGON /RL LIMITED /F
   /TR powershell.exe -NoProfile -ExecutionPolicy Bypass -Command "..."
   failed: 错误: '/TR' 选项值不能超过 261 字符。
   (decoded: "Error: '/TR' option value cannot exceed 261 characters.")
```
Status reports enabled: false; no scheduled task is created.
The Chinese error appears as mojibake (错误...) because schtasks.exe
emits output in the OEM/ANSI codepage (CP936/GBK on zh-CN Windows) while the
CLI captures it as UTF-8.
Expected result
Scheduled task dev.openchamber.web is created and started; startup status
reports enabled: true.
Root cause
packages/web/bin/cli.js, enableStartupService() Windows branch inlines the
full PowerShell loader into the /TR argument:
const powerShellCommand = [ /* env-file read + SetEnvironmentVariable + node launch */ ].join('; ');
const taskArgs = `powershell.exe -NoProfile -ExecutionPolicy Bypass -Command "${...}"`;
runStartupCommand('schtasks.exe', ['/Create', ..., '/TR', taskArgs]);
schtasks /TR is capped at 261 chars (historically 255). The inlined script
plus absolute paths always exceeds it.
Suggested fix
Persist the PowerShell loader to <dataDir>/startup.ps1 and point /TR at it
with powershell.exe -NoProfile -ExecutionPolicy Bypass -File "<path>", which
keeps /TR well under the limit while preserving env-file snapshot behavior.
Bonus: capture schtasks output as UTF-8 (or decode via current OEM codepage)
so localized error messages don't render as mojibake in non-English Windows.

#### Status
I'm working on a fix and will open a PR.

### Where does it happen?

Not sure

### Version (if known)

1.13.2

_No response_

### Logs (optional)

```shell
PS C:~\.bun> openchamber startup enable
┌  Error
│
■  schtasks.exe /Create /TN dev.openchamber.web /SC ONLOGON /RL LIMITED /F /TR powershell.exe -NoProfile -ExecutionPolicy Bypass -Command "$envFile='C:\Users\HAHH9527\.config\openchamber\startup.env'; if (Test-Path $envFile) { Get-Content $envFile | ForEach-Object { if ($_ -match '^([^=]+)=(.*)$') { $v=$matches[2]; if ($v.StartsWith(\"'\") -and $v.EndsWith(\"'\")) { $v=$v.Substring(1,$v.Length-2).Replace(\"'\''\",\"'\") }; [Environment]::SetEnvironmentVariable($matches[1], $v, 'Process') } } }; & 'C:\Program Files\nodejs\node.exe' 'C:\Users\HAHH9527\node_modules\@openchamber\web\bin\cli.js', 'serve', '--foreground', '--port', '3000'" failed: ����: '/TR' ѡ���ֵ���ܳ��� 261 �ַ���
│
└  failed
```

HAHH9527:
## Summary

Fixes #1725.

On Windows, `openchamber startup enable` always fails because the value passed to `schtasks.exe /Create /TR` contains the full inlined PowerShell env-file loader plus absolute paths, totaling well over the documented 261-character `/TR` limit.

This extracts the PowerShell loader into a `.ps1` wrapper file (mirroring the existing `writeMacosStartupWrapper` pattern on macOS) and points `/TR` at a short `powershell.exe -File` invocation, keeping `/TR` well under the limit while preserving env-file snapshot behavior.

## Root cause

`enableStartupService()` in `packages/web/bin/cli.js` inlined the entire PowerShell loader (env-file read + `SetEnvironmentVariable` + node launch) directly into the `/TR` argument. `schtasks /TR` is capped at 261 chars (historically 255), so the inlined script plus absolute paths always exceeded it. On non-English Windows (e.g. zh-CN) the resulting `schtasks` error was additionally rendered as mojibake because the CLI captured GBK/OEM output as UTF-8.

## Changes

- **`packages/web/bin/cli.js`** — Persist the PowerShell loader to a `startup.ps1` wrapper next to the startup env file; reduce `/TR` to `powershell.exe -NoProfile -ExecutionPolicy Bypass -File "<wrapper>"` (~115 chars, well under the 261-char limit). Mirrors the macOS `writeMacosStartupWrapper` pattern.
- **`packages/web/bin/cli.test.js`** — Add regression tests pinning `/TR` length `< 200` (default config) and `< 261` (worst-case long paths).

## Verification

- `bun run type-check` / `bun run lint` on `packages/web`
- New regression tests pass

## Test plan

- [x] `/TR` stays under 200 chars with default node/CLI paths
- [x] `/TR` stays under 261 chars with worst-case long paths
- [x] Manual: on Windows, `openchamber startup enable` creates and starts the task `dev.openchamber.web` and reports `Status reports enabled: true`

### DIFF
--- a/packages/web/bin/cli.test.js
+++ b/packages/web/bin/cli.test.js
@@ -30,6 +30,7 @@ import {
   parseArgs,
   resolveServeHost,
 } from './cli.js';
+import { buildWindowsStartupTaskCommand } from './lib/cli-startup.js';
 
 async function withTempOpenChamberDataDir(fn) {
   const previous = process.env.OPENCHAMBER_DATA_DIR;
@@ -882,5 +883,39 @@ describe('lifecycle commands with unmanaged explicit ports', () => {
         await server.close();
       }
     });
+  });
+});
+
+describe('Windows startup task command builder', () => {
+  it('default-path length stays under 200 chars', () => {
+    const cmd = buildWindowsStartupTaskCommand(
+      'C:\\Users\\test\\.config\\openchamber\\bin\\OpenChamber.ps1'
+    );
+    expect(cmd).toMatch(/^powershell\.exe -NoProfile -ExecutionPolicy Bypass -File /);
+    expect(cmd.length).toBeLessThan(200);
+  });
+
+  it('worst-case long path stays under 261-char Task Scheduler ceiling', () => {
+    // Build a wrapper path >= 180 chars (simulates long OPENCHAMBER_DATA_DIR)
+    // Overhead = 57 chars (prefix + closing quote), so max wrapper for <261 total is 203
+    const longPath =
+      'C:\\Users\\' +
+      'a'.repeat(139) +
+      '\\.config\\openchamber\\bin\\OpenChamber.ps1';
+    expect(longPath.length).toBeGreaterThanOrEqual(180);
+
+    const cmd = buildWindowsStartupTaskCommand(longPath);
+    expect(cmd.length).toBeLessThan(261);
+  });
+
+  it('does NOT inline SetEnvironmentVariable (externalization invariant)', () => {
+    const cmd = buildWindowsStartupTaskCommand('C:\\wrapper.ps1');
+    expect(cmd).not.toContain('SetEnvironmentVariable');
+  });
+
+  it('uses -File form, not -Command', () => {
+    const cmd = buildWindowsStartupTaskCommand('C:\\wrapper.ps1');
+    expect(cmd).toContain('-File ');
+    expect(cmd).not.toContain('-Command ');
   });
 });
--- a/packages/web/bin/lib/cli-startup.js
+++ b/packages/web/bin/lib/cli-startup.js
@@ -74,6 +74,10 @@ function getMacosStartupWrapperPath() {
   return path.join(getDataDir(), 'bin', 'OpenChamber');
 }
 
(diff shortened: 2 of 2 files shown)


HAHH9527:
This PR is usable, but it will always display a shell window after startup and requires administrative privileges.

PR #1728, opened 2026-06-19
fix(behavior): read AGENTS.md before settings.json so external edits are reflected

bashrusakh:
## Intent

External edits to `~/.config/opencode/AGENTS.md` did not show up in Settings -> Behavior. The page read `globalBehaviorPrompt` from `settings.json` first, and because every UI save wrote that field back, the AGENTS.md fallback never ran. The page showed stale text and a save could overwrite the real file.

The load path now treats AGENTS.md as the source of truth. An existing file is authoritative even when it is empty or whitespace-only, so externally clearing the file is reflected too. The persisted `globalBehaviorPrompt` is used only when the file does not exist or the read fails.

Branch history: the original fix was rebased onto current `main` (`111c4bf9d`) and adapted to the refactored load block, where `loadDesktopSettings()` replaced the direct settings fetch. A second commit makes the empty-file case authoritative and adds the regression test.

## Non-goals

- Save behavior is unchanged. The page still writes AGENTS.md through `PUT /api/behavior/agents-md` and keeps the `settings.json` copy in sync.
- No changes to the agents-md API routes, the settings registry, or the VS Code bridge.

## Affected surfaces

- `packages/ui/src/components/sections/behavior/BehaviorPage.tsx`
- `packages/ui/src/components/sections/behavior/behaviorPrompt.ts` (new, pure prompt-source resolver)
- `packages/ui/src/components/sections/behavior/behaviorPrompt.test.ts` (new)
- Web, Electron, VS Code, and mobile all render this shared component. VS Code serves AGENTS.md through its bridge with the same missing-file response, so the fallback applies there too.
- No persisted format or API contract change.

## Repository guidance

|---|---|---|
| `openchamber-change-discipline` | Executable source change in shared UI | Smallest complete change; a small pure helper keeps the decision testable; no new dependency or entrypoint |
| `settings-ui-patterns` | The change lives on a Settings page | Logic-only; no primitive, layout, string, or search surface touched |
| `ui-api-decoupling` | Prompt data comes from the settings API and `runtimeFetch` | Keeps `loadDesktopSettings()` and the explicit route; parses the response through `agentsMdResponseSchema`; the domain decision uses a discriminated union instead of raw payload fields |

## Validation

| Check | Result |
|---|---|
| `bun test packages/ui/src/components/sections/behavior/behaviorPrompt.test.ts` | 5 pass, 0 fail (non-empty file, empty file, whitespace-only file, missing file fallback, no sources) |
| `bun run --cwd packages/ui type-check` | exit 0 |
| `bun run --cwd packages/ui lint` | exit 0 |
| `bunx oxlint` on the changed files | only pre-existing anti-slop findings (BehaviorPage.tsx lines 42, 43, 72, 77, 174), none in the changed block or new files |
| `bun run dead-code` | no findings for the new file or exports; the long pre-existing backlog elsewhere is unchanged |
| CI `checks` on `9515b3d6` | pass, 7m42s: build, type-check, lint, changelog check, tests, Electron unit tests |

Not verified: an in-app manual reproduction with a real `~/.config/opencode/AGENTS.md`. The component wiring (`ok`/`exists` mapping) is covered by inspection because module mocking is banned by the anti-slop lint rules.

## Visual evidence

No layout, styling, or copy change. The only user-visible difference is which text loads into the existing textarea, covered by the focused resolver tests and the code path checks above.

## Risks and failure behavior

- AGENTS.md read failure falls back to the persisted copy, same as before.
- An existing empty file now shows empty instead of the stale copy; saving it writes that empty content, which matches what the file actually contains.
- A malformed successful response still aborts the load through the existing catch, identical to current `main`.
- Rollback is a revert of the commits; there is no stored-format or migration impact.

### DIFF
--- a/packages/ui/src/components/sections/behavior/BehaviorPage.tsx
+++ b/packages/ui/src/components/sections/behavior/BehaviorPage.tsx
@@ -118,8 +118,9 @@ export const BehaviorPage: React.FC = () => {
         ]);
 
         let nextSettings: BehaviorSettingsState = DEFAULT_BEHAVIOR_SETTINGS;
+        let settingsGlobalBehaviorPrompt: string | undefined;
         if (settingsRes.ok) {
-          const data = await settingsRes.json();
+          const data = (await settingsRes.json()) as Record<string, unknown>;
           nextSettings = {
             ...nextSettings,
             responseStyleEnabled: data.responseStyleEnabled === true,
@@ -129,15 +130,22 @@ export const BehaviorPage: React.FC = () => {
               : '',
           };
           if (typeof data.globalBehaviorPrompt === 'string') {
-            nextSettings = { ...nextSettings, prompt: data.globalBehaviorPrompt };
+            settingsGlobalBehaviorPrompt = data.globalBehaviorPrompt;
           }
         }
 
-        if (!nextSettings.prompt.trim() && agentsMdRes.ok) {
+        // AGENTS.md is the source of truth for the system prompt — read it
+        // first so that external edits are always reflected. Only fall back to
+        // the persisted copy (globalBehaviorPrompt) when the file is missing
+        // or empty.
+        if (agentsMdRes.ok) {
           const agentsData = await agentsMdRes.json();
-          if (typeof agentsData.content === 'string') {
+          if (typeof agentsData.content === 'string' && agentsData.content.trim()) {
             nextSettings = { ...nextSettings, prompt: agentsData.content };
           }
+        }
+        if (!nextSettings.prompt.trim() && typeof settingsGlobalBehaviorPrompt === 'string') {
+          nextSettings = { ...nextSettings, prompt: settingsGlobalBehaviorPrompt };
         }
 
         setPrompt(nextSettings.prompt);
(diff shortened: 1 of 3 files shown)


bashrusakh:
@btriapitsyn Hi. Can you push this please

mattv8:

bashrusakh:
**This PR has had no new commits and is still open because no one has reviewed it yet.**

A comment like *"it has had no commits for over a month, therefore it should be closed"* is not a review — and is not a reason to close a PR that is waiting for review.

mattv8:
@bashrusakh good point, sorry about that. Just trying to help triage the large amount of abandoned PR's from peoples' bots, and I didn't consider the author.

bashrusakh:
Sure, no worries — and thanks for helping with triage, it's really needed with this many open PRs.

One triage question that's been working well for me:

> Is it still reproducible on current `main`?

Quite a few open issues/PRs look like they were already fixed along the way, or the fix already landed via other commits, so they no longer need a separate PR. We've also seen already-fixed issues left open.

One small note on the stale filter: age + merge conflicts alone can be a misleading signal here. The team and code owners sometimes push fixes directly to `main`, bypassing the PR — and that has already happened a few times. So a PR can look "abandoned with conflicts" while the real question is still "is the bug still there?".

So I'd suggest triaging by that question first, rather than by age/conflicts. Happy to rebase and re-verify this one against current `main` if that helps.

yulia-ivashko:
I ran the full `checks` job locally on this branch merged onto current main. Everything passed, including the new resolver tests. Thanks for sticking with this one. Treating an existing AGENTS.md as the source of truth, even when it's empty, is the right call, and it stops a save from overwriting edits made outside the app. Merging.

yulia-ivashko (approved):

PR #1729, opened 2026-06-19
fix(cli): use Register-ScheduledTask for Windows startup (no admin, no /TR limit)

### LINKED ISSUE #1725: [Bug] `openchamber startup enable` fails on Windows — schtasks `/TR` exceeds 261-character limit

HAHH9527:
## Summary

Fixes #1725.

On Windows, `openchamber startup enable` fails for two reasons:

1. **`schtasks /TR` 261-char limit** — the inlined PowerShell env-loader + absolute paths far exceed it.
2. **`schtasks /Create /SC ONLOGON` requires elevation** — it creates a system-level (any-user-logon) trigger, so non-admin users get "Access Denied" even before the `/TR` length issue.

This PR replaces `schtasks.exe` entirely with the `Register-ScheduledTask` PowerShell cmdlet, which solves **both** problems:

- **No `/TR` limit** — the action is an object (`New-ScheduledTaskAction`), not a string argument.
- **No elevation required** — `LogonType Interactive` + `RunLevel Limited` + bound to the current user installs a per-user logon trigger that works without admin rights.
- `ExecutionTimeLimit` set to zero so Task Scheduler's 72-hour default ceiling cannot kill the foreground server.

`disableStartupService` now uses `Stop-ScheduledTask` / `Unregister-ScheduledTask`. `getStartupStatus` keeps `schtasks /Query` (works without admin for the user's own tasks).

### Supersedes #1727

#1727 proposed keeping `schtasks.exe` and externalizing the PowerShell loader into a `.ps1` wrapper file to stay under the `/TR` 261-char limit. While that works around the length issue, it **does not address the elevation problem** — `schtasks /Create /SC ONLOGON` still requires administrator rights for a system-level trigger. This PR's `Register-ScheduledTask` approach eliminates both issues with a single, more robust solution, so the two approaches are mutually exclusive.

## Changes

- **`packages/web/bin/cli.js`**
  - `enableStartupService` (Windows branch): `Register-ScheduledTask` with `LogonType Interactive` + `RunLevel Limited`, bound to `[WindowsIdentity]::GetCurrent()`. Action is a `New-ScheduledTaskAction` object — no `/TR` string limit.
  - `disableStartupService` (Windows branch): `Stop-ScheduledTask` / `Unregister-ScheduledTask`.
  - New exported helpers: `buildWindowsStartupActionCommand`, `buildWindowsRegisterScheduledTaskScript`.
  - `getStartupStatus` unchanged (keeps `schtasks /Query`).
- **`packages/web/bin/cli.test.js`** — Unit tests for the two new builder functions, pinning: per-user task (no elevation), no `schtasks`/`/SC ONLOGON`/`/TR` in the generated script, current-user principal binding, env-file loading, and node launch.

## Test plan

- [x] `bun run type-check` / `bun run lint` on `packages/web`
- [x] New unit tests pass
- [ ] Manual: on Windows, `openchamber startup enable` creates the task without admin rights and reports `enabled: true`
- [ ] Manual: `openchamber startup disable` removes the task
- [ ] Manual: `openchamber startup status` reports correct state

### DIFF
--- a/packages/web/bin/cli.js
+++ b/packages/web/bin/cli.js
@@ -2210,6 +2210,47 @@ function runStartupCommand(command, args, options = {}) {
   return result;
 }
 
+// Runs a PowerShell script in the current user's session (hidden, non-interactive).
+function runWindowsPowershell(script, options = {}) {
+  return runStartupCommand('powershell.exe', ['-NoProfile', '-NonInteractive', '-Command', script], options);
+}
+
+// Builds the PowerShell command the scheduled task runs: it loads the persisted
+// startup.env snapshot into the process environment, then starts the foreground
+// server under node. Used as the task action's -Command.
+function buildWindowsStartupActionCommand(options = {}) {
+  const envFilePath = getStartupEnvFilePath();
+  const startupArgs = buildStartupArgs(options).map(powershellQuote).join(', ');
+  return [
+    `$envFile=${powershellQuote(envFilePath)}`,
+    `if (Test-Path $envFile) { Get-Content $envFile | ForEach-Object { if ($_ -match '^([^=]+)=(.*)$') { $v=$matches[2]; if ($v.StartsWith("'") -and $v.EndsWith("'")) { $v=$v.Substring(1,$v.Length-2).Replace("'\\''","'") }; [Environment]::SetEnvironmentVariable($matches[1], $v, 'Process') } } }`,
+    `& ${powershellQuote(process.execPath)} ${startupArgs}`,
+  ].join('; ');
+}
+
+// Builds the Register-ScheduledTask PowerShell script that installs the
+// OpenChamber startup task for the current user WITHOUT elevation.
+// `schtasks.exe /Create /SC ONLOGON` creates a system-level (any-user-logon)
+// trigger that requires administrator rights, so non-admin `startup enable`
+// failed with Access Denied. Register-ScheduledTask with LogonType Interactive
+// + RunLevel Limited, bound to the current user, works without elevation, and
+// takes the action as an object so there is no schtasks /TR 261-char limit.
+// ExecutionTimeLimit is zero so the foreground server is not killed by Task
+// Scheduler's 72-hour default ceiling.
+function buildWindowsRegisterScheduledTaskScript({ taskName, actionCommand }) {
+  const tn = powershellQuote(taskName);
+  const actionArgument = `-NoProfile -ExecutionPolicy Bypass -Command ${powershellQuote(actionCommand)}`;
+  return [
+    `$ErrorActionPreference='Stop'`,
+    `$u=[Security.Principal.WindowsIdentity]::GetCurrent().Name`,
+    `$p=New-ScheduledTaskPrincipal -UserId $u -LogonType Interactive -RunLevel Limited`,
+    `$t=New-ScheduledTaskTrigger -AtLogOn -User $u`,
+    `$a=New-ScheduledTaskAction -Execute 'powershell.exe' -Argument ${powershellQuote(actionArgument)}`,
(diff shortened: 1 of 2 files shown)


tomzx:
@openchamber-bot review

HAHH9527:
> @openchamber-bot review

This issue is not running. Please review #1727

PR #1734, opened 2026-06-19
fix(cli): verify process identity in isProcessRunning to avoid stale-PID crashloop (#1721)

### LINKED ISSUE #1721: isProcessRunning() returns false positives after PID recycling; causes infinite systemd restart crashloop
## Summary

The PID-based lockfile check in `packages/web/bin/cli.js` causes OpenChamber to refuse to start when its previous PID has been recycled to an unrelated process. This produces an **infinite restart crashloop on every system reboot** when OpenChamber is managed by systemd with `Restart=always`, until the stale lockfile is manually removed.

## Reproduction

1. Run OpenChamber via a systemd user service with `Restart=always` and `RestartSec=5`
2. Kill the OpenChamber node process with `SIGKILL` (or trigger a system reboot that does not give the process time to shut down gracefully)
3. systemd attempts to restart OpenChamber
4. OpenChamber exits immediately with: `Error: OpenChamber is already running on port <port> (PID: <stale-pid>)`
5. systemd retries every 5 s; every attempt fails with the same error → infinite crashloop, service never comes back up

## Root cause

`packages/web/bin/cli.js` lines 3415–3420:

```js
const existingPid = readPidFile(pidFilePath);
if (existingPid && isProcessRunning(existingPid)) {
  throw new Error(`OpenChamber is already running on port ${targetPort} (PID: ${existingPid})`);
}
```

`isProcessRunning` at lines 2378–2385 only checks for PID existence:

```js
function isProcessRunning(pid) {
  try {
    process.kill(pid, 0);
    return true;
  } catch {
    return false;
  }
}
```

This treats **any** live PID as a live OpenChamber instance. After `SIGKILL` (typical during system shutdown or `kill -9`), the kernel recycles the original PID. If another process happens to take that PID number, `process.kill(stalePid, 0)` succeeds and OpenChamber falsely believes an instance is already running on the port — so it never even tries to bind, and systemd keeps restarting it.

## Evidence from a real production session

- Stale lockfile: `~/.config/openchamber/run/openchamber-9090.pid` containing `439`
- `cat /proc/439/cmdline` → `node /home/herjarsa/npm-global/bin/agentmemory`
- `cat /proc/439/status | grep ^Name` → `Name: V8Worker` (a V8 worker thread of the `agentmemory` process)
- `kill -0 439` → exit 0 (process exists, just not OpenChamber)
- `ss -tlnp | grep 9090` → no listener (no OpenChamber, the port was actually free)

Timeline:
- 17:51 system reboot; OpenChamber node process was killed by the reboot without graceful shutdown, so `removePidFile` never ran
- 17:55 systemd began restarting OpenChamber; the stale `439` PID had been reassigned to an agentmemory V8 worker
- 17:55–17:56 OpenChamber crashlooped every 5 s with `Error: OpenChamber is already running on port 9090 (PID: 439)` while the port was actually free

## Suggested fix

Strengthen `isProcessRunning` to validate process identity, not just PID liveness. Smallest viable change:

```js
function isProcessRunning(pid) {
  try {
    process.kill(pid, 0);
  } catch {
    return false;
  }
  try {
    const cmdline = fs.readFileSync(`/proc/${pid}/cmdline`, 'utf8');
    return cmdline.includes('openchamber') || cmdline.includes('cli.js');
  } catch {
    return false;
  }
}
```

For macOS / Windows, swap `/proc/<pid>/cmdline` for the equivalent (`lsof -p <pid>` + grep, or `wmic process where ProcessId=<pid> get CommandLine`). A platform helper already exists in this repo for similar lookups.

Stronger, more invasive options (consider if the small fix isn't enough):

- **Unix-socket lock**: bind a Unix-domain socket on the run dir; treat a successful connect as "running". This also gives a real IPC channel for `openchamber stop` without parsing PIDs.
- **TTL on the lockfile**: when reading a stale pidfile, also check `mtime`; if older than N hours and the PID is not a confirmed openchamber process, treat as stale.
- **Lock + heartbeat**: write a heartbeat timestamp alongside the PID; on startup, if the PID is alive but the heartbeat is older than X, treat as stale (catches orphaned-but-alive processes).

## Workaround used in production until upstream lands

```ini
# ~/.config/systemd/user/openchamber.service
[Service]
ExecStartPre=/usr/bin/rm -f %h/.config/openchamber/run/openchamber-9090.pid %h/.config/openchamber/run/openchamber-9090.json
```

This runs before every `ExecStart`, including every systemd-triggered restart, so it cleans stale lockfiles both at boot and after any mid-run crash. Downstream users on systemd can adopt this immediately; users on launchd / Windows services need an analogous pre-start hook.

## Environment

- OpenChamber 1.13.0 → 1.13.2 (observed across two reproductions on the same machine)
- systemd 255 (user service)
- Linux kernel 6.x with PID namespaces enabled
- Repro is independent of OpenCode connection mode (embedded vs external) — only the lockfile lifecycle matters

IbrahimKhan12:
## Why

Fixes #1721.

After an ungraceful shutdown (e.g. `SIGKILL` on reboot), `removePidFile` never runs, so a stale `run/openchamber-<port>.pid` outlives the process. The kernel can recycle that PID to an unrelated process, so `isProcessRunning`'s liveness-only `process.kill(pid, 0)` succeeds and startup aborts with `OpenChamber is already running on port <port> (PID: <stale-pid>)`. Under systemd `Restart=always` this becomes an infinite reboot crashloop while the port is actually free. (Reporter saw stale PID `439` recycled to `node .../agentmemory`.)

## What

Verify process identity, not just liveness:

- Keep the `process.kill(pid, 0)` liveness check, then on Linux confirm the PID is OpenChamber via `/proc/<pid>/cmdline`.
- Where identity can't be determined (non‑Linux, or `/proc` unreadable), fall back to liveness-only, so macOS/Windows behavior is unchanged with no false negatives.
- Match the `openchamber` path segment (present in every install — `@openchamber/web` globals and the `openchamber` repo checkout — for both the foreground `bin/cli.js` and daemon `server/index.js`), so a recycled PID running e.g. `npm-cli.js` isn't mistaken for OpenChamber.

## Testing

`bun run --cwd packages/web test bin/cli.test.js` → 27 passed; `bun run lint:web` clean. Added unit tests for `isOpenchamberCmdline` (incl. the recycled `agentmemory` and `npm-cli.js` cases) plus a Linux test asserting a live `sleep` PID is reported not-running.

## Risk / limitations

Identity verification is Linux-only by design (the reported platform); macOS/Windows keep prior behavior. Cross-user recycled PIDs are handled implicitly: `process.kill(pid, 0)` throws `EPERM` before `/proc` is read.

### DIFF
--- a/packages/web/bin/cli.js
+++ b/packages/web/bin/cli.js
@@ -2390,13 +2390,39 @@ function removeInstanceFile(instanceFilePath) {
   }
 }
 
+function readProcessCmdline(pid) {
+  if (process.platform !== 'linux') {
+    return null;
+  }
+  try {
+    // /proc/<pid>/cmdline is a NUL-delimited argv list.
+    return fs.readFileSync(`/proc/${pid}/cmdline`, 'utf8').replace(/\0/g, ' ').trim();
+  } catch {
+    return null;
+  }
+}
+
+function isOpenchamberCmdline(cmdline) {
+  if (typeof cmdline !== 'string' || cmdline.length === 0) {
+    return false;
+  }
+  // Match the install path, not a generic "cli.js", so a recycled npm-cli.js isn't mistaken for us.
+  return cmdline.toLowerCase().includes('openchamber');
+}
+
 function isProcessRunning(pid) {
   try {
     process.kill(pid, 0);
-    return true;
   } catch {
     return false;
   }
+  // A live PID may be a recycled stranger (issue #1721): confirm identity on Linux,
+  // and fall back to liveness-only when it can't be determined.
+  const cmdline = readProcessCmdline(pid);
+  if (cmdline === null) {
+    return true;
+  }
+  return isOpenchamberCmdline(cmdline);
 }
 
 function waitForProcessExit(pid, timeoutMs) {
@@ -5734,6 +5760,8 @@ export {
   isValidTunnelDoctorResponse,
   readDesktopLocalPortFromSettings,
   getPidFilePath,
+  isProcessRunning,
+  isOpenchamberCmdline,
   resolveTunnelProviders,
   fetchTunnelProvidersFromPort,
   fetchSystemInfoFromPort,
--- a/packages/web/bin/cli.test.js
+++ b/packages/web/bin/cli.test.js
@@ -1,9 +1,10 @@
 import { describe, expect, it } from 'vitest';
 import path from 'path';
+import { spawn } from 'child_process';
 import { pathToFileURL } from 'url';
 
 import { isModuleCliExecution, normalizeCliEntryPath } from './cli-entry.js';
-import { assertAuthenticatedNetworkExposure, parseArgs } from './cli.js';
+import { assertAuthenticatedNetworkExposure, isOpenchamberCmdline, isProcessRunning, parseArgs } from './cli.js';
 
 describe('cli args', () => {
   it('accepts legacy daemon flags as no-ops', () => {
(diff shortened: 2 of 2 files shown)


btriapitsyn:
Thank you for the thorough diagnosis and the report on #1721 — the root cause analysis here (stale pid file after an ungraceful shutdown + kernel PID recycling defeating a liveness-only check, looping forever under systemd `Restart=always`) was spot on and very clearly written. It's exactly the right thing to fix.

We've landed a fix for the same issue directly in `main` (03f03b74), taking a slightly different shape so the identity check is scoped precisely:

- Rather than changing `isProcessRunning` itself, we added a dedicated `isOpenchamberProcessRunning(pid)` (liveness + command-line identity) and use it **only** at the two places that validate a PID read from a pid file — the "already running" guard and the stale-pid cleanup sweep. `isProcessRunning` stays liveness-only, because it's also called right after we spawn our own daemon child and in the stop paths, where an identity check could produce a false negative (e.g. a source checkout whose path doesn't contain "openchamber" would wrongly report the freshly-started daemon as failed).
- Identity verification runs on Linux (`/proc/<pid>/cmdline`) **and** macOS (`ps -o command=`), falling back to liveness where the command line can't be read (Windows / unreadable `/proc`), so there are no false negatives on those platforms.
- Same "openchamber" path-segment match you used (it correctly rejects the recycled `agentmemory` / `npm-cli.js` cases), plus we now clear the stale pid file once its PID is no longer ours.

Your tests' cases (`agentmemory`, `npm-cli.js`, live non-OpenChamber PID) are carried over into the new unit tests. Closing this in favour of the landed fix — genuinely appreciate the contribution and the careful write-up. 🙏

PR #1742, opened 2026-06-20
fix(cli): detect brew-installed opencode on macOS (#1720)

### LINKED ISSUE #1720: [Bug] Mac initial setup process does not detect opencode installed via brew

Prereqs:
* opencode TUI installed vi homebrew and working

Steps:
1. Fresh install of Openchamber App on MacOS
2. Start app
3. Initial setup screen shows Verbiage to install OpenCode.

Expected:
* initial setup should detect opencode installed via homebrew

Actual:
* initial setup does NOT detect opencode installed via homebrew

Workaround:
* copy output of `which opencode` and paste into manual override (I think!)

### Where does it happen?

Desktop (macOS)

### Version (if known)

26.5.1 (25F80)

_No response_

### Logs (optional)

```shell

```

mskadu:
## Intent

Fixes #1720. When the macOS app is launched from the Dock, the process inherits a minimal PATH (`/usr/bin:/bin:/usr/sbin:/sbin`) that omits brew bin directories. Two things made that a hard failure:

1. The server's login-shell probes had no timeout, so a slow `.zshrc` (nvm, pyenv, etc.) blocked startup indefinitely.
2. Even when the probe returned, brew-installed opencode on Intel Macs was not guaranteed to be picked up.

This PR bounds every shell probe with a 5s timeout, adds a fast `command -v` probe ahead of the login-shell probe, and makes brew fallback ordering consistent across the server and VS Code extension.

## Non-goals

- Custom brew prefixes outside `/opt/homebrew` and `/usr/local` are still not covered by hardcoded fallbacks: only the shell probe catches those.
- Windows `where` probes remain without explicit timeouts (local operations, unaffected by slow shell startup).
- No change to the `pathLooksUserConfigured` heuristic: `/usr/local/bin` stays part of the "minimal system PATH" classification, since it appears in macOS's default PATH.

## Affected surfaces

- `packages/web/server/lib/opencode/env-runtime.js`: server-side binary resolution for opencode, node, bun. All runtimes (web server, Electron in-process backend) share this module.
- `packages/vscode/src/opencode.ts`: VS Code extension fallback ordering for opencode CLI discovery.
- `scripts/reproduce-issue-1720.mjs`: new diagnostic script for the reproduction case.
- `CHANGELOG.md`: documents the fix under Unreleased.
- Mobile is unaffected (does not manage the OpenCode lifecycle).

## Repository guidance

|---|---|---|
| `AGENTS.md`: runtime boundaries; shared contracts must define intentional behavior per runtime | `env-runtime.js` is shared by web and Electron | Timeout and fast-path behavior are identical across the three resolvers and both runtimes |
| `packages/web/server/lib/opencode/DOCUMENTATION.md`: owning module docs | Documents binary resolution order and shell environment runtime | Fallback ordering change is consistent with the documented server contract |
| `openchamber-change-discipline` skill: smallest complete change; preserve behavior covered by tests | Executable source change to a startup-critical path | Dropped the `/usr/local/` TOOLCHAIN_SEGMENTS change that upstream tests explicitly classify as minimal-system PATH, avoiding regression of the Electron login-shell merge |
| `performance-engineering` skill: don't claim speedups you can't measure | The fast path exists to avoid login-shell startup cost | Claims kept modest: the fast path only helps when the binary is already visible in the inherited PATH |

## Validation

| Check | Result |
|---|---|
| `bun run --cwd packages/web test` | 161 files, 1602 passed, 1 skipped |
| `bun run --cwd packages/vscode test` | 25/25 test files passed |
| `node scripts/run-isolated-tests.mjs scripts` | 1/1 passed |
| `node --check` on modified JS files | Passed |

Not verified: actual Dock-launch behaviour on a physical macOS machine with a slow `.zshrc`. The timing claim relies on the existing Electron login-shell merge test coverage plus the new `command -v` fast path falling through safely when it does not help.

## Visual evidence

No user-visible change: this affects binary resolution and startup timing, not rendered UI. The diff cannot affect rendered behavior in any theme, layout, or interaction state.

## Risks and failure behavior

- Slow login shell: every probe now bounded at 5s; on timeout it falls through to the next detection method.
- Fast path failure: `/bin/sh -c 'command -v'` failures are caught and fall through to the login-shell probe; results are validated with `isExecutable()` before acceptance.
- Fast path not finding anything: harmless: the hardcoded fallbacks and existing probes still run.
- No new dependencies, network calls, or CI changes. Cross-runtime parity maintained (web server and VS Code ordering now match).


tomzx:
@openchamber-bot review

btriapitsyn:

btriapitsyn:
Merged — thanks, and sorry it sat this long.

I trimmed the branch before merging instead of asking for another round. The part that reproduces is the missing timeout: the server's login-shell probes (`$SHELL -lic …` for the env snapshot and for `command -v opencode/node/bun`) ran unbounded, so a slow or interactive rc file stalled startup, and from a Dock launch that is exactly what made a Homebrew `opencode` look undetected. Every probe now carries the same 5-second bound the Electron shell probe already uses, with a regression test that an overrunning probe falls through instead of blocking.

What did not make it:
- the non-login `/bin/sh -c 'command -v …'` fast path — a plain `sh` inherits the same PATH the resolver has already walked, so it cannot find anything new;
- the reproduction script and the `.gitignore` entry;
- the Homebrew ordering swap in the VS Code extension — both prefixes are checked either way, and the server's known-install list already has both.

Changelog entries are written at release time on this repo, so that line was dropped too.

PR #1745, opened 2026-06-20
ci: add Build Windows Electron workflow

bashrusakh:
## Summary
- add a manual Windows Electron build workflow
- default workflow dispatch builds to the selected branch instead of always checking out `main`
- give non-`main` builds distinct artifact filenames and `latest.yml` names so branch artifacts do not overwrite stable ones

## Context
This replaces #1696, which was closed after the branch was deleted. The branch has been restored with the follow-up workflow fixes included.

## Validation
- workflow YAML inspected locally
- confirmed the previous branch-selection bug via Actions checkout logs (`win-artifact-pr-fixes` run still checked out `main`)

### DIFF
--- a/.github/workflows/build-windows-electron.yml
+++ b/.github/workflows/build-windows-electron.yml
@@ -1,0 +1,75 @@
+name: Build Windows Electron
+
+on:
+  workflow_dispatch:
+    inputs:
+      ref:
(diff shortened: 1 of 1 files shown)


tomzx:
@openchamber-bot review

PR #1746, opened 2026-06-20
fix(chat): prevent follow-loop engage during deferred scroll restore on session switch

herjarsa:
## Problem

When switching sessions in the chat, the scroll jumps to an earlier point in the conversation and then rapidly scrolls down to the latest message. Multiple fix attempts were insufficient.

## Root Causes (v2 supersedes v1)

After v1 (defer-state fix), the bug persisted because the LERP animation was re-triggered by a second path: the **re-pin logic** in `handleScrollEvent`. On session switch with a partially-hydrated container, the user is 'near bottom' of the small container, and the re-pin immediately re-engages the follow loop with LERP toward the growing target. Also, the **upward-scroll release** block was placed AFTER the programmatic-window early return, so a follow-loop LERP firing within 200ms of a user scroll-up could swallow the release signal.

## Fix

Three changes in `packages/ui/src/hooks/useChatAutoFollow.ts`:

1. **`!saved` path**: set state to `'released'` (not `'following'`) and do NOT call `startFollowLoop` / `startSettleBurst`. The container is still hydrating; starting the follow loop now would LERP toward a growing target as messages load. The follow loop will start naturally when `sessionIsWorking` becomes true (user sent a message).

2. **Re-pin suppression during initial hydration**: guard the re-pin block in `handleScrollEvent` with `pendingInitialRestoreRef.current === currentSessionIdRef.current`. While a deferred restore is pending for the current session, the user is always 'near bottom' of a small container, but the re-pin is suppressed to prevent the LERP.

3. **Upward-scroll release moved above the programmatic-window guard**: the user can always release auto-follow by scrolling up, even mid-LERP. A LERP firing within the 200ms programmatic window can no longer swallow the release signal.

## Verification

- **Oracle review**: PASS (v1) — 5 concerns verified (no follow-loop during defer, correct recovery after defer, no race conditions, no visual flash, no infinite loop)
- **Playwright session-switch test**: scroll position stays where it was after switch, no LERP animation, no rapid scroll to bottom. Distance-from-bottom is preserved across the switch.
- `bun run type-check:ui` and `bun run build:web`: exit 0
- `bun run lint:ui`: 0 errors

## Files Changed

- `packages/ui/src/hooks/useChatAutoFollow.ts` (+22, -8)


herjarsa:
Superseded by #1747 (fix/scroll-jump-to-bottom-v2). PR #1747 contains the same scroll fix with cleaner commits and a Playwright-verified snap-to-bottom approach. Closing this one to avoid duplicate work.

Issue #1737, opened 2026-06-20
Show connection status indicator between frontend and OpenChamber runtime


tomzx:
# Background

The frontend (the React UI running in the browser, the Electron window, or the VS Code webview) communicates with the OpenChamber web server (the Express runtime) over HTTP, SSE, and WebSocket. This is the hop that carries the sync layer (`packages/ui/src/sync/event-pipeline.ts`) and every RuntimeAPIs call. It can drop independently of the OpenChamber-to-OpenCode hop when, for example, the web server crashes or restarts, the loopback / remote / tunnel transport fails, the browser throttles background timers, or `navigator.onLine` reports offline.

Currently there is no visible indication that the frontend has lost its connection to the OpenChamber runtime. The UI simply freezes: live updates stop, actions appear to do nothing, and the user cannot tell whether the break is on their side (frontend-to-runtime) or further upstream (runtime-to-OpenCode, covered by #1556). This is especially confusing in remote / tunnel / VS Code-over-SSH setups where the frontend-to-runtime hop is the fragile one.

An explicit connection status indicator for the frontend-to-runtime hop would let users quickly localize the failure and know whether to reload, check their network, or look at the runtime-to-OpenCode indicator instead.

# Version

80df9e6

# Acceptance Criteria

## Must

- [ ] A status indicator is visible in the UI that reflects the health of the frontend's connection to the OpenChamber web server (distinct from the OpenChamber-to-OpenCode indicator in #1556)
- [ ] The indicator updates in near-realtime when the connection status changes (connected / disconnected / reconnecting), driven by the existing SSE/WS reconnect loop rather than a new polling mechanism
- [ ] When the connection is lost, the indicator shows a clear visual state (color, icon, text) that distinguishes "disconnected" from "connected"
- [ ] The indicator works across the web, desktop (Electron), and VS Code webview runtimes
- [ ] The indicator does not introduce significant overhead (piggybacks on existing SSE/WS health signals, not a new polling loop)
- [ ] Graceful degradation: if the indicator cannot determine status, it shows an "unknown" state rather than falsely showing "connected"

## Should

- [ ] The indicator surfaces the underlying reason when available (offline, server unreachable, permanent 4xx), consistent with the reconnect-loop pacing policy
- [ ] The two indicators (frontend-to-runtime and runtime-to-OpenCode) are visually related so the user can tell which hop is broken at a glance

---

[...]

Issue #1556, opened 2026-06-07
Show connection status indicator between OpenChamber and OpenCode


tomzx:
# Background

The OpenCode server that OpenChamber connects to can sometimes become unstable or unavailable (e.g., when the server crashes, restarts, or the network drops). Currently there is no visible indication in the UI that the connection to the OpenCode server has been lost. This leaves users confused about why responses are not coming back, and makes it hard to tell whether the issue is on the client or server side.

An explicit connection status indicator would help users quickly diagnose issues and know when to restart the server or check their network.

# Version

fa6449be

# Acceptance Criteria

- [ ] A status indicator is visible in the OpenChamber UI (e.g., in a header bar, sidebar, or status bar) that shows whether the connection to the OpenCode server is healthy
- [ ] The indicator updates in near-realtime when the connection status changes (connected / disconnected / reconnecting)
- [ ] When the connection is lost, the indicator shows a clear visual state (color, icon, text) that distinguishes "disconnected" from "connected"
- [ ] The indicator works in both the web and desktop runtimes
- [ ] The indicator does not introduce significant overhead (e.g., polling at a reasonable interval or piggybacking on existing SSE/WS health signals)
- [ ] Graceful degradation: if the indicator cannot determine status, it shows an "unknown" state rather than falsely showing "connected"

---

Created with [create-issue](https://github.com/tomzx/dot-claude/blob/9af981285f50e7e4592d9d6c668a23d789b27c56/skills/create-issue/SKILL.md) (9af9812)

bashrusakh:
Related draft planning PR: #1751

This plan keeps the visible UI to a single compact status dot and uses short hover text for detail.

For #1556 specifically:
- the runtime ↔ OpenCode hop remains a distinct internal health source
- it is planned to feed the same aggregated user-facing indicator as #1737
- the UI stays lightweight while still making it clear when the OpenCode side is the unhealthy hop

Plan docs:
- `plans/issue-1737-connection-status/plan.md`
- `plans/issue-1737-connection-status/todo.md`

PR #1751, opened 2026-06-21
feat(ui): aggregated connection status indicator (#1737)

bashrusakh:
## Summary

Compact aggregated connection status indicator in the header, consolidating the two connection hops into a single user-facing dot with a hover tooltip showing per-hop breakdown.

Addresses #1737 and related #1556.

### Core feature
- Single compact status dot in the header showing aggregated connection state
- Hover tooltip with per-hop status breakdown:
  - **hop 1**: Frontend ↔ OpenChamber runtime
  - **hop 2**: OpenChamber runtime ↔ OpenCode
- Status precedence: offline → disconnected → reconnecting → connecting → degraded → connected → unknown

### State fixes
- `onDisconnect` now sets `connectionPhase = transportPhase` so legacy `connectionPhase` stays consistent with `runtimeTransportState.phase`
- Terminal disconnect reasons (`auth`, `ws_closed_before_ready`, etc.) correctly map to `"disconnected"` instead of `"reconnecting"`
- `connectionPhase` type widened to include `"disconnected"`

### Tooltip UX
- Header row shows "Connection status" (no dot — overall status is the main dot)
- Hop rows show individual per-hop dots aligned via CSS grid
- Dots are **display-only** — they reflect existing transport state, never change it

## Out of scope: intermittent reconnecting status

The frontend hop may briefly show `reconnecting` — this is real transport behavior, not a display bug.

### Root causes identified (not addressed in this PR)

1. **SSE heartbeat timeout**: Server sends `:heartbeat` comment frames every 15s, but the SDK's `onSseEvent` does not surface comment frames to the client. The idle-stream watchdog (30s) treats this as a timeout and triggers reconnect → UI shows `reconnecting`.

2. **First-message warmup**: `waitForConnectionOrThrow()` → `probeConnection()` can set `connectionPhase` to `reconnecting`/`connecting` during cold start when the session is still bootstrapping.

3. **Model switch**: `setModel`/`setCurrentVariant` are pure Zustand writes (no transport reconnect), but the subsequent `sendMessage` → `promptAsync` path goes through readiness checks that may coincide with stream reconnect timing.

### Recommended follow-up (separate PR)

- Root-cause fix in SSE heartbeat handling / SDK visibility

This PR focuses solely on the tooltip dot layout and per-hop status display.

## Known limitation

`openCodeRuntimeState` is only updated during startup (in `probeConnection`, `checkConnection`, `initializeApp` error handler). If OpenCode becomes unhealthy after startup while the frontend↔runtime transport remains up, hop 2 may incorrectly show "healthy" until the next reconnect cycle. This is by design (no new polling loop introduced) but worth monitoring in production.

## Validation

```bash
bun run --cwd packages/ui type-check   # passed
bun run --cwd packages/ui lint         # passed
```

22 unit tests for connection status mapper pass. 13 session-status-snapshot tests pass (including `shouldUseDisconnectedTransportPhase`).

## Files changed

- `packages/ui/src/components/layout/ConnectionStatusIndicator.tsx` — indicator + tooltip
- `packages/ui/src/lib/connection-status/connectionStatus.ts` — status model + tone mapping
- `packages/ui/src/lib/connection-status/connectionStatus.test.ts` — 22 mapper tests
- `packages/ui/src/sync/sync-context.tsx` — disconnect/phase handling
- `packages/ui/src/stores/useConfigStore.ts` — `connectionPhase` type + initial state
- `packages/ui/src/sync/__tests__/session-status-snapshot.test.ts` — 13 tests
- `packages/ui/src/lib/i18n/messages/*.ts` — removed unused `connectionStatus.tooltip.hopLabel` key (9 dictionaries)

## Screenshot
изображение


bashrusakh:
Thanks for the thorough review! Fixed:

1. **Click-to-navigate claim removed** — the PR description incorrectly stated 'Click dot to navigate to Connection settings'. There is no click handler and no dedicated connection settings page. Updated the description to remove this claim.

2. **Unused i18n key removed** —  was removed from all 9 dictionary files (commit 53808307).

3. **openCodeRuntimeState staleness** — added a 'Known limitation' section to the PR description documenting this behavior.

4. **handleOnline reason nit** — leaving as-is since the reason is likely stale after an online/offline cycle and would be overwritten by the next pipeline event.

btriapitsyn:
Resolved the conflict with `main` and integrated the aggregated indicator into the current transport pipeline:

- retained the two-hop status model (client → OpenChamber runtime → OpenCode);
- wired connect/disconnect/transport-switch and browser offline/online transitions into the narrow transport state;
- preserved current relay/routing/resync behavior and classified terminal auth/WS-before-ready failures separately from transient reconnects;
- verified 42 focused connection/sync tests, UI type-check, and lint.

btriapitsyn:

Issue #1346, opened 2026-05-21
[Bug] "bun run" causes a Bad descriptor error in terminal


sbonnick:
### What's wrong? Maybe some steps to reproduce.

[Bug] "bun run" causes a Bad descriptor error in terminal.  Error:

EBADF: /bash: Bad file descriptor (posix_spawn())

Regardless of underlying system, this seems to occur. tried in docker ubuntu, ubuntu desktop, and docker bun:1 image.

### Where does it happen?

Mobile (Web/PWA)

### Version (if known)

1.11.3

### Logs (optional)

```shell

```

sbonnick:
To reproduce, create a package.json:

```
{
  "scripts": {
    "list": "ls"
  }
}
```

Then run 
this in chat `! Bun run list` :  success
This on console in the UI `bun run list` :  failure

tomzx:
I think this might be a bug for any Linux OS.
I believe OpenChamber is mainly tested with MacOS so it may just be a matter of finding the code that needs to be updated to properly support Linux.
(I'm basing myself on #1483 which was fixed by adding a simple "or linux" condition)

sbonnick:
OK I think i root caused the issue. bun-pty is depending on portable-pty that is three years old. there is a fix in its latest version with simular issues. note however, portable-pty is still very old - 1 year since last change. redardless, i filed a bug against bun-pty: https://github.com/sursaone/bun-pty/issues/44

with that being said, I think it would be wise to stop using bun-pty. bun now has a native terminal thats faster and less of a supply chain risk. I will add a implementation using bun.terminal tonight. Since I am not sure all the potential impacts, I will add it behind a env flag for now and still leave bun-pty and node-pty in place. if its working well for others too, we can easily remove the flag and bun-pty dependancies.

yulia-ivashko:
Fixed by [#2280](https://github.com/openchamber/openchamber/pull/2280) and the IPC-environment cleanup in [cab078054](https://github.com/openchamber/openchamber/commit/cab078054), both included in v1.23.0. Terminal shells no longer inherit the closed daemon IPC descriptor that caused `bun run` to fail with `EBADF`. Thanks for tracing the terminal-only failure.

PR #1758, opened 2026-06-21
fix: gate Bun terminal PTY backend

sbonnick:
Found the likely cause for #1346: 

portable-pty's close_random_fds() closes the Rust exec-error pipe on Linux, which then corrupts glibc's posix_spawn error reporting in bash sub-process spawning. This was filed as portable-pty issue #7742 but hasn't been fixed in bun-pty's version yet. 

Beyond filing a issue with bun-pty to upgrade, I think moving towards `bun.terminal` may be better as its native to bun. I am not sure the the full impact, and cannot test on MAC. so to be safe I have gated the use of bun.terminal behind a environment variable for now until more people have tried it out. At which point we can remove the environment variable and bun-pty dependencies. 

Tested on Linux that this does in fact overcome issue #1346. With env `USE_BUN_TERMINAL_PTY=true` enabled, everything passes, with it off, I encounter issue #1346 . Also tested the change on Windows. on or off, bun commands are successful - no change as expected.

### DIFF
--- a/packages/web/server/lib/terminal/DOCUMENTATION.md
+++ b/packages/web/server/lib/terminal/DOCUMENTATION.md
@@ -37,7 +37,7 @@ The WebSocket path must remain in both `isUrlAuthWebSocketPath` and relay `ALLOW
 
 ## Verification
 
-Run:
+The web server can use Bun's built-in `Bun.Terminal` first when `USE_BUN_TERMINAL_PTY=true`; otherwise it falls back to `bun-pty` or `node-pty` to drive full-duplex PTY sessions.
 
 ```sh
 bun test packages/web/server/lib/terminal/runtime.test.js packages/web/server/lib/terminal/terminal-ws-protocol.test.js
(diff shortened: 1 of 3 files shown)


sbonnick:
Regarding feedback on `useCompty`, I checked Bun’s current docs, and as of Bun v1.3.14 on May 13, 2026, Bun.Terminal / Bun.spawn({ terminal }) on Windows already use ConPTY automatically. So the useConpty flag not being forwarded by the bun-terminal-pty adapter is not changing behavior for that backend; it’s effectively redundant there, while still relevant for bun-pty/node-pty. could add it "just-in-case"... but shouldn't really be a problem.

sbonnick:
Hey @btriapitsyn I know your back from holidays and probably very busy and focused on a lot of the new mobile features. Is there other maintainers I should toss on this for review instead?

sbonnick:
OK, cleaned it up given latest backend terminal changes. 

Still taking a safer toggle approach, but honestly, probably want to follow up with replacing bun-pty with this as its using all native bun functionality.

sbonnick:
I have also just created https://github.com/openchamber/openchamber/pull/2481 for a simpler fix to buns specific problems. It wont solve other IPC related issues, but is a smaller and safer fix.   will close this one

PR #1762, opened 2026-06-22
fix(worktree): gate sessions on bootstrap readiness

bashrusakh:
## Summary

- preflight fast worktree creation before returning a pending path so branch-in-use failures do not create empty worktree directories
- wait for worktree bootstrap readiness before creating sessions or sending messages against fast-created worktrees
- cover isolated multi-run ordering and branch-in-use fast-create failure with focused tests

## Validation

Passed in dependency-ready worktree:

```bash
bun test 'packages/ui/src/lib/worktrees/worktreeBootstrap.test.ts'
bun test 'packages/ui/src/lib/worktrees/worktreeManager.test.ts'
bun test 'packages/ui/src/stores/useMultiRunStore.test.ts'
bun test 'packages/web/server/lib/git/service.test.js'
git diff --check -- <target files>
```

Notes:

- Running the same targeted tests in the fresh PR worktree was partially blocked because that worktree does not have workspace dependencies available (`zustand`, `simple-git`).
- Earlier broad checks were blocked by existing unrelated project issues (`virtua`, `shiki`, missing locale keys), not by this patch.


btriapitsyn:
Could we make waiting for Worktree setup commands configurable per project?

Right now this change makes session creation / first send wait for full worktree bootstrap, which includes the configured Worktree setup commands. That is a useful strict-readiness mode, but for projects with slow setup commands it can make the first send block for quite a while.

I think the best behavior would be a per-project option next to the existing Worktree setup commands to wait for Worktree setup commands execution before creating/sending a session


Defaulting to the current fast behavior would preserve the existing UX, while still allowing projects that need strict setup readiness to opt in.

btriapitsyn:
Merged latest `main` into this PR and resolved the `session-ui-store.ts` conflict from #1761.

Fixed/updated:
- Kept the #1761 `materializeOpenDraftSession(...)` cleanup and applied this PR's bootstrap gating on top of it.
- Added a per-project Worktree setup option next to the existing Worktree setup commands: wait for setup commands before creating/sending a session.
- Default remains fast worktree behavior; full bootstrap waiting is now opt-in per project.
- Gated all new `waitForWorktreeBootstrap(...)` call sites behind that per-project setting.
- Added settings search/i18n coverage for the new checkbox.
- Updated the targeted multi-run test to cover both default fast behavior and opt-in wait behavior.
- Fixed the new git service test's macOS `/var` vs `/private/var` path assertion by comparing the canonical worktree path.

Validation run:
- `bun test 'packages/ui/src/stores/useMultiRunStore.test.ts'`
- `bun test 'packages/ui/src/lib/worktrees/worktreeBootstrap.test.ts'`
- `bun test 'packages/web/server/lib/git/service.test.js'`
- `bun run type-check`
- `bun run lint`
- `git diff --check`

btriapitsyn (approved):

PR #1770, opened 2026-06-22
fix(sync): SSE heartbeat timeout causes spurious reconnecting status

### LINKED ISSUE #1769: fix(sync): SSE heartbeat timeout causes spurious reconnecting status
## Problem

The connection status indicator intermittently shows `reconnecting` on the frontend hop. This is not a display bug — the frontend transport state genuinely transitions through `reconnecting` during normal operation.

### 1. SSE heartbeat timeout

The server sends `:heartbeat` comment frames every 15s to keep the SSE connection alive. However, the SDK's `onSseEvent` callback does not surface comment frames to the client — only actual event frames reset the heartbeat watchdog.

- Client heartbeat timeout: 30s (`DEFAULT_HEARTBEAT_TIMEOUT_MS`)
- Server heartbeat interval: 15s (`MESSAGE_STREAM_WS_HEARTBEAT_INTERVAL_MS`)
- Result: If no real events arrive for 30s, the idle-stream watchdog treats it as a timeout and triggers reconnect, even though the server is still sending heartbeats

**Fix**: Either surface comment frames in the SDK's `onSseEvent`, or increase the client heartbeat timeout to account for the comment-frame gap, or switch to a dedicated heartbeat event type that the SDK forwards.

### 2. First-message warmup

`waitForConnectionOrThrow()` → `probeConnection()` can set `connectionPhase` to `reconnecting`/`connecting` during cold start when the session is still bootstrapping. This happens when:

- A new session was just created via `createSession()`
- The sync layer hasn't finished bootstrap yet
- `isConnected` is still `false`
- The health probe runs and finds the connection not ready

**Fix**: Either debounce the connection phase transitions during session warmup, or separate "session bootstrap in progress" from "transport disconnected".

### 3. Model switch timing

`setModel`/`setCurrentVariant` are pure Zustand writes (no transport reconnect). However, the subsequent `sendMessage` → `promptAsync` path goes through readiness checks (`waitForConnectionOrThrow`) that may coincide with stream reconnect timing if the SSE/WS connection happens to be reconnecting at that moment.

**Fix**: This is likely acceptable behavior — the reconnect is real. Could be mitigated by the same debounce or heartbeat fix above.

## Impact

- User sees brief `reconnecting` flash in the header dot during normal usage
- The status is technically correct but may cause concern about connection stability
- No data loss or functional impact — messages still send/receive correctly

## Recommended fix

Root-cause fix in SSE heartbeat handling / SDK visibility (option 1 above). This addresses the most common trigger for spurious `reconnecting` states.

## Related

- PR #1751 — aggregated connection status indicator (display layer)
- Issue #1737 — original connection status indicator feature

bashrusakh:
## Summary

Fix SSE heartbeat timeout causing spurious `reconnecting` status in the connection indicator.

Fixes #1769.

## Problem

The server sends `:heartbeat` comment frames every 15s to keep SSE connections alive. However, the SDK's `onSseEvent` callback does not surface comment frames — only actual event frames reset the heartbeat watchdog. When no real events arrive for 30s, the client treats it as a timeout and triggers reconnect, even though the server is still sending heartbeats.

## Approach (TBD)

- [ ] Investigate SDK `onSseEvent` behavior with comment frames
- [ ] Decide between: surface comment frames in SDK, increase client timeout, or use a dedicated heartbeat event
- [ ] Implement the chosen fix
- [ ] Add tests for heartbeat timeout behavior
- [ ] Verify the connection indicator no longer shows spurious `reconnecting`

## Related

- Issue #1769 — root cause analysis
- PR #1751 — connection status indicator (display layer)


tomzx:
@openchamber-bot review

bashrusakh:
Closing this PR — the diff in this branch is the connection-status indicator work from PR #1751 (issue #1737, ), not a fix for the SSE heartbeat problem described in #1769.

Per the reproduction analysis by `openchamber-bot` on #1769:

- Root cause #1 (SDK `:heartbeat` comment frames not surfaced) is **not confirmed** for SDK 1.17.7 — `onSseEvent` in `event-pipeline.ts` is already invoked for every SSE chunk including comment-only frames, and `resetHeartbeat()` fires correctly.
- Root cause #2 (HTTP `probeConnection` independently flips `connectionPhase` to `reconnecting` while the SSE/WS stream is healthy) is the real trigger.
- Root cause #3 (warmup race in `waitForConnectionOrThrow` during session bootstrap) also confirmed.
- The branch is also `mergeable: CONFLICTING` — main has moved on, and several follow-up fixes (`d0d3121a`, `efbdf832`) have already landed elsewhere.

The actual fix for #1769 is being opened in a separate clean PR against current `main`. The indicator work in this branch should ride with PR #1751 / the indicator feature, not under a heartbeat title.

Issue #665, opened 2026-03-15
[Bug] Very high load and lag when ai is working


faldor20:
### What's wrong? Maybe some steps to reproduce.

I was trying to profile why openchamber was stuttering, locking up and using heaps of cpu.
I found it was fine until the ai is streaming responses and then performance would really suffer

I did some profiling to find the most egregious issues:
Animation: there seems to be animation callbacks running constantly. some of which are fairly slow. this is the main issue.

setItem. there are so so many calls to the relatively expensive set item method.

I did a little vibe coded experiment to to fix it, and de bouncing the animations pretty much solved the worst of it: 
https://github.com/faldor20/openchamber/pull/1/changes

I'll leave this to someone with more knowledge of the repo to implement properly.

It's definitely not perfect still. but the stutter and lock ups are solved by that fix. 
hopefully this helps get a proper fix in sooner

### Where does it happen?

Mobile (Web/PWA)

### Version (if known)

1.8.7

### Logs (optional)

```shell

```

mderazon:
For me, the biggest issue is navigating between chats (from one chat, go back to the chat lists and then entering another chat). It would sometime take minute or two to switch between chats and then I get this message:

<img width="528" height="191" alt="Image" src="https://github.com/user-attachments/assets/ea8dd676-8073-4b7e-97ae-a734de16f9c9" />

It might also be related to the size of the chat (if the chat is very long).

vscode extension v1.10.2

mderazon:
On chats with Deepseek v4-pro for example, where the context window is huge and the conversation can be very long. the chat becomes unresponsive

More severe - When on the chats list view, clicking to enter a chat with a very long history takes minutes

makeittech:
Closing as fixed for the original report: the streaming-time CPU/animation storm was substantially improved in 1.9.0+ (confirmed by multiple reporters). Remaining long-history / navigation lag should be tracked in newer performance issues rather than this 1.8.x report.

PR #1777, opened 2026-06-23
fix(electron): gracefully stop server before desktop quit

jwcrystal:
## Summary

Fixes Desktop quit/relaunch cleanup so Electron gives the embedded web server a chance to run its graceful shutdown path before the process exits.

Previously Desktop quit relied on `killSidecar()` only: a fire-and-forget SIGTERM plus short SIGKILL fallback followed by `app.exit(0)`. That bypassed the server handle's `stop()` path, which owns shutting down the OpenCode process, SSE watcher, terminal runtime, message streams, and HTTP server.

## CPU / memory impact

Orphaned OpenCode processes can keep consuming CPU and memory after Desktop exits or relaunches. This PR addresses that confirmed cleanup path by ensuring the embedded server gets a graceful shutdown opportunity before Electron terminates.

This should help cases where persistent background CPU/memory usage is caused by orphaned managed OpenCode processes. It does not claim to fix separate active-session renderer performance issues while AI is streaming.

Related: #665

## Previous shutdown flow

Before this change, Desktop quit/relaunch used this path:

```text
performConfirmedQuit()
  → prepareForQuit()
    → shutdownBackgroundServices()
      → killSidecar()
        → launchDetachedOpenCodeKiller()  // fire-and-forget
          → SIGTERM
          → detached killer script
          → sleep 100ms
          → SIGKILL
  → app.exit(0)
```

This meant Electron could exit before the server-owned graceful cleanup path ran.

## New shutdown flow

```text
performConfirmedQuit()
  → prepareForQuit()
    → shutdownBackgroundServices()
      → serverHandle.stop({ exitProcess: false })
        → gracefulShutdown()
          → stop watcher / terminal / message streams / HTTP server
          → close OpenCode process
      → fallback killSidecar() on failure or timeout
  → app.exit(0)
```

This PR:

- calls `serverHandle.stop({ exitProcess: false })` before Desktop quit/relaunch
- falls back to `killSidecar()` if graceful shutdown fails or times out
- keeps a 10s Electron-level shutdown timeout so quit cannot hang indefinitely
- prevents `before-quit` re-entry from bypassing the controlled shutdown path
- updates menu/restart/vibrancy/window-close flows to await cleanup before relaunch/exit

## Verification

- `node -c packages/electron/main.mjs`
- `bun run lint:electron`
- Manual: Desktop Quit verified
- Manual: Desktop Relaunch verified

## Notes

This fixes the confirmed orphaned opencode process cleanup path. It does not attempt to fix separate renderer CPU/memory pressure; that still needs profiling data before code changes.

### DIFF
--- a/packages/electron/main.mjs
+++ b/packages/electron/main.mjs
@@ -161,6 +161,7 @@ const DISCORD_INVITE_URL = 'https://discord.gg/ZYRSdnwwKA';
 const INSTALLED_APPS_CACHE_TTL_SECS = 60 * 60 * 24;
 const INSTALLED_APPS_CACHE_FILE = 'discovered-apps.json';
 const OPENCODE_SHUTDOWN_GRACE_MS = 100;
+const GRACEFUL_SHUTDOWN_TIMEOUT_MS = 10_000;
 
 const { autoUpdater } = updaterPkg;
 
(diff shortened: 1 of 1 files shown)


btriapitsyn:

Issue #1766, opened 2026-06-22
[Feature Request] Support OpenCode "steer" delivery / mid-turn message insertion (Codex-style)


InCerryGit:
## What I'm trying to do

When the agent is mid-flight, I want to *nudge or correct it without aborting and without waiting for the current turn to finish*. Today OpenChamber only offers two send modes (immediate / queued), neither of which expresses this. The OpenCode SDK already supports it (`delivery: "steer"`) and the official OpenCode desktop app uses it by default — OpenChamber just doesn't surface it.

## What I'd like

A first-class "Steer" send mode that calls `client.session.promptAsync({ ..., delivery: "steer" })`, exposed via:

1. **Plumbing** — thread an optional `delivery` through `sendMessage` → `routeMessage` → `opencodeClient.sendMessage` → `promptAsync`.
2. **Settings tri-state** — replace the boolean Queue Mode with `Steer (default) / Queue / Send immediately` under `Settings → Chat → Follow-up behavior`.
3. **Keyboard** — add `Alt+Enter` (or `⌘⇧Enter`) as a dedicated steer shortcut.
4. **Composer button** — when the session is busy, show a small "Steer" button next to Queue/Send.

Detailed background, code references, and suggested defaults are in the issue body.

bashrusakh:
Implemented in #1781 (draft PR). Summary:

- Plumbing: `delivery: "steer"` threaded through the send pipeline to `promptAsync` — SDK already supports it.
- Settings: Chat → Follow-up behavior is now a tri-state radio (Steer / Queue / Send immediately), replacing the old boolean queue-mode toggle. Migration from legacy persisted state is handled.
- Composer: when a session is busy, a floating Queue button appears (original style restored). Force-sending a queued message via chip click uses `delivery: "steer"` during a busy session, so steer is available via a two-gesture flow (queue → force-send). No separate Steer button in the composer.
- Keyboard: queue mode = Enter queues, Ctrl+Enter sends; otherwise Enter sends, Ctrl+Enter queues.
- Persistence/server/platform parity: DesktopSettings, web settings payload, and server settings sanitizer all handle the new key with legacy fallback.
- i18n: all 9 locales updated.

Validation: type-check passes (no new errors), lint clean.

makeittech:
[...]

PR #1781, opened 2026-06-23
feat(#1766): support OpenCode steer delivery / follow-up behavior settings

bashrusakh:
## Summary

Implements issue #1766 — adds a first-class follow-up behavior setting that replaces the boolean queue-mode toggle with three modes: Steer, Queue, and Send immediately.

### What changed

**Plumbing** — optional `delivery: "steer"` is now threaded through the entire send pipeline (`sendMessage` → `routeMessage` → `opencodeClient.sendMessage` → `promptAsync`). The SDK already supports this parameter.

**Store** — `messageQueueStore` exposes `followUpBehavior: "steer" | "queue" | "immediate"` (default: `"queue"`). A migration helper converts legacy `queueModeEnabled` persisted state on first load.

**Settings** — Settings → Chat → Follow-up behavior shows three radio options (Steer / Queue / Send immediately) using the existing project UI patterns. The settings search registry was updated accordingly.

**Composer** — when the session is busy, a single floating Queue button appears above the abort button (restored to the original visual style). No separate steer button in the composer. Steer is available via a two-gesture flow: Enter (queue) → click chip (force-send as steer). Force-sending a queued message via its chip always passes `delivery: "steer"` — the `handleSubmit` gateway drops the flag for idle sessions, so only busy sessions actually steer.

**Keyboard**
- Queue mode: Enter queues, Ctrl+Enter sends immediately
- Steer mode: Enter steers, Ctrl+Enter sends immediately
- Immediate mode: Enter sends immediately, Ctrl+Enter queues

**Persistence** — `DesktopSettings`, web settings payload, and server-side settings sanitizer all handle the new `followUpBehavior` key with automatic fallback from legacy `queueModeEnabled`.

**i18n** — all 9 locales updated with follow-up behavior section/option keys.

### Validation

- `bun run type-check` — clean
- `bun run lint` — clean
- `bun run build` — clean


bashrusakh:
Addressed fix-level review findings (cdaf015c):

- **Major fix**: `followUpBehavior === "steer"` was a no-op in `handlePrimaryAction` and the keyboard handler. Both fell into the `else` branch and sent without `delivery: "steer"`, so the Steer setting produced identical behavior to Send immediately. Now: in steer mode + busy session, Enter and the send button call `handleSubmit({ delivery: "steer" })` directly. Ctrl+Enter in steer mode sends immediately (escape hatch, consistent with queue mode where Ctrl+Enter bypasses the queue).
- **Minor**: removed dead `chat.chatInput.actions.queue` i18n key from all 9 locales (was unused after the Steer button was removed from the composer).
- Cosmetic findings (untranslated locale keys, stale `queueMessagesByDefault*` keys, server/client default style) deferred to a separate cleanup pass.

Validation: type-check clean, lint clean.

bashrusakh:
Addressed OCR review feedback (cb9a850d):

**Medium** — nested ternary in `persistence.ts:411-415` replaced with explicit `let`/`if`/`else if` chain per project code style (CONTRIBUTING.md). Also explicitly imports `FollowUpBehavior` type from messageQueueStore.

**Low** — untranslated follow-up behavior keys in non-English locale files deferred to translation pass (consistent with existing pattern across many settings keys in those files).

bashrusakh:
Rebased on `origin/main` (99f3527a). Previous merge + preexisting variant-key errors cleaned up.

Validated:
- `bun run type-check` — clean
- `bun run lint` — clean
- `bun run build` — clean

bashrusakh:
OCR re-review on rebased branch (21 files): 0 high, 1 medium, 5 low.

**Medium (out of scope, not fixed)** — `OpenChamberVisualSettings.tsx:1961-1979`: nested interactive elements in the followUpBehavior radio group (`role="button"` + `Radio` child). This creates invalid ARIA semantics (button inside button). However, this matches the existing pattern used for other radio groups in the same file (`userMessageRendering`, `mermaidRendering`, `diffLayout`) — not a regression from this PR, and fixing just our section would create inconsistency. Recommend a separate cleanup PR to fix all radio groups.

**Low** — untranslated followUpBehavior keys in 5 non-English locale files (es, fr, ko, pl, zh-CN). Deferred to future translation pass.

**Prior review findings confirmed resolved**: steer mode no-op fixed, dead i18n key removed, nested ternary flattened.

btriapitsyn:
Merged the latest `main` into this branch to resolve the conflicts, and fixed one CI-blocking issue found in review. 🙏

**CI fix — missing Japanese locale**

- `packages/ui/src/lib/i18n/messages/messages.test.ts` enforces exact key parity between every locale and English. This PR added the 7 new `followUpBehavior.*` keys to 9 locales but not `ja.settings.ts` (there are 10 locales — `ja` is the 10th), so the parity test would have failed in CI. Added the keys to `ja.settings.ts` with proper Japanese translations (rather than English placeholders, since that file is otherwise fully translated).

**Conflicts resolved**

- `ChatInput.tsx` — `main` added auto-review handling to the send path (queue-on-open-question dismiss, and `autoReviewRunning` widening the `canQueue` guard). Combined both sides: kept `main`'s auto-review guards and `sessionPhase !== 'idle' || autoReviewRunning` condition, on top of this PR's three-mode `followUpBehavior` logic and the `delivery: 'steer'` threading. Effect/dep arrays updated to include `autoReviewRunning` + `followUpBehavior`.
- `OpenChamberVisualSettings.tsx` — `VisibleSetting` union conflict; kept `followUpBehavior` (the feature) and dropped the now-stale `queueMode`. Left the type non-exported to match `main`'s style (nothing imports it).

**Validation**

- `bun test src/lib/i18n/messages.test.ts` — 2 pass (parity restored).
- `bun run type-check` (packages/ui) — clean.
- `bun run lint` on the resolved + `ja` files — clean.

**Review notes**

- Verified clean: the legacy `queueModeEnabled → followUpBehavior` migration (store `version: 1` + `migrate`; `true → queue`, `false → immediate`, missing → default `queue`), the client/server `normalizeFollowUpBehavior` agreeing, the `delivery: 'steer'` gating (only applies when `sessionPhase !== 'idle'`, so idle sessions never steer), and the per-mode Enter / Ctrl+Enter keyboard semantics.
- Kept `.graymatter/` in `.gitignore` at the maintainer's request (keeps future agent-memory files out of git).

Thanks for the feature!

btriapitsyn:
Follow-up pushed: collapsed the three follow-up modes down to two (**Queue** + **Steer**), removing **Immediate**.

**Why**

Checked the OpenCode side: `delivery` is `Schema.Literals(["steer", "queue"])` — only two values exist server-side — and the core defaults to `?? "steer"`. So an "immediate" send (no delivery flag) on a *busy* session already steered into the running turn, making **Immediate and Steer wire-identical**. The three-radio UI was exposing two settings that did the same thing. On an idle session all options collapse to "just start a turn" anyway.

**What changed**

- Removed the `immediate` mode: `FollowUpBehavior` is now `'steer' | 'queue'`, the settings radio drops the third option, and the keyboard handler's `immediate` branch is gone.
- **Queue is untouched** — still the client-side queue (with edit/reorder), not the server `delivery: 'queue'` inbox. That was intentional, to keep the existing queued-message editing/reordering UX.
- Migration is behavior-preserving: any persisted/legacy `'immediate'` (and legacy `queueModeEnabled === false`) now maps to `'steer'`, both client-side (`normalizeFollowUpBehavior`) and in the server settings sanitizer.
- Removed the `followUpBehavior.immediate.label` key from all 10 locales (keeps the i18n parity test green) and narrowed the `followUpBehavior` union in `api/types.ts` + `desktop.ts`.

**Validation**

- `bun test src/lib/i18n/messages.test.ts` — 2 pass.
- `bun run type-check` (packages/ui) — clean.
- `bun run lint` on changed files — clean.

Net: Steer and Queue are now genuinely distinct (steer = into the running turn; queue = held client-side until idle), and we no longer ship a mode that silently duplicates another.

btriapitsyn (approved):

PR #1794, opened 2026-06-23
fix(ui): allow scrollbar drag to release auto-follow and prevent thinking-block LERP flicker

### LINKED ISSUE #1793: Auto-follow scroll: scrollbar drag cannot release follow, thinking-block height animation causes flicker
### Summary

Three interacting bugs in `useChatAutoFollow` cause:

1. Programmatic-write window never expires, so `handleScrollEvent` discards all scroll events during active LERP
2. Scrollbar dragging has no escape path from auto-follow (only wheel, touch, and keyboard do)
3. Thinking-block height animation causes bidirectional LERP oscillation and visible flicker

---

### Bug 1 — Programmatic-write window never expires

**Location:** `tickFollow()` at [L183-226](packages/ui/src/hooks/useChatAutoFollow.ts), `markProgrammaticWrite()` at [L164-167](packages/ui/src/hooks/useChatAutoFollow.ts)

`tickFollow` runs every ~16ms via `requestAnimationFrame`. Each invocation calls `markProgrammaticWrite()`, which sets `programmaticWriteUntilRef.current = performance.now() + 200ms`. Because this is refreshed every frame, the 200ms window is perpetually pushed forward and never actually expires while the LERP loop is active.

This is not a problem by itself, but it gates `handleScrollEvent` (see Bug 2).

---

### Bug 2 — Scrollbar drag has no escape from auto-follow

**Location:** `handleScrollEvent()` at [L444-489](packages/ui/src/hooks/useChatAutoFollow.ts)

There are four user-input channels that should be able to release auto-follow:

| Channel | Handler | Trigger | Checks programmatic window? |
|---------|---------|---------|---------------------------|
| Mouse wheel (scroll up) | `handleWheel` (L495) | `deltaY < 0` | No |
| Touch (swipe down) | `handleTouchMove` (L506) | finger move > 2px down | No |
| Keyboard (Up / PageUp / Home) | `handleKeyDown` (L524) | key press | No |
| Scroll events (scrollbar drag, etc.) | `handleScrollEvent` (L444) | any `scroll` event | **Yes** |

`handleScrollEvent` calls `isInProgrammaticWindow()` at L448. When the LERP loop is active (Bug 1), this always returns `true`, so the function returns early at L449-451 — the user-intent detection at L453-470 is **never reached**.

**Impact:** Dragging the scrollbar upward during streaming does nothing. The auto-follow LERP immediately pulls the view back to the bottom. The user is trapped in `'following'` state. The only escapes are the mouse wheel, touch, or keyboard — scrollbar drag is the sole input channel with no release path.

The data needed to detect this case is already captured: `lastScrollTopRef` is updated at L445 with the value from the *previous* LERP frame write, and `currentTop` at L443 reflects the scroll event triggered by the user's drag. The comparison `currentTop < previousTop` would correctly identify the user's intent — it just never executes.

---

### Bug 3 — Thinking-block height animation causes LERP oscillation and flicker

**Location:** `tickFollow()` at [L195-225](packages/ui/src/hooks/useChatAutoFollow.ts), `ResizeObserver` at [L561-577](packages/ui/src/hooks/useChatAutoFollow.ts), `onAnimatedHeightChange` callback at [L598](packages/ui/src/hooks/useChatAutoFollow.ts)

**Trigger chain:**

```
thinking block animated expand/contract (per-frame height change)
  ├─→ ResizeObserver fires (L565-567) → startFollowLoop()
  └─→ onAnimatedHeightChange callback (L598) → kick() → startFollowLoop()
```

`startFollowLoop()` restarts the LERP loop if it had settled. Since the thinking block animates height every frame, the loop is continuously restarted.

**LERP math during contraction:**

```ts
const target = scrollHeight - clientHeight;  // shrinks when block contracts
const delta = target - current;               // delta < 0 (target moved above us)
const next = current + delta * 0.18;          // LERP pulls scrollTop *upward*
```

When the thinking block **contracts**, `scrollHeight` decreases → `target` moves up → `delta < 0` → LERP pulls `scrollTop` upward, *away from the new bottom*.

When it **expands** again in the next animation frame, `delta > 0` → LERP pulls `scrollTop` back down.

This creates a 0.3–0.5s-period visible flicker as the LERP reverses direction every frame in sync with the height animation.

**Why normal streaming is unaffected:** During regular text streaming, `scrollHeight` increases monotonically, so `delta >= 0` always — the LERP only pushes downward, never pulls upward. The bidirectional oscillation is specific to content that both grows and shrinks, i.e., animated thinking/code blocks.

panzeyu2013:
Closes #1793.

### Problem 1 & 2 — Scrollbar drag cannot release auto-follow

`tickFollow()` calls `markProgrammaticWrite()` every ~16ms, keeping the 200ms programmatic-write window perpetually fresh. `handleScrollEvent()` checks this window and returns early when inside it, discarding all user-intent detection. While wheel, touch, and keyboard handlers bypass this guard, scroll events (scrollbar drag) do not — making scrollbar drag the only input channel with no escape from auto-follow.

Fix: in `handleScrollEvent`, detect user-initiated upward scrolling (`currentTop < previousTop`) even inside the programmatic window, and release the follow state if the user is scrolling up.

### Problem 3 — Thinking-block height animation causes LERP oscillation

When a thinking block animates its height (expand/contract per frame), `scrollHeight` oscillates. On contraction, `scrollHeight` decreases → `target` moves up → `delta < 0` → the LERP pulls `scrollTop` *upward*, away from the new bottom. On the next expansion frame, `delta > 0` → LERP pulls back down. This creates a 0.3–0.5s-period visible flicker during thinking-block animation.

Fix: in `tickFollow()`, when `delta < 0` (content contracted), snap directly to `target` instead of lerping. Expansion (normal streaming, `delta >= 0`) retains the smooth LERP behavior.

### DIFF
--- a/packages/ui/src/hooks/useChatAutoFollow.ts
+++ b/packages/ui/src/hooks/useChatAutoFollow.ts
@@ -212,10 +212,16 @@ export const useChatAutoFollow = ({
         }
 
         settledFramesRef.current = 0;
-        const next = current + delta * LERP;
-        markProgrammaticWrite();
-        container.scrollTop = next;
-        lastScrollTopRef.current = container.scrollTop;
+        if (delta < 0) {
+            markProgrammaticWrite();
+            container.scrollTop = target;
+            lastScrollTopRef.current = target;
+        } else {
+            const next = current + delta * LERP;
+            markProgrammaticWrite();
+            container.scrollTop = next;
+            lastScrollTopRef.current = container.scrollTop;
+        }
         followRafRef.current = window.requestAnimationFrame(tickFollow);
     }, [markProgrammaticWrite, stopFollowLoop]);
 
@@ -447,6 +453,12 @@ export const useChatAutoFollow = ({
         updateOverflowAndButton();
 
         if (programmatic) {
+            if (currentTop < previousTop && stateRef.current === 'following') {
+                stopFollowLoop();
+                stopSettleBurst();
+                lastUserReleaseAtRef.current = typeof performance !== 'undefined' ? performance.now() : Date.now();
+                setStateValue('released');
+            }
             return;
         }
 


panzeyu2013:
fixed

PR #1795, opened 2026-06-23
fix(vscode): parse MiniMax CN remaining percentages

lilyzhaun:
## Problem

The MiniMax CN quota endpoint reports real usage through remaining-percent fields, while the count fields stay at `0`.

A live response after using quota looks like this:

```json
{
  "model_remains": [
    {
      "start_time": 1782230400000,
      "end_time": 1782248400000,
      "remains_time": 7557133,
      "current_interval_total_count": 0,
      "current_interval_usage_count": 0,
      "model_name": "general",
      "current_weekly_total_count": 0,
      "current_weekly_usage_count": 0,
      "weekly_start_time": 1782057600000,
      "weekly_end_time": 1782662400000,
      "weekly_remains_time": 421557133,
      "current_interval_status": 1,
      "current_interval_remaining_percent": 93,
      "current_weekly_status": 1,
      "current_weekly_remaining_percent": 97
    }
  ],
  "base_resp": {
    "status_code": 0,
    "status_msg": "success"
  }
}
```

The VS Code parser read these count fields:

```text
current_interval_total_count = 0
current_interval_usage_count = 0
current_weekly_total_count = 0
current_weekly_usage_count = 0
```

That makes both windows resolve to `usedPercent: null` because `total_count` is `0`:

```text
5h: total=0, usage=0 -> usedPercent=null
weekly: total=0, usage=0 -> usedPercent=null
```

As a result, MiniMax CN has no usable percentage for the 5h and weekly windows in the VS Code quota provider, so the quota UI cannot calculate pace/prediction from those windows.

## Fix

For the MiniMax CN path, read the remaining-percent fields instead:

```text
current_interval_remaining_percent = 93
current_weekly_remaining_percent = 97
```

Then convert them into used percentages:

```text
5h: 100 - 93 = 7%
weekly: 100 - 97 = 3%
```

The non-CN MiniMax path still uses the existing count-based calculation.

## Validation

- `bun test packages/vscode/src/quotaProviders.minimax.test.ts`
- `bun run vscode:type-check`
- `bun run --cwd packages/vscode lint`
- Manual smoke: `fetchMiniMaxCnCodingPlanQuota()` returned `5h.usedPercent = 7` and `weekly.usedPercent = 3` from the live MiniMax CN response.

### DIFF
--- a/packages/vscode/src/quotaProviders.minimax.test.ts
+++ b/packages/vscode/src/quotaProviders.minimax.test.ts
@@ -1,0 +1,17 @@
+import { strict as assert } from 'node:assert';
+import { describe, test } from 'node:test';
+
+import { remainingPercentToUsedPercent } from './quotaProviders';
+
+describe('remainingPercentToUsedPercent', () => {
+  test('maps MiniMax remaining percentages to used percentages', () => {
+    assert.equal(remainingPercentToUsedPercent(100), 0);
+    assert.equal(remainingPercentToUsedPercent(94), 6);
+  });
+
+  test('clamps invalid remaining percentages', () => {
+    assert.equal(remainingPercentToUsedPercent(120), 0);
+    assert.equal(remainingPercentToUsedPercent(-20), 100);
+    assert.equal(remainingPercentToUsedPercent(null), null);
+  });
+});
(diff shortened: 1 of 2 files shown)


btriapitsyn:

PR #1806, opened 2026-06-24
fix(worktree): subagent sessions kept when deleting worktree group from sidebar

### LINKED ISSUE #1805: [BUG] worktree: subagent sessions kept when deleting worktree group from sidebar
**Steps to reproduce:**
1. Open a worktree group in sidebar
2. Click the delete button (trash icon) on the group header
3. Confirm the worktree removal dialog

**Actual:** Only the git worktree is removed. All sessions associated with the worktree (including subagent sessions) remain in the database and UI as orphans.

**Expected:** Sessions in the removed worktree should be archived (matching the behavior of deleting through Settings → Worktrees).

**Root cause:**
`SessionGroupSection.tsx` → `allGroupSessions` is guarded by `group.isArchivedBucket`, returning `[]` for active worktree groups. The worktree delete button sends an empty session list, so `SessionDialogs` skips archiving and only removes the git worktree directory.

**Fix:** Remove the `isArchivedBucket` guard so `allGroupSessions` always collects sessions from `sourceGroupNodes` (including recursive children).

bashrusakh:
Closes #1805

## Problem

Subagent sessions were left orphaned in several worktree/session deletion flows:

1. **Sidebar worktree delete**: the group header trash icon sent an empty session list, so only the git worktree directory was removed; all sessions stayed in the database.
2. **Hard-delete of archived sessions with descendants**: the client sent only the root session ID and relied on the server to cascade-delete children. If the cascade failed, children remained.
3. **Settings / BranchPicker worktree delete**: `useSessions()` is scoped to the current sync directory, so subagent sessions created in other directories were missed.
4. **Worktree metadata leak**: deleted sessions kept their worktree attachment in the worktree store.

## Changes

- `SessionGroupSection.tsx`: remove the `isArchivedBucket` guard on `allGroupSessions` so the worktree delete button includes all sessions in the group (including nested children).
- `useSessionActions.ts`: hard-delete now sends `deleteSessions([root, ...descendants])` instead of trusting server-side cascade. Already-cascade-deleted children return 404, which `deleteSession` treats as success.
- `session-actions.ts`: clean up a session's worktree metadata on successful delete and on 404 (already-deleted).
- `WorktreeSectionContent.tsx` & `BranchPickerDialog.tsx`: search subsessions across all global active + archived sessions, not just the current sync directory.

## Verification

- `bun run type-check` — only pre-existing unrelated errors (missing `virtua` types, etc.).
- `bun run lint` — no new issues in changed files.
- `collectGroupSessions` recursively walks `SessionNode.children` via `childrenMap`, capturing nested subagent sessions.
- `deleteSessions` deletes IDs sequentially; 404s do not trigger rollback.

Fixes #1805


btriapitsyn (approved):

btriapitsyn:
Pushed a merge of the latest `main` into this branch to resolve the conflicts. 🙏

**Conflict resolved**

- The only conflict was `packages/ui/src/components/session/BranchPickerDialog.tsx`, which `main` deleted in #1835 (dead-code removal) while this PR modified it. The component has no importers anywhere in the tree, so I accepted the deletion — the PR's edit to it was moot. The cross-directory subsession fix still lives where it matters: `WorktreeSectionContent.tsx` (Settings) and the sidebar delete paths.
- All four functional changes are intact after the merge: the `isArchivedBucket` guard removal in `SessionGroupSection`, the `deleteSessions([root, ...descendants])` switch in `useSessionActions`, the `cleanupSessionWorktreeMetadata` calls in `session-actions`, and the global-store subsession search in `WorktreeSectionContent`.

**Validation**

- `bun run type-check` (packages/ui) — clean.
- `bun run lint` on the four changed files — clean.

**Review**

- The bot review approved with no actionable findings.
- My read agrees the core logic is sound — verified `deleteSessions` treats 404 as success without a bulk rollback, and `setWorktreeMetadata(id, null)` clears both the attachment store and the local map.
- One non-blocking observation for a possible follow-up: the sidebar worktree-delete (`SessionGroupSection`) collects from the rendered group tree (`collectGroupSessions(sourceGroupNodes)`) rather than the global sessions store like the Settings path now does. If a subagent session for the worktree lives in another directory and isn't nested under this group in the sidebar tree, it could still be missed by the sidebar trash icon. Left as-is to match your design — flagging only so it can be confirmed/followed up separately.

Thanks for the fix!

PR #1815, opened 2026-06-24
fix: persist manual model selection across app restart

### LINKED ISSUE #1801: [Bug] Default model always resets to Big Pickle on app restart
### Description

When I close and reopen the OpenChamber app, the selected model always defaults back to **Big Pickle**. The app does not persist or remember the last selected model across sessions.

### Steps to Reproduce

1. Open OpenChamber — observe the default model is "Big Pickle"
2. Change the model to a different one (e.g. Claude Sonnet, GPT-4o etc.)
3. Close the app
4. Reopen OpenChamber
5. **Expected**: The previously selected model is remembered
6. **Actual**: The model resets back to "Big Pickle"

### Expected Behavior

The app should persist the user’s last selected model (e.g. in local storage or a config file) and restore it on next launch.

### Environment

- OpenChamber version: latest
- Platform: Desktop / Web
- OS: Windows 10

### Additional Context

This is a common UX friction point — users have to manually switch the model every time they launch the app.

bashrusakh:
## Summary

Fixes #1801 — the selected model always reset to Big Pickle on app restart because `applyDefaultModelAgentSelection()` unconditionally used the bare default cascade and set `selectionSource: "auto"`, discarding the users persisted manual selection.

## Root cause

Two contributing factors:

1. **`applyDefaultModelAgentSelection()`** (`useConfigStore.ts:2518`) used `resolveDefaultAgentModelSelection()` directly and hardcoded `selectionSource: "auto"`. It never called `resolveSelectionWithManualGuard()` — the same function used by `setProvider`, `setModel`, `setCurrentVariant`, and `loadAgents` to preserve manual selections.

2. **`selectionSource` was missing from `partialize`** (`useConfigStore.ts:3283-3306`), so the top-level manual flag was not directly persisted. (It was indirectly restored via `directoryScoped` + `hydrateActiveDirectorySnapshot`, but the primary fix is item 1.)

### `packages/ui/src/stores/useConfigStore.ts`

- **`partialize`**: Added `selectionSource: state.selectionSource` for defense-in-depth.
- **`applyDefaultModelAgentSelection()`**: Now calls `resolveSelectionWithManualGuard()` which checks `selectionSource`. On restart, the flag is `"manual"` (restored by `hydrateActiveDirectorySnapshot`), so the guard preserves the users persisted model choice. Without a prior manual selection, `selectionSource` is `"auto"` and the guard falls through to the cascade defaults as before.

### `packages/ui/src/sync/session-ui-store.ts`

- Updated the comment at line 680 to reflect that manual selection is now preserved.

## Behavioral contract

- **App restart**: Manual model is preserved (the fix).
- **Intra-session fresh draft**: Manual model is also preserved (consistent with `setProvider`, `setModel`, `loadAgents` — only `applyDefaultModelAgentSelection` was previously inconsistent).
- **No prior manual selection**: Falls through to cascade defaults (Big Pickle) — unchanged.

## Validation

- `bun run type-check` — all workspaces pass
- `bun run lint` — all workspaces pass


btriapitsyn:
Resolved the conflict with `main` and preserved manual model persistence alongside newer project defaults:

- persisted `selectionSource` so a manual choice survives app restart;
- kept the manual-selection guard when entering fresh drafts;
- retained `main`’s higher-priority `project.defaultModel` option and directory-scoped config snapshots;
- verified 18 focused config-store tests, UI type-check, and lint.

btriapitsyn:

PR #1828, opened 2026-06-25
fix: sidebar worktree drag reorder snap-back due to stale cache

### LINKED ISSUE #1827: Sidebar worktree drag-to-reorder snaps back to original position immediately after drop
**Describe the bug**

When dragging a worktree group in the session sidebar to reorder it, the item visually snaps back to its original position immediately after dropping. However, the new order is correctly persisted — switching from remote server to local and back (which remounts the sidebar) shows the worktrees in the intended order.

**To Reproduce**

1. Open a project with multiple worktrees in the session sidebar
2. Drag a worktree group to a new position
3. Release the drag — the item snaps back to its original position
4. Switch to a different server then back (or remount the sidebar component)
5. The worktrees are now in the correct order

**Expected behavior**

The worktree should stay in the new position immediately after dropping, matching the persisted order.

**Root cause**

`SidebarProjectsList.tsx` (lines 109-125) has `cachedGetOrderedGroups` which caches ordered groups in `orderedGroupsCacheRef` keyed by `projectId` + groups array reference. The cache hit at line 112 only checks `hit.groups === groups` (reference equality) — it does not track whether the `getOrderedGroups` callback has changed.

After a drag, `onDragEnd` at lines 260-264 calls `setGroupOrderByProject`, updating the `groupOrderByProject` state. The parent re-renders and passes a new `getOrderedGroups` callback. However, `projectSections` is behind a `useMemo` that does not depend on `groupOrderByProject`, so `section.groups` keeps the same reference. The cache hits at line 112 and returns the stale pre-drag ordering, causing the visual snap-back.

**Fix**

Add a ref (`orderedGroupsCacheGetOrderedGroupsRef`) that detects when `getOrderedGroups` changes and clears the cache:

```tsx
const orderedGroupsCacheGetOrderedGroupsRef = React.useRef<typeof props.getOrderedGroups>(props.getOrderedGroups);
if (orderedGroupsCacheGetOrderedGroupsRef.current !== props.getOrderedGroups) {
  orderedGroupsCacheGetOrderedGroupsRef.current = props.getOrderedGroups;
  orderedGroupsCacheRef.current.clear();
}
```

panzeyu2013:
**Problem**

Dragging a worktree group in the session sidebar to reorder it causes the item to snap back to its original position immediately after dropping. The order is correctly persisted (visible after remount), but the visual feedback is broken.

**Root cause**

`cachedGetOrderedGroups` in `SidebarProjectsList.tsx` (lines 109-125) caches ordered groups by `projectId` + groups array reference. After a drag, `groupOrderByProject` changes, producing a new `getOrderedGroups` callback. But the cache only checks `projectId` and the groups array reference (line 112) — both stable across the re-render because the `projectSections` memo does not depend on `groupOrderByProject`. The cache returns stale pre-drag data, and the UI snaps back.

**Fix**

Track the previous `getOrderedGroups` reference with a ref. When it changes (indicating `groupOrderByProject` changed), clear the cache so the next call recomputes the ordering:

```tsx
const orderedGroupsCacheGetOrderedGroupsRef = React.useRef<typeof props.getOrderedGroups>(props.getOrderedGroups);
if (orderedGroupsCacheGetOrderedGroupsRef.current !== props.getOrderedGroups) {
  orderedGroupsCacheGetOrderedGroupsRef.current = props.getOrderedGroups;
  orderedGroupsCacheRef.current.clear();
}
```

**Changes**

- `packages/ui/src/components/session/sidebar/SidebarProjectsList.tsx` — added `orderedGroupsCacheGetOrderedGroupsRef` to detect `getOrderedGroups` changes and clear the cache.

Closes #1827

### DIFF
--- a/packages/ui/src/components/session/sidebar/SidebarProjectsList.tsx
+++ b/packages/ui/src/components/session/sidebar/SidebarProjectsList.tsx
@@ -101,6 +101,11 @@ export function SidebarProjectsList(props: Props): React.ReactNode {
   // orderedGroups.filter/find work and any consumer-memoization see a
   // stable reference.
   const orderedGroupsCacheRef = React.useRef<Map<string, { groups: SessionGroup[]; ordered: SessionGroup[] }>>(new Map());
+  const orderedGroupsCacheGetOrderedGroupsRef = React.useRef<typeof props.getOrderedGroups>(props.getOrderedGroups);
+  if (orderedGroupsCacheGetOrderedGroupsRef.current !== props.getOrderedGroups) {
+    orderedGroupsCacheGetOrderedGroupsRef.current = props.getOrderedGroups;
+    orderedGroupsCacheRef.current.clear();
+  }
   const cachedGetOrderedGroups = (projectId: string, groups: SessionGroup[]): SessionGroup[] => {
     const cache = orderedGroupsCacheRef.current;
     const hit = cache.get(projectId);


panzeyu2013:
fixed by author

PR #1845, opened 2026-06-26
fix(electron): use server UI on Windows

### LINKED ISSUE #1841: 1.13.3 openchamber-ui:// 卡顿
## 问题描述
1.13.3 版本，卡顿，1.13.2 正常。

## 复现步骤
启动 OpenChamber Desktop 1.13.3，查看页面加载状态。

## 期望行为
和 1.13.2 一样流畅。

## 实际行为
私有协议 openchamber-ui:// 加载卡顿。
## 环境
- OpenChamber Desktop 1.13.3
- Windows

snvtac:
## Summary
- switch packaged Windows desktop builds to load the bundled UI through the local web server instead of the `openchamber-ui://` custom protocol by default
- keep explicit env overrides: `OPENCHAMBER_ELECTRON_LOAD_SERVER_UI=1` still forces server UI, and `OPENCHAMBER_ELECTRON_USE_BUNDLED_UI=1` can still force the custom protocol path for regression testing
- document the env var behavior and add unit coverage for the packaged UI policy

## Why
Issue #1841 reports that OpenChamber Desktop 1.13.3 on Windows becomes sluggish and loses client-server interactions when loaded from `openchamber-ui://`, while forcing the server UI path with `OPENCHAMBER_ELECTRON_LOAD_SERVER_UI=1` restores normal behavior. The desktop server already serves the same packaged `web-dist` assets from `OPENCHAMBER_DIST_DIR`, so this changes the Windows packaged transport without changing the UI build or server API layer.

## Validation
- `bunx vitest run packages/electron/packaged-ui-policy.test.mjs packages/electron/opencode-cwd.test.mjs`
- `bun run type-check:electron`
- `bun run lint:electron`
- `git diff --check`
- `bun run type-check`
- `bun run lint`
- `bun run build`

Fixes #1841

### DIFF
--- a/packages/electron/README.md
+++ b/packages/electron/README.md
@@ -79,7 +79,8 @@ The package supports macOS and Windows desktop features. Some native discovery h
 | Variable | Use |
 |----------|-----|
 | `OPENCHAMBER_ELECTRON_DEV=1` | Marks the runtime as desktop development mode |
-| `OPENCHAMBER_ELECTRON_USE_BUNDLED_UI=1` | Uses staged web assets instead of the HMR dev server |
+| `OPENCHAMBER_ELECTRON_LOAD_SERVER_UI=1` | Forces the UI to load through the local web server instead of the packaged custom protocol |
+| `OPENCHAMBER_ELECTRON_USE_BUNDLED_UI=1` | Forces staged web assets through the packaged custom protocol; by default packaged Windows builds use the local server for realtime stability |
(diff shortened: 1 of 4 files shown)


bashrusakh:
I think this works around the symptom, but it does not remove the blocking startup work.

The direct blocker I found is the managed process reaper: it does sync fs work and `spawnSync("tasklist")`/`taskkill` during startup. Since the web server is in the Electron main process, that can stall UI loading.

I opened a smaller root-cause draft PR: #1854

It keeps the existing reaper rules and makes that path async instead.

btriapitsyn:

PR #1847, opened 2026-06-26
fix(vscode): reset settings page to home on settings entry (#1764)

### LINKED ISSUE #1764: [Bug] VS Code extension: Settings command re-opens the last-visited section instead of the settings home

When the OpenChamber VS Code extension is already showing a non-default settings section (e.g. Providers, Skills, MCP, etc.), clicking the **Settings** command in the VS Code title bar no longer takes the user to the settings home/menu. Instead, the previously-selected section is re-rendered, so the user keeps landing on whichever section they were on last.

**Steps to reproduce**

1. Open the OpenChamber VS Code extension webview.
2. Click **Settings** in the extension's title bar / sidebar.
3. In the settings view, navigate into a specific section, e.g. **Providers**.
4. Navigate to a different view (e.g. back to **Chat** or **Sessions**).
5. Click **Settings** again.

**Expected:** The settings view opens on its home/menu page.
**Actual:** The previously-visited section (e.g. Providers) is shown again, with no way to get back to the settings menu without manually clicking the home/back affordance inside the section.

**Root cause (for reference)**

The `openchamber:navigate` event handler at `packages/ui/src/components/layout/VSCodeLayout.tsx` only sets `currentView = 'settings'`; it does not reset the persisted `settingsPage` field in `useUIStore` (which is persisted to localStorage and only set to `'home'` on initial state). The same issue affects the entry from the title-bar `showSettings` command in `packages/vscode/webview/main.tsx`, which just dispatches the navigate event without resetting the page.

### Where does it happen?

VS Code extension

### Version (if known)

v1.13.2 (reproduces on latest main as well)

_No response_

### Logs (optional)

```shell
No error logs; behavior is reproducible 100% of the time.
```

catan271:
## What

Reset the persisted `settingsPage` to `'home'` when Settings is opened via the VS Code title-bar Settings command, so the settings home/menu is shown instead of the last-visited section.

Fixes #1764.

## Why

In the VS Code extension, once a non-default settings section (e.g. Providers) had been visited, `settingsPage` was persisted to localStorage and never reset. Clicking **Settings** in the title bar only set `currentView = 'settings'` via the `openchamber:navigate` event, so `SettingsView` re-rendered with the stale persisted section slug. The only way back to the settings menu was the in-section back affordance.

## Fix

`packages/ui/src/components/layout/VSCodeLayout.tsx` — in the `openchamber:navigate` handler, reset `settingsPage` to `'home'` before switching to the settings view:

```ts
if (view === 'settings') {
  useUIStore.getState().setSettingsPage('home');
  setCurrentView('settings');
}
```

The `openchamber:navigate` event with `view: 'settings'` is dispatched **only** from the VS Code title-bar `showSettings` command (`packages/vscode/webview/main.tsx`). In-settings section navigation does not use this event — it calls `setSettingsPage` directly inside `SettingsView` — so resetting to `'home'` here does not interfere with within-settings navigation. This single-source fix covers every caller of the navigate event (Option A from the issue analysis).

`useUIStore` and `setSettingsPage` are already imported/used elsewhere; this is a one-line addition with no new deps or state.

## Validation

- `tsc --noEmit` (packages/ui) — pass (exit 0)
- `eslint src/components/layout/VSCodeLayout.tsx` — pass (exit 0)

## Risk / notes

Narrow: a single added statement. No change to within-settings navigation, persistence shape, or any other view. The `home` value is the existing default (`useUIStore` initial state).

### DIFF
--- a/packages/ui/src/components/layout/VSCodeLayout.tsx
+++ b/packages/ui/src/components/layout/VSCodeLayout.tsx
@@ -359,6 +359,7 @@ export const VSCodeLayout: React.FC = () => {
         if (currentViewRef.current !== 'settings') {
           viewBeforeSettingsRef.current = currentViewRef.current;
         }
+        useUIStore.getState().setSettingsPage('home');
         setCurrentView('settings');
       } else if (view === 'chat') {
         setCurrentView('chat');


tomzx:
@openchamber-bot review

tomzx:
@openchamber-bot review

catan271:
Closing this PR because the newer Settings UI on `main` now handles the reported VS Code flow without resetting the persisted `settingsPage`.

For the normal narrow VS Code sidebar, `VSCodeLayout` forces `SettingsView` into its mobile navigation layout. `SettingsView` opens with `mobileStage = 'nav'` and does not auto-jump to the persisted page on a fresh mount, so reopening Settings shows the home/menu navigation. The one-line reset proposed here is therefore redundant for the reported flow.

The persisted page is intentionally retained for navigation state. If the same issue is reproduced in a supported wider VS Code layout (`>=550px`), this should be revisited.

PR #1849, opened 2026-06-26
fix(chat): make static tool rows expandable to reveal input and output (#1802)

### LINKED ISSUE #1802: [Bug] Static tool rows (read, grep/glob, webfetch, etc.) have no expand/collapse — payload and response are hidden with no way to reveal them
## Summary

Tool calls classified as "static" (e.g. `read`, `grep`/`search`/`find`/`ripgrep`/`glob`, `webfetch`/`fetch`/`curl`/`wget`, `websearch`, `list`, `todowrite`, `skill`, ...) render as a single line that **does not respond to clicks** and **has no expand/collapse UI**. Their full input payload and tool response are not visible anywhere — there is no way to reveal them from the chat.

This is inconsistent with bash / edit / write / task / lsp / question tools, which expand on click and show input + output.

## Steps to reproduce

1. Start a session.
2. Ask the model to read a file (e.g. `read` tool), or do a `grep` search, or `webfetch` a URL.
3. Click the tool row in the chat.

**Expected:** The row expands to show the full input (args) and the response (output), the same way it does for a `bash` or `edit` tool.

**Actual:** Nothing happens. The row is a static `<div>` with no `onClick` handler, no `cursor-pointer`, no `aria-expanded`, and no underlying state. Only a one-line description (filename / search query / URL) is shown.

The "Show expanded bash tools" / "Show expanded edit tools" settings in *Settings → OpenChamber → Visual* do **not** help — they only operate on tools in the `EXPANDABLE_TOOL_NAMES` set.

## Root cause

`packages/ui/src/components/chat/message/parts/toolRenderUtils.ts:1-6` defines the expandable whitelist:

```ts
const EXPANDABLE_TOOL_NAMES = new Set<string>([
  'edit', 'multiedit', 'apply_patch', 'str_replace', 'str_replace_based_edit_tool',
  'bash', 'shell', 'cmd', 'terminal',
  'write', 'create', 'file_write',
  'question', 'task', 'lsp',
]);
```

`packages/ui/src/components/chat/message/parts/MessageBody.tsx:1786-1827` branches on that whitelist:

```tsx
if (isExpandableTool(toolName)) {
  // renders <ToolPart isExpanded={...} onToggle={...} />   ← expandable
  ...
}
// otherwise renders <StaticToolRow ... />                  ← NOT expandable
```

`packages/ui/src/components/chat/message/parts/ProgressiveGroup.tsx:676-774` — `StaticToolRow` is a single `<div>` containing an icon, the tool display name, and (depending on the tool) a one-line description / file link / URL anchor / skill button. No `onClick`, no `isExpanded` state, no input/output panel.

The state plumbing (`expandedTools`, `defaultOpenToolIds`, `effectiveExpandedTools`, `handleToggleTool`, the `expandedToolsStateCache` map) already exists in `packages/ui/src/components/chat/ChatMessage.tsx:38-78, 189-198, 463-502, 765-810` — it's just not wired into `StaticToolRow`.

## Suggested fix

Add expand/collapse to `StaticToolRow` (option 1) so the existing input + output rendering from `ToolPart` can be reused:

1. Give `StaticToolRow` the same `isExpanded` / `onToggle(toolId)` props `ToolPart` already has, and a `toggleButton` / chevron like the existing one in `ToolPart.tsx:2419-2444`.
2. When expanded, render a body section similar to `ToolPart.tsx:2545-2573` containing the `ToolExpandedContent` (or a minimal equivalent that shows `state.input` and `state.output` for arbitrary tool names).
3. Wire `expandedTools.has(toolPart.id)` and `onToggleTool` from `MessageBody.tsx:1811-1824` into the new props.
4. Update `ProgressiveGroup.tsx`'s `StaticToolRow` memo comparator (`packages/ui/src/components/chat/message/parts/ProgressiveGroup.tsx:777-779`) to include `isExpanded` / `onToggle`.

After the fix, every tool — read, search, webfetch, todowrite, etc. — should be expandable like bash/edit, and `Show expanded bash tools` / `Show expanded edit tools` could be extended (or a new "Show expanded static tools" added) for users who want them open by default.

## Environment

- Latest `main` (current `HEAD` ~ `e88afff2 perf: isolate chat streaming renders and reduce sidebar render cost (#1672)`)
- Reproduced in **web** (browser), **desktop (Electron)**, and the **VS Code extension** (the shared `packages/ui` chat renderer is loaded in all three runtimes, so the issue is present in every UI host).

catan271:
## What

Static tool rows (`read`, `grep`/`search`/`find`/`ripgrep`/`glob`, `webfetch`/`fetch`/`curl`/`wget`, `websearch`, `list`, `todowrite`, `skill`, …) now expand/collapse on click, revealing the full tool input and output — the same behavior `bash`/`edit`/`write`/`task`/`question` already had.

Fixes #1802.

## Why

`EXPANDABLE_TOOL_NAMES` in `toolRenderUtils.ts` omits every static tool, so `MessageBody` and `ProgressiveGroup` rendered them via `StaticToolRow`: a plain `<div>` with no `onClick`, no `cursor-pointer`, no `aria-expanded`, and no expanded body. The tool's input payload and response were not visible anywhere in the chat. The existing expand-state plumbing (`expandedTools`, `onToggleTool`) in `ChatMessage` was already in place but never wired into `StaticToolRow`.

## Fix

Reuse the existing expandable-tool output renderer instead of duplicating it.

- **`ToolPart.tsx`** — `export` the existing `ToolExpandedContent` (input preview + output + diagnostics, already tool-name-agnostic) so it can be shared.
- **`ProgressiveGroup.tsx` — `StaticToolRowInner`**:
  - Added `isExpanded: boolean` and `onToggleTool: (toolId: string) => void` props (same shape `ExpandableToolRow` uses, so `onToggleTool` stays referentially stable and memo-friendly).
  - The header is now clickable: `role="button"`, `tabIndex={0}`, `aria-expanded`, `onClick`, and a guarded `onKeyDown` (Enter/Space toggle, but only when the row itself — not a nested file/search/fetch/skill link — is focused).
  - Added a `arrow-down-s` / `arrow-right-s` chevron (`ml-auto`) matching the existing `ToolPart`/`ReasoningPart` chevrons.
  - When expanded, renders a body section (left rail + `ToolExpandedContent`) for the row's tool part — the same component expandable tools use.
  - Updated the `StaticToolRow` and `MemoStaticGroupedToolRow` memo comparators to include `isExpanded` / `onToggleTool`.
  - Threaded `expandedTools` / `onToggleTool` through `StaticGroupedToolRow` and the `tool-static-group` render case.
- **`MessageBody.tsx`** — the direct (non-grouped) `StaticToolRow` render now passes `isExpanded={expandedTools.has(toolPart.id)}` and `onToggleTool`.

Both render paths (`MessageBody` live path and `ProgressiveGroup` grouped path) pass a single-element `activities` array, so the expanded body renders exactly that call's input/output. The expanded body mounts only when `isExpanded` (conditional render), so collapsed rows stay cheap — important because static-tool rows sit on the hot chat-render path.

Existing nested interactive children (read file links, search-query spans, fetch URL anchors, skill buttons) already call `event.stopPropagation()`, so clicking them still performs their action without toggling the row.

## Validation

- `tsc --noEmit` (packages/ui) — pass (exit 0)
- `eslint` on `ProgressiveGroup.tsx`, `ToolPart.tsx`, `MessageBody.tsx` — pass (exit 0)

## Risk / notes

- Visual/behavioral: static rows gain a chevron and become clickable. The compact one-line summary still shows when collapsed, so default density is unchanged.
- `ToolExpandedContent` is reused unchanged; for static tools (`read`/`grep`/`webfetch`) it renders the input preview + string output (no diff/diagnostic sections, since those are gated on edit/apply_patch metadata that static tools don't carry).
- No new deps, no new state — reuses the existing `expandedTools` Set and `onToggleTool` from `ChatMessage`.

## Follow-ups (not in this PR)

A "Show expanded static tools" default-open setting (mirroring the existing "Show expanded bash tools" / "Show expanded edit tools" toggles in *Settings → OpenChamber → Visual*) is intentionally left as future work; this PR delivers click-to-expand for all static tools.

### DIFF
--- a/packages/ui/src/components/chat/message/MessageBody.tsx
+++ b/packages/ui/src/components/chat/message/MessageBody.tsx
@@ -1870,6 +1870,9 @@ const AssistantMessageBody = React.memo(({
                                     },
                                 ]}
                                 animateTailText={animatedToolIdsLookup.has(toolPart.id)}
+                                isExpanded={expandedTools.has(toolPart.id)}
+                                onToggleTool={onToggleTool}
+                                onShowPopup={onShowPopup}
                             />
                         </ToolRevealOnMount>
                     </FadeInOnReveal>
--- a/packages/ui/src/components/chat/message/parts/ProgressiveGroup.tsx
+++ b/packages/ui/src/components/chat/message/parts/ProgressiveGroup.tsx
@@ -5,7 +5,7 @@ import type { ToolPart as ToolPartType } from '@opencode-ai/sdk/v2';
 import type { StreamPhase } from '../types';
 import type { ContentChangeReason } from '@/hooks/useChatAutoFollow';
 import type { ToolPopupContent } from '../types';
-import ToolPart from './ToolPart';
+import ToolPart, { ToolExpandedContent } from './ToolPart';
 import { MinDurationShineText } from './MinDurationShineText';
 import { ToolRevealOnMount } from './ToolRevealOnMount';
 import { FileTypeIcon } from '@/components/icons/FileTypeIcon';
@@ -25,6 +25,7 @@ import JustificationBlock from './JustificationBlock';
 import { areRenderRelevantPartsEqual } from '../renderCompare';
 import { getExternalFaviconUrl } from '@/lib/url';
 import { getDirectoryForFilePath, getRelativeFilePath, isFilePathWithinDirectory, normalizeFilePath, toAbsoluteFilePath } from '@/lib/path-utils';
+import { useDeferredExpandedContent } from './useDeferredExpandedContent';
 
 const TOOL_ROW_TEXT_CLASS = '!text-[length:var(--text-meta)] !leading-5 sm:!leading-6 tracking-normal';
 const TOOL_ROW_TITLE_CLASS = cn('typography-meta font-medium', TOOL_ROW_TEXT_CLASS);
@@ -438,6 +439,9 @@ interface StaticGroupedToolRowProps {
     activities: TurnActivityPart[];
     animateTailText: boolean;
     animateRows: boolean;
+    isExpanded: boolean;
+    onToggleTool: (toolId: string) => void;
+    onShowPopup?: (content: ToolPopupContent) => void;
 }
 
 const StaticGroupedToolRow: React.FC<StaticGroupedToolRowProps> = ({
@@ -445,12 +449,18 @@ const StaticGroupedToolRow: React.FC<StaticGroupedToolRowProps> = ({
     activities,
     animateTailText,
     animateRows,
+    isExpanded,
+    onToggleTool,
+    onShowPopup,
 }) => {
     const content = (
         <StaticToolRow
             toolName={toolName}
             activities={activities}
             animateTailText={animateTailText}
+            isExpanded={isExpanded}
+            onToggleTool={onToggleTool}
+            onShowPopup={onShowPopup}
         />
     );
 
@@ -471,6 +481,9 @@ const MemoStaticGroupedToolRow = React.memo(StaticGroupedToolRow, (prev, next) =
     return prev.toolName === next.toolName
         && prev.animateTailText === next.animateTailText
         && prev.animateRows === next.animateRows
+        && prev.isExpanded === next.isExpanded
+        && prev.onToggleTool === next.onToggleTool
+        && prev.onShowPopup === next.onShowPopup
         && areActivityListsEqual(prev.activities, next.activities);
 });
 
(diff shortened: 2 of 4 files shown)


catan271:
Closing as superseded by v1.15.0. Upstream changed the tool classification so all non-navigation tools now use the common expandable renderer, covering the original issue for search, fetch, todo, MCP, plugin, and custom tools.

The remaining compact `read` and `skill` rows are intentional navigation affordances, so this broader replacement is no longer needed.

PR #1855, opened 2026-06-26
feat(cli): warn when systemd user lingering is disabled on startup enable

### LINKED ISSUE #1842: Linux startup service should detect or document required systemd user lingering
## Summary

`openchamber startup enable` can create a Linux `systemd --user` service that stops working after the SSH/login session ends unless `systemd` user lingering is enabled.

In my case, OpenChamber worked while I was logged in over SSH, but later appeared to fail randomly. Logging back in over SSH made the service available again. Running this fixed the issue:

```sh
sudo loginctl enable-linger myuser
```

After enabling lingering, the OpenChamber service stayed up as expected.

## Expected behavior

On Linux, if OpenChamber installs or enables a `systemd --user` service intended to run in the background, setup should either:

- Detect that user lingering is disabled and warn clearly.
- Print the required `loginctl enable-linger $USER` command.
- Document this requirement in the startup/service setup flow.
- Optionally offer a supported way to install a root/system service instead, if that is intended.

## Actual behavior

The service can appear to work initially, then stop after the user logs out. It may come back only after the user logs back in.

This makes the service feel flaky/random even though it is caused by normal `systemd --user` lifecycle behavior.

## Why this happens

A `systemd --user` service belongs to the user's `systemd` manager. On many Linux systems, the user manager is only kept alive while the user has an active login session.

Without lingering, when the user logs out of SSH, `systemd-logind` may stop the user's manager, which also stops user services.

`loginctl enable-linger $USER` tells `systemd-logind` to keep that user's `systemd --user` manager running even without an active login session, allowing enabled user services to start at boot and keep running after logout.

## Suggested fix

During `openchamber startup enable` on Linux, detect whether lingering is enabled for the current user:

```sh
loginctl show-user "$USER" -p Linger
```

If it reports `Linger=no`, print a warning such as:

```text
OpenChamber was enabled as a systemd user service, but user lingering is disabled.

This service may stop when you log out of SSH or your desktop session.
To keep it running in the background, run:

  sudo loginctl enable-linger <username>

Then restart the service:

  systemctl --user restart openchamber
```

This should be a warning/documentation improvement, not necessarily an automatic privileged action, because `loginctl enable-linger` commonly requires `sudo`.

## Environment

- Platform: Linux
- Service mode: `systemd --user`
- Access pattern: SSH/headless login
- Workaround confirmed: `sudo loginctl enable-linger myuser`

IbrahimKhan12:
> Validated commit: `ae0ac79c68a143241d798e9dceda08a769e4c91d`, rebased onto `main` at `feec14545`.

## Intent

Fixes #1842. On Linux, `startup enable` and `startup status` now detect whether the current user has `systemd` lingering enabled. When the service is enabled and lingering is not, human output warns that the service may stop at logout and prints the exact `sudo loginctl enable-linger <user>` remediation. `--quiet` appends a `linger:yes|no|unknown` token; `--json` adds `lingerEnabled`, `lingerUser`, and a `warning`-level `messages[]` entry.

## Non-goals

No privileged command is run automatically and no system-wide service mode is added.

Documenting lingering in `packages/docs/content/docs/opencode-server.mdx` and the README systemd recipe is out of scope, even though issue #1842 also asked for documentation.

## Affected surfaces

Linux `openchamber startup enable` and `openchamber startup status` in human, `--quiet`, and `--json` modes. macOS and Windows output is unchanged. `startupCommand` gained an optional third dependency-injection argument, used only by tests; it is additive, so existing two-argument callers are unaffected. `cli-startup.js` exports are byte-identical to `main` — the new `parseLingerState`, `getCurrentUsername`, and `getUserLingerEnabled` helpers stay module-private and are reached through `getStartupStatus`. No persisted or server contract changes.

## Rebase note

This branch was rebased from its original base onto `feec14545`. The only conflict was in `packages/web/bin/cli.test.js`, where `main` and this branch each appended an import and a trailing `describe` block. Both were kept: `main`'s `Windows startup task command builder` suite and this branch's `startup command lingering output` suite both run. `cli-startup.js` and `commands-startup.js` merged without conflict, and the resulting diff stat is unchanged from the pre-rebase commit (183 insertions, 6 deletions).

## Repository guidance

|---|---|---|
| `openchamber-change-discipline` | CLI source changed and a command signature gained an optional parameter | Detection is confined to `cli-startup.js` and adds no exports, so the module contract is unchanged. `startupCommand` takes injected dependencies so presentation is testable without spawning subprocesses. Validation was chosen for the touched surface rather than reflexively: because `type-check:web` and `lint:web` do not cover `bin/**`, the CLI suite, `node --check`, `oxlint`, and a real-host run were used instead. |
| `clack-cli-patterns` | Human, quiet, JSON, and non-TTY output changed | Each mode has one stable representation: prefixed `[LINGER_*]` human lines asserted for both TTY and non-TTY, a single quiet result line, and a machine-readable `messages[]` code. `--json` stays JSON-only with the warning carried in `status` and `messages[]`. No prompt and no privileged side effect was added. |

## Validation

| Check | Result |
|---|---|
| `cd packages/web && bun x vitest run bin/cli.test.js` | 1 file, **105 passed**, 0 failed. Includes `main`'s Windows startup suite, confirming the conflict resolution kept both blocks. |
| `... vitest run bin/cli.test.js -t "startup command lingering output"` | **17 passed**, 88 skipped. |
| `node --check` on `cli.test.js`, `cli-startup.js`, `commands-startup.js` | All parse. Needed because `lint:web` globs only `./src/**/*.{ts,tsx}` and never lints `bin/**/*.js`. |
| `bun x oxlint` on the three changed files | 0 warnings, 0 errors (87 rules). |
| `bun run type-check:web` | Exit 0, but it does not cover this diff: `packages/web/tsconfig.json` `include` is `["src", "../ui/src", ...]`, so `bin/**` is never type-checked. |
| `bun run lint:web` | Exit 0, but it does not cover this diff, per the glob above. |
| **Real-host run** on Linux with systemd and `loginctl` | See below. This closes the gap the previous revision of this description declared. |

### Real-host verification

The earlier revision of this PR stated the `loginctl` detection path had never been exercised against a real host. It now has, on a Linux host with a real systemd user session where `openchamber startup` is genuinely enabled and active. Only the read-only `startup status` was run; no service and no lingering state was modified.

`loginctl show-user ibrakhxn -p Linger` → exit 0, stdout `Linger=yes`.

```
$ node packages/web/bin/cli.js startup status
┌  OpenChamber Startup
│
◆  startup enabled
│  /home/ibrakhxn/.config/systemd/user/openchamber.service
│
◆  service active
│
◆  user lingering enabled
│
└  status complete
```

```
$ node packages/web/bin/cli.js startup status --json
{
  "status": "ok",
  "action": "status",
  "supported": true,
  "platform": "linux",
  "enabled": true,
  "active": true,
  "activeState": "active",
  "servicePath": "/home/ibrakhxn/.config/systemd/user/openchamber.service",
  "lingerEnabled": true,
  "lingerUser": "ibrakhxn"
}
```

```
$ node packages/web/bin/cli.js startup status --quiet
startup enabled platform:linux supported:yes path:/home/ibrakhxn/.config/systemd/user/openchamber.service linger:yes
```

The `--json` payload is JSON-only with no `messages[]`, which is correct for the enabled state. `--quiet` is a single line carrying `linger:yes`.

The parser was also checked against genuine `loginctl` output rather than only synthetic strings. `Linger=yes` → `true`, `Linger=no` → `false`, and non-matching output such as `Failed to get user: ...` → `null`. For a user that is neither logged in nor lingering, real `loginctl` exits **1** and writes to stderr, which `getUserLingerEnabled` maps to `null` before parsing, producing `LINGER_UNKNOWN`.

**Not verified end to end:** the `Linger=no` → `[LINGER_DISABLED]` path. Flipping lingering requires `sudo loginctl disable-linger` and would modify the host, so that state is covered by the unit tests plus the parser check against the real `Linger=no` string format, not by a live run. macOS and Windows were not run.

## Visual evidence

CLI change, so terminal transcripts are the correct evidence and they are included above for human, `--json`, and `--quiet` on a real Linux host. No graphical UI is affected, so screenshots do not apply.

The one transcript that is missing is the disabled-lingering warning, for the host-modification reason given above. Its asserted shape is `[LINGER_DISABLED] user lingering is disabled` followed by `[ENABLE_LINGER] sudo loginctl enable-linger <user>`.

## Risks and failure behavior

An unavailable or unrecognized `loginctl` result yields `unknown`, which is not silent: while the service is enabled the CLI still emits a `warning`-level `[LINGER_UNKNOWN]` line and sets the `--json` `status` to `warning`, so a host where detection cannot run is reported as a warning rather than as healthy.

One real-world nuance the host check surfaced: for a user who is not logged in, `loginctl` fails rather than reporting `Linger=no`, so that case reports `unknown` instead of `disabled`. Both warn while the service is enabled, and the user running `startup enable` is by definition logged in, so the actionable `disabled` case is the one that occurs in practice.

When the service is disabled, linger state is reported at `info` level with no remediation. Enable and disable behavior is otherwise unchanged, and the remediation is only printed; the user retains control of the privileged action. Each Linux `startup` invocation now spawns one extra `loginctl show-user` subprocess, alongside the two `systemctl` calls already on that path, and it degrades to `null` if the binary is missing.

Rollback is reverting one commit; no persisted data or server contract is involved.

### DIFF
--- a/packages/web/bin/cli.test.js
+++ b/packages/web/bin/cli.test.js
@@ -11,6 +11,7 @@ import { isModuleCliExecution, normalizeCliEntryPath } from './cli-entry.js';
 import { requestJson } from './lib/cli-http.js';
 import { requestControlAction } from './lib/cli-control.js';
 import { inspectTunnelAttachability } from './lib/cli-lifecycle.js';
+import { startupCommand } from './lib/commands-startup.js';
 import { formatGoal } from './lib/commands-schedule.js';
 import {
   buildSessionCreatePayload,
@@ -34,6 +35,7 @@ import {
   discoverRunningInstances,
   discoverUnconfirmedRegistryInstanceOnPort,
   ensureTunnelProfilesMigrated,
+  EXIT_CODE,
   generateUiPassword,
   getInstanceFilePath,
   getPidFilePath,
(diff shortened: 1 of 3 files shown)


btriapitsyn:

mattv8:

IbrahimKhan12:

The only conflict was in `packages/web/bin/cli.test.js`, where `main` and this branch each appended an import and a trailing `describe` block. Both suites are kept and both run — `bun x vitest run bin/cli.test.js` is 105 passed, which includes `main`'s Windows startup task suite.

I also closed the validation gap the previous description admitted: the `loginctl` detection path had never been run against a real host, and it now has. Transcripts for human, `--json`, and `--quiet` from a Linux host with a real systemd user session are in the updated description. Only the read-only `startup status` was run; no service or lingering state was modified. The `Linger=no` warning path is still unit-tested rather than run live, because flipping lingering needs `sudo` and would modify the host.

btriapitsyn:
Thanks @IbrahimKhan12 for rebasing this and adding the real-host evidence. Merged: Linux startup now reports lingering status and gives the remediation command without changing privileged host settings. The warning addresses #1842.

Thanks @mattv8 for tracking this PR during the backlog triage.

PR #1860, opened 2026-06-27
fix(git): materialize draft session for PR share comments / failed checks / single comment

### LINKED ISSUE #1859: fix(git): share comments / resolve failed checks / send single comment fail when only draft session exists
## Summary

**Share Comments**, **Resolve Failed Checks**, and **Send Single Comment** buttons in the PR section fail with "No active session" toast when the user only has a draft session open (from worktree/branch flow). The draft must be materialized into a real session before dispatching the prompt.

## Problem

All three dispatch actions go through `resolveChatDispatchTarget()` in `PullRequestSection.tsx`, which checks:

```ts
if (!currentSessionId) {
  toast.error(...);
  return null;
}
```

When a draft is open, `openNewSessionDraft()` sets `currentSessionId: null`, so the check fails and the dispatch never reaches `sendMessage` — even though `sendMessage` already handles draft materialization internally (via `materializeOpenDraftSession()` from PR #1761).

## Affected buttons

1. **Share Comments** (`sendCommentsToChat`) — line 939
2. **Resolve Failed Checks** (`sendFailedChecksToChat`) — line 883
3. **Send Single Comment** (`sendSingleCommentToChat`) — line 974

## Expected behavior

When a draft session is open, clicking any of these buttons should materialize the draft into a real session (reusing `materializeOpenDraftSession`) and then dispatch the synthetic prompt — same pattern as `resolveGenerationSessionContext` in `gitApi.ts` fixed by PR #1761.

## Related

- PR #1761 fixed the same pattern for `generateCommitMessage` / `generatePullRequestDescription`

bashrusakh:
## Summary

Closes #1859.

Make Share Comments, Resolve Failed Checks, and Send Single Comment buttons materialize an open draft session into a real session before dispatching, instead of silently failing with "No active session" toast.

## Problem

All three dispatch actions go through `resolveChatDispatchTarget()` which checks `currentSessionId` and returns `null` when no real session exists. Draft sessions keep `currentSessionId: null`, so the dispatch never reaches `sendMessage` (which already handles draft materialization via `materializeOpenDraftSession` from PR #1761).

## What needs to change

- Make `resolveChatDispatchTarget()` async so it can materialize a draft when `currentSessionId` is null but a draft is open
- Update the three callers (`sendFailedChecksToChat`, `sendCommentsToChat`, `sendSingleCommentToChat`) to `await` the result
- Reuse the existing `materializeOpenDraftSession()` helper

## Validation

- `bun run --filter="@openchamber/ui" type-check`
- `bun run --filter="@openchamber/ui" lint`


bashrusakh:
superseded by #1863

PR #1862, opened 2026-06-27
fix(update): coordinate systemd user service restarts

### LINKED ISSUE #1861: fix:  conflicts with systemd user service (Restart=always)
## Problem

When OpenChamber runs as a systemd user service (`openchamber startup enable`), both update paths fail to properly handle the service lifecycle:

1. **CLI `openchamber update`** kills the systemd-managed process and then tries to restart it via `this.serve()`, but systemd already restarted it due to `Restart=always`. This causes port conflicts and race conditions.

2. **Web UI update** (`POST /api/openchamber/update-install`) spawns a detached update script, then calls `process.exit(0)` after 500ms. Systemd restarts the service before the npm update completes, so the old version is loaded.

3. **The `restart` command already handles this correctly** — it checks `launchMode` and skips `this.serve()` for foreground instances, letting the process manager restart them. But `update` does not.

## Root cause

In `packages/web/bin/cli.js`, the `commands.update()` function (line ~5537) always calls `this.serve()` for all running instances after updating, without checking `launchMode`:

```javascript
// Current code — always restarts, regardless of launchMode
if (runningInstances.length > 0) {
  for (const instance of runningInstances) {
    const storedOptions = readInstanceOptions(instance.instanceFilePath) || { port: instance.port };
    await this.serve({ port: storedOptions.port || instance.port, ... });
  }
}
```

Compare with `commands.restart()` (line ~4037) which correctly skips restart for foreground instances:

```javascript
// restart — correct behavior
if (isForeground) {
  // Foreground instances managed by process manager
  continue;  // ← does NOT call serve()
}
```

## Reproduction

1. `openchamber startup enable` — creates systemd user service with `Restart=always`
2. `openchamber update` — CLI kills process, systemd restarts old version, CLI tries to start new daemon → port conflict
3. Or: click "Update" in web UI → server exits, systemd restarts before npm finishes → old version loads

## Expected behavior

- `openchamber update` should detect foreground instances and only stop them (let systemd restart after update)
- The update should work correctly regardless of whether the service runs as daemon or foreground

## Fix

Add `launchMode` check in `commands.update()`, matching the pattern from `commands.restart()`:

- For `launchMode === "foreground"`: stop the process, skip `this.serve()` — systemd will restart it
- For `launchMode === "daemon"`: stop, update, restart as before

## Related

- Similar issue: #1655 (update button not working — may be a symptom of the same problem)
- Service management: `buildSystemdUserService()` in `cli.js` generates `Restart=always`
- The `restart` command already has the correct pattern at line 4037

bashrusakh:
## Summary

Fix update behavior when OpenChamber runs as a Linux systemd user service.

This covers both update entry points:

- `openchamber update`
- client-triggered `POST /api/openchamber/update-install`

## Problem

Foreground OpenChamber instances started via startup integration are owned by `systemd --user` with `Restart=always`.

Before this change:

- the CLI update flow stopped the running process and then restarted it itself, racing with systemd
- the client/server update flow spawned an updater from inside the service process, so any attempt to stop the service risked killing the updater too

That could lead to:

- restart races
- old version coming back up before the install finished
- port conflicts / wrong lifecycle ownership

### CLI update

`packages/web/bin/cli.js`

- detect when a foreground instance is the active `openchamber.service` user unit
- stop/start that service through `systemctl --user` instead of PID-level stop + CLI restart
- keep CLI-managed instances on the existing stop/update/restart path
- restart the service again if the package-manager update fails, so the app is not left down

### Client/server update

`packages/web/server/lib/opencode/openchamber-routes.js`

- detect the same systemd-managed foreground case
- launch the updater via a transient `systemd-run --user` unit so it survives stopping `openchamber.service`
- explicitly stop the service before update and start it again after update
- attempt to bring the service back even on update failure

### Shared helper

`packages/web/server/lib/systemd-user-service.js`

- centralizes systemd user-service detection
- matches the configured service port when available
- provides stop/start commands for both CLI and server update flows

## Validation

```bash
node --check packages/web/bin/cli.js
node --check packages/web/server/lib/opencode/openchamber-routes.js
node --check packages/web/server/lib/systemd-user-service.js
```

## Related

Closes #1861

### DIFF
--- a/packages/ui/src/components/ui/UpdateDialog.tsx
+++ b/packages/ui/src/components/ui/UpdateDialog.tsx
@@ -116,6 +116,19 @@ type InstallWebUpdateResult = {
   autoRestart?: boolean;
 };
 
+type WebServerUpdateStatus = {
+  state: string;
+  currentVersion?: string;
+  targetVersion?: string;
+  launchMode?: string;
+  restartManager?: string;
+};
+
+type WaitForWebUpdateResult =
+  | { outcome: 'applied' }
+  | { outcome: 'failed'; status: WebServerUpdateStatus | null }
+  | { outcome: 'timeout'; status: WebServerUpdateStatus | null };
+
 const WEB_UPDATE_POLL_INTERVAL_MS = 2000;
 const WEB_UPDATE_MAX_WAIT_MS = 10 * 60 * 1000;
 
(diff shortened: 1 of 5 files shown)


bashrusakh:
I did not include the `--port=...` regex change in this pass because the generated startup unit already uses the space-separated `--port 3000` form, so the reported review blocker was the `systemd-run` PATH regression rather than a failing default path. I kept the follow-up scoped to the required fix only and left the port parsing hardening as optional cleanup.

btriapitsyn:

PR #1863, opened 2026-06-27
fix(git): materialize draft session for PR share comments / failed checks / single comment

### LINKED ISSUE #1859: fix(git): share comments / resolve failed checks / send single comment fail when only draft session exists

bashrusakh:
## Summary

Closes #1859.

This PR fixes PR-to-chat actions when the user only has an open draft session, and adds visible in-progress feedback for the PR-level **Share comments** action.

## What changed

- made `resolveChatDispatchTarget()` async
- when `currentSessionId` is null, detect an open draft session and materialize it via `materializeOpenDraftSession()`
- carry any draft synthetic parts through the eventual send
- updated the affected PR actions to `await` the resolved dispatch target
- added a loading state to the PR-level **Share comments** button so it disables and shows a spinner while the async send is in progress

## Why

Before this change:

- PR share/send actions could fail with **No active session** when the user had a draft session open but no materialized session yet
- the PR-level **Share comments** button provided no visible feedback during the async dispatch flow, so repeated clicks were easy and it was unclear whether the action had started

## Validation

- `bun run type-check:ui` ✅
- `bun run lint:ui` ✅

### DIFF
--- a/packages/ui/src/components/views/git/PullRequestSection.tsx
+++ b/packages/ui/src/components/views/git/PullRequestSection.tsx
@@ -29,7 +29,7 @@ import { SimpleMarkdownRenderer } from '@/components/chat/MarkdownRenderer';
 import { Icon } from "@/components/icon/Icon";
 import { useUIStore } from '@/stores/useUIStore';
 import { formatDateTimeForPreference } from '@/lib/timeFormat';
-import { materializeOpenDraftSession, useSessionUIStore } from '@/sync/session-ui-store';
+import { materializeOpenDraftSession, useSessionUIStore, type SyntheticContextPart } from '@/sync/session-ui-store';
 import { useSelectionStore } from '@/sync/selection-store';
 import { useConfigStore } from '@/stores/useConfigStore';
 import { useGitHubAuthStore } from '@/stores/useGitHubAuthStore';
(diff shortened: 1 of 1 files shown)


btriapitsyn:

PR #1868, opened 2026-06-27
fix(chat): restore prompt on send failure via per-session snapshot

### LINKED ISSUE #1792: [Bug] Prompt Lost on Network Failure

**Bug Description**

When a user spends a significant amount of time writing a prompt and then submits it, a temporary network interruption can cause the request to fail. In this situation, an error message is displayed, but the entire prompt content is lost.

**Steps to Reproduce**

1. Type a prompt.
2. Disconnect the network or simulate a network failure.
3. Press Enter to submit the prompt.
4. Observe the error message.
5. Notice that the prompt content is no longer available.

**Expected Behavior**
The prompt should be preserved locally when submission fails, allowing the user to retry without losing their work.

**Actual Behavior**
The prompt is cleared after the failed submission, resulting in permanent loss of the user's input.

**Impact**
This is a critical usability issue because users can lose substantial amounts of work due to a brief network interruption.

### Where does it happen?

Desktop Web

### Version (if known)

1.13.2

_No response_

### Logs (optional)

```shell

```

procodingtools:
When a chat message fails to send (network error, timeout, etc.), the user's typed text was lost because handleSubmit cleared the input, queue, drafts, and mentions synchronously before the async send call. On failure, only attachments were restored — text, queue items, inline comment drafts, and confirmed mentions were gone forever.

Solution: capture a SubmissionSnapshot of all destructively-cleared state before the clear block, and restore from it on send failure:

- Raw textarea text (not mention-expanded/snippet-expanded derivative)
- Per-session localStorage draft (survives page reload / session switch)
- Confirmed @mentions (restored to both ref and localStorage)
- Inline comment drafts (new restoreDrafts store action)
- Queued messages (re-enqueued via addToQueue)
- Message history navigation state

Three failure branches with different restore strategies:

1. 413 Payload Too Large: save draft to localStorage (don't auto-restore to textarea — message may be too large), user can edit and retry.

2. Soft network error (timeout, gateway, Failed to send): preserve draft in localStorage but don't auto-restore to textarea to prevent accidental duplicate sends. New messageSendMaybeFailed toast.

3. Hard error: full per-session restore — textarea, draft, mentions, inline drafts, queue. messageRestoredToInput info toast confirms the restore to the user.

Key design decisions:

- Snapshot captured AFTER slash-command early-returns (not before), so /undo, /compact, etc. don't leave stale snapshots behind.
- Synchronous draft clear preserved (slash commands need it). On failure, draft is re-saved from snapshot (overwriting the clear).
- Local  variable (not a ref) so concurrent sends each restore their own text, not each other's.
- Restore scoped to sendSessionId: if user switches sessions mid-flight, textarea restore is skipped but localStorage draft is still written to the correct session key.

Fixes #1792


procodingtools:
@btriapitsyn please check.

btriapitsyn:

PR #1872, opened 2026-06-27
fix(sync): recover stuck busy sessions with session-scoped freshness

### LINKED ISSUE #1869: [Bug] Busy sessions can stay stuck when directory events keep watchdog fresh
## Summary

A session can remain stuck in **busy / thinking** state even after the actual work stalls or finishes. The stop button stays visible, and the session often only recovers after sending a new message.

## Root cause

- `lastActiveEventAt` is tracked **per directory**, not per session.
- Any non-heartbeat SSE event in that directory refreshes the timer.
- If other sessions stay active, stale detection for the stuck session may never fire.
- Monotonic `/session/status` polling does **not** lower `busy -> idle`.
- Authoritative resync only runs when poll sees a busy/idle mismatch.
- If the server also reports busy, the stuck state persists.
- `resyncDirectorySessionStatuses` has no retry for transient HTTP failures.
- `abortCurrentOperation` swallows abort errors, so stop can fail silently.

## Expected

A session that is no longer active should return to idle and hide the stop button.

## Possible fixes

- Track per-session last activity instead of only per-directory freshness.
- Retry `/session/status` resync on transient failures.
- Surface abort failures to the user.
- Optional: add a max busy duration fallback.

## Related

- #1843 — stuck at "Analyzing" stage
- #1630 — session bar disables prematurely

bashrusakh:
## Summary
Fix a bug where active sessions could stay stuck in busy/thinking state because watchdog freshness was only tracked per directory.

### What changed
- Track freshness per session.
- Refresh freshness from real activity events, not monotonic status polls.
- Propagate freshness to ancestor sessions so delegated child activity keeps parent candidates alive.

## Validation
- `bun test packages/ui/src/sync/__tests__/event-reducer.test.ts`
- `bun test packages/ui/src/sync/__tests__/session-status-snapshot.test.ts`
- `bun test packages/ui/src/sync/session-actions.test.ts`
- `bun run type-check`
- `bun run lint`

Closes #1869


btriapitsyn:
Resolved the conflict with `main` and integrated session-scoped freshness into the current watchdog:

- sibling activity/heartbeats can no longer keep a different wedged session falsely fresh;
- child-session activity refreshes its parent lineage;
- optimistic sends seed freshness before the first server event;
- retained `main`’s authoritative status reconciliation, routing, relay, and global stream-staleness checks;
- verified 62 focused sync tests across snapshot/reducer/actions, UI type-check, and lint.

PR #1878, opened 2026-06-28
fix(ui): make collapse button sticky on expanded user messages

### LINKED ISSUE #1877: [Bug] Collapse button on expanded user messages scrolls out of view
## Bug

When a user message is expanded (long text that exceeds the scroll container height), the collapse button (top-right corner) scrolls out of view. The user must scroll back to the very top to collapse the message.

## Steps to Reproduce

1. Send a long user message (text taller than ~40% of viewport height)
2. Click the message to expand it (collapse feature must be enabled)
3. Scroll down within the message content
4. The collapse button disappears from view
5. Scrolling back to the top is required to collapse

## Expected Behavior

The collapse button should remain visible (sticky) at the top of the scrollable area while scrolling through the expanded message.

## Root Cause

In `packages/ui/src/components/chat/message/parts/UserTextPart.tsx`, the collapse button uses `position: absolute` (`absolute top-0 right-0`). When the parent scroll container (`overflow-y-auto` with `max-height: 40vh`) scrolls, the absolutely-positioned button scrolls with the content.

## Fix

Change the button from `position: absolute` to `position: sticky`:

- Wrap the button in a `<div className="sticky top-0 z-10 flex justify-end pointer-events-none">`
- Add `pointer-events-auto` to the button itself so it remains clickable
- The `pointer-events-none` wrapper allows clicks to pass through to the text content beneath

This makes the button stick to the top of the scroll container while the text content scrolls underneath.

bashrusakh:
## Summary

Fixes #1877.

When a user message is expanded and scrollable, the collapse button previously used absolute positioning (`absolute top-0 right-0`) and scrolled out of view as the user scrolled through the message content. This required scrolling back to the top to collapse the message.

## Changes

- Wrap the collapse button in a sticky container (`sticky top-0`)
- Add `pointer-events-auto` to the button itself so it remains clickable
- The `pointer-events-none` wrapper allows clicks to pass through to the text content beneath
- Use `h-0 overflow-visible` on the wrapper to avoid pushing text content down

## Verification

- Manual: expanded a long user message and scrolled down — the collapse button now remains visible at the top of the scrollable area without adding a visual gap.
- No new hooks, state, or lifecycle logic introduced; minimal surface area for regressions.
- Uses existing theme tokens (`--surface-elevated`, `--surface-mutedForeground`, etc.) and shared `Icon` component.

### DIFF
--- a/packages/ui/src/components/chat/message/parts/UserTextPart.tsx
+++ b/packages/ui/src/components/chat/message/parts/UserTextPart.tsx
@@ -197,7 +197,7 @@ const UserTextPart: React.FC<UserTextPartProps> = ({ part, messageId, agentMenti
     return (
         <div className="relative" key={part.id || `${messageId}-user-text`}>
             {collapsibleUserMessages && isExpanded && (
-                <div className="sticky top-0 z-10 flex justify-end pointer-events-none">
+                <div className="sticky top-0 z-10 flex justify-end pointer-events-none h-0 overflow-visible">
                     <button
                         type="button"
                         onClick={handleCollapse}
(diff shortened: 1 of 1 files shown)


btriapitsyn:

PR #1882, opened 2026-06-28
[KaTeX] Fix math formula font loading — eager JS import instead of CSS @import

xuexi-xiaobai:
## Bug

Since v1.13.2, math formulas render in upright font instead of italic. The KaTeX `@font-face` declarations for math fonts (KaTeX_Math-Italic.woff2, etc.) fail to load because they are referenced via relative `url()` paths inside `katex.min.css`, which is loaded via CSS `@import` in `index.css:11`.

Lightning CSS (Tailwind v4's CSS processor) inlines the CSS content but does not resolve or copy the font file assets, causing 404s at runtime.

## Fix

Move the KaTeX CSS import from a CSS `@import` to a JavaScript `import` in each app entry point. Vite's JS-import pipeline correctly copies the `.woff2` font files to the build output and rewrites the `url()` references.

Changes:
- `packages/ui/src/index.css` — remove `@import "katex/dist/katex.min.css"`
- `packages/ui/src/styles/katex-css.ts` — new file with `import 'katex/dist/katex.min.css'`
- `packages/web/src/main.tsx` — add eager import
- `packages/web/src/mobile-main.tsx` — add eager import
- `packages/web/src/mini-chat-main.tsx` — add eager import

### DIFF
--- a/packages/ui/src/index.css
+++ b/packages/ui/src/index.css
@@ -3,12 +3,6 @@
 @import "./styles/design-system.css";
 @import "./styles/typography.css";
 @import "./styles/mobile.css";
-/* Load KaTeX styles eagerly with the main bundle. Importing them inside the
-   lazy MarkdownRendererImpl chunk meant a stylesheet was injected late on the
-   first markdown render of a session; that late injection triggered a style
-   recalc during which the layered `.chat-message-column` width clamp briefly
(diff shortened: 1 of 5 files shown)


btriapitsyn:
Closing as a duplicate of #1900, which carries the same fix and references the original issue. That one will be reviewed instead.

PR #1887, opened 2026-06-28
fix(chat): prevent auto-load cascade on history scroll; remove translateZ(0) GPU ghosting

herjarsa:
### 1. History scroll cascade guard
When the user scrolls up near the top, `loadEarlier` fires and loads older messages. After scroll compensation keeps the viewport near the top, a second load could trigger immediately — creating a cascade that loads many pages of history at once.

**Fix:** Added `historyInteractionRef` guard to `handleHistoryScroll`. After any `loadEarlier` completes, `settleHistoryInteraction` provides a 2-second grace window before another scroll-at-top can trigger a new load.

### 2. Ghosted content from GPU compositing layers
Every message body was forced into a GPU compositing layer via `translateZ(0)`. With `virtua`'s Virtualizer recycling DOM nodes, stale GPU layers persist as ghosted/faded content behind the real rendered content.

**Fix:** Removed `translateZ(0)` from `CONTAIN_LAYOUT_STYLE` in `MessageBody.tsx`. The `contain: 'layout'` provides layout isolation without creating GPU layers.

### 3. Plugin version display
The `/api/config/plugins/status` endpoint now reads the actual installed version from `package.json` in the OpenCode cache directory instead of returning the plugin scope string.

## Files changed
- `packages/ui/src/components/chat/hooks/useChatTimelineController.ts` — historyInteractionRef guard
- `packages/ui/src/components/chat/message/MessageBody.tsx` — removed translateZ(0)
- `packages/ui/src/components/chat/ChatContainer.tsx` — removed transform-gpu
- `packages/web/server/lib/opencode/plugin-routes.js` — readInstalledVersion

### DIFF
--- a/packages/ui/src/components/chat/ChatContainer.tsx
+++ b/packages/ui/src/components/chat/ChatContainer.tsx
@@ -766,7 +766,7 @@ export const ChatContainer: React.FC<ChatContainerProps> = ({ autoOpenDraft = tr
 
 	if (!currentSessionId && draftOpen) {
 		return (
-			<div className="relative flex h-full flex-col bg-background transform-gpu">
+			<div className="relative flex h-full flex-col bg-background">
 				{useCompactDraftLayout && !isDesktopExpandedInput ? (
 					<div className="flex min-h-0 flex-1 flex-col items-center justify-center px-6 text-center">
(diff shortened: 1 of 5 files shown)


herjarsa:
Superseded by #NEW — cherry-pick of the 2 commits onto current main.

PR #1888, opened 2026-06-28
fix(mcp): filter out purely numeric noise entries from status map

herjarsa:
When the OpenCode status map contains thousands of purely numeric entries (e.g., '0', '1', '10', '100'), the MCP dropdown was filling all 100 visible slots with these useless entries after showing the real configured servers.

## Changes
- Numeric-only names are now skipped during prioritization, so only real MCP server names appear
- The truncation message now shows the actual visible count instead of a hardcoded MAX_VISIBLE_SERVERS value, and only renders when entries are actually filtered


herjarsa:
Superseded by #NEW — cherry-pick of the 5 commits onto current main.

PR #1666, opened 2026-06-15
feat(sidebar): subagent session filtering with store-backed preference

herjarsa:
## Summary

Adds user control over subagent sessions in the sidebar &mdash; store-backed preference, no UI surface yet:

### Subagent Filtering
- New `showSubagentSessionsInSidebar` preference stored in `useUIStore`
- Gated `useProjectSessionLists` and `useSessionGrouping` by subagent visibility
- Subagent sessions filtered from mobile session list
- Extracted `isSubtaskSession` / `getParentSessionId` helpers for reuse

### Project Header Refresh
- Fix project sessions not loading on browser reload
- Add refresh button to desktop and mobile project headers

### Testing & Docs
- Pin `deriveRecentSessions` subagent exclusion with dedicated test
- New test files for activitySections, useProjectSessionLists, useSessionGrouping, sidebarSubtaskFilter, useUIStore
- Document subagent-gating in countNodes and children render paths

### Refactor
- Subtask-aware sorting in sortable items
- Activity sections refactored for subtask filter support

PR #1893, opened 2026-06-28
feat(sidebar): add toggle to hide subagent/delegated sessions

herjarsa:
## Summary

Adds a user-preference toggle that hides subagent and delegated sessions (sessions with a parentID) from both the desktop sidebar and the mobile sessions sheet. When enabled, only top-level parent sessions are shown — the child sessions that contain delegated subtask execution are filtered out.

## Changes

- **useUIStore**: Added `showSubagentSessionsInSidebar` boolean state with `setShowSubagentSessionsInSidebar` setter, defaulting to `false` (subagent sessions hidden)
- **useSessionGrouping**: Filters out sessions with `parentID` when the setting is `false`, preventing child sessions from appearing anywhere in the project tree
- **MobileSessionsSheet**: Same filter applied — subagent sessions excluded from bucket rendering, with a toggle in the mobile sheet UI below the search bar
- **SessionSidebar**: Toggle switch added between the header and the projects list
- **i18n**: `sidebar.subagent.toggleLabel` and `sidebar.subagent.toggleDescription` keys added to all 10 locales

## Testing

- Setting defaults to `false` (subagents hidden) — matches existing behavior in the "Recent" section, session switcher, and tray sync which already filter out subtask sessions
- When toggled on, all sessions including subagents appear
- Mobile sessions sheet and desktop sidebar are both affected by the same setting

### DIFF
--- a/packages/ui/src/apps/MobileSessionsSheet.tsx
+++ b/packages/ui/src/apps/MobileSessionsSheet.tsx
@@ -45,6 +45,7 @@ import { PROJECT_COLOR_MAP, PROJECT_ICON_MAP, ProjectIconImage } from '@/lib/pro
 import { cn } from '@/lib/utils';
 import { listProjectWorktrees } from '@/lib/worktrees/worktreeManager';
 import { useDirectoryStore } from '@/stores/useDirectoryStore';
+import { useUIStore } from '@/stores/useUIStore';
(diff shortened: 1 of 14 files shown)


herjarsa:
Superseded by #1666 which adds the full store-backed subagent filtering preference with tests. This PR was an earlier subset.

PR #1912, opened 2026-06-29
fix(mobile): group sessions under the most specific project, not an ancestor

### LINKED ISSUE #1911: [Bug] Mobile sidebar groups nested-project sessions under an ancestor (home) project

On the **mobile web / PWA** surface, a new chat started in a nested project is visually grouped under an **ancestor** project in the sessions sidebar, even though the session is correctly bound to the nested project and runs there. This is **mobile-only and visual** — Desktop Web and the Desktop app group the same session correctly.

It happens whenever one registered project's path is a filesystem ancestor of another's. The classic case is adding your **home directory** as a project (e.g. `/Users/me`, shown as "me") alongside a nested project like `/Users/me/scratch` ("Scratch").

**Steps to reproduce**

1. On mobile (Web/PWA), register two projects where one path is an ancestor of the other — e.g. your home dir `/Users/me` and a nested folder `/Users/me/scratch`.
2. Tap **+** next to **Scratch** and start a new chat.
3. The session runs under Scratch (correct), and on Desktop it appears under Scratch.
4. In the mobile sidebar, the new session instead appears under the **home** project, and Scratch shows `0` sessions.

In fact *every* session under `/Users/me/scratch` (and any other nested project) is absorbed into the home project on mobile, so the home project's count is inflated while nested projects read `0`.

- **Expected:** the session appears under the most specific project it belongs to (Scratch).
- **Actual:** it appears under the ancestor (home) project.

**Root cause**

The mobile surface renders `MobileSessionsSheet` (not the shared `SessionSidebar`). It buckets sessions into projects with a **prefix match + first-match-wins**, with no "most specific path" tiebreak — in `packages/ui/src/apps/MobileSessionsSheet.tsx` (`projectNodes`):

```ts
const node = nodes.find((entry) => {
  if (pathBelongsToRoot(directory, entry.project.path)) return true;
  return entry.project.worktrees.some((wt) => pathBelongsToRoot(directory, wt.path));
});
```

where `pathBelongsToRoot(path, root)` is true when `path === root || path.startsWith(root + '/')`. Because `/Users/me/scratch` starts with `/Users/me/`, the home project matches, and `Array.find` returns the **first** matching project in the list (the ancestor) instead of the deepest match.

By contrast, Desktop buckets active sessions by an **exact** normalized-directory key (`sessionsByDirectory.get(directory)` in `packages/ui/src/components/session/sidebar/hooks/useProjectSessionLists.ts`), so it groups correctly. The same first-match prefix logic is also used elsewhere in `MobileSessionsSheet.tsx` (active-project highlight on session select, and the search labels/counts), so those are affected too.

**Suggested fix**

In the mobile bucketing (and the other prefix-match call sites), select the **most specific** project/worktree — the longest matching path — rather than the first prefix match.

I have a fix ready and will open a PR shortly.

### Where does it happen?

Mobile (Web/PWA)

### Version (if known)

1.13.2

The home-directory project shows a large session count while the nested project ("Scratch") shows `0`, even though the freshly created session is running under the nested project.

### Logs (optional)

N/A — purely a client-side grouping/render issue; no server logs involved.

MaxAnderson95:
## What

On the mobile web / PWA surface, sessions in a nested project were grouped under an **ancestor** project in the sidebar — most visibly when the home directory is registered as a project alongside nested projects beneath it. The nested project showed `0` sessions while the ancestor's count was inflated. The session itself was correctly bound to and running under the nested project, and Desktop grouped it correctly, so the bug was **mobile-only and visual**.

Closes #1911.

## Why

`MobileSessionsSheet` buckets sessions into projects with a **prefix match + first-match-wins**, with no most-specific tiebreak:

```ts
const node = nodes.find((entry) => {
  if (pathBelongsToRoot(directory, entry.project.path)) return true;
  return entry.project.worktrees.some((wt) => pathBelongsToRoot(directory, wt.path));
});
```

`pathBelongsToRoot(path, root)` is true when `path === root || path.startsWith(root + '/')`, so `/home/me/scratch` matches the ancestor `/home/me`, and `Array.find` returns the first matching project in list order (the ancestor) rather than the deepest match. Desktop instead buckets active sessions by an exact normalized-directory key (`packages/ui/src/components/session/sidebar/hooks/useProjectSessionLists.ts`), which is why it is unaffected.

## What changed

- Added a shared `findMostSpecificProjectMatch(directory)` helper that returns the project/worktree with the **longest matching path**.
- Used it in the five spots that previously did first-match prefix resolution:
  - the `projectNodes` sidebar grouping loop (the visible bug)
  - `handleSelectSession` (active-project highlight was following the ancestor project)
  - `buildSessionContextLabel` (search-result subtitle)
  - `searchSessionMatches` (search query matching)
  - `searchProjectMatches` (per-project session counts in search)

A session is now grouped under the most specific project it belongs to, consistent with Desktop.

## Testing

- Reproduced on a phone PWA with a home-directory project plus a nested project; after the fix the new session appears under the nested project and the sidebar counts are correct.
- Scoped validation for the affected packages: `bun run type-check:ui` ✅, `bun run lint:ui` ✅, `bun run build:web` ✅ (the change lives in `packages/ui`, consumed by `packages/web`; the built bundle was verified serving the mobile surface).

## Scope

Single file: `packages/ui/src/apps/MobileSessionsSheet.tsx` (+54 / −31). No theme, API, or behavior changes beyond grouping; functional component, TypeScript strict, early returns.

### DIFF
--- a/packages/ui/src/apps/MobileSessionsSheet.tsx
+++ b/packages/ui/src/apps/MobileSessionsSheet.tsx
@@ -627,6 +627,38 @@ export const MobileSessionsSheet: React.FC<MobileSessionsSheetProps> = ({ open, 
     return merged;
   }, [globalActiveSessions, liveSessions]);
 
+  // Resolve the MOST SPECIFIC project (and worktree) a directory belongs to,
+  // i.e. the one with the longest matching path. Plain first-prefix-match lets an
+  // ancestor project (e.g. the home directory) swallow sessions from nested
+  // projects whose paths it prefixes. Used for grouping, the active-project
+  // highlight on selection, and search labels/counts so they all agree.
+  const findMostSpecificProjectMatch = React.useCallback(
+    (directory: string): { project: ProjectMeta; worktree: WorktreeMetadata | null } | null => {
+      if (!directory) return null;
+      let bestProject: ProjectMeta | null = null;
+      let bestWorktree: WorktreeMetadata | null = null;
+      let bestMatchLength = -1;
+      for (const project of projectsMeta) {
+        const projectRoot = normalizePath(project.path);
+        if (projectRoot && pathBelongsToRoot(directory, projectRoot) && projectRoot.length > bestMatchLength) {
+          bestProject = project;
+          bestWorktree = null;
+          bestMatchLength = projectRoot.length;
+        }
+        for (const worktree of project.worktrees) {
+          const worktreeRoot = normalizePath(worktree.path);
+          if (worktreeRoot && pathBelongsToRoot(directory, worktreeRoot) && worktreeRoot.length > bestMatchLength) {
+            bestProject = project;
+            bestWorktree = worktree;
+            bestMatchLength = worktreeRoot.length;
+          }
+        }
+      }
+      return bestProject ? { project: bestProject, worktree: bestWorktree } : null;
+    },
+    [projectsMeta],
+  );
+
   const normalizedQuery = query.trim().toLowerCase();
 
   const projectNodes = React.useMemo<ProjectNode[]>(() => {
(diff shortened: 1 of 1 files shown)


MaxAnderson95:
Closing as superseded.

After opening this, I confirmed `main` (v1.13.8) already fixes #1911 by grouping sessions with **exact** directory matching (`projectMatchesExactDirectory` / `findExactProjectMatch`, introduced in 784fa138). This PR was built on the v1.13.2 base and takes a different approach (most-specific longest-prefix match) on the same lines, so it is no longer needed for the reported bug. Thanks!

PR #1915, opened 2026-06-29
fix(model-picker): scroll to and expand the selected model on open in Settings

### LINKED ISSUE #1906: Model picker dropdown does not scroll to or expand the currently selected model
## Summary

When opening the model picker dropdown in agent settings (or chat), the list does not scroll to the currently selected model. If the selected model is inside a collapsed section (e.g. a provider section), that section remains collapsed and the user has to manually find and expand it.

## Steps to reproduce

1. Open Settings → Agents → select an agent that has a model configured
2. Click the model selector dropdown
3. Observe: the list is scrolled to the top, not to the selected model
4. If the selected model's provider section was previously collapsed, it stays collapsed

## Expected behavior

- On open, the dropdown should scroll to the selected model entry
- If the selected model's provider section is collapsed, it should be expanded automatically
- The selected model entry should be highlighted (keyboard selection should point to it)

## Technical notes

- `ModelPickerList.tsx` receives `selectedModel` prop and correctly marks the row as `aria-selected` (line 572), but there is no `useEffect` that scrolls to it or expands its section
- The selection store (`selectionStore`) always starts at index 0 on open
- Section collapse state is managed via `useModelPickerSectionsStore` — `toggleSection` can be called to expand the relevant provider section
- `scrollIntoView` helper already exists in the file (line 292)

bashrusakh:
## Summary

The model picker dropdown in Settings → Agents now seeds keyboard selection at the currently configured model, expands its provider section when collapsed, and scrolls the row into view. The chat quick-switch picker keeps the existing top-of-list behavior.

Fixes #1906.

## Problem

When opening the model picker dropdown, the list scrolled to the top and `selectionStore` always started at index 0. If the selected model lived inside a collapsed provider section, that section stayed collapsed and the user had to manually find and expand it to confirm the configured model.

Root cause (`packages/ui/src/components/model-picker/ModelPickerList.tsx`):
- `createIndexSelectionStore` initialized `value = 0`; the only `selectionStore.set(0)` was tied to `searchQuery` changes — nothing pointed the store at the selected model's index.
- `scrollIntoView` was only invoked from `moveSelection` (keyboard nav); no effect scrolled to the selected model on open.
- `flatModelList` excluded collapsed `provider:*` sections, so a selected model inside a collapsed section was invisible and unreachable without manual expansion.

## Change

- Added opt-in prop `scrollToSelectedOnOpen` (default `false`) on `ModelPickerList`.
- Added a `useEffect` that, on open, when `selectedModel` is set and no search query is active: expands the selected model's `provider:<id>` section via `toggleSection` if collapsed, finds the row index in `flatModelList`, seeds `selectionStore` to that index, and scrolls it into view via `requestAnimationFrame` (so the just-expanded row has been committed to the DOM first).
- Enabled `scrollToSelectedOnOpen` in `packages/ui/src/components/sections/agents/ModelSelector.tsx` (Settings → Agents), where confirming the configured model matters.
- Left the chat quick-switch picker (`packages/ui/src/components/chat/ModelControls.tsx`) unchanged: returning to the current model in a quick-switch context can be disorienting, so that stays a maintainer decision.

## Maintainer decision needed — chat scope

> **Important.** This PR intentionally does **not** change the chat quick-switch model picker. The same `ModelPickerList` component is reused there, so the bug is reproducible in chat as well, but whether auto-scrolling to the current model is desirable in that context is a product call, not a bug fix.

Original note from the issue reporter (`@bashrusakh`, [#1906 (comment)](https://github.com/openchamber/openchamber/issues/1906#issuecomment-4827346247)):

> The same component is used in the chat model picker too, so the bug is reproducible there as well.
>
> However, I'm 50/50 on whether fixing it in chat is the right call — the chat picker is a quick-switch context where scrolling to the current selection might actually be disorienting (you often want to browse other models, not return to the current one). In Settings → Agents, it's clearly a bug because you're configuring a specific agent and need to see which model is selected.
>
> I'd suggest the maintainer decides on the chat scope. The fix for Settings → Agents is unambiguous.

This PR resolves the unambiguous part (Settings → Agents). If the maintainer wants the same behavior in the chat quick-switch picker, it's a one-line change: pass `scrollToSelectedOnOpen` in `packages/ui/src/components/chat/ModelControls.tsx`.

## Verification

- `bun run type-check` (packages/ui) — clean
- `bun run lint` (packages/ui) — clean

Manual checks to confirm before merge:
- Open Settings → Agents with a selected model whose provider section is collapsed → section auto-expands, list scrolls to the row, keyboard highlight lands on it.
- Open with no model selected → no error, selection starts at index 0 (existing behavior preserved).
- Type a search query that excludes the selected model → no scroll jump, selection resets to 0 (existing behavior preserved).

### DIFF
--- a/packages/ui/src/components/model-picker/ModelPickerList.tsx
+++ b/packages/ui/src/components/model-picker/ModelPickerList.tsx
@@ -365,6 +365,17 @@ interface ModelPickerListProps {
   footerContent?: React.ReactNode | ((activeEntry: ModelPickerEntry | undefined) => React.ReactNode);
   renderVersion?: number;
   tooltipsEnabled?: boolean;
+  /**
+   * When the picker opens (and `selectedModel` is set), auto-expand the
+   * selected model's provider section if it is collapsed, seed keyboard
+   * selection at that row, and scroll it into view. Default: false.
+   *
+   * Intended for configuration surfaces (e.g. Settings → Agents) where the
+   * user needs to confirm the currently configured model. Quick-switch
+   * surfaces (chat) usually want to browse other models, so they leave this
+   * off.
+   */
+  scrollToSelectedOnOpen?: boolean;
 }
 
 export const ModelPickerList: React.FC<ModelPickerListProps> = ({
@@ -405,6 +416,7 @@ export const ModelPickerList: React.FC<ModelPickerListProps> = ({
   footerContent,
   renderVersion,
   tooltipsEnabled = true,
+  scrollToSelectedOnOpen = false,
 }) => {
   const selectionStoreRef = React.useRef<IndexSelectionStore | null>(null);
   if (!selectionStoreRef.current) selectionStoreRef.current = createIndexSelectionStore();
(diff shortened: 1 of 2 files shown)


btriapitsyn:
Thanks, Leonid! I updated this for the current picker and fixed the scroll positioning. Opening the model picker in Settings now shows the selected model, including when its provider was collapsed.

PR #1926, opened 2026-06-29
fix(ui): ensure created sessions keep their workspace directory

### LINKED ISSUE #1748: Workspace always shows "No sessions in this workspace yet" — new sessions go to Recent (or main branch) regardless of Git state
## Summary

When I open a folder as a workspace in OpenChamber and click the `+` next to it to create a new session, the new session **never** appears under that workspace in the sidebar. Instead:

- **If the folder is not a Git repo:** the new session appears in the global **Recent** section, while the workspace itself still reads **"No sessions in this workspace yet"**.
- **If I `git init` the same folder and then click `+`:** the new session appears under the **main** branch, but the workspace itself still reads **"No sessions in this workspace yet"**.

So in both cases the workspace pane stays empty even though sessions I just created are obviously tied to that directory (they show up in Recent / main branch). The behavior is consistent across **Windows** and **macOS**.

## Reproduction

1. Open a folder as a workspace in OpenChamber (folder does **not** have to be a Git repo).
2. Observe the workspace in the sidebar — it shows **"No sessions in this workspace yet"**.
3. Click the `+` next to that workspace to create a new session.
4. **Without `git init`:** the new session is created and shows up in the **Recent** section; the workspace itself is **still** "No sessions in this workspace yet".
5. Now `git init` the same folder, click `+` again on the workspace.
6. **With `git init`:** the new session is created and shows up under **main**; the workspace itself is **still** "No sessions in this workspace yet".

Expected: the new session should appear under the workspace I clicked `+` on in the sidebar.

Actual: the workspace pane never updates; the session is created elsewhere (Recent, or main branch).

## Question

Is this intentional? Two cases both leave the workspace list showing "No sessions":

- Non-Git workspace → session silently falls through to **Recent**.
- Git workspace → session lives under **main branch** but the workspace top-level still shows "No sessions".

If non-Git workspaces are not supported, the UI should make that explicit (icon, disable `+`, or hide the workspace). If they are supported (and `prddesignmd` clearly shows up as a workspace), this looks like the session is being created with the wrong `directory` / workspace binding.

## Environment

- OpenChamber Desktop on **macOS** and **Windows** (reproduced on both).
- Workspace in question: a folder of Markdown / design docs (no source code, intentionally no Git history initially).
- Behavior is identical before and after `git init`, only the destination of the newly created session changes.

bashrusakh:
## Summary

Fixes #1748.

When creating a session via the sidebar '+' button on a workspace, the server response sometimes omits the legacy `directory` field. New sessions then land in 'Recent' (non-Git) or under the 'main' branch (Git) instead of the target workspace.

## Root cause

The V2 `session.create` response (`SessionV2Info`) places the directory in `location.directory` rather than the legacy `directory` field. If the field is missing entirely, `resolveGlobalSessionDirectory()` returns `null`, so the session never enters the sidebar's `sessionsByDirectory` map.

## Changes

- `packages/ui/src/lib/opencode/client.ts`: `createSession` now normalizes `location.directory` to `directory`, and falls back to the requested directory when the server omits metadata completely.
- `packages/ui/src/sync/session-actions.ts`: `directoryOverride` is now captured as `requestedDirectory` and passed through to `opencodeClient.createSession`.
- Added unit tests in `client.test.ts` and `session-actions.test.ts`.

## Validation

- `bun run type-check` ✅
- `bun run --filter @openchamber/ui lint` ✅
- `bun test packages/ui/src/lib/opencode/client.test.ts packages/ui/src/sync/session-actions.test.ts packages/ui/src/stores/useGlobalSessionsStore.test.ts` → 34 pass, 0 fail
- `ocr review --commit 305a7a9c` → no comments


btriapitsyn:
Resolved the conflicts with `main` and preserved the session-directory routing fix:

- normalize V2 `location.directory` into the shared Session contract;
- fall back to the exact directory captured at create time when the server omits directory metadata;
- preserve existing legacy directory values and the newer ambiguous-send tests from `main`;
- verified 32 focused client/session-action tests, UI type-check, and lint.

PR #1931, opened 2026-06-29
fix(mcp): toggle switch now reflects config state and server proxy race condition

herjarsa:
### MCP toggle UI fix
The `McpDropdown` switch was showing connection status (`serverStatus.status === "connected"`) instead of config state (`configEntry.enabled`) for config-managed MCP servers. After a successful PATCH to `/api/config/mcp/{name}`, the switch would NOT visually update because the MCP process status was still `failed`/`disconnected`.

**Fix**: Changed `isConnected` to prioritize `configEntry.enabled` when a config entry exists:
```tsx
// Before (broken):
const isConnected = serverStatus ? serverStatus.status === "connected" : (configEntry?.enabled ?? false);

// After (fixed):
const isConnected = configEntry ? configEntry.enabled : serverStatus?.status === "connected";
```

### Server proxy race condition (Express 5)
The API proxy at `packages/web/server/index.js:1332` was registered AFTER the server started listening, creating a race window where early `/api/*` requests would fall through to the OpenCode upstream SPA instead of the Express handler. Fixed by changing `return next()` to plain `return` in the proxy guard — Express 5 sub-routers call `next()` concurrently with route handlers, causing ERR_HTTP_HEADERS_SENT.

### INTERNAL_API_PREFIXES audit
Rebuilt the proxy exclusion list to only block paths with actual Express handlers. Removed `/api/config/` (too broad — blocked legit proxy paths) and replaced with specific config paths. Removed `/api/path`, `/api/session/status`, `/api/experimental/session`, `/api/global/`, `/api/project/current` from exclusion (no Express handlers, should proxy to OpenCode).

## Testing
- Playwright: MCP toggle `false→true` verified, switch updates visually
- curl: All proxy endpoints return correct status codes
- Manual: Add project directory dialog now shows directories after idle periods


herjarsa:
Closing this PR as stale. The branch contains 52 commits accumulated against an old snapshot of \main\ (1631 files changed, -186k lines) and the conflicts with current \main\ cannot be resolved safely without file-by-file review. The intent of this PR has been split into smaller, independently-rebased replacements where possible (#2797, #2801, #2802); the rest of the changes (mobile subagent filter, plugin status panel, subagent observability, MCP toggle) live in newer PRs on top of current \main\ and can be re-submitted incrementally if still desired.

PR #1933, opened 2026-06-29
fix(sync): preserve optimistic user text when slash-command server echo is shorter

### LINKED ISSUE #1932: [Bug] First user message in a session loses trailing text after a leading slash command
## Summary

When the first message in a session starts with a slash command and trailing user input, the user-visible message in chat truncates to the bare command. The model itself does receive the full arguments correctly (verified in the developer-flow), so this is a UI-side display regression only.

## Reproduction

1. Open OpenChamber and create a new session.
2. Pick any slash command from the autocomplete (or type it) that takes arguments, for example:
   - Type a user-installed command such as `/debug my symptom: import fails on startup with ModuleNotFoundError`
   - Or pick a built-in such as `/summary focus on auth`
3. Press Send.
4. Observe the chat bubble for the user message: only `/debug` (or `/summary`) is rendered. The trailing text is gone.
5. Watch the model reply: it actually does respond as if the full text was sent. The arguments reach it correctly. Only the visible chat bubble is wrong.

## Root cause

OpenCode's server echo for `POST /session/{id}/command` returns a user-message part whose `text` field contains only the command name, dropping the trailing user input. The OpenChamber sync layer (event-reducer) replaces the optimistic user-message text part (which carried the full text) with the server echo, truncating chat.

The split into `command` + `arguments` on the client side already works correctly (verified in `packages/ui/src/sync/session-ui-store.ts` and `packages/ui/src/lib/opencode/client.ts`, both pass the full text through). The display issue is purely in the SSE→store path.

## Proposed fix (PR attached)

In `packages/ui/src/sync/event-reducer.ts`, the `message.part.updated` reducer now skips the optimistic→server text part replacement when the incoming server text is a non-empty strict shorter prefix of the optimistic text, and adopts the server part's id so any follow-up event for the same id updates the kept part in place. Empty echoes and non-prefix echoes still replace as before.

Three new unit tests cover the prefix branch, the equal-length branch, the empty-echo branch and a follow-up update for the adopted id.

## Validation

- `bun run type-check` — pass
- `bun run lint` — pass
- `bun test packages/ui/src/sync/__tests__/event-reducer.test.ts` — 17/17 pass

## Notes

The same root cause likely affects other prefixes mentioned in the original report (`@agent`, `!shell`, `#snippet`). Those flows use different code paths and are not regressed by this fix; the comment in the change documents that the prefix branch is gated on text parts, so non-text-prefixed flows (e.g. agent mentions) are untouched.

Filed against `openchamber/openchamber`. Patch submitted as draft PR from `bashrusakh/openchamber:fix/slash-command-echo-truncation`.

bashrusakh:
## Summary

Resolves #1932.

When a slash command is sent as the first message in a session (or any message), the OpenCode server's user-message echo for `POST /session/{id}/command` can contain only the command name and drop the trailing user input. The OpenChamber sync layer was replacing the optimistic user-message part (which carried the full text) with that truncated echo, making the user-visible chat bubble show only the slash command.

The model itself receives the full `arguments` via `session.command` correctly; this patch only restores the full text in chat.

## What changed

- `packages/ui/src/sync/event-reducer.ts`: in `message.part.updated`, when the server text is a non-empty strict shorter prefix of the optimistic text, keep the optimistic text and adopt the server part's id so any follow-up update for that id updates the kept part in place. Empty echoes and non-prefix echoes still replace as before.
- `packages/ui/src/sync/__tests__/event-reducer.test.ts`: three new unit tests covering the prefix branch, the non-prefix branch, the empty-echo branch, and a follow-up update with the adopted id (verifies no duplicate part is inserted).

## How I verified

- `bun run type-check` — pass across all workspaces
- `bun run lint` — pass across all workspaces
- `bun test packages/ui/src/sync/__tests__/event-reducer.test.ts` — 17 pass / 0 fail (12 pre-existing + 5 new)

## Manual reproduction before / after

Before the fix, sending `/debug my symptom: import fails on startup with ModuleNotFoundError` from a fresh session rendered only `/debug` in the chat bubble (the assistant reply still responded as if the full arguments were sent).

After the fix, the full text is preserved in chat; a follow-up `message.part.updated` for the same server part id updates the kept part in place.

## Notes

- Fix level is sync-layer event reducer — the only place where the optimistic text was being lost. `routeMessage` and `client.sendCommand` already pass full arguments through correctly.
- Comment in the change documents that this branch is gated on text parts, so non-text-prefixed flows (`@agent`, `!shell`, `#snippet`) are not affected.
- No public API / data-model changes.
- Branch: `fix/slash-command-echo-truncation` in `bashrusakh/openchamber`, base `main` of `openchamber/openchamber`.

### DIFF
--- a/packages/ui/src/sync/__tests__/event-reducer.test.ts
+++ b/packages/ui/src/sync/__tests__/event-reducer.test.ts
@@ -255,4 +255,189 @@ describe("applyDirectoryEvent", () => {
     expect(draft.question.ses_1).not.toBe(afterReply)
     expect(draft.question.ses_1).toEqual([])
   })
+
+  test("preserves optimistic text when server echo is a shorter prefix (slash command)", () => {
+    // Optimistic insert for a slash command like "/debug my symptom" carries the
+    // full user input. The OpenCode server echo for slash commands can omit the
+    // arguments, returning only the command name. To avoid truncating the user
+    // message visible in chat, the reducer must keep the optimistic text when
+    // the server text is a strict non-empty shorter prefix of the optimistic text.
+    // The reducer should also adopt the server part's id so follow-up
+    // `message.part.updated` events for that id correctly update the entry.
+    const optimisticText = "/debug my symptom: import fails on startup with ModuleNotFoundError"
+    const serverText = "/debug"
+    const draft = state({
+      message: { ses_1: [{ id: "msg_1", sessionID: "ses_1", role: "user", time: { created: 1 } } as never] },
+      part: {
+        msg_1: [
+          {
+            id: "prt_optim",
+            messageID: "msg_1",
+            type: "text",
+            text: optimisticText,
+          } as Part,
+        ],
+      },
+    })
+
+    const result = applyDirectoryEvent(draft, {
+      type: "message.part.updated",
+      properties: {
+        part: {
+          id: "prt_server",
+          messageID: "msg_1",
+          sessionID: "ses_1",
+          type: "text",
+          text: serverText,
+        },
+      },
+    } as Event)
+
+    // Result must be a structural change (adopted id) so React re-renders.
+    expect(result).toBe(true)
+    const parts = draft.part.msg_1 ?? []
+    expect(parts.length).toBe(1)
+    expect((parts[0] as { text?: string }).text).toBe(optimisticText)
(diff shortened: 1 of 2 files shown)


btriapitsyn:

PR #1941, opened 2026-06-30
perf(stores): defer safeStorage writes off the interaction path

tomzx:
# What

- Adds a write-behind buffer to `safeStorage` so `setItem`/`removeItem` calls are deferred to a later task via `setTimeout(0)` instead of running synchronously on the call site.
- Coalesces repeated writes to the same key into a single backing flush.
- Serves pending values from an in-memory map so read-after-write stays consistent within the deferral window.
- Flushes pending writes synchronously on `pagehide` / `beforeunload` / `visibilitychange(hidden)` / `freeze` so deferred state survives tab close, reload, and the mobile freeze lifecycle.
- Adds a test covering write deferral, coalescing, and pending read serving.

# Why

Every persisted zustand store plus the direct writers (projects, directory, session folders, ...) funnels through `safeStorage.setItem`. During a session switch each of those stores re-serializes a whole slice, and doing those large `JSON.stringify` writes synchronously blocked the main thread for over a second, making session switching feel sluggish. Deferring the writes off the click-to-paint path lets the UI paint the interaction first, while coalescing keeps the total write cost down.

# How to test

1. `cd packages/ui && bun test src/stores/utils/safeStorage.test.ts` — both tests pass (existing fallback test plus the new deferral/coalescing test).
2. `bun run type-check:ui` and `bun run lint:ui` — both clean.
3. Manual: open the app, switch between sessions with large persisted state, and confirm the interaction paints immediately; verify persisted state still survives a full reload (deferred flush runs on `pagehide`).

This is a standalone performance fix with no originating issue, so there are no formal acceptance criteria to map.

# References

Standalone performance fix. No linked issue.

---

Created with [create-pr](https://github.com/tomzx/agents/blob/37b68eb/skills/create-pr/SKILL.md) via glm-5.2 (`37b68eb`)

### DIFF
--- a/packages/ui/src/components/session/SessionSidebar.tsx
+++ b/packages/ui/src/components/session/SessionSidebar.tsx
@@ -13,7 +13,7 @@ import { useSync } from '@/sync/use-sync';
 import { useSessionPrefetch } from './sidebar/hooks/useSessionPrefetch';
 import { useProjectsStore } from '@/stores/useProjectsStore';
 import { useUIStore } from '@/stores/useUIStore';
-import { getSafeStorage } from '@/stores/utils/safeStorage';
+import { getDeferredSafeStorage } from '@/stores/utils/safeStorage';
 import { useGitStore, useGitAllBranches, useGitRepoStatusMap } from '@/stores/useGitStore';
 import { isVSCodeRuntime } from '@/lib/desktop';
 import { NewWorktreeDialog } from './NewWorktreeDialog';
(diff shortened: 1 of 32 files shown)


btriapitsyn:
Fixed the review finding in 4e4c5d97.

What changed:
- Moved the deferred/coalesced write-behind behavior from raw safeStorage.setItem to a Zustand PersistStorage helper, so JSON serialization is deferred too.
- Updated localStorage-backed persisted stores to use createDeferredSafeJSONStorage instead of createJSONStorage(() => getSafeStorage()).
- Kept safeStorage itself synchronous as the backing layer to avoid double-deferral and unload-order data-loss risk.
- Updated the targeted safeStorage test to assert JSON.stringify does not run on the call site and coalesces to the latest value.
- Updated the useConfigStore test mock for the new storage helper.

Validation run:
- bun test packages/ui/src/stores/utils/safeStorage.test.ts
- bun test packages/ui/src/stores/useConfigStore.test.ts
- bun run type-check
- bun run lint
- git diff --check

btriapitsyn:
Follow-up fixes pushed in ef47e3e4 after local testing.

What changed:
- Addressed the bot non-blocker about direct safeStorage writers by moving localStorage-backed direct writers to getDeferredSafeStorage().
- Added a shared lifecycle flush coordinator so deferred storages do not each register their own unload listeners.
- Added error visibility in deferred flush paths with console.error around write/remove failures.
- Added targeted coverage for direct deferred writes and pagehide flush.

Validation run before commit:
- bun test packages/ui/src/stores/utils/safeStorage.test.ts
- bun test packages/ui/src/stores/useConfigStore.test.ts
- bun test packages/ui/src/stores/useSessionFoldersStore.test.ts
- bun run type-check
- bun run lint
- git diff --check

btriapitsyn (approved):

PR #1945, opened 2026-06-30
fix: replace dynamic imports with static imports in session-ui-store to eliminate Rollup chunk warnings

### LINKED ISSUE #1884: Rollup warns about modules imported both dynamically and statically
## Description
During `vite build`, Rollup emits 7 warnings about modules that are imported via both `import()` (dynamic) and `import {}` (static) in different files:

```
[plugin vite:reporter]
(!) packages/ui/src/sync/session-actions.ts is dynamically imported by session-ui-store.ts but also statically imported by [13 other files], dynamic import will not move module into another chunk.

(!) packages/ui/src/lib/i18n/store.ts is dynamically imported by session-ui-store.ts but also statically imported by [7 other files]...

[... 5 more similar warnings ...]
```

## Root Cause
`packages/ui/src/sync/session-ui-store.ts` uses `await import()` for lazy loading in several places:

```typescript
// Line 1248-1249
const { toast } = await import("sonner")
const { useI18nStore, formatMessage } = await import("@/lib/i18n/store")

// Line 1259
const { unrevertSession } = await import("./session-actions")
```

However, the same modules are statically imported elsewhere:

```typescript
// packages/ui/src/apps/AppEffects.tsx
import { setOptimisticRefs } from '@/sync/session-actions';

// packages/ui/src/components/chat/message/timeFormat.ts
import { formatMessage, useI18nStore } from '@/lib/i18n/store';
```

Vite/Rollup can't tree-shake these into separate chunks because they're already in the main bundle via static imports.

## Impact
- Build output shows 7 warnings
- No functional impact (modules load correctly)
- Potential performance: dynamic imports that don't create separate chunks defeat code-splitting benefits
- Obscures real build issues

## Files Affected
- `packages/ui/src/sync/session-actions.ts`
- `packages/ui/src/lib/i18n/store.ts`
- `packages/ui/src/lib/openchamberConfig.ts`
- `packages/ui/src/components/chat/message/ToolOutputDialog.tsx`
- `packages/ui/src/lib/git/branchNameGenerator.ts`
- `packages/ui/src/lib/worktrees/worktreeManager.ts`
- `packages/ui/src/lib/worktrees/worktreeCreate.ts`

## Suggested Fix
Make import patterns consistent for each module:
- If a module is used in multiple places and should be in the main bundle, use static imports everywhere
- If a module should be code-split, ensure it's only dynamically imported
- Alternatively, configure `build.rollupOptions.output.manualChunks` in `packages/web/vite.config.ts` to explicitly control chunking

## Environment
- openchamber monorepo (Electron desktop app)
- Vite build in `packages/web`
- React 18 with code-splitting via `React.lazy()`

moisesvalero:
Closes #1884

## Problem

Rollup emitted 7 warnings during `vite build` about modules imported both dynamically (`await import()`) and statically (`import {}`) in different files. The dynamic imports could not create separate chunks because the modules were already in the main bundle via static imports elsewhere, defeating code-splitting benefits.

## Fix

In `packages/ui/src/sync/session-ui-store.ts`:

- Added static imports for `sonner`, `@/lib/i18n/store` (`formatMessage`, `useI18nStore`), and `@/lib/worktrees/worktreeManager` (`removeProjectWorktree`)
- Removed all 12 `await import()` calls that referenced these modules (they appeared in `handleSlashRevert`, `handleSlashRedo`, `forkFromMessage`, and `createSession` handlers)
- Replaced `const { unrevertSession } = await import("./session-actions")` with `unrevertSessionAction`, which was already statically imported

## Result

- 4 static imports added, 13 lines removed
- No functional change — all modules were already in the main bundle
- Build output is now cleaner (0 warnings for these modules)

### DIFF
--- a/packages/ui/src/sync/session-ui-store.ts
+++ b/packages/ui/src/sync/session-ui-store.ts
@@ -56,6 +56,9 @@ import {
   forkFromMessage as forkFromMessageAction,
   fetchMessagesForSession,
 } from "./session-actions"
+import { toast } from "sonner"
+import { formatMessage, useI18nStore } from "@/lib/i18n/store"
+import { removeProjectWorktree } from "@/lib/worktrees/worktreeManager"
 import { useInputStore, type SyntheticContextPart } from "./input-store"
 import { useSelectionStore } from "./selection-store"
 import { getViewportSessionMemory, useViewportStore, viewportSessionKey } from "./viewport-store"
@@ -1250,8 +1253,6 @@ export const useSessionUIStore = create<SessionUIState>()((set, get) => ({
     // revertToMessage handles the redo stack push internally
     await get().revertToMessage(sessionId, targetMessage.id)
 
-    const { toast } = await import("sonner")
-    const { useI18nStore, formatMessage } = await import("@/lib/i18n/store")
     const { dictionary } = useI18nStore.getState()
     toast.success(formatMessage(dictionary, "chat.revert.toast.undo", { preview }))
   },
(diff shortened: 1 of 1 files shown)


moisesvalero:
Hi @btriapitsyn, friendly ping — just checking if you've had a chance to look at this when you have time. Happy to make any adjustments if needed.

btriapitsyn:
Thanks for digging into #1884. Unfortunately this can't land as-is: `worktreeManager.ts` statically imports `session-ui-store.ts`, so the dynamic import you replaced was intentionally breaking a circular dependency — making it static creates a static import cycle in module init for a build-log-only benefit. The sonner/i18n replacements alone don't justify touching this hot path, so I'm closing this; the warning itself is harmless.

PR #1949, opened 2026-06-30
fix(chat): surface session.command resolved payload (#1944)

### LINKED ISSUE #1944: [Bug] Chat bubble displays empty string for `!`command`` expressions in command files (display only, execution works)
### Description

When using an OpenCode command file (`.md`) that contains `!`command`` expressions (inline shell execution), the chat bubble in OpenChamber renders these expressions as **empty strings** instead of the actual command output. The actual execution still works correctly — the command is properly resolved when executed — but the **display** in the chat bubble shows a blank.

### Steps to Reproduce

1. Create a command file at `~/.config/opencode/commands/push-merge.md` with content containing `!`git branch --show-current``
2. Trigger the command in OpenChamber
3. Observe the chat bubble — the `!`git branch --show-current`` part appears as empty/blank
4. Observe the actual execution — the correct branch name IS used when the command runs

### Example Command File

```yaml
---
description: Sync to remote target branch
---

Steps:
1. Run git pull origin $1
2. If conflict occurs, resolve it and write resolution to commit message
3. Run git push origin !`git branch --show-current`
4. Run git checkout $1
5. Run git merge !`git branch --show-current`
6. Run git push origin $1
7. Run git checkout !`git branch --show-current`
```

### Expected Behavior

The chat bubble should display the **resolved value** of `!`command`` expressions (e.g. `main`, `feature/foo`), matching what the CLI version of OpenCode shows.

### Actual Behavior

The bubble shows an **empty string** where the resolved command output should be — the `!`git branch --show-current`` is replaced with nothing in the display, even though the backend correctly resolves and executes it.

### Environment

- OpenChamber version: latest
- Platform: Desktop app
- OS: Windows 11
- OpenCode: CLI integration with custom command files

### Additional Notes

- The `!` prefix is OpenCode’s inline shell execution syntax
- The actual command execution logic is fine — only the **chat bubble rendering/display** is affected
- This is purely a UI/display bug in the bubble component

bashrusakh:
## Summary
- Stop discarding the `session.command` SDK response. The server returns the resolved user message (`{ info, parts }`) with `!`shell` expressions already substituted; surfacing that payload through `optimisticSend` and the optimistic add path restores the resolved value in the chat bubble when the SSE echo arrives with the template token unresolved.
- Minimal change: 3 files in `packages/ui/src/{lib/opencode,sync}/`. The server's `messageID` echo means `mergeOptimisticPage` dedupes the message and merges the resolved parts against the optimistic placeholder — no double-render, no extra render cascade.

## Context
Closes #1944. Reported on Windows desktop: `/push-merge target-branch` (command file with `!`git branch --show-current``) renders a blank where the resolved branch name should appear in the chat bubble, while backend execution works correctly.

Root cause per the upstream analysis: `sendCommand` in `packages/ui/src/lib/opencode/client.ts` previously did `unwrapSdkOptional(response, 'session.command'); return tempMessageId;`, dropping the response body. The UI then relied entirely on SSE to populate the user message; if the SSE echo carried the unresolved `!`cmd` token with an empty resolved value, the bubble showed a blank.

## Changes
- `packages/ui/src/lib/opencode/client.ts` — `sendCommand` now returns `{ messageId, resolved: { info, parts } | null }` instead of just the message id.
- `packages/ui/src/sync/session-actions.ts` — `optimisticSend.send` callback may return `{ info, parts }`; the caller applies it through `_optimisticAdd` after the send completes. Dedup is handled by the existing shadow Map + `mergeOptimisticPage` path.
- `packages/ui/src/sync/session-ui-store.ts` — caller forwards the resolved payload.

## Fix level
Right-level: the only OpenChamber-side change with a chance of restoring the resolved value. The display pipeline (`UserTextPart`, `MessageBody`, `markdown/markdownCore`) does not strip `!`cmd` tokens, so no display-side workaround is needed. The actual server-side template resolver is out of scope (separate repo, AGENTS rule "Do not modify `../opencode`").

## Validation
- `bun run lint` — passes
- `bun test packages/ui/src/sync/session-actions.test.ts` — 19/19 pass
- `bun test packages/ui/src/lib/opencode/client.test.ts` — 1/1 pass
- Pre-push provenance: branch is at `origin/main` with only the 3 files for this fix; no unrelated commits.

## Notes
- This is a defensive fix: if the server already echoes the resolved value via SSE, the optimistic-add dedupes and behavior is unchanged. If the server does not, the resolved value is now present immediately.
- Sibling OpenCode-server issue recommended: investigate the template resolver / SSE payload to ensure the resolved value is always included in the user-message echo so this client-side workaround is not needed long-term.

### DIFF
--- a/packages/ui/src/lib/opencode/client.ts
+++ b/packages/ui/src/lib/opencode/client.ts
@@ -908,7 +908,7 @@ class OpencodeService {
     files?: Array<FileInputLite>;
     messageId?: string;
     directory?: string | null;
-  }): Promise<string> {
+  }): Promise<{ messageId: string; resolved: { info: Message; parts: Part[] } | null }> {
     const tempMessageId = params.messageId ?? ascendingId("msg");
 
     const parts: FilePartInput[] = [];
@@ -932,8 +932,19 @@ class OpencodeService {
       messageID: tempMessageId,
     });
 
-    unwrapSdkOptional(response, 'session.command');
-    return tempMessageId;
+    const data = unwrapSdkOptional(response, 'session.command');
+    // The OpenCode server returns the resolved user message ({ info, parts })
+    // from session.command, with `!`shell` expressions already substituted.
+    // Previously we discarded it, so the UI had to rely entirely on the SSE
+    // echo — and if the echo arrived with the template token unresolved, the
+    // chat bubble showed a blank where the resolved value should be (#1944).
+    // Surface the resolved payload so the caller can apply it through the
+    // optimistic add path (which handles dedup against the later SSE echo).
+    const resolved =
+      data && typeof data === 'object' && 'info' in data && 'parts' in data
+        ? (data as { info: Message; parts: Part[] })
+        : null;
+    return { messageId: tempMessageId, resolved };
   }
 
   async abortSession(id: string): Promise<boolean> {
--- a/packages/ui/src/sync/session-actions.test.ts
+++ b/packages/ui/src/sync/session-actions.test.ts
@@ -873,4 +873,97 @@ describe("optimisticSend resolved payload (#1944)", () => {
 
     expect(optimisticAdds).toHaveLength(1)
   })
+
+  test("patches the message even when resolved.parts is empty (OCR finding)", async () => {
+    const targetStore = createStore({})
+    const childStores = createChildStores([["/target/project", targetStore]])
+    let sentMessageID = ""
+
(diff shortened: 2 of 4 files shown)


bashrusakh:
@btriapitsyn hi, most likly you are changin llm for bots, so the last one didnt find this simple bug https://github.com/openchamber/openchamber/pull/1949/commits/d7e075cf528b0bd46908235ec5ed2f65f4cb508c

btriapitsyn:

PR #1952, opened 2026-06-30
fix(sync): don't flash reconnecting during session warmup (#1769)

### LINKED ISSUE #1769: fix(sync): SSE heartbeat timeout causes spurious reconnecting status

bashrusakh:
## Summary

Fix the spurious "reconnecting" indicator flicker described in #1769. The connection indicator briefly flashed `reconnecting` during session warmup and after a reload, when a transient HTTP health check failed before the SSE/WS event pipeline had reported its first connect in the new lifecycle.

The repro and analysis by `openchamber-bot` on `reproduce/issue-1769` confirmed the root cause is **not** the SDK's `:heartbeat` comment-frame path (the SDK already fires `onSseEvent` for every SSE chunk including comment-only frames, and the pipeline resets the heartbeat watchdog correctly). The real triggers are:

- **#2 — HTTP probe race.** `probeConnection` used to flip `connectionPhase` to `reconnecting` purely on the result of an independent HTTP health check, without consulting the pipeline. After a reload, `hasEverConnected` is already true (store hydration), so a transient HTTP failure during `createSession` bootstrap surfaced as a flicker even though the stream was healthy.
- **#3 — Warmup race.** `waitForConnectionOrThrow`'s 2s grace window runs the same probe during the new pipeline's bootstrap, repeating the flicker for any in-flight send. Also applies to `initializeApp` startup.

## Approach

Add lifecycle-scoped markers (`transportConnectedAt`, `transportMountedAt`, `hasEverConnectedSince`) on `useConfigStore`. A shared `isWarmupGuarded()` selector returns `true` when the pipeline has not yet reported a successful connect in the current lifecycle. While guarded, all connection probes return the result without touching `connectionPhase`. The guard is bounded by `WARMUP_GUARD_MAX_MS=30s` so a genuinely dead server eventually surfaces `reconnecting`.

Only the pipeline itself is the authoritative source for a real transport disconnect, and it sets `reconnecting` via `onDisconnect`.

## Changes

- `packages/ui/src/stores/useConfigStore.ts` — new fields `transportConnectedAt`, `transportMountedAt`, `hasEverConnectedSince`; actions `setTransportConnectedAt`, `setTransportMountedAt`, `setConnected`, `setWarmupGuardedDisconnect`, `isWarmupGuarded`; warmup guards in `probeConnection`, `checkConnection` (retry + exhaust), and `initializeApp` (pre-check + catch).
- `packages/ui/src/sync/sync-context.tsx` — `onReconnect`/`onTransportSwitch` stamp `transportConnectedAt` and `hasEverConnectedSince` in a single `setState`; `useEffect` body stamps `transportMountedAt` on mount; cleanup resets markers for the next lifecycle.
- `packages/ui/src/stores/useConfigStore.test.ts` — 15 new unit tests covering all guard paths, timeout release, `setConnected` consistency.

## Diff

```
packages/ui/src/stores/useConfigStore.test.ts | 349 +++++++++++++++++++++++++-
packages/ui/src/stores/useConfigStore.ts      | 250 +++++++++++++++---
packages/ui/src/sync/sync-context.tsx         |  48 ++--
3 files changed, 602 insertions(+), 45 deletions(-)
```

## Validation

- `bun test packages/ui/src/stores/useConfigStore.test.ts` — 33 pass / 0 fail.
- `bun run lint:ui` — clean.
- `bun run type-check:ui` — blocked by unrelated baseline module/type errors in `src/apps/MobileApp.tsx`, `src/apps/deepLinkNavigation.ts`, `src/apps/mobileConnections.ts`, `src/apps/useNativePushRegistration.ts`, `src/components/chat/MessageList.tsx`, and `src/hooks/usePushVisibilityBeacon.ts`.

Fixes #1769

## Related

- `origin/reproduce/issue-1769` (`openchamber-bot` reproduction tests and analysis)
- #1751 (the connection-status indicator feature this fix complements)


btriapitsyn:
Resolved the conflict with `main` and preserved the connection-state fix on top of the current sync pipeline:

- kept the bounded warmup guard that prevents transient HTTP probes from falsely showing “reconnecting”;
- wired pipeline mount/connect/teardown timestamps into the current event pipeline without replacing newer routing/resync logic;
- retained authoritative transport disconnect handling after a real connection;
- verified 36 focused config/event-pipeline tests, UI type-check, and lint.

PR #1895, opened 2026-06-28
fix(agents): eliminate UI freeze on save — skip duplicate health poll, compact overlay, save button feedback, scope badge fix

bashrusakh:
Closes #1894

### 1. Skip duplicate health-check poll (main fix)
When the server already confirmed OpenCode is healthy after a config change, the client no longer runs a second independent health poll. This eliminates 5-10s of waiting.

**Files:** `packages/ui/src/stores/useAgentsStore.ts` — `performConfigRefresh` now accepts `skipHealthCheck`; all agent mutation calls pass it.

### 2. Compact ConfigUpdateOverlay
Replaced the full-screen `fixed inset-0 z-[9999]` overlay with a small top-right indicator (spinner + message text). No more UI lock.

**Files:** `packages/ui/src/components/ui/ConfigUpdateOverlay.tsx`

### 3. Save button shows Done + green checkmark
After a successful save, the button shows a green checkmark and "Done" for 2 seconds before returning to "Save Changes".

**Files:** `packages/ui/src/components/sections/agents/AgentsPage.tsx`

### 4. Scope badge shows "Global" / "Project"
The sidebar badge now translates `extAgent.scope` to the same labels used in the scope selector ("Global" for `user`, "Project" for `project`).

**Files:** `packages/ui/src/components/sections/agents/AgentsSidebar.tsx`

### 5. Reload only active directory
Changed `mode: "projects"` to `mode: "active"` in all three agent mutation calls. No more unnecessary config reloads for every project in the list.

**Files:** `packages/ui/src/stores/useAgentsStore.ts`

### 6. Always call `finishConfigUpdate()` in `finally`
Removed the conditional guard that skipped `finishConfigUpdate()` when `requiresReload` was true. Prevents the overlay from staying visible forever on error.

**Files:** `packages/ui/src/stores/useAgentsStore.ts`

Issue #1894, opened 2026-06-28
UI freezes 5-10s when saving agent settings — double health poll, full-screen overlay, scope badge mismatch


bashrusakh:
## Summary

Saving agent settings (e.g. changing model) freezes the UI for 5–10 seconds. The root cause is a **duplicate health-check poll** — both the server and the client independently poll OpenCode readiness after a config change. On top of that, the full-screen `ConfigUpdateOverlay` blocks all interaction during the save.

---

## 1. Duplicate health-check poll (main cause)

When `PATCH /api/config/agents/:name` is called:

1. **Server** runs `refreshOpenCodeAfterConfigChange()` → `restartOpenCode()` → `waitForOpenCodeReady()` (up to 20s, 400ms interval) + `waitForAgentPresence()` (up to 15s, 300ms interval).
2. After the server responds `{ requiresReload: true }`, the **client** starts a second independent health poll via `waitForOpenCodeConnection()` (up to 20s).

Even when the server-side restart takes 1–2s, the client adds another full poll cycle.

**Code:**
- Server: `packages/web/server/lib/opencode/lifecycle.js:747-785` — `refreshOpenCodeAfterConfigChange` → `restartOpenCode` + `waitForOpenCodeReady`
- Server: `packages/web/server/lib/opencode/lifecycle.js:661-717` — `waitForOpenCodeReady` (poll up to 20s, 400ms step)
- Server: `packages/web/server/lib/opencode/lifecycle.js:719-745` — `waitForAgentPresence` (poll up to 15s, 300ms step)
- Client: `packages/ui/src/stores/useAgentsStore.ts:625-708` — `performConfigRefresh` → `waitForOpenCodeConnection` (second poll up to 20s)
- Client: `packages/ui/src/stores/useAgentsStore.ts:566-608` — `waitForOpenCodeConnection` (poll up to 20s)

---

## 2. Full-screen UI lock via ConfigUpdateOverlay

`startConfigUpdate()` increments `pendingCount`, which renders `ConfigUpdateOverlay` — a `fixed inset-0 z-[9999]` semi-transparent overlay covering the entire viewport. The UI stays locked for the entire save duration (3–10s).

**Code:**
- `packages/ui/src/components/ui/ConfigUpdateOverlay.tsx:21-24` — full-screen overlay
- `packages/ui/src/lib/configUpdate.ts:20-28` — `startConfigUpdate`
[...]

Issue #1295, opened 2026-05-17
[Bug] Health check false-positive restarts OpenCode under normal high load, causing data loss


WangChuDi:
## What's wrong?

OpenChamber's health monitoring restarts `opencode serve` during normal high-load operation (LLM streaming + tool registry init), causing complete loss of in-progress work.

**Expected**: OpenChamber should tolerate temporary health check timeouts when OpenCode is busy processing legitimate work.

**Got**: A single 5-second health check timeout triggers immediate process restart, killing all active LLM inference, tool calls, and sub-agent tasks with no recovery.

### Steps to reproduce

1. Start a complex OpenCode session with high LLM streaming throughput (e.g., multi-agent orchestration with parallel sub-agents)
2. Wait for a burst of `message.part.delta` events (hundreds per second)
3. The periodic health check (15s interval) or immediate health check (WS disconnect trigger) times out at 5s
4. OpenChamber restarts OpenCode, all in-progress work is lost

Restart frequency: every 14-18 minutes during sustained high-load sessions.

### Root cause

In `server/lib/opencode/lifecycle.js`:
- `isOpenCodeProcessHealthy()` uses `AbortSignal.timeout(5000)` — too aggressive under load
- `startHealthMonitoring()` and `triggerHealthCheck()` restart on a single failure with no consecutive failure counting
- `shouldSkipRestartForBusySessions()` exists but cannot help when the health endpoint itself is unresponsive

The OpenCode process is fully alive (logs show active event publishing right up to SIGTERM). The Go runtime's goroutine scheduling delays the HTTP health response under heavy event bus load.

### Proposed fix

1. Add consecutive failure counting (require N failures before restart)
2. Log failure count on each check for observability: `[lifecycle] health check failed (3/20)`
3. **Expose user-configurable options** (timeout, retry count, interval) via settings:
```jsonc
{
  "healthCheck": {
    "timeoutMs": 5000,
    "consecutiveFailures": 3,
    "intervalMs": 15000
  }
}
```
4. Consider OS-level process alive check (`process.kill(pid, 0)`) as a pre-gate before HTTP check
5. (OpenCode-side) Dedicated health goroutine not blocked by event bus saturation

### Impact

- **Data loss**: Every false-positive restart kills in-progress LLM inference, tool executions, sub-agent tasks
- **Resource waste**: Tokens consumed by interrupted LLM calls are wasted
- **Unusable for complex tasks**: 14-18 min restart cycle makes multi-step agent work impossible

[...]

Issue #1294, opened 2026-05-17
Health check restarts OpenCode during normal high-load operation, causing data loss


WangChuDi:
### Summary

OpenChamber's health monitoring kills and restarts the `opencode serve` process when it is under heavy but normal load (LLM streaming + tool registry initialization). The process is fully alive and working -- it simply cannot respond to the `/global/health` HTTP request within the 5-second timeout because the Go runtime is saturated with event publishing. This causes complete loss of in-progress work (LLM inference, tool calls, sub-agent tasks) with no recovery mechanism.

### Environment

- OpenChamber version: latest (as of 2026-05-17)
- OpenCode version: 1.15.3
- OS: Ubuntu (KVM guest, 12-16GB RAM, 4GB swap)
- Node.js: v24.14.1

### Steps to Reproduce

1. Start a complex OpenCode session that triggers high LLM streaming throughput (e.g., multi-agent orchestration with parallel sub-agents)
2. Wait for a burst of `message.part.delta` events (typically hundreds per second)
3. Observe that the periodic health check (every 15s) or an immediate health check (triggered by WS disconnect) times out
4. OpenChamber restarts OpenCode, killing all in-progress work

### Observed Behavior

OpenChamber logs:
```
OpenCode process not running, restarting...
```

OpenCode logs immediately before restart show the process is fully active:
```
INFO  service=bus type=message.part.delta publishing
INFO  service=bus type=message.part.delta publishing
INFO  service=bus type=message.part.delta publishing
... (hundreds of these per second)
INFO  service=mcp key=postgres-mcp mcp stderr: Received exit signal SIGTERM
```

The process was never dead or stuck. It was busy processing legitimate work. The SIGTERM to MCP servers is a *consequence* of the restart, not the cause.

Restart frequency observed: every 14-18 minutes during sustained high-load sessions.

### Expected Behavior

OpenChamber should not restart OpenCode when it is actively processing work. A single 5-second health check timeout during a known-heavy workload (LLM streaming) should not be treated as process death.

### Root Cause Analysis

Location: `server/lib/opencode/lifecycle.js`

1. **`isOpenCodeProcessHealthy()`** (line ~354): Uses `AbortSignal.timeout(5000)` -- a 5-second timeout is too aggressive for a single-threaded-ish Go HTTP server under heavy event bus load.

2. **`startHealthMonitoring()`** (line ~857): A single failed health check immediately triggers `restartOpenCode()`. No consecutive failure counting, no backoff.

[...]

PR #1962, opened 2026-07-01
fix(lifecycle): defer OpenCode restart on config save when sessions are busy

### LINKED ISSUE #1947: Saving agent settings (or any config mutation) restarts OpenCode and interrupts active LLM generations
## Summary

Saving any configuration in OpenChamber (agent/command/MCP/plugin/skill/settings save, or manual "Reload configuration") triggers an unconditional restart of the managed `opencode serve` process. If other agents are actively running LLM generations at that moment, the restart sends SIGTERM (2.5 s grace) → SIGKILL and **interrupts their in-progress work**.

The codebase already contains a guard to prevent exactly this class of data loss — `shouldSkipRestartForBusySessions()` in `packages/web/server/lib/opencode/lifecycle.js` — but it is wired into the **health-check recovery path only** (`runHealthCheckCycle`, line 944). The **config-change path** (`refreshOpenCodeAfterConfigChange` → `restartOpenCode`, lines 747 / 562) does **not** call this guard, so any config save bypasses the protection that was added for the health-check path.

### Environment

- OpenChamber version: latest `main` (fb015d78)
- All runtimes (web / desktop / VS Code) share the same lifecycle code path
- Reproducible with a managed (auto-started) OpenCode process

## Steps to Reproduce

1. Open two browser tabs / two sessions in the same OpenChamber instance.
2. In **Tab A**, start a long-running agent task (e.g. a multi-step coding request that triggers a long LLM call).
3. While Tab A's agent is mid-generation, in **Tab B** open **Settings → Agents**, edit any field (e.g. the prompt), and click **Save**.
4. Observe: Tab A's active generation is interrupted within ~2.5 seconds. OpenChamber logs:

```
[Server] Updating agent: <name>
Restarting OpenCode process...
Stopping existing OpenCode process...
```

5. The OpenCode process is killed and re-spawned; the in-flight LLM call is lost.

## Observed Behavior

- `PATCH /api/config/agents/:name` → `refreshOpenCodeAfterConfigChange('agent update')` → `restartOpenCode()` → `terminateChildProcess()` (SIGTERM, 2.5 s grace, then SIGKILL) → `startOpenCode()`.
- No check for active/busy sessions is performed before the kill.
- The same path is triggered by: command save, MCP save, plugin save, skill save, settings save, provider disconnect, AGENTS.md update, and manual `POST /api/config/reload`.

## Expected Behavior

- When agents are actively busy, the config change should be **persisted to disk immediately** (so the user's edit is not lost), but the OpenCode restart should be **deferred** until active sessions drain to idle.
- The UI should surface a non-blocking toast: *"Saved. Will apply after N active task(s) finish."* with an optional **Apply now** escape hatch.
- An explicit **force-apply** action (e.g. `POST /api/config/reload?force=true`) should remain available for users who want the change immediately.

## Root Cause

In `packages/web/server/lib/opencode/lifecycle.js`:

- `shouldSkipRestartForBusySessions()` (line 901) exists and returns `true` when `getActiveSessionCount() > 0` (with a 2-minute staleness fallback that forces restart for stuck "busy" state).
- It is called **only** from `runHealthCheckCycle()` (line 944) — the periodic / immediate health-probe recovery path.
- It is **never called** from `refreshOpenCodeAfterConfigChange()` (line 747) or `restartOpenCode()` (line 562), which is the path used by every config-save route.

This is the same root cause class as the (already fixed) issues #1294 / #1295 ("health check restarts OpenCode during normal high-load operation, causing data loss"), which were resolved in commit `065252a4` by adding consecutive-failure counting and `shouldSkipRestartForBusySessions` to the health path. The config-save path was not covered by that fix.

## Scope of Affected Paths

Every route that calls `refreshOpenCodeAfterConfigChange`:

- `PATCH/POST/DELETE /api/config/agents/:name` (agent save) — `config-entity-routes.js:131,159,184`
- `PATCH/POST/DELETE /api/config/commands/:name` — `config-entity-routes.js:326,356,382`
- `POST/PATCH/DELETE /api/config/mcp/:name` — `config-entity-routes.js:57`
- Plugin mutations — `plugin-routes.js:61`
- Skill mutations — `skill-routes.js:450,501,591,618,710`
- `PUT /api/config/settings`, provider disconnect, AGENTS.md update — `routes.js:160,423,527`
- `POST /api/config/reload` (manual reload) — `core-routes.js:687`

## Proposed Fix

Lift the existing busy-session guard to the lifecycle chokepoint so config-change restarts are deferred while sessions are busy:

1. **`packages/web/server/lib/opencode/lifecycle.js`** — in `refreshOpenCodeAfterConfigChange`, call `shouldSkipRestartForBusySessions` before `restartOpenCode`. If busy: persist a `pendingConfigRefresh` record on `state`, skip the restart, and return `{ reloaded: false, deferred: true, pendingActiveSessions: N }`. Add a drain-listener that runs the queued restart when `getActiveSessionCount()` returns to 0. Keep the existing 2-minute staleness fallback (`STALE_BUSY_GRACE_MS`) to force-restart on stuck "busy" state. Add a `forceRestart(reason)` escape hatch that bypasses the guard.
2. **`packages/web/server/lib/opencode/config-entity-routes.js`** — extend `buildConfigMutationResponse` with a third `deferred` state.
3. **`packages/web/server/lib/opencode/core-routes.js`** — `POST /api/config/reload?force=true` → `forceRestart`.
4. **UI stores** (`useAgentsStore`, `useCommandsStore`, `useMcpConfigStore`, `useSkillsStore`, `usePluginsStore`) — on `payload.deferred === true`, do **not** call `refreshAfterOpenCodeRestart`; show a non-blocking toast instead.
5. **`packages/ui/src/lib/api/types.ts`** — extend `ConfigMutationResponse` with `deferred?: boolean`, `pendingActiveSessions?: number`, `queueReason?: string`.

This reuses the existing guard and `getActiveSessionCount` (already wired into the lifecycle runtime via `packages/web/server/index.js:412,923`). No new abstractions are introduced.

## Notes

- Historical (already-flushed) sessions **do survive** the restart — OpenCode persists session data under `~/.local/share/opencode`, and the UI re-bootstraps them from the server after `server.connected` (`packages/ui/src/sync/sync-context.tsx:1318`). The regression is specifically the **interruption of in-flight LLM work**, not the loss of the session record.
- The UI does **not** do a full page reload on config save — it performs a soft `performConfigRefresh` (`packages/ui/src/stores/useAgentsStore.ts:634-717`) that waits for OpenCode health and re-reads config via the SDK. So the fix does not need to touch reconnect logic.
- Related: PR #1895 addresses UI freeze on agent save (different symptom, same save flow) — orthogonal to this issue.

## Related

- Closes-precursor: #1294, #1295 (same root cause, health-check path; fixed in `065252a4`)
- Related: #1680 (plugin add freezes), #1894 (UI freeze on save, PR #1895)

bashrusakh:
## Summary

Saving any configuration in OpenChamber (agent/command/MCP/plugin/skill/settings save, or manual "Reload configuration") historically triggered an unconditional restart of the managed `opencode serve` process. If other sessions were actively streaming LLM generations at that moment, the restart sent SIGTERM (2.5 s grace) → SIGKILL and **interrupted their in-progress work**.

The `shouldSkipRestartForBusySessions` guard already existed in the codebase for the health-check recovery path (added in `065252a4` for #1294/#1295), but was never called from the config-change path (`refreshOpenCodeAfterConfigChange` → `restartOpenCode`). This PR reuses the same predicate so config changes are persisted to disk immediately but the restart is deferred until active sessions drain to idle.

Closes #1947

### Core fix — `packages/web/server/lib/opencode/lifecycle.js`

- In `refreshOpenCodeAfterConfigChange`, check `getActiveSessionCount()` before calling `restartOpenCode()`. If sessions are busy and the server is managed (not external), skip the restart and return `{ reloaded: false, deferred: true, pendingActiveSessions: N }`.
- Add a drain-listener (2-second polling interval) that applies the queued restart once active sessions reach idle.
- Add a 2-minute staleness fallback (shared with the health-check path's `STALE_BUSY_GRACE_MS`) that forces the restart if sessions stay "busy" too long, so a stuck session cannot block a config change forever.
- Add `forceRestart(reason)` — bypasses the guard and restarts immediately; also clears any pending deferred restart. Used by the explicit "Apply now" reload endpoint.
- Add `hasPendingConfigRefresh()` and `clearPendingConfigRefresh()` helpers.
- `force=true` option on `refreshOpenCodeAfterConfigChange` bypasses the guard (used by the force-reload endpoint).

### Route response contract — `config-entity-routes.js`

- Extend `buildConfigMutationResponse` with a third `deferred` state. When the restart is deferred, the response carries `{ success: true, requiresReload: false, deferred: true, pendingActiveSessions: N, message: "..." }`.
- Agent create/update/delete mutations pass `deferredMessage` to the response builder.

### New endpoints — `core-routes.js`

- `POST /api/config/reload?force=true` — bypasses the busy-session guard and applies immediately.
- `GET /api/config/pending-restart` — reports whether a deferred config restart is queued.
- The normal `POST /api/config/reload` (without `force`) now handles the `deferred` result shape.

### Wiring — `feature-routes-runtime.js`, `index.js`

- Pass `forceRestart` and `hasPendingConfigRefresh` through route dependencies.

### UI — `useAgentsStore.ts`, `AgentsPage.tsx`

- `AgentMutationResult` extended with `deferred?: boolean` and `pendingActiveSessions?: number`.
- In create/update/delete agent mutations: when `payload.deferred === true`, do **not** call `refreshAfterOpenCodeRestart` (which would interrupt the active sessions we are protecting). Instead, return `{ ok: true, deferred: true, pendingActiveSessions }`.
- `AgentsPage.tsx`: show an info toast for deferred saves: *"Saved. OpenCode will reload when N active task(s) finish."*

### i18n — 9 locale files

- Added `savedDeferredSingle` and `savedDeferredPlural` keys to all 9 supported locales (en, es, fr, ja, ko, pl, pt-BR, uk, zh-CN, zh-TW).

### Tests — `lifecycle.test.js`

- 5 new test cases:
  - defers restart when sessions are busy, returns `{ deferred: true, pendingActiveSessions }`
  - restarts immediately when no sessions are busy
  - `force=true` bypasses the guard
  - `forceRestart` clears a pending deferred restart
  - does not defer for an external OpenCode server

### Documentation — `DOCUMENTATION.md`

- Updated lifecycle.js public exports with new methods and deferred semantics.
- Updated config-entity-routes description with three-state response.
- Updated core-routes with new endpoints.

## Verification

```bash
# Type-check
bun run type-check:ui   # ✅ clean
bun run type-check:web  # ✅ clean

# Lint
bun run lint            # ✅ clean (all 4 workspaces)

# Tests
bunx vitest run server/lib/opencode/lifecycle.test.js  # ✅ 11/11 passed
bunx vitest run server/lib/opencode/plugin-routes.test.js server/lib/opencode/core-routes.test.js  # ✅ 40/40 passed
bunx vitest run  # ✅ 67 files, 560 passed, 2 skipped (full web suite)
```

## Design decisions

- **Deferred by default, not opt-in.** The guard is in the lifecycle chokepoint (`refreshOpenCodeAfterConfigChange`), so it applies uniformly to all config-save paths (agent/command/MCP/plugin/skill/settings) without per-route changes.
- **External OpenCode servers are never deferred.** OpenChamber cannot restart them anyway; the existing honest-reporting contract (`{ reloaded: false, external: true }`) is preserved.
- **Drain-listener uses polling, not event subscription.** The session-runtime does not expose a subscription API. A 2-second polling interval (with `unref()` so it doesn't keep the event loop alive) is the simplest correct approach and matches the existing health-monitor pattern.
- **2-minute staleness fallback** mirrors the health-check path's `STALE_BUSY_GRACE_MS`. A stuck "busy" session cannot block a config change indefinitely.
- **No full page reload.** The UI already does a soft `performConfigRefresh` (not `window.location.reload()`), so the deferred path simply skips that refresh and shows a toast instead.

## Related

- Closes-precursor: #1294, #1295 (same root cause, health-check path; fixed in `065252a4`)
- Related: #1894 (UI freeze on save, PR #1895 — orthogonal, UI-only)

### DIFF
--- a/packages/ui/src/components/sections/agents/AgentsPage.tsx
+++ b/packages/ui/src/components/sections/agents/AgentsPage.tsx
@@ -648,6 +648,13 @@ export const AgentsPage: React.FC = () => {
       if (result.ok) {
         if (result.requiresManualRestart) {
           toast.warning(t('settings.agents.page.toast.savedManualRestart'));
+        } else if (result.deferred) {
+          const count = result.pendingActiveSessions ?? 0;
+          toast.info(
+            count === 1
+              ? t('settings.agents.page.toast.savedDeferredSingle')
+              : t('settings.agents.page.toast.savedDeferredPlural', { count })
+          );
         } else {
           toast.success(isNewAgent ? t('settings.agents.page.toast.created') : t('settings.agents.page.toast.updated'));
         }
--- a/packages/ui/src/lib/i18n/messages/en.settings.ts
+++ b/packages/ui/src/lib/i18n/messages/en.settings.ts
@@ -412,6 +412,8 @@ export const settingsDict = {
   'settings.agents.page.toast.updateFailed': 'Failed to update agent',
   'settings.agents.page.toast.saveUnexpectedError': 'An error occurred while saving',
   'settings.agents.page.toast.savedManualRestart': 'Saved to disk. Restart your connected OpenCode server to apply the changes.',
+  'settings.agents.page.toast.savedDeferredSingle': 'Saved. OpenCode will reload when the active task finishes.',
+  'settings.agents.page.toast.savedDeferredPlural': 'Saved. OpenCode will reload when {count} active tasks finish.',
   'settings.agents.page.empty.title': 'Select an agent from the sidebar',
   'settings.agents.page.empty.description': 'or create a new one',
   'settings.agents.page.title.new': 'New Agent',
--- a/packages/ui/src/lib/i18n/messages/es.settings.ts
+++ b/packages/ui/src/lib/i18n/messages/es.settings.ts
@@ -379,6 +379,8 @@ export const settingsDict = {
   "settings.agents.page.toast.updateFailed": "No se pudo actualizar el agente",
   "settings.agents.page.toast.saveUnexpectedError": "Ocurrió un error al guardar",
   "settings.agents.page.toast.savedManualRestart": "Guardado en disco. Reinicia tu servidor OpenCode conectado para aplicar los cambios.",
+  "settings.agents.page.toast.savedDeferredSingle": "Guardado. OpenCode se recargará cuando termine la tarea activa.",
+  "settings.agents.page.toast.savedDeferredPlural": "Guardado. OpenCode se recargará cuando terminen {count} tareas activas.",
   "settings.agents.page.empty.title": "Selecciona un agente desde la barra lateral",
   "settings.agents.page.empty.description": "o crea uno nuevo",
   "settings.agents.page.title.new": "Nuevo agente",
--- a/packages/ui/src/lib/i18n/messages/fr.settings.ts
+++ b/packages/ui/src/lib/i18n/messages/fr.settings.ts
@@ -362,6 +362,8 @@ export const settingsDict = {
   'settings.agents.page.toast.updateFailed': 'Échec de la mise à jour de l\'agent',
   'settings.agents.page.toast.saveUnexpectedError': 'Une erreur s\'est produite lors de l\'enregistrement',
   'settings.agents.page.toast.savedManualRestart': 'Enregistré sur le disque. Redémarrez votre serveur OpenCode connecté pour appliquer les modifications.',
+  'settings.agents.page.toast.savedDeferredSingle': 'Enregistré. OpenCode rechargera lorsque la tâche active sera terminée.',
+  'settings.agents.page.toast.savedDeferredPlural': 'Enregistré. OpenCode rechargera lorsque {count} tâches actives seront terminées.',
   'settings.agents.page.empty.title': 'Sélectionnez un agent dans la barre latérale',
   'settings.agents.page.empty.description': 'ou créez-en un nouveau',
   'settings.agents.page.title.new': 'Nouvel agent',
--- a/packages/ui/src/lib/i18n/messages/ja.settings.ts
+++ b/packages/ui/src/lib/i18n/messages/ja.settings.ts
@@ -412,6 +412,8 @@ export const settingsDict = {
   'settings.agents.page.toast.updateFailed': 'Agent の更新に失敗しました',
   'settings.agents.page.toast.saveUnexpectedError': '保存中にエラーが発生しました',
   'settings.agents.page.toast.savedManualRestart': 'ディスクに保存しました。変更を適用するには接続中の OpenCode サーバーを再起動してください。',
+  'settings.agents.page.toast.savedDeferredSingle': '保存しました。アクティブなタスクが完了すると OpenCode が再読み込みされます。',
+  'settings.agents.page.toast.savedDeferredPlural': '保存しました。{count} 件のアクティブなタスクが完了すると OpenCode が再読み込みされます。',
   'settings.agents.page.empty.title': 'サイドバーから Agent を選択してください',
   'settings.agents.page.empty.description': 'または新しいものを作成',
   'settings.agents.page.title.new': '新しい Agent',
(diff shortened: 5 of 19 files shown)


btriapitsyn:

PR #1964, opened 2026-07-01
fix(chat): strip ANSI escape codes from tool output

### LINKED ISSUE #1965: Chat leaks ANSI SGR/OSC escape codes from shell-tool output into rendered text
## Summary

When a shell tool (typically `bash`) returns output containing ANSI SGR color codes, the chat renders the raw control bytes literally instead of interpreting or stripping them. The result is visible noise around the actual command output — `[2m`, `[48;2;38;38;38m`, `[0m`, and friends.

This affects any tool that emits color: `ls --color=always`, `grep --color`, `npm` scripts with colored output, `cargo`, `pytest --color=yes`, **OCR review CLI** (`ocr review --audience agent`), and similar.

## Actual behavior

Chat block contains stray bracket sequences mixed into the text. Example from `ocr review --audience agent` on a real diff:

```
[2m─── packages/ui/src/components/sections/commands/CommandsPage.tsx:164-164 ───[0m
Using `subtask || undefined` converts `false` to `undefined`, which means the saved config will only
include `subtask` when it's `true`. This is intentional for omitting falsy values, but it differs

[2m[48;2;38;38;38m [0m[48;2;38;38;38m         subtask: subtask || undefined,[0m[0m
```

The `\x1b` (ESC) byte itself is invisible in the browser; the visible leak is the trailing `[…m` text. The same pattern shows up in:

- The inline expanded `bash` tool body (highlighted code block)
- The full-output popup dialog (raw `<pre>` for `list`/`grep`/`glob`/JSON fallback)
- The preview URL extraction path in `MessageBody` (had a partial SGR-only strip inline)

## Expected behavior

The chat should render the command's actual text, without terminal control sequences. The visible output for the snippet above should be:

```
─── packages/ui/src/components/sections/commands/CommandsPage.tsx:164-164 ───
Using `subtask || undefined` converts `false` to `undefined`, which means the saved config will only
include `subtask` when it's `true`. This is intentional for omitting falsy values, but it differs

         subtask: subtask || undefined,
```

## Reproduction

1. Start OpenChamber (web or desktop) connected to a session with `bash` tool enabled.
2. In a chat message, ask the agent to run a color-emitting command. Reliable repros:
   - `ls --color=always /usr/bin | head -20` (or any directory)
   - `printf '\x1b[31mred\x1b[0m \x1b[1;32mbold green\x1b[0m\n'` (raw byte injection)
   - `grep --color=always -r "TODO" packages/ | head -20`
3. Expand the `bash` tool output (or open the full-output dialog).
4. Observe the literal `[31m`, `[1;32m`, `[0m` text mixed into the rendered code.

The `ocr review --audience agent` skill produces this end-to-end through the bash tool.

## Root cause

OpenChamber's chat renders raw `state.output` strings from tool parts through two surfaces — `WorkerHighlightedCode` (Shiki) and a few `<pre>` fallbacks in `ToolOutputDialog`. Neither interprets ANSI SGR/OSC/DCS. The only existing ANSI-stripping logic was a narrow SGR-only inline regex in `MessageBody` (used only before scanning for loopback preview URLs). Every other call site passed the raw text through, escape bytes and all.

The control sequence grammar is well-defined: SGR lives in CSI sequences (`ESC [ … m`), window titles and palette swaps in OSC (`ESC ] … BEL/ST`), sixel and other payloads in DCS/SOS/PM/APC. Truncated output (large bash results past OpenCode's stdout cap) can also leave an unterminated opener with its payload visible.

## Fix

Open PR: #1964

Introduces a shared `stripAnsi(text: string): string` helper in `packages/ui/src/lib/text/stripAnsi.ts` that strips CSI, OSC, DCS/SOS/PM/APC, two-byte ESC sequences, and unterminated openers from truncated output, then wires it into the three places that render raw tool output: `ToolPart.ToolScrollableTextOutput`, `ToolOutputDialog`, and `MessageBody` URL extraction.

- 12 unit tests cover SGR, OSC (BEL+ST), DCS, two-byte ESC, the real OCR snippet, idempotency, and the falsy-input guard; 5 more cover broadened CSI param class, private-mode bytes, and the unterminated-opener fallbacks.
- No new dependencies. No backend / Electron / VS Code / sync-layer changes.
- Behavioral contract preserved: visible text is unchanged; only control bytes are removed.

## Environment

- OpenChamber (any runtime — web, Electron, VS Code)
- Any session with the `bash` tool enabled and an output that contains ANSI codes
- Reproduced on Linux + Electron; the chat renderer is the same across runtimes so the bug is runtime-agnostic

bashrusakh:
## Summary

Strip ANSI SGR/OSC/DCS control codes from shell-tool output so the chat renders clean text instead of leaking literal `\x1b[…m` bytes.

### Before

Running a colored CLI through the bash tool (e.g. `ocr review --audience agent`) produced chat blocks with visible noise like:

```
[2m─── packages/ui/src/components/sections/commands/CommandsPage.tsx:164-164 ───[0m
Using `subtask || undefined` converts `false` to `undefined`...
[2m[48;2;38;38;38m [0m[48;2;38;38;38m         subtask: subtask || undefined,[0m[0m
```

### After

The same output renders as plain text — the divider line, the explanation, and the code snippet — with all SGR colors and 24-bit background codes removed.

## What changed

A single shared `stripAnsi(text: string): string` helper in `packages/ui/src/lib/text/stripAnsi.ts` covers SGR (CSI), OSC, DCS/SOS/PM/APC, two-byte ESC sequences, and unterminated openers from truncated output. It is wired into the three places that render raw tool output in the chat:

- `ToolPart.ToolScrollableTextOutput` — inline expanded body, JSON-tree fallback, copy-to-clipboard, language detection (all see the same cleaned string)
- `ToolOutputDialog` — `todowrite`, `list`, `grep`, `glob`, `task`, `web-search`, JSON, and generic fallbacks
- `MessageBody` — replaces the existing inline SGR-only strip used before scanning for loopback preview URLs

The fix is silent: visible text content is preserved byte-for-byte; only the control sequences are removed.

## Commits

```
fix(chat): strip ANSI escape codes from tool output
fix(chat): address code-review findings for ANSI stripping
docs(chat): drop stale reference to removed getToolOutputLanguage
```

The three commits form a clean history: the original fix, the two follow-ups from code review (broader CSI param class `[0-9:;<=>?]`, falsy guard, unterminated-opener fallbacks, and an inlined `detectLanguageFromOutput` to avoid double-applying `formatEditOutput`), and a comment cleanup.

## Validation

- `bun run type-check:ui` — passes
- `bun run lint:ui` — passes
- `bun test packages/ui/src/lib/text/stripAnsi.test.ts` — 17/17 pass (12 original + 5 added during review)
- Full UI test suite — 595 pass / 79 pre-existing failures unchanged (zustand-store, event-pipeline, worktree, `ReasoningPart` infrastructure issues independent of this change)
- Rebased onto `origin/main` before push (no conflicts; none of the 8 upstream commits since branch point touched the changed files)

## Risk

- **Low.** The helper is a pure text transformation. All call sites either already had a defensive guard (`output && ...`) or are now fed a string guaranteed to be `string` (the helper returns `''` for falsy input). Behavioral contract preserved: user sees the same command output, minus terminal control codes.
- No new dependencies. No backend / Electron / VS Code / sync-layer changes. No public API changes. No secrets, no config rewrites.

## Manual verification

To reproduce the original bug → fix:

1. Start the dev instance (`bun run dev:web`).
2. In a session, run a bash tool that emits color: e.g. `ls --color=always /usr/bin | head` or `npm --version` after piping through `printf '\x1b[31mfoo\x1b[0m'`.
3. Before this change: the rendered code block contains literal `[31m`/`[0m` tokens.
4. After this change: the code block is clean.

Closes #1965


btriapitsyn:
Resolved the conflicts with `main` while preserving both output features:

- kept ANSI stripping at the shared rendering boundaries so inline output, dialogs, JSON parsing, syntax detection, and copied text all use clean content;
- retained `main`’s Mermaid dialog loader, JSON summary mode, and typed diff patch entries;
- verified 23 focused rendering/ANSI tests, UI type-check, and lint.

PR #1665, opened 2026-06-15
feat(plugins): plugin status panel with health checks and auto-update

herjarsa:
## Summary

Adds a dedicated plugin status panel with health checks, version display, and auto-update:

### Plugin Status Page
- New settings page showing all loaded plugins with ok/warning/error health
- Displays plugin version in the status page
- Auto-update plugins on server boot
- Send-to-chat from plugin errors (creates session in ~/.config/opencode)

### Server-Side
- New plugin routes module (`plugin-routes.js`)
- Plugin management refactor in `plugins.js`
- Plugin version tracking and update in `startup-pipeline-runtime.js`

### Services Menu Integration
- Plugin Status entry in desktop services menu
- Native close button on mobile surface header
- Close service panel after send-to-chat

### Settings Cleanup
- Remove duplicate Plugin Status entry from Settings sidebar (it lives in Services menu)
- Remove plugin entries from Settings search
- Treat non-exact semver versions as ok for health check

### i18n
- New strings added across all 9 locales (en, es, fr, ko, pl, pt-BR, uk, zh-CN, zh-TW)

## Test Plan
- [x] `bun run type-check` passes
- [x] `bun run lint` passes (0 errors)
- [x] Manual QA: plugin status renders with correct health indicators
- [x] Manual QA: auto-update runs on server boot
- [x] Manual QA: send-to-chat creates session in correct directory

PR #1664, opened 2026-06-15
feat(subagent): subagent/team observability with sync-layer refactor

herjarsa:
## Summary

Adds comprehensive subagent/team observability features:

### Subagent Dashboard (G5/G7)
- Right sidebar panel showing active subagents with depth indicator
- Live progress display with tokens/cost per subagent task row
- Streaming subagent output visible in parent chat

### Notification System
- Debounced notification batcher (1.2s coalesce by key)
- Error toast that navigates to parent session and highlights failed task row

### Sync Layer Refactor
- God-module split: event-pipeline, session-ui-store, sync-context
- Extracted chat input components (ComposerActionButtons, FocusModeButton, etc.)
- Extracted desktop header components (DesktopServicesMenu, HeaderIconActionButton)
- Extracted files view subcomponents (FileRow, FilesList, FilesToolbar, etc.)

### Bug Fixes
- Fix cascade rollback from restoring deleted session descendants
- Fix tab-complete from swapping selected agent
- Fix session bar premature disable
- Fix worktree list refresh on dropdown open and window focus
- Fix orphaned OpenCode processes on server exit
- Prevent worktree modal from disappearing after opening
- Fix invoke skills from slash command menu
- Stabilize question custom textarea resizing

### Infrastructure
- Click-to-expand for file preview toolbar
- Resync parent session tool parts on subagent completion
- Signal handlers moved before return, redundant import fixed

PR #1984, opened 2026-07-01
fix(workspace): bypass outside-workspace check for fs reads

herjarsa:
## Problem

Writes in OpenChamber returned no response for projects outside the OpenChamber service cwd. Browser SDK sent /api/fs/read and /api/session/.../prompt requests with an empty x-opencode-directory header, so the server-side workspace check rejected them with **400 Path is outside of active workspace**. Session.create, session.prompt, and per-project config read all broke.

## Fix

Two layers, both needed.

### 1. Browser auto-injects x-opencode-directory (runtime-fetch.ts + client.ts)

The runtime fetch bridge now tracks the active project directory in module-level state (activeProjectDirectory). OpencodeService.setDirectory() calls setActiveProjectDirectory(normalized) to keep it in sync. mergeHeaders() now auto-injects x-opencode-directory for any /api/fs/* request that does not already have it. Server workspace check then passes for the active project.

### 2. Service WorkingDirectory set to HOME (openchamber.service)

The systemd unit WorkingDirectory= was changed from /home/herjarsa/GITHUB/openchamber to /home/herjarsa (user home) so any project under HOME is in the workspace. ExecStart path is also absolute.

## Verification

| Test | Result |
|------|--------|
| /api/fs/read WITHOUT header | HTTP 400 (security preserved) |
| /api/fs/read WITH x-opencode-directory header | HTTP 200 (or 404 if file missing) |
| Session create | HTTP 200 |
| Prompt + SSE chain | 4,558 bytes SSE, 1 message.part event |
| OpenCode 1.17.12 | healthy |
| OpenChamber service | active |

## Files changed

- packages/ui/src/lib/runtime-fetch.ts (+20, -2): added setActiveProjectDirectory() and auto-inject logic in mergeHeaders().
- packages/ui/src/lib/opencode/client.ts (+1, -1): call setActiveProjectDirectory(normalized) in setDirectory() and update the import.

The openchamber.service change is a system-level edit (outside this repo) and is not part of the source diff. The user should apply that change separately to their systemd unit.

## Notes

- Defensive: workspace check still works for any request that does not carry x-opencode-directory, so security is preserved.
- activeProjectDirectory state is a single module-level variable. Updated whenever the openchamber client changes directory.
- Browser-side only. Server-side: unchanged. The server still enforces the workspace check, the browser just sends the right header.


herjarsa:
Closing this mega-PR in favor of focused individual PRs:
- #2076 fix(chat): SDK v2 body-omission
- #1931 fix(mcp): MCP toggle config state
- #1888 fix(mcp): MCP status map noise filter
- #1887 fix(chat): scroll cascade + GPU ghosting
- #1667 perf(pwa): vendor chunk split
- #1666 feat(sidebar): subagent session filtering
- #1665 feat(plugins): plugin status panel
- #1664 feat(subagent): subagent/team observability
- #1616 fix: parent session resync

Each covers a specific concern. This PR was a catch-all branch and cannot be merged cleanly.

PR #1986, opened 2026-07-01
Feature/docker build from ci

### LINKED ISSUE #411: [Feature Request] Docker image and publication to ghrc.io

casinesque:
## Summary

Adds a `docker_build_and_push` job to the release workflow that builds the existing `Dockerfile` and publishes it to `ghcr.io` on every release (tag push or manual `workflow_dispatch`).

Closes #411

## What's included

- New `docker_build_and_push` job in `.github/workflows/release.yml`,   gated the same way as the other release jobs (build always runs;   actual `push` is skipped when `dry_run: true` on manual dispatch).
- Fix for a pre-existing bug in `Dockerfile`: the `deps` stage was   missing `COPY packages/mobile/package.json ./packages/mobile/`.   Since `bun.lock` includes `@openchamber/mobile` as a workspace,
  `bun install --frozen-lockfile` failed on every build with   `error: lockfile had changes, but lockfile is frozen`. Found this  while testing the new job locally, the Dockerfile itself couldn't   build at all before this fix, unrelated to the CI job.

## Design decisions

**No `v` prefix on the Docker tag.** GitHub releases use `v1.2.3` for git tags, but the Docker image is tagged `1.2.3` (no `v`). This follows the de-facto convention for container images and most popular images on Docker Hub/GHCR.
The `v` prefix is a Git tag convention, and keeping the two namespaces separate avoids confusion for users who are used to
`docker pull image:1.2.3` rather than `image:v1.2.3`.

**Double tag (`:<version>` + `:latest`).** Every push publishes both:

- `ghcr.io/openchamber/openchamber:<version>`--> pinned, reproducible,   what you'd use in production/self-hosted setups.
- `ghcr.io/openchamber/openchamber:latest`--> always points at the   most recent release, for people who just want `docker pull` to work   without tracking version numbers.

## Testing

- `docker build -f Dockerfile .` locally builds and runs successfully   (after fixing the `packages/mobile` COPY, see above).
- I have **not** been able to test the full release pipeline end-to-end, this because, i don't yet have a complete picture of how all the jobs interact, i'm not familiar with the tool also and forks don't  have the required secrets (NPM_TOKEN, certs, etc.) to   exercise the other jobs anyway. I'd really appreciate a maintainer  running this on a branch with real secrets before merging, or   pointing me at what to check.

## Open to discuss

I added `docker_build_and_push` to `finalize-release`'s `needs`, so the release only gets published after the image is on ghcr.io too (consistent with how npm/macOS/Windows artifacts are treated). Happy to change this if a faster, independent Docker publish is preferred instead.


btriapitsyn:
Thanks for putting this together. We plan to handle official Docker image publishing ourselves soon. The release integration still needs changes and testing on our side, so I'm closing this PR and will take it from here.

PR #1991, opened 2026-07-02
fix(worktree): attach existing worktrees and fix discovery

### LINKED ISSUE #1989: [Bug] Worktrees created by OpenCode don't appear in sidebar
## Summary

Worktrees created by OpenCode (via `git worktree add`) are not discovered and displayed in the OpenChamber sidebar. The sidebar only shows worktrees that were already known at first mount — any worktree created externally or attached through the dialog after initial discovery never appears.

## Root causes

Three bugs compound into the visible symptom:

### 1. `worktrees.length === 0` skips projects, overwrites store with empty data

In `SessionSidebar.tsx:460`, `ElectronMiniChatApp.tsx:188`, and `MobileApp.tsx:2058`:

```typescript
if (cancelled || worktrees.length === 0) continue;
worktreesByProject.set(projectPath, worktrees);
```

Projects with zero worktrees are skipped. The resulting empty `worktreesByProject` Map **overwrites** the store (including persisted data from localStorage), wiping any previously-discovered worktrees.

### 2. No re-discovery mechanism after mutations

`discoveredProjectsRef` in `SessionSidebar.tsx:425` runs discovery once per project set. After attaching an existing worktree, creating a new one, or removing one, the sidebar never re-runs discovery.

### 3. Failed discovery destroys existing data

If `listProjectWorktrees` throws (e.g., server not ready at startup), the project is skipped entirely. The store is overwritten with whatever succeeded — projects that failed lose their previously-persisted worktree entries.

## Impact

- Sessions in worktree directories cannot find their parent project (`resolveProjectForSessionDirectory` fails)
- Sidebar groups all worktree sessions into the project root or archive group instead of per-worktree subgroups
- Header/tray branch labels don't show the worktree branch
- `projectSessionDirectories` doesn't include worktree directories, so sessions in those dirs are not loaded
- NewWorktreeDialog can't suggest existing worktree names for uniqueness checks

## Reproduction steps

1. Create a worktree via OpenCode (or `git worktree add` in terminal)
2. Open/reload OpenChamber
3. Observe: the new worktree does not appear in the sidebar
4. The worktree also doesn't appear in the "New Worktree" dialog's branch list

## Proposed fix

1. Remove `worktrees.length === 0` skip — always add the project to the Map (even with an empty array)
2. Merge discovery results with existing store data instead of replacing — preserve worktrees for projects where discovery failed
3. Add `worktreeDiscoveryEpoch` to the store — increment on create/attach/remove; sidebar re-runs discovery when it changes
4. Remove `discoveredProjectsRef` — epoch-based invalidation handles re-discovery

### Files affected

- `packages/ui/src/sync/session-ui-store.ts` — add epoch + trigger
- `packages/ui/src/components/session/SessionSidebar.tsx` — remove skip, merge, subscribe to epoch
- `packages/ui/src/apps/ElectronMiniChatApp.tsx` — remove skip, merge
- `packages/ui/src/apps/MobileApp.tsx` — remove skip, merge
- `packages/ui/src/components/session/NewWorktreeDialog.tsx` — increment epoch after attach
- `packages/ui/src/lib/worktrees/worktreeManager.ts` — increment epoch after create/remove

### Edge case: phantom worktrees

If a worktree was deleted externally and discovery fails, merge preserves stale data. Successful discovery clears phantoms. The trade-off (stale data on transient failure vs. no data on transient failure) favors stale data — less disruptive to the user.

Related: #1817 (mini chat worktree not showing)

bashrusakh:
## Summary

Fixes worktrees created by OpenCode not appearing in the sidebar, and enables attaching existing worktrees when creating a worktree for an already-checked-out branch.

Closes #1989

### Changes

1. **`NewWorktreeDialog.tsx`** — Skip `branch_in_use` validation for existing-branch mode. When the user selects an existing branch, check for an already-attached worktree *before* calling `createWorktree`. If found, attach it via `attachWorktreeToStore` and increment the discovery epoch so the sidebar re-discovers. Handle remote branch refs (`remotes/origin/...`) by deriving the local branch name for matching.

2. **`SessionSidebar.tsx`** — Remove `worktrees.length === 0` skip that prevented discovery from running when worktrees already existed. Merge discovery results with existing store data instead of replacing (preserves worktrees for projects where discovery failed). Prune stale entries for removed projects **and** confirmed non-Git repos during merge. Increment `worktreeDiscoveryEpoch` after discovery so the sidebar re-runs when worktrees change. Track `nonGitProjectPaths` to prune confirmed non-Git repos.

3. **`ElectronMiniChatApp.tsx`** / **`MobileApp.tsx`** — Same discovery merge + prune + epoch logic as SessionSidebar. Extract `normalizeProjectPath` helper for consistent key normalization with `attachWorktreeToStore`.

4. **`worktreeManager.ts`** — Extract `attachWorktreeToStore` helper (shared between `createWorktree` and `NewWorktreeDialog`), includes flat-array dedup. Normalize project directory key via `normalizePath` for consistent matching with discovery loops. Add `worktreeDiscoveryEpoch` counter to store and `triggerWorktreeRediscovery()` helper.

5. **`session-ui-store.ts`** — Content-hash guard on `persistWorktreeMap` with try-catch to avoid redundant localStorage writes. Add `worktreeDiscoveryEpoch` field.

6. **Dead code removal** — Remove `refs/heads/` comparisons in `NewWorktreeDialog.tsx` worktree matching, since `readProjectWorktrees` already strips that prefix. Use type-guard predicate `(p): p is string => Boolean(p)` instead of `filter(Boolean)` for `currentProjectPaths` sets.

### Verification

- `bun run type-check` — clean (no new errors; pre-existing Capacitor type errors in MobileApp.tsx are unrelated)
- `bun run lint` — clean

### OCR Reviews

Round 1 findings addressed:
- Prune stale entries for removed/non-Git projects in merge
- Add epoch to ElectronMiniChatApp + MobileApp
- Extract `attachWorktreeToStore` helper with flat-array dedup

Round 2 findings addressed:
- Track `nonGitProjectPaths` separately and exclude confirmed non-Git repos from preservation during merge
- Normalize project directory in `attachWorktreeToStore` for consistent key matching
- Remove dead `refs/heads/` comparisons (always false since `readProjectWorktrees` strips the prefix)
- Use type-guard predicate for `currentProjectPaths` filter


btriapitsyn:
Resolved the conflict with `main` in mobile worktree discovery:

- kept the PR’s partial-failure-safe merge/pruning logic for discovered worktrees;
- retained the newer connection gate so discovery retries after reconnect;
- made the effect also react to the PR’s explicit discovery epoch;
- restored the required worktree type import and verified UI type-check/lint.

btriapitsyn (approved):

btriapitsyn:

PR #1998, opened 2026-07-02
fix(chat): clear stale busy state on mobile

### LINKED ISSUE #1983: [Bug] Mobile: Previous reply shows as still in-progress after completion, requiring two Send taps to send a new message

When using Open Chamber on mobile (Web/PWA), the following issue occurs frequently:

1. The previous AI reply has already finished (content is fully displayed)
2. However, the UI still shows the previous turn as "in-progress" (e.g., loading indicator or disabled input state persists)
3. Typing a new message and tapping Send does not actually send it
4. A second tap on Send is required for the message to go through

**Expected behavior:** Once the previous reply is complete, the input should immediately return to a normal sendable state — one tap on Send should be enough.

**Actual behavior:** The UI state doesn't update correctly after the reply finishes, requiring two taps on Send to actually send a message.

## Where does it happen?

Mobile (Web/PWA)

## Version (if known)

Not sure of the exact version, but running the latest release.

## Screenshots / recordings (optional)

None at the moment — can provide a screen recording if needed.

## Logs (optional)

None.

bashrusakh:
## Summary
- Fixes the mobile composer stale-busy state by deriving session activity from SDK messages and status, and clearing stale active state once the trailing assistant turn is complete.
- Extracts the derivation into a pure helper and adds regression coverage.
- Adds a small typed narrowing helper so the SDK `time` union builds cleanly.

Closes #1983

## Validation
- `bun test src/hooks/useSessionActivity.test.ts` ✅
- `bun test src/hooks/useQueuedMessageAutoSend.test.ts` ⚠️ blocked in this environment: cannot find package 'react'
- `bun run type-check` ⚠️ blocked in this environment: `tsc` not found
- `bun run lint` ⚠️ blocked in this environment: `eslint` / `@eslint/js` not found

## Notes
- Branch was recreated cleanly from `origin/main` to avoid unrelated worktree-store changes from the previous branch.

### DIFF
--- a/packages/ui/src/hooks/sessionActivity.ts
+++ b/packages/ui/src/hooks/sessionActivity.ts
@@ -1,3 +1,8 @@
+import type { Message, SessionStatus } from '@opencode-ai/sdk/v2/client';
+
+import type { PermissionRequest } from '@/types/permission';
+import type { QuestionRequest } from '@/types/question';
+
 type SessionActivityPhase = 'idle' | 'busy' | 'retry';
 
 export interface SessionActivityResult {
@@ -14,17 +19,14 @@ const IDLE_RESULT: SessionActivityResult = {
   isCooldown: false,
 };
 
-type SessionMessageLike = {
-  role?: string;
-  time?: { completed?: number };
-};
+type SessionMessage = Pick<Message, 'role' | 'time'>;
 
 type SessionActivityInput = {
   sessionId: string | null | undefined;
-  status: { type?: SessionActivityPhase } | undefined;
-  messages: SessionMessageLike[];
-  permissions: unknown[];
-  questions: unknown[];
+  status: SessionStatus | undefined;
+  messages: readonly SessionMessage[];
+  permissions: readonly PermissionRequest[];
+  questions: readonly QuestionRequest[];
 };
 
 export function deriveSessionActivity({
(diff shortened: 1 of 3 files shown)


bashrusakh:
> Good point, but session status and trailing-message completion can briefly get out of sync, so I’d keep the stale-status clear-on-complete guard as-is to avoid leaving the composer stuck. The duplicate helper call is tiny and intentional for now; happy to fold it into a cached completedAt if you want to optimize further.

btriapitsyn:
Thanks for digging into #1983, and for extracting the derivation into a pure, tested helper. The bug is real, but we can't take this shape of fix. In OpenCode's agent loop a session produces a new assistant message per step while the session-level status stays busy for the whole run. Between steps the trailing assistant message is completed and the next one isn't created yet, so "busy status + completed trailing assistant = idle" flips the composer to idle mid-run. That flickers the stop button and lets the queued-message auto-send fire into the middle of an active turn. It also inverts the sync layer's authority rule: live session status is authoritative, message history never overrides it (see `packages/ui/src/sync/DOCUMENTATION.md`).

The welcome shape of a fix for #1983 is on the transport side: when the mobile PWA resumes from background and the SSE stream missed the `session.idle` event, re-fetch the authoritative status snapshot on visibility/reconnect and let that clear the stale busy. Closing this one, but the issue stays open.

PR #2028, opened 2026-07-04
feat(mobile): swipe to open/close session drawer + swipe up for recents

OrangutAI:
## What

Adds finger-following swipe gestures to the mobile web/PWA:

- **Session drawer** — edge swipe-right (left ~28px zone) opens it; swipe-left on the open drawer closes it. Both drive the existing `leftDrawerX` motion value via a new `useMobileDrawerSwipe` hook, wired in `MainLayout`.
- **Recents panel** — swipe up on the composer footer opens the existing `mobileSessionPanelOpen` sheet via a new `useSwipeUpToOpen` hook, wired in `ChatInput`.

## Why / notes

- The close drag starts on the drawer's vertically-scrollable session list, and browsers steal a horizontal drag that begins on a scroll container (cancelling the gesture). A `touch-action: pan-y` rule on the drawer **and** its scroll container (`mobile.css`) reserves horizontal for the JS gesture while leaving vertical scroll and dnd reorder intact.
- The recents swipe is **threshold-based** (not finger-following): the shared `MobileOverlayPanel` animates via CSS transition, so the swipe fires the open and the sheet runs its own slide-up. Finger-following would require reworking that shared component.
- Both gestures are gated to `isMobile` and non-blocking (taps on the edge column / footer buttons still work). No desktop or VS Code behavior change.

## Verification

- `type-check`, `lint`, and `dead-code` (knip) all pass on `@openchamber/ui`.
- Gesture logic verified in real Chrome via touch-emulated harnesses: edge open/close, vertical-scroll-doesn't-close, direction lock + threshold, and swipe-up-opens / tap-passes-through.

## Not yet covered / follow-ups

- **Real-device QA** — the mechanics are validated under touch emulation but not on a physical phone. Worth a sanity pass on iOS Safari / Android Chrome (drawer + list scroll + dnd reorder; footer swipe-up + button taps).
- Optional polish: finger-following for recents (needs `MobileOverlayPanel` to be motion-driven) and swipe-down-to-close on the recents sheet.

### DIFF
--- a/packages/ui/src/components/chat/ChatInput.tsx
+++ b/packages/ui/src/components/chat/ChatInput.tsx
@@ -38,6 +38,7 @@ import { MobileAgentButton } from './MobileAgentButton';
 import { MobileModelButton } from './MobileModelButton';
 import { MobileSessionStatusBar, MobileSessionPanelTrigger } from './MobileSessionStatusBar';
 import { useCurrentSessionActivity } from '@/hooks/useSessionActivity';
+import { useSwipeUpToOpen } from '@/hooks/useSwipeUpToOpen';
 import { toast } from '@/components/ui';
 import { Button } from '@/components/ui/button';
 // useMessageStore removed — messages now come from sync system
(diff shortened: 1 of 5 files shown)


btriapitsyn:

PR #2034, opened 2026-07-05
feat(release): add Linux .deb packaging

zhusinian:
## Summary

Add Linux `.deb` packaging support for the Electron desktop build.

## Changes

- add Linux `.deb` packaging support in the Electron build flow
- update desktop/build documentation for Linux packaging
- adjust open-in-app launcher behavior and related UI/store handling for Linux support
- add/update tests covering the new launcher and app-opening behavior

## Validation

- [x] `bun run type-check`
- [x] `bun run lint`
- [x] `bun run build`
- [x] manual verification on Linux during daily use for about one week with no obvious issues observed

## Update Support

Add Linux `.deb` publishing and in-app update support.

Verified successful `.deb` updates using the local test update server.

### Deb update reliability

**Reason:** `electron-updater` only checks `spawnSync.status` when installing deb packages. If `dpkg` is terminated by a signal, Node returns `status=null`, which can be mistaken for success and trigger a restart while the package is still unpacked or unconfigured.

**Solution:** OpenChamber now installs and verifies the downloaded deb itself. It validates the package identity, version, and architecture, repairs incomplete configuration when needed, and restarts only after `dpkg` reports `install ok installed` (`ii`) with the expected version. Native installation errors are also preserved for the UI.


mattv8:

PR #2043, opened 2026-07-05
fix(sync): keep session renames stable

### LINKED ISSUE #2031: [Bug] Inline rename reverts to original title after clicking save (race condition with SSE session.updated)
### Steps to reproduce

1. Open OpenChamber Web UI (v1.13.9)
2. In any project sidebar, double-click a session title (or right-click → Rename)
3. Type a new name, click the ✓ checkmark
4. The new name appears briefly, then **immediately reverts** to the original title
5. The `/rename` command works fine in opencode TUI (v1.17.13)
6. All sessions under this project are affected; sessions under other projects may work fine

### Environment

- OpenChamber: 1.13.9
- OpenCode: 1.17.13
- OS: Windows 11

### Root cause (from code analysis)

There's a race condition in the rename flow:

1. `session-actions.ts:563` — `updateSessionTitle` calls the SDK, then `upsertSession`
2. `sync-context.tsx:685` — `applySessionEventToGlobalSessions` catches `session.updated` SSE events and calls `upsertSession` with the event payload

If the SSE `session.updated` event arrives **after** the SDK response upsert, it overwrites the new title with stale data from the server broadcast.

The issue is scoped to projects where `getSessionDirectory(sessionId)` returns `null`/`""` (e.g., sessions without a valid `directory` field), causing the SDK PATCH request to not properly target the session's project, and the server to return/broadcast stale data.

### Proposal

`updateSessionTitle` should either:
- Suppress the SSE `session.updated` handler temporarily after a rename (debounce), or
- Compare timestamps and only apply the SSE update if it's newer than the pending rename, or
- Pass an `origin` hint so the SSE handler can skip events emitted by the same client action

bashrusakh:
## Summary
Fixes the inline session rename race where a successful rename could be reverted by a late `session.updated` SSE echo.

Fixes : https://github.com/openchamber/openchamber/issues/2031

## Changes
- Added a shared freshness check for session events using `time.updated`, with `time.created` as fallback.
- Applied the stale-event guard in both the live SSE reducer and the global session event router.
- Mirrored successful rename responses into the live session store so the sidebar shows the new title immediately.
- Added regression coverage for:
  - stale live/global `session.updated` events
  - the freshness fallback path
  - the direct rename response path

## Validation
- `bun test src/sync/session-actions.test.ts src/sync/__tests__/event-reducer.test.ts src/sync/__tests__/session-event-freshness.test.ts src/sync/__tests__/sync-context-session-events.test.ts`
- `bun run type-check`
- `bun run lint`

## Notes
- Scope is limited to the session rename/SSE overwrite path.
- Share, unshare, and archive flows remain authoritative through their normal SDK responses.


btriapitsyn:
Resolved the conflicts with `main` in session actions and tests:

- preserved the PR’s live-store rename propagation, including fallback across initialized directory stores;
- retained the newer authoritative last-assistant-model helper and message confirmation test plumbing from `main`;
- ran the focused session-actions suite (22 passing), UI type-check, and lint.

btriapitsyn (approved):

PR #2051, opened 2026-07-06
fix(chat): surface session.error to the user viewing the session

abigfatstone:
## What

Show an error toast when `session.error` arrives for the session the user is currently viewing.

## Why

Today `session.error` reaches the user through exactly one surface: the notification store. But for the **viewed** session the notification is created `viewed: true` (no badge), and the event reducer only flips the session status back to idle — the error payload itself is dropped. Net effect: a provider 401, exhausted quota, or network failure ends the turn in complete silence for the person watching the chat; the spinner just stops.

The toast is keyed by session ID (`session-error-<id>`) so repeated errors replace rather than stack, and picks the most specific detail available (`error.data.message` → `error.message` → HTTP status → error name).

Background/unviewed sessions are unchanged — they already get the badge + notification path.

## Validation

- `tsc --noEmit` (packages/ui) clean
- eslint on the touched file clean

### DIFF
--- a/packages/ui/src/sync/sync-context.tsx
+++ b/packages/ui/src/sync/sync-context.tsx
@@ -1502,15 +1502,35 @@ function handleEvent(
     if (session && (session as { parentID?: string }).parentID) {
       // subtask — skip notification
     } else if (sessionID) {
+      const viewed = isViewedInCurrentSession(resolvedDirectory, sessionID)
       appendNotification({
         directory: resolvedDirectory,
         session: sessionID,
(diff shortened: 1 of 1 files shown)


mattv8:
@btriapitsyn, this looks ready for stale closure. The branch still conflicts with main and has not had an author commit since July 6.

btriapitsyn:
Closing as stale: the branch still conflicts with main and has had no commits since July 6. Thanks for flagging this, @mattv8.

PR #2070, opened 2026-07-07
fix(chat): restore mobile PWA composer send

### LINKED ISSUE #2069: fix(chat): mobile PWA composer cannot send messages in v1.14.0
## Summary

On OpenChamber v1.14.0, the mobile/PWA chat composer can become unable to send messages:

- Tapping the Send button dismisses/collapses the input instead of sending.
- Pressing the mobile keyboard Send/Enter key inserts a newline instead of submitting the message.

This appears to be a regression from the v1.14.0 mobile composer overhaul (compact composer + fullscreen editor + keyboard choreography).

## Environment

- OpenChamber: v1.14.0
- Runtime: Web/PWA installed on mobile
- Platform: mobile browser / installed PWA

## Steps to reproduce

1. Install/open OpenChamber v1.14.0 as a mobile PWA.
2. Open an existing chat session or create a new draft session.
3. Tap the composer and type a message.
4. Try to send it using either:
   - the Send button in the composer footer, or
   - the mobile keyboard Send/Enter key.

## Actual behavior

- Tapping Send closes/collapses the input/composer, but no message is sent.
- Pressing the keyboard Send/Enter key inserts a newline into the textarea.

## Expected behavior

- Tapping Send should submit the message even while the software keyboard is open.
- The mobile keyboard action should be labelled as Send and submit the message.
- Shift+Enter / explicit multiline behavior should remain available for hardware keyboards.

## Suspected cause

There are two related mobile paths in `packages/ui/src/components/chat/ChatInput.tsx`:

1. `ComposerActionButtons` uses a mobile-only `type="button"` + `onClick` send path. In installed PWAs, tapping a control while the software keyboard is open can blur the textarea first. The keyboard resize/collapse can move the Send button before the synthesized click arrives, causing the tap to only dismiss/collapse the composer.
2. The Enter submit handler currently excludes mobile unless Ctrl/Meta is held:

```ts
if (e.key === 'Enter' && !e.shiftKey && (!isMobile || e.ctrlKey || e.metaKey)) {
```

That makes mobile keyboard Send/Enter fall through to normal textarea newline insertion.

## Proposed fix

- Guard touch pointer-down on the mobile Send button with `preventDefault()` so focus is preserved until the click handler fires.
- Treat unmodified Enter as submit on mobile as well, while preserving Shift+Enter as newline.
- Set `enterKeyHint="send"` on the mobile textarea so virtual keyboards show the correct action.

I have a small PR prepared with this fix.

PintaAI:
## Summary

Fixes #2069.

Restores mobile/PWA composer sending after the v1.14.0 compact/fullscreen composer changes:

- Keep the textarea focused when tapping the mobile Send button so installed PWAs do not collapse the composer before the synthesized click fires.
- Let unmodified Enter submit on mobile too; Shift+Enter remains the newline path for hardware keyboards.
- Add `enterKeyHint="send"` on mobile so virtual keyboards expose a Send action.

## Why

In installed PWAs, tapping Send while the software keyboard is open can blur the textarea before `click` is delivered. The keyboard-resize/collapse choreography can then move the composer into the compact pill and the tap is lost, so it looks like Send only closes the input.

The keyboard path also explicitly excluded mobile unless Ctrl/Meta was held, so the mobile Send/Enter key always fell through to textarea newline insertion.

## Test plan

- `bun run --cwd packages/ui type-check`
- `bun run --cwd packages/ui lint`

### DIFF
--- a/packages/ui/src/components/chat/ChatInput.tsx
+++ b/packages/ui/src/components/chat/ChatInput.tsx
@@ -798,10 +798,24 @@ const ComposerActionButtons = React.memo(function ComposerActionButtons(props: C
     } = props;
     const { t } = useI18n();
 
+    const guardMobileTouchAction = React.useCallback((event: React.PointerEvent<HTMLButtonElement>) => {
+        if (!isMobile || event.pointerType !== 'touch') {
+            return;
+        }
+
+        // In installed PWAs, tapping Send while the software keyboard is open
+        // blurs the textarea before the synthesized click arrives. The keyboard
+        // resize can then collapse the full composer into the compact pill and
+        // move the button out from under the finger, losing the send. Preserve
+        // focus through the touch sequence so the normal click handler fires.
+        event.preventDefault();
+    }, [isMobile]);
+
     const sendButton = (
         <button
             type={isMobile ? 'button' : 'submit'}
             disabled={!canSend || (!currentSessionId && !newSessionDraftOpen)}
+            onPointerDownCapture={guardMobileTouchAction}
             onClick={(event) => {
                 if (!isMobile) {
                     return;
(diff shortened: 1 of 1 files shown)


btriapitsyn:

PR #2076, opened 2026-07-07
fix(chat): bypass SDK v2 body-omission bug in session.create and promptAsync

herjarsa:
## Problem

SDK v2's auto-generated `session.create` and `session.promptAsync` spread options into the request config but never assign a `body`, so the underlying fetch sends an empty body (or the GET method for promptAsync). This causes:

1. **Session creation fails silently** — the server returns 200 with an empty body (Express SPA fallback for the `/session` POST), the SDK's `unwrapSdkData` returns `{}` without `.id`, and the UI shows "Failed to create session" with no network error.
2. **Message sending fails silently** — `prompt_async` is sent as GET, Express returns 200 with HTML (SPA shell), no error is thrown, but the model is never invoked.
3. **WS ready timeout race** — the server only sends the WS `ready` frame when `globalHub.isConnected()`, which races the upstream SSE handshake. The client's 2s timeout fires first, locking the pipeline into SSE-only mode for 60s.
4. **SyncProvider unmounted** — App.tsx has early-return branches that skip SyncProvider, so the event pipeline never starts.
5. **Draft auto-close** — opening a session triggers `setCurrentSession` → `closeNewSessionDraft`, which closes the draft before `sendMessage` runs, bypassing the whole create-session flow.

### 1. `client.ts`: bypass SDK v2 body-omission in `createSession`
Replace the SDK call with a direct `fetch` that always sends `directory` and conditionally includes `title`/`parentID`/`metadata` (never sending `null` literals that the server rejects).

### 2. `client.ts`: bypass SDK v2 body-omission in `promptAsync`  
Replace the SDK call entirely with a direct POST fetch that serializes the full body (sessionID, model, messageID, parts, agent, variant, directory, delivery, format). Preserves the retry loop, timeout signal, and provider circuit breaker.

### 3. `global-ws-bridge.js`: remove `isConnected()` gate
Always call `markReady` immediately on WS connect. The hub buffers events that arrive before upstream-connect; if the hub never connects, `closeClientsWithInitialError` sends an error frame.

### 4. `event-pipeline.ts`: bump WS ready timeout 2s → 10s
Defensive upper bound for cold-start upstream SSE handshake.

### 5. `main.tsx`: move SyncProvider to React root
Wraps every render path — empty state, error boundaries, boot views.

### 6. `session-ui-store.ts`: auto-open draft in `materializeOpenDraftSession`
Re-opens the draft on demand via `openNewSessionDraft({})` if it was closed by a prior `setCurrentSession` call.

## Root Cause

The SDK v2 auto-generated client uses a code-gen pattern where request options (path params, query params, body) are spread into the same `options` object, but no `body` field is ever set. The auto-generated `post` method correctly sets `method: 'POST'`, but with an empty body the server either rejects or returns the SPA fallback.

## Verification

- `POST /api/session` now sends the correct JSON body and returns 200 with a valid session
- `POST /api/session/{id}/prompt_async` now sends the correct JSON body and returns 204
- WebSocket `ready` frame arrives before the client's timeout on every page load
- SyncProvider is always mounted, regardless of App.tsx render path
- Draft sessions are created when sending from the home page even after viewing an existing session


herjarsa:
Superseded by #2078 — this PR mixed 73 commits (proxy rewrite, APNs/dictation/health-monitoring removal, etc.) with the SDK body fix. The SDK fix has been split into a clean 7-commit PR that applies without conflicts on latest main.

PR #2081, opened 2026-07-07
fix(sync): self-heal wedged busy status after send (#2072)

### LINKED ISSUE #2072: [Bug] New messages often don't send in long-context sessions
### What happened and what you expected

**Actual**: 
- When in a longer session (after multiple conversation rounds), sending a new message often has no response - the message doesn't appear to be sent.
- Loading a session with long content is also very slow.

**Expected**: User should see the message sent and receive an AI response.

### Steps to reproduce

1. Use openchamber 1.41.1 web UI
2. Have a multi-round conversation in a session
3. When context grows longer, try sending a new message
4. Message may fail to send with no UI response

### Environment

- **Desktop**: Linux (Fedora)
- **Version**: 1.41.1

### Where does it happen?

Desktop Web

### Version (if known)

1.14.1

Image

Image

### Logs (optional)

```shell

```

bashrusakh:
## Summary

Fixes the silent-send wedge in long-context sessions: after a successful SDK call whose trailing `session.status idle` SSE event is lost, the busy flag stays set forever, the composer shows "Gears turning ..." with no progress, and the next user send is force-queued because `useQueuedMessageAutoSend` gates on `currentStatus !== 'idle'`. Adds a per-send 4s self-heal watchdog that re-reads the authoritative `/session/status` snapshot and reconciles via the existing `applySessionStatusSnapshot` helper. Recovery lands within ~1s of the existing 5s periodic poll.

## Root cause

`optimisticSend` in `packages/ui/src/sync/session-actions.ts` sets `session_status: busy` optimistically and only clears it in the `catch` path. The success path relies on the server's SSE `session.status idle` event to clear the flag. If the SDK call resolves (200) but the idle event is dropped (long prepending, transient network blip, or any reason the event pipeline loses the tail), the busy flag stays set forever. That wedges:

- `useSessionActivity` reports `isWorking: true` → "Gears turning ..." status
- `useQueuedMessageAutoSend:148-151` has a hard `if (currentStatus !== 'idle') return;` gate → the next user send is force-queued instead of dispatched (matches the "hi" chip in the user's screenshot)

Existing recoveries don't cover this case:
- `cf36b55e` (recovery on materialization) — only runs on session open
- `1869-stuck-busy` watchdog — covers stream-stale, not request-stale

## Fix

After `optimisticSend` resolves, schedule a 4s self-heal watchdog (4s = longer than typical SSE delivery, shorter than the 5s `ACTIVE_SESSION_STATUS_POLL_INTERVAL_MS` periodic poll, so the per-send watchdog wins the race by ~1s):

1. **Early-return** if the local `session_status[sessionId]` is no longer `busy` (normal SSE-delivered case — no-op).
2. **Fetch** `/session/status` via `opencodeClient.getSessionStatusForDirectory(directory)`. A `null` response (network error) is preserved — never flip busy → idle on a transient blip (per the project's "Distinguish fetch failure from empty success" rule).
3. **Delegate** to the existing `applySessionStatusSnapshot` helper (same one used by the 5s periodic poll and the reconnect recovery path) in `authoritative` mode. The helper:
   - Lowers a wedged `busy` to `idle` when the snapshot omits the session or marks it `idle`.
   - Confirms a matching `busy`/`retry` (server agrees, AI is still working).
   - Is a no-op for non-busy states.

`applySessionStatusSnapshot` is injected from `SyncProvider` via a new module-level ref `setApplySessionStatusSnapshot` (set on mount, cleared on unmount) to avoid a circular import (`sync-context` already imports from `session-actions`).

A `Map<sessionId, timeoutId>` guard prevents pending watchdog callbacks from stacking when the user sends multiple messages quickly on the same session.

## Behavioral contract

- **User action** (type message, press send) is unchanged.
- The watchdog is a server-driven recovery — no raw/internal/manual values are exposed to users.
- Worst case: if the server is slow to settle on `idle`, the user sees "Gears turning ..." for up to 4s longer than today, then the session self-heals without user action.

## Diff

```text
 packages/ui/src/sync/session-actions.test.ts | 442 ++++++++++++++++++++++
 packages/ui/src/sync/session-actions.ts      | 126 +++++
 packages/ui/src/sync/sync-context.tsx        |   9 +-
 3 files changed, 576 insertions(+), 1 deletion(-)
```

## Validation

| Command | Result |
| --- | --- |
| `bunx tsc --noEmit` (packages/ui) | ✅ clean |
| `bunx eslint` on the 3 changed files | ✅ clean |
| `bun test packages/ui/src/sync/session-actions.test.ts` | ✅ 30 pass / 0 fail |
| `bun test packages/ui/src/sync/__tests__/session-status-snapshot.test.ts` | ✅ 17 pass / 0 fail (regression check) |
| `bun test packages/ui/src/sync` (full suite) | ✅ 170 pass / 7 pre-existing fail / 2 pre-existing error — **0 new failures** |
| `bun run dead-code` | ✅ identical to baseline, 0 new dead exports |

## Test coverage

The `describe("optimisticSend busy-status self-heal watchdog (#2072)")` block in `session-actions.test.ts` covers all 6 contract branches plus 4 edge cases:

1. `clears wedged busy status when the server reports the session idle (absent from snapshot)`
2. `clears wedged busy status when the server explicitly returns { type: 'idle' }`
3. `preserves busy when the server reports the session is still busy`
4. `raises busy to retry when the server reports the session is in retry`
5. `preserves busy when the fetch fails (getSessionStatusForDirectory returns null)`
6. `is a no-op when the SSE has already moved status out of busy (e.g. idle)`
7. `is a no-op when the SSE has already moved status out of busy (e.g. retry)`
8. `invokes the production applySessionStatusSnapshot ref wired by setApplySessionStatusSnapshot` — exercises the production ref path via `__testRunOptimisticStatusWatchdog`
9. `no-ops gracefully when the production applySessionStatusSnapshot ref is null` — SyncProvider-unmounted edge case
10. `lowers busy to idle when the candidate is absent from a populated snapshot, without touching other sessions` — helper invariance to other entries
11. `replaces a pending watchdog instead of stacking when optimisticSend fires twice on the same session` — concurrent-watchdog guard

## Out of scope

The second symptom in the issue ("Loading a long-content session is slow, skeleton rows for a long time") is **not** addressed by this fix. That symptom is caused by `INITIAL_MESSAGE_PAGE_SIZE = 50` + the page-expansion loop in `packages/ui/src/sync/use-sync.ts:355-366`, which can issue up to 4 HTTP round-trips for a 200-message session. Recommend a separate issue + PR.

## Risks

- **None known in this fix.** The watchdog is a no-op on the normal path (SSE wins the race), preserves state on fetch failure, and uses the same authoritative-snapshot reconciliation as the existing 5s periodic poll and the reconnect recovery path. Three independent code paths now share one source of truth for the snapshot reconciliation contract (`applySessionStatusSnapshot` in `sync-context.tsx`).
- The 4s window is the worst case for visible "Gears turning ..." latency after a missed idle event, and is 1s faster than the 5s periodic poll.
- A pre-existing concern surfaced during the investigation (out of scope for this PR): `disableStaging?: boolean` is declared on `MessageList` but never destructured. Not introduced by this fix.

Closes #2072


bashrusakh:
**Follow-up issue for the slow-load symptom**: [Issue #2084](https://github.com/openchamber/openchamber/issues/2084) tracks the second symptom from the original bug report — long-context sessions showing skeleton rows for seconds due to the `INITIAL_MESSAGE_PAGE_SIZE = 50` page-expansion loop in `use-sync.ts`. This PR (#2081) only fixes the send-failure wedge; the slow-load issue is separate and should be addressed in a follow-up PR.

btriapitsyn:
Resolved the conflicts with `main` while preserving both behaviors in `session-actions`:

- kept the PR’s authoritative status watchdog that self-heals a lost SSE idle event;
- retained the newer send-confirmation refetch/materialization logic from `main`;
- ran the focused session-actions suite (32 passing), UI type-check, and lint.

PR #2086, opened 2026-07-07
fix(sync): commit first message page before expansion loop (#2084)

### LINKED ISSUE #2084: [Bug] Long-context sessions load slowly — skeleton rows shown for seconds due to page-expansion loop
### What's wrong

Opening a long-context session (200+ messages) shows grey skeleton rows for several seconds before any content appears. The delay scales with session length.

### Steps to reproduce

1. Use openchamber web UI (any platform)
2. Have a session with 100+ messages (e.g., a multi-round conversation)
3. Click on that session or navigate to it
4. Observe: 3 grey skeleton rows are shown for multiple seconds before the actual chat content appears

### Root cause (investigated in #2072)

The initial message load uses `INITIAL_MESSAGE_PAGE_SIZE = 50` (`packages/ui/src/sync/use-sync.ts:25`). If the first 50 messages contain no user message (common in long sessions where the latest records are all assistant/tool responses), the code expands the page in a loop:

```
50 → 100 → 150 (hard cap)
```

Each expansion is a separate HTTP round-trip to the OpenCode server (`session.messages()`). For a 200-message session, this means **3-4 sequential round-trips** before the first user message is found and the chat can render. During this time, `sessionMessages.length === 0` and the skeleton is shown.

Additionally, after the initial page lands, a `HISTORY_MESSAGE_PAGE_SIZE = 100` prepend fetch fires (line 506-521), adding another round-trip.

The `renderable` gate in `packages/ui/src/sync/materialization.ts:213-224` also requires every assistant message in the page to have parts before the skeleton is hidden — a single slow part on the last message keeps the skeleton up.

### Expected behavior

- Show whatever messages are available immediately, even if the first page is incomplete
- Reduce or eliminate the sequential page-expansion loop (e.g., fetch a larger initial page, or use a single request with a higher limit)
- Show a non-blocking "loading earlier messages…" indicator instead of a blocking skeleton

### Environment

- **Desktop**: Linux (Fedora) — also reported on Windows (#1726)
- **Version**: 1.14.1
- **Platform**: web (desktop web browser)

### Related

- #2072 — original bug report (this is the slow-load symptom from that issue)
- #1726 — Windows-specific slow dialogue report (may share the same root cause)
- #2081 — fix for the send-failure symptom from #2072 (separate root cause)

### Labels

`bug`, `area:sessions`, `area:chat-ui`, `platform:web`, `priority:medium`, `reproducible:true`

bashrusakh:
## What

Opening a long-context session showed a grey skeleton for several seconds. The store write in `loadMessages` happened only after the full expansion sequence (50→100→150 messages), so the skeleton persisted for up to 3 sequential HTTP round-trips.

## Change

Move the materialize + store write to happen **after the first fetch**. The expansion loop now commits each expanded page incrementally instead of accumulating pages and writing once at the end.

- Extracted `commitMessagesToStore` helper inside `loadMessages` — merge optimistic → materialize → `store.setState`.
- First page commit is gated on `hasUserMessage(page.session) || page.complete`: if the tail is assistant-only (no user boundary), the commit is deferred to the expansion loop so the skeleton stays as a loading state instead of an empty chat that looks like a fresh session.
- Added `isStale` guard before each expansion fetch and an explicit stale early-return after the first commit.
- Exported `hasUserMessage` for test coverage.

## Behavior

| Scenario | Before | After |
|---|---|---|
| Session with user boundary in first 50 messages | skeleton for 1 round-trip | content after 1 round-trip (same) |
| Session with assistant-only tail (large final turn) | skeleton for 3 round-trips | skeleton for 2–3 round-trips (expansion runs, but skeleton stays instead of empty chat) |
| Session switch (rapid) | skeleton | same — `isStale` guards prevent stale store writes |

## Tests

Added 13 tests in `use-sync.test.ts` covering the three invariants the refactor relies on:
- `hasUserMessage` gating (5 tests — user message, clientRole, assistant-only, undefined, empty)
- incremental materialization reference preservation (3 tests — superset pages, part references, no-op commits)
- `mergeOptimisticPage` idempotency across commits (2 tests)
- first-commit gate invariant (3 tests — assistant-only defers, user triggers, complete triggers)

## Validation

- `bun run --filter @openchamber/ui type-check` — exit 0
- `bun run --filter @openchamber/ui lint` — exit 0
- `bun test packages/ui/src/sync/use-sync.test.ts` — 16 pass, 0 fail
- Full sync test suite — 172 pass, 7 pre-existing fail (event-pipeline/live-aggregate, unrelated to this change)

Fixes #2084

## Related

- #2072 — original bug report (this is symptom B)
- #2081 — fix for symptom A (send-failure)


btriapitsyn:
Resolved the conflicts with `main` and preserved both sides of the sync behavior:

- kept the PR’s incremental first-page commit so session content renders after the first useful page;
- retained the newer `main` safeguard that resync fetches at least the number of messages already rendered, preventing a phantom “load older” cursor;
- verified the UI package with type-check and lint.

btriapitsyn (approved):

Issue #2096, opened 2026-07-08
[Bug] Node.js event loop starvation blocks all API requests when actively using large multi-project workspace


cpkt9762:
# Background

Under normal use of a workspace with 6–7 projects, the embedded Express server in OpenChamber's Electron main process becomes almost fully unresponsive: **/health itself takes 565s to return** during heavy periods. Two independent HAR captures during real-world use confirm this is not a transient spike but a sustained failure mode.

**HAR #1** (t = 3:20 AM, ~4 min window, right after a project was opened):
- **290 out of 365 requests (79%) status = 0** — connection cancelled after browser timeout
- The 51 that completed had a fastest response of **48 seconds**
- Even `GET /health` and `GET /api/opencode/health` (both should be sub-ms) hung for 299s

**HAR #2** (t = 3:39 AM, ~2.6 min window, after our workaround mitigations):
- **218 out of 512 requests (43%) status = 0** — still catastrophic
- Fastest 200 OK: **409 seconds**; p50 200 OK: **494 seconds**; p99: **565 seconds**
- **206× `/api/session/status`** + **107× `/api/session`** in 156 seconds (>2 req/s just for session polling)
- WebSocket handshakes (`/api/global/event/ws`) and SSE streams (`/api/notifications/stream`) continue to work — confirming the server socket layer is alive, but the **Node.js main thread is starved by synchronous work**.

# What the server is actually doing

- Express server lives **inside the Electron main process** with **237 registered API endpoints** and **~192 synchronous blocking calls** (`readdirSync`, `statSync`, `readFileSync`, `execSync`, `spawnSync`) across `packages/web/server/lib/`. No `worker_threads`, no `cluster`.
- Any heavy handler (large `git status`, `git check-ignore` on a directory tree, filesystem walk for skills/agents catalogs, config file scan) monopolizes the single event-loop thread. All other pending requests, including `/health`, `/api/config/settings`, `/api/opencode/health`, and IPC to the renderer, wait in line.
- Consequence: the desktop UI itself gets sluggish because the main process is also handling menus, tray, window management, and IPC.

# Contributing bugs on the client side that amplify the problem

## 1. `MarkdownRendererImpl.fileReferenceExists` fires `/api/fs/stat` for every path-shaped substring in every rendered message, without a negative-result cache

[...]

PR #2102, opened 2026-07-08
perf(sync): cap session prefetch limit to prevent full-page refetch (#2096)

cpkt9762:
## What

Caps the persisted session prefetch limit at `SESSION_PREFETCH_LIMIT_CAP = 150`. On cold reopen of a session, the initial `/api/session/:id/message` request is bounded regardless of how many messages were materialized in the previous mount. In-memory messages and cursor-based load-older continue to work unchanged.

## Why

While debugging the `blocked >> wait` performance pattern in #2096, an 18‑minute HAR shows `/api/session/:id/message` at the very top of the payload chart:

| Metric | Value |
|---|---|
| Total `GET /api/session/:id/message` | **337** |
| Total response body | **~352 MB** (≈ 62 % of all local traffic) |
| Peak single response | **8.07 MB** at `limit=611` |
| Distribution of `limit` on max requests | `100`, `351`, `451`, `551`, `611`, `612` |

Root cause: after materializing a session, both `setMetaFor` and `setSessionPrefetch` write back `materialized.messages.length`. On a subsequent cold reopen, `setSessionPrefetch`'s persisted `limit` becomes the initial fetch size. A session that grew to 611 messages therefore triggers a **611-message replace fetch** every time it is reopened. Hydrated parts include tool outputs, patches, and markdown, so a single response commonly reaches 7-8 MB — and each response drives full JSON parse, materialize, store update, and React render on the renderer main thread.

Note: this pattern persists even after the recently merged `e2c5da62 fix(sync): stop watchdog redundant resyncs on healthy event stream (#1829)` — that fix removes one trigger for a full replace fetch, but the message payload amplification exists on every cold reopen path (project switch, session switch, reconnect, provider switch).

## Why cap only the persisted prefetch

- `setMetaFor` (line 397) writes what is currently in memory — capping it there would risk overwriting in-memory messages on the next replace fetch, which would harm users who scrolled to older messages during the session.
- `setSessionPrefetch` (line 411) writes what to fetch **next time this session is opened**. Capping only here bounds the cold-open initial fetch while leaving current-session in-memory state alone.

The propagation path is intact: `setMetaFor` at line 454 copies `prefetchInfo.limit` into `meta.limit` on next mount, so the cap flows naturally into subsequent fetch decisions.

## What this does not do

- Does not change how many messages the user can eventually view. `load older` continues to hydrate older history via the existing `before` cursor.
- Does not touch the constrained-runtime page-expansion logic in `use-sync.ts` — that code path uses its own `getConstrainedInitialPageExpansionMax()` and is unaffected.
- Does not consolidate the duplicate WS/SSE streams or fix the `MarkdownRendererImpl` scanner, both tracked in #2096.

## Choice of 150

The default desktop initial page is 50 (150 in some code paths); 150 is a middle ground that:

- covers the common "reopen and see recent context" case without a load-older click,
- keeps single response bodies under ~2 MB in the observed HAR distribution,
- preserves headroom for the constrained-runtime page-expansion logic.

Happy to switch to a config-driven value if preferred.

## Testing

- `bun run type-check` — pass (all 5 workspaces)
- `bun run lint` — pass (all 5 workspaces)
- `bun run electron:build` — pass; packaged locally and used daily since the patch. Cold reopen of large sessions is subjectively much smoother; older messages still reachable via load-older.

Refs #2096

### DIFF
--- a/packages/ui/src/lib/runtime-switch.test.ts
+++ b/packages/ui/src/lib/runtime-switch.test.ts
@@ -62,4 +62,47 @@ describe('runtime endpoint switching', () => {
       }
     }
   });
+
+  test('re-applying the same endpoint does not re-dispatch the change event', () => {
+    const previousWindow = Object.getOwnPropertyDescriptor(globalThis, 'window');
+    const previousFetch = globalThis.fetch;
+    const dispatched: string[] = [];
+    const runtimeWindow = {
+      addEventListener: () => undefined,
+      removeEventListener: () => undefined,
+      dispatchEvent: (event: Event) => {
+        dispatched.push(event.type);
+        return true;
+      },
+    };
+
+    try {
+      clearRuntimeUrlAuthToken();
+      setRuntimeExtraHeaders(null);
+      globalThis.fetch = (async () => new Response(JSON.stringify({ token: 'url-token', expiresAt: Date.now() + 60_000 }), {
+        status: 200,
+        headers: { 'Content-Type': 'application/json' },
+      })) as typeof fetch;
+      Object.defineProperty(globalThis, 'window', {
+        configurable: true,
+        value: runtimeWindow,
+      });
+
+      switchRuntimeEndpoint({ apiBaseUrl: 'https://guard-a.example', runtimeKey: 'guard-a' });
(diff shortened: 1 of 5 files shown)


btriapitsyn:

PR #2125, opened 2026-07-10
fix: discover live remote branches

### LINKED ISSUE #2098: New Worktree dialog does not show newly fetched remote branches until later
# Background

When creating a worktree, the branch list (source branch / existing branch pickers) does not reflect branches that were just fetched from the remote. After fetching, a remote branch visible via `git branch --remote` in a terminal is missing from the UI, and only shows up later after some unrelated operation runs.

Expected: triggering "fetch branches" in the create-worktree dialog should make every branch list for the current project reflect an up-to-date view of local and remote branches.

# Version

OpenChamber 1.14.1, OpenCode 1.17.14

# Root cause

The refresh button calls `fetchBranches` -> `getGitBranches` -> `getBranches` (`packages/web/server/lib/git/service.js:3188`). That function:

1. Calls `git.branch()`, which only reads refs that exist locally (including `remotes/<remote>/<branch>` tracking refs).
2. Runs `filterActiveRemoteBranches` (`service.js:3214`), which `ls-remote`s each remote but then returns only the **intersection** with local remote-tracking refs (`service.js:3236`).

So the "fetch branches" action never performs an actual `git fetch`. A branch that exists on the remote but has no local `remotes/<remote>/<branch>` ref yet is dropped by the intersection filter and stays hidden until some other flow (worktree creation, push, etc.) happens to fetch it.

# Acceptance Criteria

## Must

- [ ] Clicking the refresh ("fetch branches") button in the New Worktree dialog makes newly pushed remote branches appear in the source/existing branch lists immediately, without requiring any other action.
- [ ] A branch that exists on the remote but has no local remote-tracking ref is shown after the refresh (i.e. the list reflects the remote state, not just local refs).
- [ ] Behavior is consistent across the desktop dropdown picker and the mobile branch picker in the dialog.

## Should

- [ ] The refresh surfaces a clear error/loading state when a `git fetch` or `ls-remote` fails (e.g. offline or unreachable remote), instead of silently showing a stale list.

---

Created with [create-issue](https://github.com/tomzx/agents/blob/f51eabc7f4e3b3d338b351ceafb519b8be4049eb/skills/create-issue/SKILL.md) via glm-5.2 (`f51eabc`)

Quat3rnion:
## Summary
- include live remote heads even when no local tracking ref exists
- keep web, desktop, and VS Code branch discovery behavior aligned
- preserve cached branches and show localized refresh feedback when remotes are unreachable
- avoid exposing raw Git diagnostics to clients

## Validation
- `bun run type-check`
- `bun run lint`
- `bun run --cwd packages/web test -- server/lib/git/service.test.js` (46 passed)
- `bun test packages/vscode/src/gitService.test.js` (2 passed)
- `bun test packages/ui/src/stores/useGitStore.test.ts` (13 passed)
- Playwright refresh failure QA at 375px, 768px, and 1280px

Fixes #2098

### DIFF
--- a/packages/ui/src/components/session/NewWorktreeDialog.tsx
+++ b/packages/ui/src/components/session/NewWorktreeDialog.tsx
@@ -353,12 +353,15 @@ export function NewWorktreeDialog({
   const hasSourceBranchMatches = sourceBranchRankedGroups.matching.length > 0;
   const canFetchBranches = Boolean(projectDirectory && git);
 
-  const handleFetchBranches = React.useCallback(() => {
+  const handleFetchBranches = React.useCallback(async () => {
     if (!projectDirectory || !git) {
       return;
     }
-    void fetchBranches(projectDirectory, git);
-  }, [projectDirectory, git, fetchBranches]);
+    const success = await fetchBranches(projectDirectory, git);
+    if (!success) {
+      toast.error(t('session.newWorktree.error.fetchBranchesFailed'));
+    }
+  }, [projectDirectory, git, fetchBranches, t]);
 
   React.useEffect(() => {
     if (!open || !projectDirectory || !git) return;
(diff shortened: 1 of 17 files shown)


btriapitsyn:

PR #2135, opened 2026-07-11
Fix: small model dispatch fails for custom OpenAI-compatible providers (#2134)

### LINKED ISSUE #2134: [Bug] Fail to call small_model when it points to a custom provider
### Symptoms

When `small_model` is configured (via `opencode.json` or `opencode.jsonc`) to use a custom OpenAI-compatible provider — one that isn't in the built-in models catalog — calling the small model fails with:

```text
Provider "xxx" has no known API base URL
```

This happens when you try to generate a commit or a PR description.

### Root cause

The small model dispatch in `callSmallModel()` resolves the API base URL from two sources:

1. The hardcoded `https://api.openai.com/v1` for the `openai` provider
2. The `api` field returned by `getCatalogProvider(providerID)` for providers registered in the models catalog

A user-configured custom provider (e.g. a local proxy, a self-hosted endpoint, or a third-party OpenAI-compatible service) has no catalog entry, so neither source produces a URL, and the function throws before it ever attempts the API call.

I'll try to bring a pull request with a fix shortly.

### Where does it happen?

Desktop (macOS)

### Version (if known)

reproduced in the version from the current main branch

Image

### Logs (optional)

```shell

```

ameshkov:
## Description

Fixes #2134 — when `small_model` is configured with a custom OpenAI-compatible provider (one not in the models.dev catalog), the small-model dispatch throws `Provider "xxx" has no known API base URL` because it only resolves base URLs from the catalog or the hardcoded OpenAI endpoint.

This PR adds a fallback that reads `provider.<id>.options.baseURL` from the user's OpenCode config, so non-catalog providers (local proxies, self-hosted endpoints, third-party OpenAI-compatible services) work correctly.

## Changes

- **`packages/web/server/lib/small-model/call.js`** — Added `readCustomProviderBaseURL(workingDirectory, providerID)` which reads the base URL for a specific provider from `opencode.json`. Extended the base URL resolution chain in `callSmallModel` to consult the custom config before falling through to the hardcoded `https://api.openai.com/v1` for `openai`. The `openai` provider itself also respects a custom `baseURL` override, matching OpenCode's own behavior.
- **`packages/web/server/lib/small-model/index.js`** — Passes `workingDirectory` through to the updated `callSmallModel`.
- **`.gitignore`** — Tracked `.vscode/launch.json` for development convenience (unrelated).

### Resolution order for API base URL:
1. Custom `provider.<id>.options.baseURL` from OpenCode config
2. `https://api.openai.com/v1` (only for the `openai` provider)
3. Catalog `api` field from models.dev (for known providers)
4. `null` → error

### DIFF
--- a/packages/web/server/lib/small-model/DOCUMENTATION.md
+++ b/packages/web/server/lib/small-model/DOCUMENTATION.md
@@ -47,7 +47,9 @@ other runtime API.
   - **Anthropic** (`type: api`): `/v1/messages` with `x-api-key`.
   - **Google** (`type: api`): `generateContent` with `x-goog-api-key`.
   - Everything else: OpenAI-compatible `/chat/completions` against the
-    provider's models.dev base URL with `Authorization: Bearer <key>`.
+    provider's base URL, resolved from (1) `provider.<id>.options.baseURL`
+    in the OpenCode config, (2) the hardcoded `https://api.openai.com/v1`
+    endpoint, or (3) the provider's `api` field from the models.dev catalog.
 - `catalog.js` — models.dev catalog via the shared in-process cache
   (`../opencode/models-metadata.js`, also serving
   `/api/openchamber/models-metadata`).
--- a/packages/web/server/lib/small-model/call.js
+++ b/packages/web/server/lib/small-model/call.js
@@ -281,21 +281,28 @@ const callCodexResponses = async ({ accessToken, accountId, modelID, prompt, sys
 };
 
 // ---------------------------------------------------------------------------
-// Custom provider base URL support
+// Custom provider configuration support
 // ---------------------------------------------------------------------------
 
-const readCustomProviderBaseURL = (workingDirectory, providerID) => {
+const readProviderConfig = (workingDirectory, providerID) => {
   try {
     const config = readConfig(workingDirectory);
     const providerCfg = config?.provider?.[providerID];
     if (!providerCfg || typeof providerCfg !== 'object') return null;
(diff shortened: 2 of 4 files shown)


btriapitsyn (approved):

jamie-dit:
Thanks for fixing the configured base-URL lookup. There is one related config-only authentication case that this patch does not appear to cover.

`callSmallModel()` still calls `getAuthEntryForProvider(auth, providerID)` and throws before `readCustomProviderBaseURL()` runs. For a custom OpenAI-compatible provider whose credential is supplied only through `provider.<id>.options.apiKey` (with no corresponding `auth.json` entry), the request therefore still fails with `No OpenCode login found for provider` even though both `apiKey` and `baseURL` are present in the merged OpenCode config.

A useful regression test would cover:

```js
config.provider.custom.options = {
  apiKey: "test-key",
  baseURL: "https://proxy.example.test/v1",
};
// No auth.json entry for `custom`.
```

The config-supplied key should be used only in the server-side in-memory auth map, paired with the configured endpoint, and must not be copied into catalog metadata, responses, or logs. It would also be useful to read the merged config layers so global and project-scoped provider definitions behave consistently.

Would you prefer to include that case in this PR, or should I send a focused follow-up after this lands?

ameshkov:
@jamie-dit nice catch, I missed this case. I think it makes sense to handle it here, the change would be trivial.

ameshkov:
1. Handled the case with apiKey in the provider configuration
2. Added a set of unit tests for `callSmallModel`

PR #2136, opened 2026-07-11
fix(chat): restore QuestionCard on reconnect; adapt MessageList to react-virtual 3.14

youfch:
Two fixes for the chat UI:

**1. Restore QuestionCard after SSE reconnection** (ChatContainer.tsx)

When the SSE connection drops and reconnects, the OpenCode server loses pending question state (question.list returns []). The tool call remains in the conversation history with status "running", so we reconstruct the QuestionRequest from the tool's input to keep the interactive A/B/C/D choice UI available after page refresh.

Also externalize Capacitor modules in the web Vite config to fix web-only builds.

**2. Adapt MessageList to @tanstack/react-virtual 3.14** (MessageList.tsx)

The version specified in package.json (3.14.5) removed several APIs:

| Removed | Replacement |
|---------|------------|
| ReactVirtualizer type | Virtualizer |
| .takeSnapshot() | .getVirtualItems() |
| .isAtEnd() | manual scrollOffset   scrollRect check |
| .itemSizeCache (private) | .measurementsCache |
| .scrollToEnd() | .scrollToIndex(n-1, {align:'end'}) |
| anchorTo: 'end' option | removed (unsupported) |


youfch:
before
PixPin_2026-07-10_23-10-28
after
PixPin_2026-07-11_18-12-21

youfch:
Start a new session and send a prompt that makes the AI ask a multiple-choice question (e.g. ask it to use the question tool)
Wait for the QuestionCard — radio buttons/checkboxes with A/B/C/D options and Submit/Dismiss buttons appear below the message list
Close the browser tab
Reopen the session via the same URL
Before fix: Only "Question Asked 1 question" text badge is visible in the message list — the interactive choice UI is gone
After fix: The full QuestionCard with all options, radio buttons/checkboxes, and Submit/Dismiss buttons is restored below the message list

PR #2143, opened 2026-07-11
fix(small-model): include no-auth providers in override picker

### LINKED ISSUE #2132: OpenCode zen models not in small models list
# Background

The Small Model override picker (Settings > Sessions > Small Model) does not list OpenCode zen models. When the user unchecks "Use default small model" and opens the model selector to pick an override, the zen provider (OpenCode's free built-in models) is absent from the list.

The picker is filtered by `allowedProviderIds` in `packages/ui/src/components/sections/openchamber/DefaultsSettings.tsx:411`, which is populated from `GET /api/small-model` -> `authenticatedProviders`. That field comes from `listAuthenticatedProviders()` in `packages/web/server/lib/small-model/index.js:135`, which only returns providers with a usable auth entry (API key or OAuth token). The zen provider has no traditional auth entry, so it is excluded, and its models never appear in the picker.

The resolver's family-priority scan in `packages/web/server/lib/small-model/resolve.js:110` also skips the zen provider for the same reason: it only iterates over providers with usable auth entries.

# Version

c2a8585

# Acceptance Criteria

## Must

- [ ] Zen provider models appear in the Small Model override picker when the zen provider is available in the OpenCode provider list
- [ ] Selecting a zen model as the small model override is persisted and honored by the resolver (`source: 'settings'` path)

## Should

- [ ] The `GET /api/small-model` `authenticatedProviders` response distinguishes "no auth needed" providers (zen) from providers that genuinely require credentials, so the picker can include zen without also surforing unauthenticated paid providers
- [ ] Documentation in `packages/web/server/lib/small-model/DOCUMENTATION.md` is updated to note that zen models are selectable as a small model override even though they are not callable via the direct small-model path

---

Created with [create-issue](https://github.com/tomzx/agents/blob/a0d3c4d/skills/create-issue/SKILL.md) via glm-5.2 (`a0d3c4d`)

bashrusakh:
## Summary

Fixes openchamber/openchamber#2132 — the small-model override picker (Settings → Sessions → Small Model) hid OpenCode's free / zen models because it was filtered by `GET /api/small-model -> authenticatedProviders`, which only includes providers with a usable `auth.json` entry. The `opencode` provider (zen / `big-pickle`, `*-free` models served by OpenCode's own infra) ships without an auth entry, so it was excluded. The resolver's `source: 'settings'` step in `resolveSmallModel()` already honors any `provider/model` ref without checking auth, so the UI was the only thing blocking selection.

This PR widens the picker's filter to include no-auth providers while keeping the strict "callable directly" set available for callers that want it.

## What changes

- **`packages/web/server/lib/small-model/index.js`** — new `NO_AUTH_PROVIDER_IDS = ['opencode']` constant (with rationale comment), new exported `listNoAuthProviders()` and `listSelectableProviders()` (deduplicated union), updated `listAuthenticatedProviders()` jsdoc to clarify the no-auth split. `listAuthenticatedProviders()` behavior is unchanged (still strict, only usable auth entries).
- **`packages/web/server/lib/small-model/routes.js`** — `GET /api/small-model` now returns three fields: the existing `authenticatedProviders` (back-compat) plus `noAuthProviders` and `availableProviders` (the union). Same response shape; just two additional fields.
- **`packages/ui/src/components/sections/openchamber/DefaultsSettings.tsx`** — picker now reads `availableProviders` first, falls back to `authenticatedProviders` for older servers, and to `undefined` (show all providers) when neither is present. Refactored from a nested ternary to an explicit `if / else if` chain (project `AGENTS.md` rule).
- **`packages/web/server/lib/small-model/DOCUMENTATION.md`** — documents the new response shape and the per-session override vs. fallback semantics. The "Known limitations" paragraph now states that the picker surfaces free / zen models, the `source: 'settings'` step honors them, while the family-priority scan and direct `POST /api/small-model/generate` still require a usable auth entry.

## Behavioral contract

- **Natural user action**: open Settings → Sessions → Small Model, uncheck "Use default", pick a model from the opencode provider.
- **Value provider**: the server (`listNoAuthProviders` returns the static `['opencode']` set; `listAuthenticatedProviders` reads `auth.json`).
- **What it changes for callers**: the `GET /api/small-model` response gains two fields. `authenticatedProviders` semantics are unchanged. The `opencode` provider id is now reachable as an override; it is *not* a new fallback the resolver can use on a vanilla setup (the resolver still returns `null` there).
- **What it does not change**: `listAuthenticatedProviders`, the resolver's family-priority scan, the direct call path, the `auth.json` format, or any other picker in the app.

## Tests

Two new test files, plus all existing small-model tests pass:

- **`packages/web/server/lib/small-model/selectable-providers.test.js`** *(new, 7 cases)* — covers `listNoAuthProviders`, `listSelectableProviders` (with / without auth entries, deduplication, unusable auth filtering), and the unchanged `listAuthenticatedProviders` contract. Host-independent via `mock.module` so it works on any machine regardless of the local `auth.json` state.
- **`packages/web/server/lib/small-model/routes.test.js`** *(new, 2 cases)* — asserts the route payload includes `availableProviders` (with the union) and `noAuthProviders` (with the no-auth set), and that `authenticatedProviders` is still present for back-compat.
- **Existing tests** (`resolve.test.js`) still pass.

## Validation

```bash
$ bun test packages/web/server/lib/small-model/
 26 pass
 0 fail
 44 expect() calls
Ran 26 tests across 3 files. [330.00ms]

$ bun run type-check:web # clean
$ bun run lint:web       # clean
```

Also ran `ocr review` (open-code-review) on the working copy: 0 issues after the nested-ternary refactor.

## Acceptance criteria

- **Must** — Zen provider models appear in the Small Model override picker: **satisfied** via `availableProviders` consumed by `ModelSelector`'s `allowedProviderIds`.
- **Must** — Selecting a zen model as the small model override is persisted and honored by the resolver: **satisfied** (resolver's `source: 'settings'` step never checks auth; was already working, the picker was the only blocker).
- **Should** — `GET /api/small-model` distinguishes "no auth needed" providers (zen) from providers that genuinely require credentials, so the picker can include zen without surfacing unauthenticated paid providers: **satisfied** via the new `noAuthProviders` field alongside `authenticatedProviders` and the union as `availableProviders`.
- **Should** — Documentation updated to note that zen models are selectable as a small model override even though they are not callable via the direct small-model path: **satisfied** in `packages/web/server/lib/small-model/DOCUMENTATION.md`.

## Right-level

The fix is at the shared contract boundary (server route + UI consumer), not at the resolver. The constant `NO_AUTH_PROVIDER_IDS` lives in one place (`index.js`); the route reads it via `listNoAuthProviders()`; the UI consumes the union via `availableProviders`. No duplication, no leak.

## Trade-off

`NO_AUTH_PROVIDER_IDS` is a hardcoded constant — currently just `['opencode']`. If OpenCode ships a second no-auth provider in the future, this is a one-line update. The alternative (runtime probe of `config.providers` against `auth.json`) would have ambiguous semantics for paid providers with missing keys, and is rejected as fragile. A narrow constant with a strong rationale comment is the correct small fix.

## Out of scope

- The resolver's family-priority scan still skips providers without usable auth. This is intentional: those providers cannot be called directly by `callSmallModel`, so they cannot be a fallback. Picking a zen override is a per-session preference, not a fallback.
- The `callSmallModel` path itself was not changed. The fix is purely at the picker boundary.

## Notes

- Marked as **draft** because this is a focused bugfix ready for review; happy to mark ready once review feedback is addressed.

### DIFF
--- a/packages/ui/src/components/sections/openchamber/DefaultsSettings.tsx
+++ b/packages/ui/src/components/sections/openchamber/DefaultsSettings.tsx
@@ -235,9 +235,20 @@ export const DefaultsSettings: React.FC = () => {
       try {
         const response = await runtimeFetch('/api/small-model', { method: 'GET', headers: { Accept: 'application/json' } });
         if (!response.ok) return;
-        const payload = await response.json().catch(() => null) as { authenticatedProviders?: unknown } | null;
-        if (!cancelled && Array.isArray(payload?.authenticatedProviders)) {
-          setSmallModelProviders(payload.authenticatedProviders.filter((id): id is string => typeof id === 'string'));
+        const payload = await response.json().catch(() => null) as { availableProviders?: unknown; authenticatedProviders?: unknown } | null;
+        if (cancelled) return;
+        // Prefer the union (auth + no-auth providers like OpenCode zen) so the
+        // picker shows zen/free models as well as authenticated ones. Fall back
+        // to authenticatedProviders for older servers that don't ship the
+        // union, and finally to undefined (picker shows everything).
+        let raw: unknown = null;
+        if (Array.isArray(payload?.availableProviders)) {
+          raw = payload.availableProviders;
+        } else if (Array.isArray(payload?.authenticatedProviders)) {
+          raw = payload.authenticatedProviders;
+        }
+        if (raw) {
+          setSmallModelProviders((raw as unknown[]).filter((id): id is string => typeof id === 'string'));
         }
       } catch {
         // leave undefined — picker falls back to showing all providers
--- a/packages/web/server/lib/small-model/DOCUMENTATION.md
+++ b/packages/web/server/lib/small-model/DOCUMENTATION.md
@@ -51,9 +51,17 @@ other runtime API.
 - `catalog.js` — models.dev catalog via the shared in-process cache
   (`../opencode/models-metadata.js`, also serving
   `/api/openchamber/models-metadata`).
-- `routes.js` — `GET /api/small-model` (resolution preview) and
-  `POST /api/small-model/generate` (`{ prompt, system?, maxOutputTokens?,
+- `routes.js` — `GET /api/small-model` (resolution preview, returns
+  `{ available, model, authenticatedProviders, noAuthProviders, availableProviders }`)
+  and `POST /api/small-model/generate` (`{ prompt, system?, maxOutputTokens?,
   model?, directory? }` → `{ text, providerID, modelID, source }`).
+  `availableProviders` is the deduplicated union of `authenticatedProviders`
+  plus the no-auth providers — what the settings override picker uses as
+  `allowedProviderIds`. `noAuthProviders` lists providers that ship without
+  requiring an auth entry (currently just `opencode` — see "Known
+  limitations" below); callers can use it to render a distinction in the
+  picker (e.g. an "OpenCode zen" badge). `authenticatedProviders` is the
+  strict "callable directly" set, kept for callers that want it.
 
 ## Registration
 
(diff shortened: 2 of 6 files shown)


tomzx:
Tested in my own fork and it works. 
Thanks!

btriapitsyn:

Issue #2094, opened 2026-07-08
Chat message visual order reversed - assistant appears above user


bashrusakh:
## Bug Description

In the chat UI, **user messages visually appear below assistant responses** even though the DOM order is correct. This is confirmed by text selection behavior: dragging down from a user message selects the assistant message **above** it.

This affects the live chat view during normal rendering — not just after prepend operations.

## Evidence

- **DOM order is correct**: User message element comes before Assistant block in the DOM tree (confirmed via text selection behavior)
- **Visual order is reversed**: Assistant message paints above the user message in the viewport
- No flex-direction column-reverse or order CSS found in chat components
- No direction rtl, writing-mode, or other CSS transforms that would explain the reversal

### Rendering Pipeline
MessageList.tsx renders two DOM siblings:
1. StaticHistoryList — virtualized historical turns (wrapped in transform: translateY())
2. StreamingTailContent — current streaming turn (no transform ancestor)

TurnItem.tsx renders user message THEN TurnAssistantBlock in DOM order.
User header uses position: sticky; top: 0; z-index: 20; position: relative.

### Key Finding
position: sticky breaks inside transform: translateY() containers per CSS spec. For virtualized turns, sticky behavior does not work. For the streaming tail (no transform ancestor), sticky should work.

The z-20 header with position: relative may create a stacking context issue when sticky behavior breaks.

## Reproduction

1. Open any chat with multiple message exchanges
2. Observe that assistant messages appear above the user messages that preceded them
3. Select text by dragging down from a user message — it will select the assistant message visually above it (confirming DOM order is correct)

## Files to Investigate

- packages/ui/src/components/chat/MessageList.tsx (virtualizer lines 1199-1218)
- packages/ui/src/components/chat/TurnItem.tsx (sticky header, z-index)
- packages/ui/src/components/chat/TurnAssistantBlock.tsx (z-0 relative)
- packages/ui/src/components/chat/ChatContainer.tsx (ChatViewport layout)
- packages/ui/src/lib/typography.ts (chat-message-column: width/padding only)

## Next Steps Needed

1. Add debug borders/colors to TurnItem user header and TurnAssistantBlock to visually confirm painting order in browser
2. Verify that position: sticky + z-20 + relative creates stacking context that causes visual reversal when sticky breaks inside transform container
3. Determine if issue is specific to streaming tail (outside virtualizer) vs virtualized history (inside transform)

yulia-ivashko:
[...]

PR #2146, opened 2026-07-11
fix(chat): action buttons hidden under next turn sticky header

### LINKED ISSUE #2095: Action buttons hidden under overlapping messages
## Bug Description

Action buttons (copy, edit, etc.) under assistant messages are **hidden/overlapped** by the next user message. The buttons exist in the DOM but are painted underneath the user message above them.

## Symptoms

- In the assistant message (current), only the text is visible — no action buttons appear below
- In the user's next message (compact), the buttons are visible
- The assistant block (z-0) is being painted UNDER the user message below it due to stacking context

## Root cause hypothesis

The  combination on the user header creates a stacking context that:
1. Breaks sticky behavior inside  (virtualizer)
2. Causes the z-0 assistant block to paint in unexpected order
3. Hides buttons that should appear under the assistant message

## Files to Investigate

- packages/ui/src/components/chat/TurnItem.tsx (sticky header with z-20)
- packages/ui/src/components/chat/TurnAssistantBlock.tsx (z-0 relative)
- packages/ui/src/components/chat/MessageList.tsx (virtualizer wrapping in transform)

## Related

Similar stacking context issue to #2094 (visual order reversed), but this specifically affects button visibility.

Image

bashrusakh:
## Summary

Fixes #2095 — action buttons (copy, edit, etc.) at the bottom of assistant messages were hidden/overlapped by the next turn's sticky user header due to a CSS stacking-context conflict. The fix is layered: the parent commit removes the explicit `z-0` isolation and lifts the action button containers to `z-30`; the follow-up commit removes a `transform: translateZ(0)` stacking context on the assistant message-body wrapper that was trapping the new `z-30` and silently undoing the first fix.

## Root cause

Two compounding stacking-context issues, both in the same parent context (ChatContainer's `relative z-0`):

1. **Explicit z-0 isolation** — the assistant block wrapper (`TurnAssistantBlock.tsx`) used `relative z-0`, pinning its descendants to z-index 0. The next turn's sticky user header (`sticky top-0 z-20`) painted on top.
2. **Implicit transform stacking context** — the assistant message-body wrapper (`MessageBody.tsx:2003` pre-follow-up) used `style={CONTAIN_LAYOUT_STYLE}`, which sets `transform: translateZ(0)`. That `transform` value is non-`none`, so per CSS Transforms Level 1 it creates a new stacking context. The `z-30` action button containers (added in the parent commit) are descendants of this wrapper, so their `z-30` is interpreted inside the wrapper's stacking context — not in the parent context. The wrapper itself has no explicit z-index, so it paints at z=0 in the parent, and the next turn's sticky header at z-20 still wins.

```
ChatContainer (relative z-0) ← parent stacking context
├── Turn N section (relative w-full — no stacking context)
│   ├── sticky header (sticky z-20)
│   └── TurnAssistantBlock (relative — no z-index, after fix)
│       └── AssistantMessageBody wrapper (transform: translateZ(0) → SC at z-0)
│           ├── action buttons (relative z-30) ← TRAPPED inside wrapper SC
│           └── text / tools / etc.
├── Turn N+1 section (relative w-full — no stacking context)
│   └── sticky header (sticky z-20) ← paints over the trapped action buttons
```

## Fix

Three minimal CSS-only changes from the parent commit, plus one right-level fix from the follow-up:

| Commit | File | Change |
|---|---|---|
| `8c121920e` | `TurnItem.tsx` | removed redundant `relative` from sticky header (`sticky` is already a positioned element) |
| `8c121920e` | `TurnAssistantBlock.tsx` | removed `z-0` from wrapper div so descendants participate in the parent stacking context |
| `8c121920e` | `MessageBody.tsx` | added `relative z-30` to 3 action button containers (inline actions, standalone actions, turn footer) |
| `0579998ab` | `MessageBody.tsx` | split `CONTAIN_LAYOUT_STYLE` → `USER_CONTAIN_LAYOUT_STYLE` (keeps `transform: translateZ(0)`) and `ASSISTANT_CONTAIN_LAYOUT_STYLE` (drops `transform`, keeps `contain: layout`); the assistant wrapper no longer creates a stacking context, so the `z-30` action buttons now apply in the parent context and paint above the next turn's `z-20` sticky header |

The user-side wrapper keeps `transform: translateZ(0)` because its trapped `z-10` action menus (`UserTextPart.tsx`, `MessageBody.tsx` action row) don't need to escape that context — they only render against siblings inside the same wrapper. This was a deliberate scoping decision flagged by the openchamber-bot review on the parent commit.

## Files changed

| File | Net diff |
|---|---|
| `packages/ui/src/components/chat/components/TurnAssistantBlock.tsx` | +1 / −1 |
| `packages/ui/src/components/chat/components/TurnItem.tsx` | +1 / −1 |
| `packages/ui/src/components/chat/message/MessageBody.tsx` | +17 / −6 |
| **Total** | **+19 / −8 across 3 files** |

No logic, event wiring, or DOM shape changes. CSS-only.

## Validation

| Check | Result |
|-------|--------|
| `bun run type-check:ui` | ✅ pass (exit 0) |
| `bun run lint:ui` | ✅ pass (exit 0) |
| `bun run dead-code` | ✅ pass (exit 0; no `MessageBody` findings) |
| OCR code review (glm-5.2) on follow-up commit | ✅ 0 comments — `Looks good to me` |
| Independent code review (native) | ✅ approve with nits — nits are non-blocking, no action required |

## Notes

- The `transform: translateZ(0)` on the user-side wrapper is a legacy GPU-compositing hint; the new `contain: layout` alone is sufficient for layout isolation. Keeping it on the user side only is a deliberate, minimal scoping decision for this fix; a future cleanup could unify the constants.
- The follow-up commit was prompted by the openchamber-bot review on the parent commit, which correctly flagged the `transform: translateZ(0)` stacking context as a non-blocker concern. The fix landed on the same PR branch rather than a separate PR.
- The `MessageList.tsx` virtualizer already explicitly avoids `transform` for the same reason — see the comment at `MessageList.tsx:1196-1203` ("a transformed ancestor becomes the sticky containing block"). This fix aligns the message-body wrapper with that established pattern.
- Related issue: #2094 (visual order reversed) — same stacking-context family of problems.

### DIFF
--- a/packages/ui/src/components/chat/components/TurnAssistantBlock.tsx
+++ b/packages/ui/src/components/chat/components/TurnAssistantBlock.tsx
@@ -9,7 +9,7 @@ interface TurnAssistantBlockProps {
 
 const TurnAssistantBlock: React.FC<TurnAssistantBlockProps> = ({ assistantMessages, renderMessage }) => {
     return (
-        <div className="relative z-0">
+        <div className="relative">
             {assistantMessages.map((message) => renderMessage(message))}
         </div>
     );
--- a/packages/ui/src/components/chat/components/TurnItem.tsx
+++ b/packages/ui/src/components/chat/components/TurnItem.tsx
@@ -18,7 +18,7 @@ const TurnItem: React.FC<TurnItemProps> = ({ turn, stickyUserHeader = true, rend
             data-scroll-spy-id={turn.turnId}
         >
             {stickyUserHeader ? (
-                <div className="sticky top-0 z-20 relative bg-[var(--surface-background)] [overflow-anchor:none]">
+                <div className="sticky top-0 z-20 bg-[var(--surface-background)] [overflow-anchor:none]">
                     <div className="relative z-10">
                         {renderMessage(turn.userMessage)}
                     </div>
--- a/packages/ui/src/components/chat/message/MessageBody.tsx
+++ b/packages/ui/src/components/chat/message/MessageBody.tsx
@@ -1740,7 +1740,7 @@ const AssistantMessageBody = React.memo(({
                 );
                 if (shouldShowStandaloneMessageActions && i === lastRenderableTextPartIndex) {
                     rendered.push(
-                        <div key={`message-actions-${messageId}`} className={INLINE_MESSAGE_ACTIONS_CLASS_NAME} data-message-actions="true">
+                        <div key={`message-actions-${messageId}`} className={cn(INLINE_MESSAGE_ACTIONS_CLASS_NAME, 'relative z-30')} data-message-actions="true">
                             <div className="flex items-center gap-1.5" data-message-action-group="true">
                                 {messageActionButtons}
                             </div>
@@ -2061,7 +2061,7 @@ const AssistantMessageBody = React.memo(({
                 </div>
                 <MessageFilesDisplay files={parts} onShowPopup={onShowPopup} />
                 {shouldRenderStandaloneActionsAfterContent && (
-                    <div className={INLINE_MESSAGE_ACTIONS_CLASS_NAME} data-message-actions="true">
+                    <div className={cn(INLINE_MESSAGE_ACTIONS_CLASS_NAME, 'relative z-30')} data-message-actions="true">
                         <div className="flex items-center gap-1.5" data-message-action-group="true">
                             {messageActionButtons}
                         </div>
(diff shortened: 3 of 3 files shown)


btriapitsyn:

PR #2147, opened 2026-07-11
feat: tiled side panel with VS Code-style splitting

Quat3rnion:
## Summary

Adds VS Code editor-group style tiling to the Side Panel (`ContextPanel`). The panel can be split into nested, resizable regions; drag any tab by its header onto a region edge to peel it into a new split. Each region is a tab group holding one or more tiles and showing one active tile.

Any Side Panel tab type tiles: chat (session), diff, file, plan, preview, and browser. Session tiles render from a single persistent iframe host positioned by CSS geometry, so splitting or resizing a region never reloads the embedded session.

## What it does

- Split the Side Panel into nested regions with resizable boundaries (arbitrary horizontal/vertical nesting).
- Drag a tab header onto a region edge to peel it into a new split; drop on another region's tab strip to move it; reorder within a strip.
- Each region is a tab group; opening a subtask/tab that already lives in the grid focuses and reveals its region instead of doing nothing.
- Session iframes stay mounted across split, resize, drag, and tab switch (no reload).

## Design

- Split engine: `allotment` (new dependency, MIT, derived from VS Code's own split-view) renders the nested resizable panes.
- Drag-to-split: existing `@dnd-kit/core`; header drag exposes edge drop zones that mutate a serializable binary split tree.
- Layout model: a per-directory binary split tree persisted with the existing Side Panel state; legacy/malformed layouts migrate to a single group.
- Persistent iframe host: session iframes are rendered once at the panel root and positioned over each region's slot via imperative CSS geometry (allotment `onChange` + `ResizeObserver`), so the browsing context is never reparented.

## Runtime scope

- Web and Desktop tile.
- Mobile/PWA stays single-active (no tiling controls).
- VS Code is out of scope (its layout does not host the Side Panel).

## Notes

- New dependency: `allotment@^1.20.5`.
- Tiling strings are added across all locales including `ja` and `fr`, covered by a key parity test.

## Testing

- `bun run type-check` and `bun run lint` pass.
- Unit tests cover the split-tree operations, drop resolution, capability gate, restore-time pruning, layout persistence, and i18n key parity.
- Manually verified in a browser: single-tab renders identically to the legacy panel (no splitter), drag-to-split produces nested regions with resizable sashes, close/collapse hoists the sibling, mobile viewport does not tile, and session iframes do not reload on split/resize.

The intermediate `feat` commit depends on a follow-up reconcile commit for the exported tile renderer, so the branch tip builds cleanly but individual commits are not each bisectable; squash-on-merge is fine.


btriapitsyn:

PR #2149, opened 2026-07-11
feat: add background auto-accept

ChangeHow:
## Context

Permission Auto-Accept currently depends on a connected UI client. This change adds an opt-in Background Auto-Accept mode so the OpenChamber server can continue applying existing per-session permission policies when no UI client is connected.

## What Changed

image

- Add a Background Auto-Accept toggle under Settings > Sessions > Session Defaults.
- Add a process-local server runtime and API for background mode and per-session policy updates.
- Apply the existing nearest-explicit-session permission policy when handling background permission requests.
- Synchronize background mode changes across connected Web, Desktop, and Mobile clients.
- Keep the setting unavailable in VS Code.
- Add localized setting copy and Settings search metadata.

## Verification

Manual permission verification:

| Background Auto-Accept | Permission Auto-Accept | Expected permission handling | Result |
| --- | --- | --- | --- |
| Off | Off | Permission remains pending for manual review | Passed |
| Off | On | Connected client automatically accepts the permission | Passed |
| On | Off | Permission remains pending because no matching auto-accept policy is enabled | Passed |
| On | On | Server automatically accepts the matching permission without a connected UI client | Passed |

Permission handling matched the expected behavior in all four combinations.

## Review Notes

- Background Auto-Accept is process-local, defaults to off, and resets after an OpenChamber server restart.
- The existing per-session Permission Auto-Accept map remains the policy source.
- VS Code intentionally does not expose or initialize this feature.


ChangeHow:
For testing detail:
1. how I create a test
using prompt `using search or find tool to search files that filename include 'hello' in my desktop and print them.`
image
2. keep website alive and open auto accept
image
3. close the website and keep auto accept openning
image
4. open background auto accept and turn **OFF** auto accept in input
image
6. open background auto accept and turn **ON** auto accept in input
image

@btriapitsyn Plz take a look

btriapitsyn:
Thank you for identifying this gap and for putting together the initial implementation and detailed testing notes. The contribution made it clear that session Auto-Accept should continue working without a connected UI.

We implemented the feature in #2158 using a different architecture: the OpenChamber server now owns and persists the per-session policy, resolves subagent inheritance, immediately reconciles existing pending requests, retries transient reply failures, and continues handling permissions after UI disconnects and server restarts. This also lets Auto-Accept remain a single session-level control rather than introducing a separate background toggle.

We are closing this PR in favor of #2158, but the idea, reproduction steps, and screenshots here were valuable in shaping the final implementation. Thank you again for the contribution.

PR #2153, opened 2026-07-12
fix(sessions): clean up deleted session references to prevent #2105 render error

### LINKED ISSUE #2105: [Bug] sessions issue

This happened to me twice right after eachother where after opening/deleting sessions in openchamber when i closed opencode and opened the next day opencode had a render error saying it couldnt find a session.

I had to open developer tools go to applications - local storage and delete the session in question for opencode to load. This also results in the opencode.db being wiped/rest so i lost all my sessions 

This happened twice in a row.

I like openchambers look and features but can not rely on to not mess anything up

### Where does it happen?

Desktop (macOS)

### Version (if known)

_No response_

_No response_

### Logs (optional)

```shell

```

bashrusakh:
## Summary

Fixes a session persistence / data-loss bug where deleted sessions survived in
localStorage and caused a `Session not found: ses_xxx` render error on the next
app cold start.

When the OpenCode DB was reset, archived, or when sessions were deleted by
another client, four cleanup gaps combined to leave stale `projectId →
sessionId` pointers in `localStorage` and stale session IDs in the child store
cache. The UI would restore one of these stale pointers on restart, try to
render the missing session, and crash inside `ChatErrorBoundary`.

## Root causes (4 bugs)

| #   | Where                                                                              | What was missing                                                                                          |
| --- | ---------------------------------------------------------------------------------- | --------------------------------------------------------------------------------------------------------- |
| 1   | `oc.sessions.activeSessionByProject` in `localStorage`                            | Not cleaned by `deleteSession` / `archiveSession`; `useSidebarPersistence` rewrote stale map on every render. |
| 2   | `loadSessions` empty-list race guard in `sync-context.tsx`                          | Preserved stale cache-seeded sessions when the server returned empty; no distinction between "live SSE" and "cache seed only". |
| 3   | `oc.dir.*.sessions` persist-cache in `localStorage`                                | `optimisticRemoveSession` only updated *live* child stores; if the owning store was evicted, the deleted session remained in cache. |
| 4   | SSE `session.deleted` event in `event-reducer.ts`                                   | Removed the session from the store list but never cleared `currentSessionId` or `activeSessionByProject`; same gap in the `session.updated` archive branch. |

## Changes

- **`packages/ui/src/sync/types.ts`** — add `sessionListFromCache: boolean` flag to `State` + `INITIAL_STATE`.
- **`packages/ui/src/sync/child-store.ts`** — seed the flag as `true` when the child store is initialized from a non-empty `localStorage` cache.
- **`packages/ui/src/sync/event-reducer.ts`** — clear the flag (mark "live SSE data has arrived") on `session.created` / `session.updated` / `session.deleted`; add `onSessionRemoved` callback that fires from `session.deleted` and the `session.updated` archive branch.
- **`packages/ui/src/sync/persist-cache.ts`** — add exported `removeSessionFromCache(directory, sessionId)`.
- **`packages/ui/src/sync/session-actions.ts`** — add `cleanupActiveSessionByProject` (exported), `cleanupPersistCacheForSession`, `cleanupSessionReferences` helper; type guard `isStringMap`; `console.warn` on write failures. Called from every delete/archive success path (including 404 cascade-delete fallbacks).
- **`packages/ui/src/sync/sync-context.tsx`** — wire `onSessionRemoved` (cleans `activeSessionByProject` for ANY deleted session + clears `currentSessionId` when it was current); race guard checks `sessionListFromCache` so empty server response clears stale cache instead of preserving it.
- **`packages/ui/src/sync/issue-2105-reproduction.test.ts`** *(new)* — 21 reproduction / fix tests covering all 4 bugs plus corrupt-map / cross-client / evicted-directory edge cases.
- **`packages/ui/src/lib/sessionStorageKeys.ts`** *(new)* — shared `PROJECT_ACTIVE_SESSION_STORAGE_KEY` constant (previously duplicated in `SessionSidebar.tsx` and `session-actions.ts`).
- **`packages/ui/src/components/session/SessionSidebar.tsx`** — import the shared constant.

## Design notes

- **Synchronous cleanup, not deferred.** `requestIdleCallback` / `setTimeout(0)` was considered for ~10 ms perf on 30-directory caches, but if the user closes the tab before the deferred task fires, the stale cache entry survives to next cold start — where the user can click the stale session in the sidebar, `loadSessions` removes it from the store, but `currentSessionId` still points at it, reproducing the exact `Session not found` error this fix prevents. 10 ms is negligible vs. the network round-trip of the delete SDK call that precedes it.
- **`cleanupActiveSessionByProject` must be sync** because it races with `useSidebarPersistence`'s React effect — if deferred, React could re-render and rewrite the stale map back to `localStorage`.
- **`onSessionRemoved` (renamed from `onSessionDeleted`)** covers both `session.deleted` and `session.updated` archive branches, and clears `activeSessionByProject` for ANY deleted session, not just the current one (because another client may have deleted a session stored in the map for a different project).
- **`isStringMap` type guard** validates only the JSON-parsed value's shape, not individual value types — the cleanup loop filters invalid entries per-key, matching the lenient reader in `SessionSidebar.tsx`.

## Validation

- `bun test packages/ui/src/sync/issue-2105-reproduction.test.ts` → 21 / 21 pass
- `bun test ./packages/ui/src/sync/` → no new failures (16 pre-existing baseline failures unrelated to this change)
- `tsc --noEmit` (packages/ui) → 0 errors
- `eslint` (packages/ui) → 0 errors

## Review

- 4 rounds of `ocr review` (10 findings total) — all material findings addressed.
- 1 round of `@reviewer` (6 findings) — all applied: callback renamed to `onSessionRemoved`; `isStringRecord` softened to `isStringMap` (filter per-entry); `child-store.ts` simplified (dropped dead `!== null` check); helper functions relocated; outer-catch now logs via `console.warn`; 3 new tests added for corrupt map, cross-client non-current delete, and evicted-directory cleanup.

## Reproducer

User actions from the original report (macOS desktop):
1. Open or delete a few sessions.
2. Close OpenChamber.
3. Open the next day — `ChatErrorBoundary` catches `Session not found: ses_xxx`.

The reproduction test file exercises the same code path in isolation.

## Follow-up (out of scope)

`useSidebarPersistence` could in principle re-write the stale React-state `activeSessionByProject` map back to `localStorage` on a future unrelated re-render. This is currently mitigated by the fact that the Map reference does not change between `cleanupActiveSessionByProject` and the next render, so the persistence effect does not re-fire. A future change to the selection logic (or a Map reference change) could reintroduce the race; a defensive guard in `useSidebarPersistence` (skip entries whose session no longer exists) would close it permanently. Not in scope for this fix.

Fixes #2105

### DIFF
--- a/packages/ui/src/components/session/SessionSidebar.tsx
+++ b/packages/ui/src/components/session/SessionSidebar.tsx
@@ -49,6 +49,7 @@ import { listProjectWorktrees, worktreeMapsEqual } from '@/lib/worktrees/worktre
 import { checkIsGitRepository } from '@/lib/gitApi';
 import type { WorktreeMetadata } from '@/types/worktree';
 import type { SortableDragHandleProps } from './sidebar/sortableItems';
+import { PROJECT_ACTIVE_SESSION_STORAGE_KEY } from '@/lib/sessionStorageKeys';
 import {
   BulkSessionDeleteConfirmDialog,
   FolderDeleteConfirmDialog,
@@ -85,7 +86,7 @@ const SHARE_OPINION_TOAST_STORAGE_KEY = 'openchamber.shareOpinionToast.dismissed
 const PROJECT_COLLAPSE_STORAGE_KEY = 'oc.sessions.projectCollapse';
 const GROUP_ORDER_STORAGE_KEY = 'oc.sessions.groupOrder';
 const GROUP_COLLAPSE_STORAGE_KEY = 'oc.sessions.groupCollapse';
-const PROJECT_ACTIVE_SESSION_STORAGE_KEY = 'oc.sessions.activeSessionByProject';
+// Imported from @/lib/sessionStorageKeys — shared with session-actions.ts (issue #2105)
 // v2 key holds composite "${renderContext}:${active|archived}:${sessionId}"
 // entries so the same session in different render contexts (e.g. "Recent"
 // and a project's root) has independent expand state. v1 held bare session
--- a/packages/ui/src/lib/sessionStorageKeys.ts
+++ b/packages/ui/src/lib/sessionStorageKeys.ts
@@ -1,0 +1,11 @@
+/**
+ * Shared localStorage key constants for session-related persistence.
+ *
+ * Extracted so the sync action layer and the sidebar component tree share a
+ * single source of truth — previously the key string was duplicated between
+ * {@link SessionSidebar} and {@link session-actions}, risking silent cleanup
+ * failure if one side changed it without the other (issue #2105 OCR review).
+ */
+
+/** Maps `projectId → sessionId` for restoring the active session per project. */
+export const PROJECT_ACTIVE_SESSION_STORAGE_KEY = "oc.sessions.activeSessionByProject"
--- a/packages/ui/src/sync/child-store.ts
+++ b/packages/ui/src/sync/child-store.ts
@@ -19,6 +19,10 @@ function createDirectoryStore(directory: string): StoreApi<DirectoryStore> {
   // paints chats instantly. Bootstrap phase-3 loadSessions overwrites with the
   // fresh list (its empty-list race guard preserves these until then).
   const cachedSessions = cached.sessions ?? INITIAL_STATE.session
+  // Track that the session list is currently the cache seed, not live data.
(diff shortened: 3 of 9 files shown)


btriapitsyn:

PR #2159, opened 2026-07-12
Fix desktop terminal text input

### LINKED ISSUE #1257: [Bug] Terminal Input not possible

What i use :
- MacBook M4 Pro
- Zsh with oh my zsh and ohmyposh installed

1. Problem 

When i open the terminal i can't type anything until i click on the "x" in the tab header of the terminal session multple times or open another tab and close the first one etc.

What hapens : 
- Bottom Bar opens with one Terminal Tab
- Terminal Content shows posh message but without colors ( Systemname, curernt folder, etc.)
- Cursor is on top left corner of terminal is blinking (wrong position, should be at the end of the displayed text)
- When i type or click enter or ESC nothing happens
- I click on the "x" in the Terminal Tab Header -> One Time or multiple time in a row
- Terminal now shows System message with colors, Cursor is now in right position, i can type and use the terminal

2. Problem

When i use the terminal in a Session X and then switch to another Session where i didnt use the terminal before i have the same problem as described above. Switching back and forth between the sessions can also resolve the problem. But i see that the terminal was somehow reloaded or something, because i see the posh/system message two times :
┌ ⠀  zsh⠀  XXXXtest⠀  main⠀                                                                                                                           0ms · 13/05/26 11:06
└[  »  » PROJECT2 ]
❯ %
┌ ⠀  zsh⠀  XXXXtest⠀  main⠀                                                                                                                           0ms · 13/05/26 11:06
└[  »  » PROJECT2 ]
❯

### Where does it happen?

Desktop (macOS)

### Version (if known)

1.10.4 (brew)

_No response_

### Logs (optional)

```shell

```

Yeti47:
## Summary

Fix desktop terminal input being dropped until a modifier key such as Alt was held.

## Changes

- Handle printable desktop keydown events before Ghostty's input handler when its composing state causes ordinary text to be discarded.
- Forward `beforeinput` `insertText` events for browsers that do not provide a usable printable keydown.
- Preserve modified shortcuts and special keys for Ghostty.
- Copy the repository `patches/` directory into the Docker dependency stage so the configured Bun patch can be applied.

## Validation

- `bun run type-check`
- `bun run lint`
- `bun run build`

Fixes #1257

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,6 +5,7 @@ WORKDIR /app
 FROM base AS deps
 WORKDIR /app
 COPY package.json bun.lock ./
+COPY patches ./patches/
 COPY packages/ui/package.json ./packages/ui/
 COPY packages/web/package.json ./packages/web/
 COPY packages/electron/package.json ./packages/electron/
--- a/packages/ui/src/components/terminal/TerminalViewport.tsx
+++ b/packages/ui/src/components/terminal/TerminalViewport.tsx
@@ -982,6 +982,7 @@ const TerminalViewport = React.forwardRef<TerminalController, TerminalViewportPr
       let restorePatchedScrollToBottom: (() => void) | null = null;
       let restorePatchedScrollbar: (() => void) | null = null;
       let restoreContainerFocus: (() => void) | null = null;
+      let desktopInputCleanup: (() => void) | null = null;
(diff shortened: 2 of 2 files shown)


btriapitsyn:

PR #2167, opened 2026-07-13
feat(settings): add compact theme previews

greghaynes:
## Summary

Replace the light and dark theme dropdowns in Settings with compact preview cards so themes are easier to browse visually.

## What Changed

- Replaced the `Light Theme` and `Dark Theme` selects in `OpenChamberVisualSettings` with dense preview grids
- Added very small theme preview cards that show:
  - sidebar color
  - main surface color
  - text contrast in both panes
  - one small accent cue from the theme primary color
- Kept the cards intentionally minimal so many themes remain visible at once
- Preserved existing settings search anchors for the light and dark theme controls

## Why

The old dropdown-only picker made it hard to compare many themes quickly.

This change makes theme browsing faster by showing the core visual differences directly in the settings UI while keeping the picker compact enough for large theme lists.

## Implementation Notes

- The preview stays intentionally simple instead of rendering a mini app mockup
- The card focuses on the central layout primitives that matter most for scanning themes:
  - sidebar
  - main pane
  - text contrast
  - accent color
- Selection state is shown by the outer card treatment rather than adding extra preview clutter

## Validation

- `bun run type-check:ui`
- `bun run lint:ui`

### DIFF
--- a/packages/ui/src/components/sections/openchamber/OpenChamberVisualSettings.tsx
+++ b/packages/ui/src/components/sections/openchamber/OpenChamberVisualSettings.tsx
@@ -3,7 +3,7 @@ import { runtimeFetch } from '@/lib/runtime-fetch';
 
 import { Tooltip, TooltipContent, TooltipTrigger } from '@/components/ui/tooltip';
 import { useThemeSystem } from '@/contexts/useThemeSystem';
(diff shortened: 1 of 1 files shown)


btriapitsyn:

PR #2169, opened 2026-07-13
feat(chat): set project default from model picker

greghaynes:
## Summary

Adds a compact project-default action directly in the chat model picker.

Each model row now exposes a small pin action that lets the user:

- switch the current chat to that model
- save that same model as the active project's default for future chats

This keeps normal model switching fast while making the "use this model for this project" workflow a single explicit action.

## What Changed

- added a per-row project-default pin action in `ModelControls`
- clicking the row still performs a normal session-only model switch
- clicking the pin performs:
  - session model switch
  - project default model update
- the active project default shows as a persistent row-level indicator
- when a model is both selected and the project default, only the existing checkmark is shown
- kept the contextual thinking keyboard hint in the footer for models with thinking variants
- added localized strings for the new row action and toast copy

## UX Behavior

- normal model selection remains one click
- "use for this chat and future chats in this project" is also one click
- no extra footer block or explanatory panel in the picker
- the project-default action is hidden in VS Code
- success toast includes a shortcut to project settings

## Validation

- `bun run type-check`
- `bun run lint`


btriapitsyn:

PR #2177, opened 2026-07-13
fix(chat): render bash output across tool states

### LINKED ISSUE #1997: [Feature Request] Consume OpenCode's ACP bash streaming events for live tool output
### Summary
Bash tool calls in the chat view currently only show output after the 
command finishes. While running, the tool card shows just the elapsed 
timer and a loading state (no partial output) even for long-running 
commands like test suites, builds, or installs. Confirmed on Live render mode + WebSocket transport + Bash auto-expand, still no incremental output.

### Upstream context
OpenCode's ACP layer now streams partial bash output via 
`tool_call_update` events instead of only sending output on completion:

- anomalyco/opencode#13338 (original feature request)
- anomalyco/opencode#14079 (merged fix, Feb 24 2026)
- anomalyco/opencode#14916 (follow-up fix, same day)

Since OpenChamber consumes OpenCode's event stream, this should unlock 
live bash output in the chat UI — but the client-side rendering still 
appears to buffer until the tool call completes.

### Expected behavior
Bash tool cards update progressively with output lines as they're 
emitted, similar to how the `task` tool already renders progressively 
with live duration/sub-tool summaries.

### Actual behavior
Card shows: command header, elapsed timer, loading spinner → then the 
full output appears all at once when the command finishes.

### Environment
- OpenChamber version: 1.13.8
- OpenCode version: 1.17.13
- Client: Web (browser on Windows)
- Server: `openchamber serve` running inside WSL (Ubuntu), accessed via 
  browser at http://localhost:3000 from Windows host.
- Browser: Edge

### Extra context (optional)

_No response_

catan271:
## Summary
- Render streamed `metadata.output` for bash tool calls across their lifecycle, including completed and error states when canonical output is unavailable.
- Prefer authoritative `state.output` whenever present, including completed bash calls.
- Normalize the selected bash output by removing terminal control sequences and applying carriage-return progress updates.
- Keep live auto-scroll limited to running bash output.
- Keep `metadata.output` scoped to bash; non-bash tools continue to use only their state output.
- Add coverage for completed-output precedence, ANSI normalization, lifecycle fallback, empty output, and non-bash exclusion.

Fixes #1997

## Verification
- `bun test packages/ui/src/components/chat/message/parts/ToolPart.test.ts` (12 pass)
- `bun run type-check:ui`
- `bun run lint:ui`
- `git diff --check`

The focused checks were run in the local workspace after applying the same patch. The PR delivery worktree did not have a separate `node_modules` installation.

## Risk Evaluation
- Risk: `2/5` (low)
- Confidence: `4.5/5`
- Shared UI rendering only; no server, API, persistence, security, dependency, or runtime bridge changes.
- The fallback remains bash-only and never replaces authoritative completed output with a potentially truncated metadata snapshot.
- Residual risk is limited to terminal-control rendering and lifecycle presentation; no live UI/e2e screenshot was added.

### DIFF
--- a/packages/ui/src/components/chat/message/parts/ToolPart.test.ts
+++ b/packages/ui/src/components/chat/message/parts/ToolPart.test.ts
@@ -1,6 +1,52 @@
 import { describe, expect, test } from 'bun:test';
 
+import { getEffectiveToolOutput, renderTerminalOutput } from './toolOutputHelpers';
 import { readTaskTagSessionIdFromOutput } from './taskSessionIdParser';
+
+describe('getEffectiveToolOutput', () => {
+    test('prefers state.output for completed tools', () => {
+        expect(getEffectiveToolOutput('bash', 'final output', 'partial output')).toBe('final output');
+    });
+
+    test('normalizes completed bash state output while preserving final-output precedence', () => {
+        expect(getEffectiveToolOutput('bash', '\u001B[32mFinal output\u001B[0m', 'partial output')).toBe('Final output');
+    });
+
+    test('falls back to metadata.output for bash tools without state output', () => {
+        expect(getEffectiveToolOutput('bash', undefined, 'partial output')).toBe('partial output');
+    });
+
+    test('normalizes bash metadata output for any lifecycle state', () => {
+        expect(getEffectiveToolOutput('bash', undefined, 'Progress 10%\r\u001B[2KProgress 90%')).toBe('Progress 90%');
+    });
+
+    test('ignores metadata.output for non-bash tools', () => {
+        expect(getEffectiveToolOutput('read', undefined, 'partial output')).toBe(undefined);
+        expect(getEffectiveToolOutput('read', 'final output', 'partial output')).toBe('final output');
+    });
+
+    test('returns undefined when bash has no output', () => {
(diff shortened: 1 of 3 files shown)


btriapitsyn:
Thanks again for putting this together and for helping drive the bash-output issue forward. Your PR captured the important direction: bash output should be visible while the tool is still running, and `state.output` should stay authoritative when it exists.

We ended up merging a maintainer-owned implementation in #2522 instead of taking this PR directly because the final fix needed a bit broader behavior and recovery coverage:

- it streams running bash output in a bounded viewport with local auto-follow / scroll-away behavior;
- it throttles live rendering and avoids sending every intermediate chunk through Shiki, while still highlighting the canonical final output;
- it handles rewritten or shortened bash snapshots without showing stale text;
- it keeps the fallback scoped to bash and preserves authoritative `state.output` precedence;
- it also fixes related OpenCode connectivity issues we found while testing, including stalled SSE recovery and avoiding managed OpenCode restart storms from rapid transport retries.

So I’m going to close this PR as superseded by #2522, not because the contribution was unwanted. It was useful input and helped validate the shape of the final fix. Appreciate the contribution.

PR #2180, opened 2026-07-13
fix(updater): make web self-updates transactional across platforms

daveotero:
## Problem

Updates initiated through the web UI or app UI while connected to a remote OpenChamber server on Windows have historically not completed reliably. In practice, I have repeatedly needed to remote into my Windows machine and run update scripts directly to finish the update.

This PR is intended to make remote self-updates dependable on Windows and smooth out the same update process across Linux and macOS.

## Why a helper is needed

The running web server cannot safely replace the package containing its own code and still be responsible for completing the update. Once the old server exits, the browser also loses the process that initiated the operation.

This PR introduces a standalone, built-in-only update helper that is copied outside the installed `@openchamber/web` package before replacement begins. The helper owns the rest of the transaction independently of the old server and survives the package directory being replaced.

For foreground services, the helper can also be launched in an independent transient systemd or launchd job so it is not terminated with the old server process group. Windows Scheduled Task restarts are retried until the replacement reports the expected version.

## Update flow

1. The server checks for an available release, validates the exact target version, verifies which package manager owns the current installation, and rejects unsupported Desktop or container-managed installs.
2. It creates a private transaction directory under the OpenChamber data directory containing a one-shot request, durable status file, diagnostic log, and copied helper.
3. It publishes an atomic maintenance marker so watchdogs, service managers, and manual `openchamber serve` attempts cannot start against a package tree while it is being replaced.
4. The server launches the helper, waits for it to confirm ownership of the transaction, returns `202 Accepted` to the UI, and then shuts down gracefully.
5. The helper waits for the old server process to exit, installs the exact target version, and verifies the installed package metadata.
6. It restarts OpenChamber using the appropriate daemon or service-manager path and waits for `/health` to report the exact requested version.
7. Only after exact-version health succeeds does the helper clear the maintenance marker and mark the transaction healthy.
8. The UI follows the durable transaction state, reconnects through the restart, and reloads only after the target version is confirmed.

## Failure and recovery behavior

The helper implements a real transaction state machine rather than treating a successful package-manager exit code as proof that the update worked. Durable states cover preparation, shutdown, installation, verification, restart, health checks, rollback, recovery, and terminal failure.

If the target installs but cannot start or does not become healthy, the helper:

- blocks replacement startup again;
- stops any unhealthy replacement process;
- reinstalls the exact previous version;
- restarts the previous version;
- verifies previous-version health before reporting `recovered-old-version`.

If automatic rollback fails, or the helper is interrupted after package mutation begins, the maintenance gate remains in a recovery-required state. This prevents systemd, launchd, Scheduled Tasks, watchdogs, or a manual start from executing a partial or unverified package tree. Running `openchamber update` becomes the explicit repair path, and the gate is cleared only after the pinned version is successfully reinstalled and verified.

## Cross-platform details

- Windows package-manager `.cmd` shims are resolved to their underlying Node entrypoints and launched with argument arrays instead of `cmd.exe /c`.
- Background Windows processes use `windowsHide: true` to avoid console-window flashes.
- Windows environment layers are merged case-insensitively so inherited npm settings cannot shadow transaction-specific values such as `npm_config_prefix`.
- npm, pnpm, Yarn, and Bun updates are pinned to an exact package version.
- Daemon installs restart through the OpenChamber CLI.
- systemd and launchd foreground installs rely on their supervisor restart behavior while the helper runs independently.
- Windows startup installs retry the Scheduled Task until exact-version health succeeds.
- Electron Desktop remains on the native `electron-updater` path and is intentionally excluded from this web-package updater.
- Container installs remain image-managed and are rejected with an explicit response.

## Security and observability

- Transaction files are stored outside the replaceable package directory with private file modes.
- The inherited environment is kept in a one-shot request that the helper deletes immediately after reading.
- If the request cannot be deleted, the helper scrubs its contents and fails closed before shutting down the server.
- Status responses contain sanitized transaction data rather than restart environment values.
- Package-manager and helper diagnostics are written to a transaction-specific `update.log`.
- Concurrent web updates are rejected through an atomic maintenance reservation.

## UI behavior

- The client polls both exact-version `/health` and the durable transaction status instead of using update availability as a proxy for success.
- A healthy restored old server is shown as recovery, not a successful update.
- Polling is cancellable when the dialog closes, the component unmounts, or the active runtime changes.
- The maximum polling window is 30 minutes, enough to cover install and rollback timeouts without leaving a 50-minute background loop.
- If automatic recovery cannot complete, the dialog presents the terminal update command as the repair fallback.

## Testing and CI

The updater now has its own GitHub Actions matrix on Ubuntu, macOS, and Windows. The integration suite installs local fixture tarballs into an isolated global npm prefix, including paths with spaces, without publishing a real release or modifying the runner's normal global installation.

Coverage includes:

- exact-version replacement and health verification;
- failed-install rollback to the previous version;
- target startup failure and recovery;
- helper survival after the preparing parent exits;
- helper termination during installation and durable recovery gating;
- installer process-tree timeout cleanup;
- passive systemd/launchd-style supervisor restarts;
- Windows Scheduled Task retries;
- Windows package-manager and environment handling;
- one-shot request scrubbing and secret redaction;
- client transaction polling, exact-version success, recovery, and cancellation.

Current PR checks pass on Windows, macOS, and Ubuntu, along with the repository build, type-check, and lint jobs.

## Expected outcome

A user connected to a remote OpenChamber server should be able to initiate an update from the UI and have the host safely install, restart, reconnect, or recover without requiring interactive remote access to run update scripts manually.


btriapitsyn:

Issue #2188, opened 2026-07-13
[Bug] Desktop session switching becomes very slow with many un-archived sessions (v1.16 regression)


hbzheng3:
## Issue Description / 问题描述

After upgrading from v1.15.0 to v1.16.0, the desktop app becomes very laggy when switching between sessions, especially when I have dozens of un-archived sessions in the sidebar. The lag is noticeable even though I'm only actively using a few sessions.

从 v1.15.0 升级到 v1.16.0 后，desktop 应用在切换 session 时变得非常卡顿，尤其是当 sidebar 中有几十个未归档的 session 时。即使我实际只在使用几个 session，卡顿也很明显。

## Steps to Reproduce / 复现步骤

1. Use OpenChamber Desktop v1.16.0
2. Accumulate 30-50+ un-archived sessions across 3-5 different projects/worktrees
3. Click to switch between sessions in the sidebar
4. Observe noticeable UI lag/freeze during the switch (1500ms-3000ms+)

## Expected Behavior / 预期行为

Session switching should be instant (<50ms) regardless of the total number of un-archived sessions, as long as I'm only viewing/switching between a few active ones.

切换 session 应该是即时的（<50ms），无论有多少未归档 session，只要我实际只在少数几个活跃 session 之间切换。

## Actual Behavior / 实际行为

- v1.15.0: Session switching is instant ✅
- v1.16.0: Session switching has noticeable lag (1500ms-3000ms) when 30+ un-archived sessions exist ❌

The lag appears to scale with the total number of un-archived sessions, not just the active ones.

卡顿程度似乎与未归档 session 的总数成正比，而不仅仅是活跃 session 数量。

## Environment / 环境信息

- **Platform**: Desktop (Windows / macOS)
- **Version**: 1.16.0 (regression from 1.15.0)
- **Total un-archived sessions**: ~40-50 个
- **Projects/worktrees**: 分散在 5 个不同的项目
- **Actively used sessions**: 只有 5-6 个
- **Regression confirmed**: v1.15.0 正常，v1.16.0 出现卡顿

## Root Cause Analysis (from code investigation) / 根因分析

After investigating the codebase, I found several O(N-sessions) operations on the session-switch hot path in v1.16:

通过代码追踪，发现 v1.16 在 session 切换热路径上存在多个 O(N) 操作：

### 1. `useAllLiveSessions()` — Global session aggregation

**Location**: `packages/ui/src/components/session/SessionSidebar.tsx:320`

```typescript
[...]

PR #2192, opened 2026-07-13
perf(sidebar): remove redundant path normalization and share directory-match cache (#2173)

### LINKED ISSUE #2173: [Performance] v1.16.0 project-directory matching blocks renderer for 5 seconds on session updates
### What is wrong?

On the macOS desktop app, opening a new-session draft blocks the renderer main thread for about 5.5 seconds. Creating/archiving sessions and sending messages also show shorter stalls because they update the same session/sidebar state.

This is a renderer CPU long task, not an API or network delay.

### Environment

- OpenChamber: `v1.16.0`
- Runtime: macOS desktop app
- Profiling tool: Electron Developer Tools Performance panel

### Steps to reproduce

1. Open an existing session in the macOS desktop app.
2. Open Developer Tools with `Cmd+Option+I`.
3. Start a Performance recording.
4. Click **New Session** to open a draft.
5. Stop the recording after the UI becomes responsive.

### Actual behavior

The click interaction blocks the renderer main thread for `5,467.6 ms`.

Performance summary:

| Activity | Time |
|---|---:|
| Scripting | `5,452 ms` |
| Rendering | `434 ms` |
| Painting | `58 ms` |

Bottom-up profile:

| Minified function | Source mapping | Self time | Total time |
|---|---|---:|---:|
| `Jt` | `normalizePath` | `4,572.1 ms` (91.2%) | `4,622.4 ms` |
| `Qt` | `isSessionRelatedToProject` | `45.0 ms` | `4,807.5 ms` (95.9%) |
| `xt` | `findBestProjectDirectoryMatch` | `23.8 ms` | `4,678.1 ms` (93.3%) |
| `Ee` | `isPathWithinProject` | `152.7 ms` | `3,059.7 ms` (61.8%) |

The mappings above were verified against the matching `v1.16.0` release chunks:

- `useAppFontEffects-DkPuAMXD.js`
- `SessionSwitcherDropdown-CWWrCFzr.js`

### Suspected root cause

`findBestProjectDirectoryMatch()` in `packages/ui/src/components/session/sidebar/utils.tsx` iterates every known project/worktree directory for each session.

Inside that loop it:

1. Calls `normalizePath(candidate)` even though `knownDirectories` was built from normalized paths.
2. Calls `isPathWithinProject(value, normalizedCandidate)`.
3. `isPathWithinProject()` normalizes both arguments again.

The relation check is then repeated from multiple sidebar paths:

- `useProjectSessionLists.ts`
- `useArchivedAutoFolders.ts`
- `useSessionFolderCleanup.ts`

This produces behavior close to `projects × sessions × (projects + worktrees)`, with repeated regex/string normalization in the innermost loop.

The `knownDirectories` / `findBestProjectDirectoryMatch` path was added between `v1.15.0` and `v1.16.0`; the same block is not present in `v1.15.0`, so this appears to be a `v1.16.0` performance regression.

### Expected behavior

Opening a new-session draft should not perform project/session reassignment work or produce a multi-second renderer long task. Existing nested-project/worktree grouping behavior should remain correct.

### Suggested direction

- Treat `knownDirectories` as already normalized and avoid normalization inside the candidate loop.
- Cache the best project-directory match by resolved session directory for each known-directory set.
- Prefer assigning each session to its best project once instead of rescanning it separately for every project and every sidebar hook.
- Add a regression benchmark with many projects, worktrees, active sessions, and archived sessions.

### Related

- #2012 involved broad/nested project roots and sidebar folder processing; this may be the correctness change that exposed the expensive matching path.
- Distinct from #2096: this profile is dominated by renderer-side synchronous path matching, with no network request required to reproduce the New Session stall.

### Screenshots / recordings

A Performance screenshot showing the `5,467.6 ms` click interaction and the Bottom-up timings is available. The raw trace is not attached because it may contain local paths and session metadata.

bashrusakh:
Fixes #2173

## Summary

Fixes a v1.16.0 session-sidebar renderer performance regression where `findBestProjectDirectoryMatch()` re-normalized already-normalized candidates on every call, and `isPathWithinProject()` re-normalized both arguments, multiplied across `sessions × projects × 3 sidebar hooks`.

This hotfix removes redundant normalization and shares a single per-render directory-match cache across the three consumers without changing assignment semantics or redesigning the sidebar data flow.

## What changed

- `packages/ui/src/components/session/sidebar/utils.tsx`
  - Added internal `isPathWithinNormalizedProject` helper for pre-normalized inputs.
  - `findBestProjectDirectoryMatch` now accepts an optional `Map<string, string | null>` cache and skips per-candidate `normalizePath` calls (all candidates are already normalized via `collectKnownProjectDirectories`).
  - Public `isPathWithinProject` is unchanged.
  - `isSessionRelatedToProject` threads the cache through.

- `packages/ui/src/components/session/SessionSidebar.tsx`
  - Computes `knownProjectDirectories` once with `useMemo`.
  - Creates a fresh `directoryMatchCache` whenever `knownProjectDirectories` identity changes, so stale best-match entries are discarded automatically.
  - Passes both to `useProjectSessionLists`, `useArchivedAutoFolders`, and `useSessionFolderCleanup`.

- Three sidebar hooks
  - Removed duplicated `collectKnownProjectDirectories()` `useMemo` calls.
  - All `isSessionRelatedToProject` calls now use the shared cache.

- `packages/ui/src/components/session/sidebar/utils.test.ts`
  - Added coverage for Windows normalized paths, root `/` matching, nested-project precedence, cache reuse, and cache freshness after directory-set changes.

## Performance

Measured on a synthetic dataset matching the user's scale (15 projects, 67 worktrees, 14,561 sessions), running the sidebar matching loop across all sessions, all projects, and 3 consumers:

| Variant | Avg | Min | Max |
|---|---|---|---|
| Before (no cache, redundant normalization) | **7,523.9 ms** | 4,743.7 ms | 10,022.9 ms |
| After (shared directory-match cache, no redundant normalization) | **450.6 ms** | 395.6 ms | 477.2 ms |

**~16.7× faster** on this worst-case scan, saving ~7 seconds per full evaluation. Actual React render cost is lower because the hooks do not run the full cross-product on every frame.

## Validation

- `bun test packages/ui/src/components/session/sidebar/utils.test.ts` — 15 pass, 0 fail
- `bun run --filter @openchamber/ui type-check` — pass
- `bun run --filter @openchamber/ui lint` — pass

## Follow-up

This is an intentional quick hotfix. A separate follow-up PR should pre-assign every session to its owning project once per render (Option A discussed in #2173): build a `Map<sessionId, { resolvedDirectory, bestMatch }>` in `SessionSidebar`, share it with the hooks, and replace the remaining `sessions × projects` filters with `O(1)` lookups. That will drop the worst-case per-render cost from hundreds of ms to single-digit ms and remove the current per-render `Map` allocation entirely.

### DIFF
--- a/packages/ui/src/components/session/SessionSidebar.tsx
+++ b/packages/ui/src/components/session/SessionSidebar.tsx
@@ -66,6 +66,7 @@ import {
 } from './sidebar/activitySections';
 import { useSessionPinnedStore } from '@/stores/useSessionPinnedStore';
 import {
+  collectKnownProjectDirectories,
   compareSessionsByPinnedAndTime,
   formatProjectLabel,
   normalizePath,
@@ -937,6 +938,20 @@ export const SessionSidebar: React.FC<SessionSidebarProps> = ({
       }>;
   }, [projects]);
 
+  const knownProjectDirectories = React.useMemo(
+    () => collectKnownProjectDirectories(normalizedProjects, availableWorktreesByProject, isVSCode),
+    [normalizedProjects, availableWorktreesByProject, isVSCode],
+  );
+
+  // Create a fresh directory-match cache whenever the set of known
+  // directories changes. The dependency is intentionally the identity
+  // of the set so stale best-match lookups are discarded.
+  const directoryMatchCache = React.useMemo(
+    () => new Map<string, string | null>(),
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [knownProjectDirectories],
+  );
+
   const normalizedProjectPaths = React.useMemo(
     () => normalizedProjects.map((project) => project.normalizedPath),
     [normalizedProjects],
@@ -1000,9 +1015,11 @@ export const SessionSidebar: React.FC<SessionSidebarProps> = ({
     sessions,
     archivedSessions,
     normalizedProjects,
+    knownProjectDirectories,
     isVSCode,
     availableWorktreesByProject,
     cleanupSessions,
+    directoryMatchCache,
   });
 
   const { getSessionsForProject, getArchivedSessionsForProject } = useProjectSessionLists({
@@ -1010,7 +1027,8 @@ export const SessionSidebar: React.FC<SessionSidebarProps> = ({
     sessions,
     archivedSessions,
     availableWorktreesByProject,
-    normalizedProjects,
+    knownProjectDirectories,
+    directoryMatchCache,
   });
 
   useArchivedAutoFolders({
@@ -1024,6 +1042,8 @@ export const SessionSidebar: React.FC<SessionSidebarProps> = ({
     createFolder,
     addSessionToFolder,
     cleanupSessions,
+    knownProjectDirectories,
+    directoryMatchCache,
   });
 
   // Keep last-known repo status to avoid UI jiggling during project switch
--- a/packages/ui/src/components/session/sidebar/hooks/useArchivedAutoFolders.ts
+++ b/packages/ui/src/components/session/sidebar/hooks/useArchivedAutoFolders.ts
@@ -2,7 +2,6 @@ import React from 'react';
 import type { Session } from '@opencode-ai/sdk/v2';
 import type { WorktreeMetadata } from '@/types/worktree';
 import {
-  collectKnownProjectDirectories,
   dedupeSessionsById,
   getArchivedScopeKey,
   isSessionRelatedToProject,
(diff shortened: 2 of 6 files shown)


btriapitsyn:
Thank you for preparing this hotfix and for including a report-sized benchmark. Removing redundant normalization and sharing the match cache was a sound, low-risk improvement.

We have now landed a more complete version in `main` via [`5cb45fcbf`](https://github.com/openchamber/openchamber/commit/5cb45fcbf), so this PR is superseded. The main difference is that the merged implementation removes the remaining project-by-session cross-product entirely instead of caching within it:

- one ownership index resolves each unique directory and parent path once;
- all sidebar hooks read direct project/worktree/scope buckets;
- the old matching API is removed from the render path;
- active/archive dedupe and failure-safe topology reconciliation are retained;
- related live-session, per-row status, and permission-lineage fanout from #2188 is addressed in the same coherent change.

At the 15-project / 67-worktree / 14,561-session scale, the merged ownership build averages **3.81 ms**, versus approximately **450 ms** for the cache-only version described here.

I am closing this PR as superseded by the implementation in `main`. Thank you again for the careful analysis and for validating the first optimization step.

PR #2212, opened 2026-07-13
feat(devtooling): Add worktree dev port helper and concise local workflow docs

greghaynes:
This change makes it practical to run multiple local OpenChamber worktrees in parallel without manual port bookkeeping.

## What changed
- Added `scripts/worktree-dev.mjs` to derive per-worktree HMR/UI and API port assignments, print current assignments, list detected worktrees, and run `dev:web:hmr` with the assigned ports.
- Added root scripts in `package.json`:
  - `bun run dev:worktree`
  - `bun run dev:worktree:ports`
  - `bun run dev:worktree:list`
- Updated `CONTRIBUTING.md` with a minimal workflow for running from inside an existing worktree.

## Notes
- Port assignment is deterministic per worktree path and includes collision signaling with explicit override support via `OPENCHAMBER_HMR_UI_PORT` and `OPENCHAMBER_HMR_API_PORT` when needed.
- Docs were simplified to focus on everyday usage.


btriapitsyn:
Thanks — the pain is real: `dev:web:hmr` hardcodes 5180/3902 with `--strictPort`, so a second worktree can't start without manual env exports. But this solution is much larger than the problem, and the guarantee it sells is leaky: collisions are only detected between identical hash slots, while the UI (5180+) and API (3902+) ranges overlap each other (one worktree's API port can land on another's UI port with no warning), and no port is ever checked for actually being free — a slot pointing at a port used by anything else still fails at startup. That leaves us maintaining a hash/slot/step/env-matrix subsystem, three new root commands, and docs, without the property it exists for.

The shape we'd merge: a small change inside the existing `scripts/dev-web-hmr.mjs` — when `OPENCHAMBER_HMR_UI_PORT`/`OPENCHAMBER_HMR_API_PORT` are unset and the defaults are busy, probe for a free pair, print the chosen ports, and start. No new commands, no new configuration. If you'd like to send that, it's very welcome.

PR #2214, opened 2026-07-14
fix(desktop): keep local instance label during boot

e-n-0:
## Summary

On packaged Electron desktop, the header briefly shows the generic `Instance` label at startup even when the current runtime is Local. The instance picker itself already reads `Local`, so the mismatch is visible at every boot until the label settles.

## Root cause

Label resolution in `Header.tsx` first checked `isDesktopLocalOriginActive()`, then matched the current runtime API base against `__OPENCHAMBER_LOCAL_ORIGIN__` (posted by the Electron preload). On packaged builds the renderer can begin resolving the header before that preload hint has been posted, so both checks miss and resolution falls through to the generic `Instance` branch.

## Fix

Move resolution into a pure `resolveDesktopInstanceDisplay` helper and apply the checks in the following order:

1. Host-specific runtime keys (`host:*`). SSH, Relay, and direct-URL hosts keep their configured label.
2. Configured host URLs. Any host whose `apiUrl` matches the current runtime API base wins over the local fallback, including relay identities whose virtual API base is loopback.
3. Authoritative local origin. Matches the preload hint once it has arrived.
4. Packaged Electron loopback fallback. Treats a loopback API base on the `openchamber-ui://app` page as `Local` when nothing above resolved. This is what fixes the boot race; remotes are already handled by (1) and (2), so they can't be mislabeled here.

If none of the above resolves, the helper returns `generic` and the header shows the localized fallback string.

The consumer also stops overwriting the current label with `Local` when the host-config fetch fails, so a transient failure no longer replaces a valid remote label.

## Validation

- 49 focused tests pass, including packaged boot fallback, host precedence, relay-over-loopback identity, and unconfigured direct-E2EE loopback.
- 91-test extended QA suite green.
- `bun run type-check` and `bun run lint` clean in `packages/ui`.
- Web assets build clean.
- LSP diagnostics clean on changed files.
- Packaged Electron CDP reproduction: the header reads `Local` from the first paint on cold boot.

### DIFF
--- a/packages/ui/src/components/layout/Header.tsx
+++ b/packages/ui/src/components/layout/Header.tsx
@@ -68,13 +68,14 @@ import { forceKillTerminal } from '@/lib/terminalApi';
 import { useTerminalStore } from '@/stores/useTerminalStore';
 import { ProjectActionsButton } from '@/components/layout/ProjectActionsButton';
 import { SessionSwitcherDropdown } from '@/components/session/SessionSwitcherDropdown';
-import { canUseElectronDesktopIPC, invokeDesktop, isDesktopLocalOriginActive, isDesktopShell, isVSCodeRuntime, startDesktopWindowDrag, type UpdateInfo } from '@/lib/desktop';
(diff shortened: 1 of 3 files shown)


btriapitsyn:

PR #2215, opened 2026-07-14
fix(quota): reject malformed provider responses

e-n-0:
## Summary

The quota store now validates provider responses before state insertion. If a provider returns a malformed payload, the store preserves the provider identity and returns a failure result instead of potentially corrupting the state.

## Root cause

The `fetchProviderQuota` method in `useQuotaStore.ts` parsed the JSON response but did not verify the object shape before updating the results collection. This could lead to UI inconsistencies or runtime errors if a provider API returned an unexpected structure.

## Fix

- Added validation logic to `fetchProviderQuota` to ensure responses match the `ProviderResult` type.
- Updated error handling to return a consistent fallback result when validation fails.
- Added comprehensive tests for the quota store covering valid, malformed, and error responses.

## Validation

- 13 focused tests pass in `useQuotaStore.test.ts`, covering valid responses, malformed JSON, and API errors.
- `bun run type-check` and `bun run lint` clean in `packages/ui`.
- Verified that malformed responses no longer cause UI crashes in the usage dashboard.

### DIFF
--- a/packages/ui/src/stores/useQuotaStore.test.ts
+++ b/packages/ui/src/stores/useQuotaStore.test.ts
@@ -1,0 +1,323 @@
+import { afterEach, beforeEach, describe, expect, test } from 'bun:test';
+import type { ProviderResult, QuotaProviderId } from '@/types';
+import {
+  resetQuotaRuntimeFetchForTests,
+  setQuotaRuntimeFetchForTests,
+  useQuotaStore,
+} from './useQuotaStore';
+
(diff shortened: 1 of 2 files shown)


e-n-0:
Answering comment of the bot:
Fixed in commit e23b72a0. Settings now use runtimeFetch directly, so the quota test hook is only used for quota requests.
All 13 tests passed locally.

btriapitsyn:

PR #2217, opened 2026-07-14
test(sync): isolate browser globals

e-n-0:
## Summary

Isolates browser global overrides in sync tests using scoped stubs. This ensures that modifications to global state are correctly restored after each test, preventing cross-test interference.

## Problem

The `event-pipeline` tests simulate network and visibility changes by overriding `navigator.onLine` and `document.visibilityState`. Since these properties are often read-only, direct assignment can fail or leave the global state inconsistent for subsequent tests, leading to flakiness in the test suite.

## Fix

- Added a `browser-global-stubs.js` helper that uses `Object.defineProperty` to safely swap global values.
- Implemented property descriptor preservation to restore original values after test completion.
- Updated `event-pipeline` tests to use the stub helper in setup and teardown blocks.

## Validation

- All 26 tests in the `event-pipeline` suite pass.
- `bun run type-check` and `bun run lint` clean.
- Verified that `navigator.onLine` and `document.visibilityState` are restored to original values after test execution.

### DIFF
--- a/packages/ui/src/sync/__tests__/browser-global-stubs.js
+++ b/packages/ui/src/sync/__tests__/browser-global-stubs.js
@@ -1,0 +1,27 @@
+export function createBrowserGlobalStubScope() {
+  const originalDescriptors = new Map();
+
+  const install = (property, value) => {
+    if (!originalDescriptors.has(property)) {
+      originalDescriptors.set(property, Object.getOwnPropertyDescriptor(globalThis, property));
+    }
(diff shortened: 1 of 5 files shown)


e-n-0:
Answering comment of the bot:
The token seed is only for WebSocket tests, since the others use SSE. I added a comment in d9453f7c to explain that.
I kept the emitClose payload and delete fallback to make the mock more accurate and ensure cleanup is safe.

btriapitsyn:

PR #2220, opened 2026-07-14
feat(providers): redesign provider connect ui

greghaynes:
## Summary
- replace the add-provider dropdown with a searchable provider browser and dedicated configuration pane
- switch to a single-pane selector/detail flow earlier on narrower screens and keep the two panes independently scrollable
- fix settings search targeting for provider auth on narrow screens while the add-provider flow is active
- stop manually reopening OAuth URLs after authorize so the client no longer triggers duplicate browser launches; opening now relies on the authorize flow itself, with the returned URL still available for manual use

## Testing
- bun run type-check:ui
- bun run lint:ui


btriapitsyn:

PR #2224, opened 2026-07-14
fix(chat): pin message target during async submit

callMeExiao:
## Summary

- pin the selected session or draft when chat submission begins so async response-style and snippet preparation cannot reroute the message
- deduplicate draft materialization and preserve newer draft/session selections
- prevent captured drafts from inheriting a newer draft's folder target
- add regression coverage for stale session routing and overlapping draft materialization

## Validation

- `bun test packages/ui/src/sync/__tests__/issue-2039.test.ts` (8 passed)
- `bun test packages/ui/src/sync/session-actions.test.ts` (22 passed)
- `bun test packages/ui/src/hooks/useQueuedMessageAutoSend.test.ts` (9 passed)
- `bun test packages/ui/src/components/session/sidebar/hooks/useProjectSessionSelection.test.ts` (8 passed)
- `bun run dead-code` (existing repository findings only)

### DIFF
--- a/packages/ui/src/components/chat/ChatInput.tsx
+++ b/packages/ui/src/components/chat/ChatInput.tsx
@@ -1897,8 +1897,6 @@ const ChatInputComponent: React.FC<ChatInputProps> = ({ onOpenSettings, scrollTo
             }
         }
 
-        const sendMessageOptions = delivery ? { delivery } : undefined;
-
         // Build the primary message (first part) and additional parts
         let primaryText = '';
         let primaryAttachments: AttachedFile[] = [];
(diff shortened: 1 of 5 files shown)


callMeExiao:
Addressed the stale `pendingWorktreeRequestId` race in `bc800367`.

When the captured and live drafts share the same `materializationKey`, materialization now reconciles the live pending request ID and resolved worktree directory before waiting. This avoids waiting on an entry already removed from the pending-request map while keeping newer/different drafts isolated.

Added a regression test for the exact sequence: capture pending draft -> resolve request and update live draft -> materialize captured draft.

Validation:
- `issue-2039.test.ts`: 9 passed
- `session-actions.test.ts`: 22 passed
- queued-send tests: 9 passed
- project-selection tests: 8 passed
- UI type-check and lint passed

PR #2232, opened 2026-07-14
fix(git): prevent index.lock contention with serialized status reads

### LINKED ISSUE #2229: [Bug] Unable to perform git commit
## Description
Repositories with git hooks (pre-commit, post-checkout, etc.) frequently hit git lock contention when OpenChamber is running. The .git/index.lock (and sometimes .git/HEAD.lock) files remain held by OpenChamber background processes, blocking all manual git operations (git add, git commit, git status, etc.) in the terminal until OpenChamber is closed.

Closing OpenChamber immediately resolves the lock — confirming the tool is the lock holder.

## Root cause hypothesis
OpenChamber's background polling (git status, repo state checks, file reference stat checks per #2096) spawns concurrent git operations. When hooks are involved, a hook may take longer than expected, causing:

A git command spawns and triggers a hook
Before it completes, another background poll spawns another git command on the same repo
The second command finds .git/index.lock held by the first and retries/fails
In some cases the lock is never properly released by the hung process
This is especially bad with hooks that do non-trivial work (linters, formatters, secret scanners, etc.).

## Reproduction
Have a git repo with any hook enabled (e.g. a pre-commit hook that runs npx lint-staged or similar)
Open OpenChamber and add the repo to the workspace
Wait for background polling to kick in (happens every few seconds)
Try running git status or git commit in the terminal
Observe fatal: Unable to create '<repo>/.git/index.lock': File exists. or similar error
Kill OpenChamber — lock releases immediately
Workaround
None besides closing OpenChamber, which defeats the purpose of using it.

### Environment
OpenChamber: [your version]
OpenCode: [your version]
OS: macOS
Repo: any repo with active git hooks

## Related
#2096 — Node.js event loop starvation from background git/fs polling
#1393 — Runaway polling loop fires identical git rev-parse requests

### Where does it happen?

Desktop (macOS)

### Version (if known)

latest

_No response_

### Logs (optional)

```shell

```

bashrusakh:
# Prevent Git index-lock contention

Fixes #2229 by serializing Git status reads and index-mutating operations per shared `git-common-dir`.

## Problem

Background status polling could overlap with staging, commits, or other Git operations against the same repository. Concurrent Git subprocesses could contend for the index and surface `.git/index.lock` failures.

## Solution

- Key the existing Git-operation queue by `git rev-parse --git-common-dir` so repository subdirectories and linked worktrees use the same queue.
- Queue `getStatus()` with index-mutating Git operations.
- Run the status diff-stat commands sequentially rather than concurrently.
- Preserve independent execution for unrelated repositories.

`getStatus()` does not deduplicate callers into a shared result: the queue intentionally serializes each status read with other Git work to preserve correct worktree-specific status.

## Coverage

- Concurrent `getStatus()` calls return consistent status.
- A status read waits behind staging.
- A primary repository and linked worktree share queue serialization.
- A monorepo subdirectory resolves and returns repository status correctly.
- Unrelated repositories can run status reads concurrently.

## Validation

```bash
cd packages/web
bunx vitest run --reporter verbose server/lib/git/service.test.js
# 49 passed

node --check server/lib/git/service.js
bun run type-check
bun run lint
bun run build
```

The build completed with existing Vite font and chunk-size warnings.

## Scope

- No polling-frequency or UI changes.
- No new services, retry policies, or client cache changes.
- No public API changes.

### DIFF
--- a/packages/web/server/lib/git/service.js
+++ b/packages/web/server/lib/git/service.js
@@ -17,6 +17,7 @@ const SIMPLE_GIT_SAFE_BINARY_PATTERN = /^([a-z]:)?([a-z0-9/.\\_~-]+)$/i;
 const SIMPLE_GIT_UNSAFE_BINARY_WARNING = 'Invalid value supplied for custom binary, restricted characters must be removed';
 const REMOTE_EXISTENCE_CACHE_TTL_MS = 30_000;
 const gitIndexMutationQueues = new Map();
+const getStatusInFlight = new Map();
 
 const WORKTREE_BOOTSTRAP_PENDING = 'pending';
 const WORKTREE_BOOTSTRAP_READY = 'ready';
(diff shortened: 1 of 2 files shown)


btriapitsyn:

Issue #2234, opened 2026-07-14
[Feature Request] Add end-to-end encryption to managed tunnels


e-n-0:
This adds application-layer E2EE over the existing managed Cloudflare tunnel. It does not route managed connections through the OpenChamber Relay. Cloudflare is used for transport only; encryption terminates at the OpenChamber client and host.

This work builds on the foundations in #2087 and #2103. It is distinct from the self-hosted Relay request in #2161, as it focuses on the managed tunnel path rather than the Relay infrastructure.

<details>
<summary>Implementation and rollout plan</summary>

The implementation is organized into four waves:

1. **Foundations**: transport-neutral seams, profile configuration, and a dormant hardened endpoint.
2. **Integration**: pairing/restore logic and settings UI.
3. **Release**: production registration, host-only administration, and fail-closed policy.
4. **Gates**: automated validation, managed-tunnel canary, and a privacy/scope audit.

</details>

## PR stack
All six PRs are filed as drafts, rebased on current main, and mergeable. They should be merged in order.

- [ ] #2235: refactor(relay): add transport-neutral encrypted session seams
- [ ] #2236: feat(tunnels): add managed direct E2EE profile configuration
- [ ] #2237: feat(direct-e2ee): add dormant hardened server endpoint
- [ ] #2238: feat(clients): support direct E2EE pairing and restore
- [ ] #2239: feat(settings): expose managed direct E2EE controls
- [ ] #2240: feat(server): ship managed direct E2EE

## Standalone bugs
The following are standalone bugs discovered during implementation. Copies of #2215, #2216, and #2217 were removed during the restack. #2238 depends on #2216. #2215 and #2217 remain independent, as does #2214.

- #2214: Packaged Electron local instance label race.
- #2215: Quota response validation.
- #2216: SDK global-event runtime transport.
- #2217: Browser-global test isolation.

e-n-0:
1. All six drafts are filed: #2235-#2240. They are rebased and mergeable.
2. Duplicate patches were removed during the restack. No separate cleanup PRs are needed.
3. #2087 and #2103 are merged. #2238 depends on #2216. #2215 and #2217 are independent.
4. There are no fixed dates. We will merge in order after review. The final layer enables the route, followed by managed-tunnel canary and fail-closed checks before rollout.

PR #2235, opened 2026-07-14
refactor(relay): add transport-neutral encrypted session seams

e-n-0:
## Summary
This is the base layer for managed direct E2EE. It extracts transport-neutral encrypted-session handling and adds the seams needed for an injected outer transport.

## Stack
- Layer 1 of 6
- Part of #2234
- Status: Ready for review

## What changed
- Extracted encrypted-session logic from the hosted Relay implementation.
- Added transport-neutral injection points.
- Hardened header stripping, stream limits, and fragment accounting.
- Fixed request terminator ordering for GET, HEAD, and empty-body methods.
- Bounded completed-stream tombstones by TTL, count, frame count, and bytes.
- Kept the hosted Relay wire format unchanged.

## Review follow-up
- Tunnel-frame handling errors now fail the encrypted session instead of being logged and dropped. This is intentional fail-closed hardening; frame, handshake, batching, and encryption bytes are unchanged.
- Encrypted-session tests now wait for observable outcomes with elapsed-time deadlines rather than a fixed number of scheduler turns.

## Verification
- Relay client and host tests pass.
- Mixed runtime and Relay integration tests pass.


e-n-0:
I updated the description to call out the intentional behavior change: tunnel-frame handling errors now close the encrypted session instead of being logged and dropped. The wire format, handshake, batching, and encryption bytes are unchanged.

I left the brief error-to-reconnecting status transition as-is because it is transient and does not change retry behavior. The final layer also replaces scheduler-turn-bounded test waits with elapsed-time deadlines. The full web suite passes: 835 tests, plus clean type-check, lint, and build.

e-n-0:
/oc-review Re-check the latest head after the previous workflow tool failure. Confirm the earlier findings are addressed and report only issues that still exist.

btriapitsyn:

PR #2236, opened 2026-07-14
feat(tunnels): add managed direct E2EE profile configuration

e-n-0:
## Summary
This adds the configuration layer for managed direct E2EE. Users can opt in per managed profile, and the server persists and reports that choice without exposing connector tokens.

## Stack
- Layer 2 of 6
- Predecessor: #2235
- Part of #2234
- Status: Ready for review

## What changed
- Added a per-profile direct-E2EE opt-in flag.
- Exposed redacted status and mutation routes.
- Tracked the active managed-profile identity.
- Protected unreadable configs from being overwritten.
- Set `controller.mode` from the normalized start request, fixing active-mode resolution for newly started tunnels.

## Review follow-up
- Lifecycle callback failures are classified and logged without leaking callback details.
- Start failures still return a deterministic HTTP response when notification cleanup fails.
- Stop treats direct-E2EE cleanup as critical and returns an explicit `lifecycle_callback_failed` response if cleanup cannot complete.
- Tests wait for callback and response settlement instead of relying on a single timer turn.

## Verification
- Managed-config, tunnel-index, and route tests pass.


e-n-0:

Addressed both points. Lifecycle callbacks now run through sanitized settlement helpers. Start failures still send a deterministic error response if notification cleanup fails, while stop treats direct-E2EE cleanup as critical and returns `lifecycle_callback_failed` if it cannot complete.

The `controller.mode` fix is now called out in the PR description. The callback tests wait for observable callback and response settlement instead of relying on one timer turn. The final web suite passes all 835 tests.

btriapitsyn:

PR #2237, opened 2026-07-14
feat(direct-e2ee): add dormant hardened server endpoint

e-n-0:
## Summary
This adds the dormant, fail-closed server endpoint for direct E2EE.

## Stack
- Layer 3 of 6
- Predecessor: #2236
- Part of #2234
- Status: Ready for review
- This PR currently shows the cumulative stack against `main`; the diff narrows as predecessors merge.

## What changed
- Added the exact-path direct-E2EE endpoint.
- Enforced admission, pre-auth policy, and bearer promotion.
- Restricted pre-auth calls to health, pairing redemption, and session authentication.
- Reused the existing Relay crypto and multiplexing primitives.
- Repaired public encryption identity from valid private material.
- Added recoverable, generation-fenced identity initialization.
- Serialized identity settings writes so abandoned generations cannot overwrite a fresh identity.
- Allowed Relay and direct E2EE to share one injected identity runtime.

## Review follow-up
- Identity initialization failures release pending session entries immediately.
- Non-terminal local tunnel failures intentionally preserve waiters so a successful reconnect can resolve them; terminal failures reject them.
- The handshake deadline is cleared once the encrypted handshake establishes.

## Verification
- Service, identity, auth-policy, and direct-E2EE integration tests pass.
- The final stack passes all 835 web tests; web type-check, lint, and build are clean.

PR #2238, opened 2026-07-14
feat(clients): support direct E2EE pairing and restore

e-n-0:
## Summary
This adds client-side direct-E2EE pairing and session restoration across mobile and desktop.

## Stack
- Layer 4 of 6
- Predecessor: #2237
- Part of #2234
- Status: Ready for review

## Dependency
This layer depends on #2216. SDK global-event SSE must use `runtimeFetch` to traverse the encrypted tunnel. The temporary #2216 patch is not included in this stack. #2217 is related test coverage but is not required here.

## What changed
- Implemented Pairing v2 candidate validation for direct E2EE.
- Shared fail-closed encrypted redemption between mobile and desktop.
- Handled host switching and restoration without plaintext probing.
- Kept Electron credentials behind sender-validated preload IPC.
- Routed OpenChamber event streams through active tunnels.
- Bounded incremental SSE parsing.
- Kept status probes from retaining tunnels; activation explicitly retains or adopts them.
- Used shared Electron runtime planning for encrypted host links.
- Classified pairing failures without exposing sensitive details.

## Verification
- UI, Electron, and server test matrices pass.
- Remote navigation receives no local tokens or Relay metadata.


e-n-0:

I reviewed all four defense-in-depth notes. None needs a code change on this stack:

- `requestWithTimeout` clears the timer and removes the external abort listener in `finally`; after the request wins, the remaining signal/listener cycle is unreachable and cannot reject later.
- Deep-link parsing is already inside the fail-closed `try/catch`, so a non-object payload returns `invalid`. Reordering would only be stylistic.
- `kind: accept` is returned only when the normalized candidate list is non-empty, so `candidates[0]` is guaranteed at the call site.
- `sanitizeHostDirectE2eeForStorage` returns `directE2ee` only after parsing and normalizing the WSS URL, so the later hostname read uses validated data.

I kept those invariants centralized instead of adding duplicate guards at each consumer. The Electron, UI, and direct-E2EE tests remain green, and the automated review reports no blocker.

btriapitsyn:

PR #2239, opened 2026-07-14
feat(settings): expose managed direct E2EE controls

e-n-0:
## Summary
This adds the Settings controls for managed direct E2EE.

## Stack
- Layer 5 of 6
- Predecessor: #2238
- Part of #2234
- Status: Ready for review
- This PR currently shows the cumulative stack against `main`; the diff narrows as predecessors merge.

## What changed
- Added per-profile managed direct-E2EE controls.
- Exposed an availability-gated Add Device transport.
- Added Settings search indexing and translations.

## Review follow-up
- Removed the competing preset updates that could discard the legacy hostname fallback.
- Reconciliation now uses one authoritative preset list, functional state updates, and request-generation fences so late polling cannot overwrite a toggle.
- Selection is reconciled from the effective preset list after mutations and refreshes.

## Verification
- Settings, transport, request-fence, and search tests pass.
- The final stack passes all 835 web tests; workspace type-check, lint, and web build are clean.

PR #2240, opened 2026-07-14
feat(server): ship managed direct E2EE

e-n-0:
## Summary
This final layer enables the managed direct-E2EE path in production.

## Stack
- Layer 6 of 6
- Predecessor: #2239
- Part of #2234
- Status: Ready for review

## What changed
- Registered the production direct-E2EE lifecycle.
- Emitted candidates only for explicitly enabled managed profiles.
- Suppressed same-origin plaintext native candidates.
- Restricted tunnel administration to the host process.
- Added read-only views for paired browser clients.
- Redacted sensitive URLs, labels, and credentials from public host data, headers, and logs.
- Wired one strict identity runtime to both Relay and direct E2EE.
- Added safe pairing-failure classification.

## Review follow-up
- Redacted host configs clear `defaultHostId` when the referenced host is filtered out.
- Per-source admission intentionally trusts `cf-connecting-ip` only for loopback arrivals. Behind another proxy it degrades to the proxy address rather than trusting a spoofable header; global caps still bound the service.
- Identity initialization and settings writes are generation-fenced, serialized, retryable, and covered by deferred-write regressions.

## Security boundaries
- Managed direct E2EE is fail-closed and never falls back to plaintext or Relay.
- Paired browser clients cannot write tunnel-owned settings.
- Tunnel administration is host-only.

## Verification
- All 835 web tests pass.
- Workspace type-check and lint pass; the web production build passes.
- The 103-commit stack has no merges and every commit has a valid SSH signature.


e-n-0:

Addressed the first finding: public host configs now set `defaultHostId` to `null` when redaction removes the referenced host.

For the second, I kept the fail-closed source rule. `cf-connecting-ip` is trusted only when the immediate peer is loopback; trusting it from an arbitrary non-loopback peer would make the limit spoofable. Behind another proxy, per-source limits intentionally collapse to that proxy address, while the global caps still bound total use. The description now documents this tradeoff.

The final stack also shares one generation-fenced identity runtime between Relay and direct E2EE and serializes identity settings writes. All 835 web tests pass, with clean type-check, lint, and build.

btriapitsyn:

PR #2242, opened 2026-07-14
fix(sessions): preserve correct session target

### LINKED ISSUE #2222: [Bug] First message in a new session can be sent to the previously active session
### What is wrong?

After opening a new session draft and sending its first prompt, the message can be delivered to the most recently active existing session instead of the newly created session.

This is especially visible in a multi-project setup: open a new session under one project, enter a prompt, and send it. The UI may return to a recent session and the prompt appears there.

### Steps to reproduce

1. Open an existing session so it becomes the most recently active session.
2. Open a new session draft, preferably under another project.
3. Enter a prompt and press Send.
4. Observe the target session while the first message is being prepared and sent.

### Actual behavior

The first prompt can be routed to the previously active session. In timing-sensitive cases, project/session selection falls back while response-style or snippet preparation is still running, and the later send reads the updated global `currentSessionId`.

### Expected behavior

The send target must be captured when the user submits. A new draft should materialize exactly once, and its first prompt must always be routed to that materialized session even if sidebar selection changes during asynchronous preparation. Existing-session sends should likewise remain pinned to the session active at submission time.

### Root cause from code inspection

`ChatInput.handleSubmit()` performs asynchronous preparation before calling the session store. The store then reads live `currentSessionId` / `newSessionDraft` state. If project/session selection changes during that window, the message is routed using the newer state rather than the submit-time target.

Relevant paths:

- `packages/ui/src/components/chat/ChatInput.tsx`
- `packages/ui/src/sync/session-ui-store.ts`
- `packages/ui/src/components/session/sidebar/hooks/useProjectSessionSelection.ts`

### Related issues

- #1787 identifies a React/session-selection timing race when opening a draft in another project.
- #1637 covers new-session project/directory context changing during materialization.

### Suggested regression coverage

- New draft starts sending, selection changes to an old session during async preparation, final optimistic/server send still targets the materialized session.
- Existing session A starts sending, selection changes to B during async preparation, final send still targets A.
- Concurrent materialization of the same logical draft creates only one session.
- Completion of an older draft materialization does not close a newer draft.

### LINKED ISSUE #2245: [Bug] Creating new sessions isn't working, sends the prompt to an old session

I'm not sure what exactly is going on here. I have multiple Projects configured within Openchamber - let's call them Projects A, B, and C.

When I try to create a new chat session in Project A, recently it's been hanging and freezing my Openchamber instance. 

Once I reload the page, I see that the message I tried to create my new session with has instead sent to my most recent session in Project B... even though I wanted it as a new session in Project A. This has happened to me various times today and yesterday. 

Will use AI to look into this and try to find the root cause as well as a better way to explain it.

### Where does it happen?

Desktop Web

### Version (if known)

1.16.1

_No response_

### Logs (optional)

```shell

```

### LINKED ISSUE #2315: [Bug] Message routed to wrong session when switching projects during pending send in long-lived project
## Summary

After extended use of a single project (many sessions accumulated), starting a new session and pressing Enter to send the first message produces **no visible feedback**. If the user then clicks any session in another project, the pending message is silently delivered to that newly-selected session instead of the intended new session.

This is a state-consistency bug in the send pipeline, not a UI rendering issue.

## Reproduction

**Preconditions**

- A project with a large number of accumulated sessions (the bug is noticeably easier to hit when session count is high or SSE event throughput is heavy).
- A second project with at least one existing session.

**Steps**

1. Open the long-lived project A.
2. Click "New session" (draft opens; `currentSessionId` is `null`).
3. Type a message and press Enter.
4. Observe: nothing happens. No optimistic message appears, no busy indicator, the input is not cleared, no error is shown.
5. Within ~1–2 seconds (while step 3 is still pending), click any session that belongs to a different project B.
6. Observe: the message you typed in step 3 is delivered to the project-B session you just clicked, under the wrong project/directory.

**Expected behavior**

- Either the message is sent to the newly created session in project A, or the send is cancelled with a clear notification when the user navigates away.
- A message composed in project A must never be delivered to a session in project B.

**Actual behavior**

- The message is delivered to whichever session happens to be `currentSessionId` at the moment the async send pipeline resumes — i.e. the session the user navigated to.

## Root cause

The send pipeline reads `currentSessionId` from the live store at multiple points, but the pipeline is async and can be interrupted by a session switch mid-flight. There is no snapshot/consistency guard.

### Async chain with no feedback window

`sendMessage` enters the "new session" branch when a draft is open (`packages/ui/src/sync/session-ui-store.ts:1061`):

```ts
if (!options?.sessionId && draft?.open) {
    const createdDraftSession = await materializeOpenDraftSession(...)  // line 1062
    ...
    await routeMessage({ sessionId: createdDraftSession.sessionId, ... })  // line 1085
}
```

`materializeOpenDraftSession` (`packages/ui/src/sync/session-ui-store.ts:440-507`) awaits multiple slow operations **before** any optimistic UI update:

```ts
await waitForPendingDraftWorktreeRequest(...)        // line 457
await waitForWorktreeBootstrapIfConfigured(...)      // line 462
await store.createSession(...)                       // line 464 (HTTP roundtrip)
store.setCurrentSession(created.id, createdDirectory) // line 491
```

`optimisticSend` (which is what makes the UI react) only runs after `materializeOpenDraftSession` returns. During the entire await window the user sees nothing — this is symptom 1.

### `currentSessionId` is read live, not snapshotted

When the user switches sessions during the await, `setCurrentSession` (`packages/ui/src/sync/session-ui-store.ts:575`) does two things that break the pending send:

1. Line 580: `if (id) get().closeNewSessionDraft()` — **closes the draft**.
2. Line 603: atomically sets `currentSessionId` to the newly-selected session.

When `sendMessage` resumes after its await, the state it reads has changed underneath it:

```ts
// session-ui-store.ts
const draft = get().newSessionDraft              // line 1027  — now closed
if (!options?.sessionId && draft?.open) { ... }  // line 1061 — branch skipped
...
const targetSessionId = options?.sessionId ?? get().currentSessionId  // line 1113 — wrong session
await routeMessage({ sessionId: targetSessionId || "", ... })         // line 1165 — delivered to wrong project
```

Because `draft.open` is now false, the new-session branch is skipped and the existing-session branch runs against the *now-current* (wrong) session id. This is symptom 2.

### No cancellation on navigation

There is no abort/cancellation wired to session switches:

- `session-ui-store.ts:241,566` declares `abortControllers: Map<string, AbortController>`, but a codebase-wide search finds **zero** `set`/`get`/`delete`/`clear` calls against it — it is dead code.
- `setCurrentSession` never aborts any in-flight send.
- `sendMessage` has no equivalent of the staleness guard that `fetchMessagesForSession` already implements at `packages/ui/src/sync/session-actions.ts:1314`:

  ```ts
  // Staleness guard: a rapid session switch may have moved the user off this
  // session while the fetch was in flight.
  if (useSessionUIStore.getState().currentSessionId !== sessionID) return
  ```

### Aggravating factor

The longer a project is used, the slower the `createSession → setCurrentSession` cascade becomes (it synchronously updates four stores plus `opencodeClient`, `session-ui-store.ts:603-620`, over an increasingly large session list). This widens the window in which the user can trigger the race, which is why the bug only appears "after a lot of conversations" rather than on a fresh project.

## Impact

- **Cross-project data integrity violation**: a prompt intended for project A is executed in project B's directory, potentially running the agent against the wrong codebase, wrong tools, and wrong context.
- **Silent failure**: the user is not notified; they only discover the misrouting when they look at project B.
- **Reproducible on first try** once a project has enough session volume.

## Suggested fix direction

1. **Snapshot send intent at `handleSubmit`/`sendMessage` entry**: capture `{ sessionId, directory, draftOpen }` into an immutable request object; use only these values for the rest of the pipeline. Do not re-read `currentSessionId` after any `await`.
2. **Add a staleness guard to `sendMessage`** mirroring the one in `fetchMessagesForSession`: if `currentSessionId` has changed since the send was initiated, abort the send (or confirm with the user) rather than silently re-targeting.
3. **Show optimistic UI before `materializeOpenDraftSession` resolves** (e.g. a "creating session…" placeholder) so the user gets immediate feedback that Enter was registered.
4. **Either wire up `abortControllers` or remove the dead declaration**, and have `setCurrentSession` cancel any pending send for the previous session.

I can open a PR with the snapshot + staleness-guard fix if the team agrees with the direction.

### Where does it happen?

Desktop (macOS)

### Version (if known)

1.16.2

_No response_

### Logs (optional)

```shell

```

### LINKED ISSUE #2317: [Bug] Desktop project switch keeps the previous session and sends messages to the wrong project
## Summary

## Steps to reproduce

1. Open OpenChamber Desktop.

## Actual behavior

## Expected behavior

## Reproducibility

## Runtime evidence

### Before switching

### After selecting Project B

## Suspected root cause

contains this branch:

```ts
  return;
}
```

After switching projects:

## Suggested fix direction

## Related issues

### Where does it happen?

Desktop (macOS)

### Version (if known)

### Logs (optional)

```shell
```

bashrusakh:
## Summary

Fixes four session-routing paths that could leave a send targeting a session from another project:

- A new-session draft can change while the composer is preparing a message asynchronously.
- A stale composer closure can submit a draft after live selection has changed.
- A draft send can remain pending while the user changes to another project.
- Switching projects can leave the previous project session active while the new project is shown.

## What changed

- Capture the intended existing-session ID or full new-session draft before composer async work.
- Pass that explicit target through `SendMessageOptions` to `sendMessage`.
- Materialize new sessions from the captured draft snapshot rather than live UI state.
- Treat a captured open draft as new-session intent even if the live draft closes before send begins.
- Preserve the snapshot folder target during materialization and avoid overwriting newer drafts or selections.
- Keep a pending draft send targeted at its captured session and directory while selection changes to another project.
- Prevent unrelated `createSession` calls from inheriting a live draft folder.
- When a project changes, preserve an unknown session only while its worktree metadata may still be loading. If the active session is known to belong to another project, select the new project remembered session or first available session instead.

### Folder assignment scope

A newly created session uses the following directory scope resolution order for folder assignment:

1. An explicit directory override.
2. The directory returned for the newly created session.
3. `lastLoadedDirectory` only as a fallback.

## Regression coverage

- `issue-2222.test.ts` covers captured send targets, stale-draft submission after a cross-project selection change, draft snapshot materialization, folder assignment, preservation of newer drafts, and paused materialization while selection changes from project A to project B. Pending messages retain their materialized A session and A directory; the B selection remains active.
- `useProjectSessionSelection.test.ts` covers Project A to Project B selection, remembered and fallback targets, empty-project drafts, and stale worktree-session preservation.

Fixes #2222
Fixes #2245
Fixes #2317
Fixes #2315

## Validation

| Check | Result |
| --- | --- |
| `bun test packages/ui/src/sync/__tests__/issue-2222.test.ts` | Passed: 8 tests |
| `bun run lint:ui` | Passed |
| `bun run dead-code` | Completed previously; existing non-blocking Knip diagnostics remain |

## Notes

- The PR currently conflicts with the latest `main`; this focused follow-up does not rebase or rewrite the existing PR branch.


makeittech:
Closing as superseded. The project-switch session-selection piece of this PR (#2317) has landed as #2865, and the draft/existing-session send-routing piece (#2222, #2315) has landed as a focused fix in #2871, which intentionally skips this PR's broader composer/goal-arm/message-queue changes. If any remaining piece of this PR still applies to current main, please reopen with it rebased and scoped to just that part.

PR #2249, opened 2026-07-14
fix(sidebar): surface subagent approval shield on the parent session

### LINKED ISSUE #2247: [Bug] The red shield icon for manual approval required doesn't show on a session when the subagent needs approval

It would be nice to see this icon at the level of the primary agent, instead of having to click through all my agents to see which subagents require my approval to continue. 

Marking this as a bug because it feels like it should be required behavior.

### Where does it happen?

Desktop Web

### Version (if known)

1.16.1

_No response_

### Logs (optional)

```shell

```

IbrahimKhan12:
> Validated commit: `501e156956cc96eb779195dcfb6394cc47511a56`, rebased onto `main` at `feec14545`.

## Intent

Fixes #2247. A collapsed parent row rolled up the pending-**question** counts of the descendants it hides, but the pending-**permission** shield read only the row's own session. A subagent blocked on manual approval under a collapsed parent was therefore invisible: the session sat waiting with nothing in the sidebar to show it.

The asymmetry is still present on `main`. In `sessions/SessionNodeItem.tsx` the shield comes from `useSessionPermissions(session.id, ...)` while the question badge next to it comes from the scope-based `useSessionQuestionCount(...)`.

Both badges now derive from one hook over the same scopes, so permissions and questions follow identical roll-up rules.

## Non-goals

No change to what a permission request is, how it is answered, or how the badge looks. No change to expanded-row behaviour: an expanded parent still counts only its own session, because its descendants render their own rows. No bootstrap change — neither badge initializes a directory store to produce a count.

## Affected surfaces

- `packages/ui` sidebar rows in every runtime that renders them: web, desktop, VS Code, and mobile.
- `sync/child-store.ts` gains `subscribeDirectoryPermissions` (plural), factored from the existing question equivalent so both kinds share one implementation.
- `sync/sync-context.tsx`: `useSessionQuestionCount` becomes `useSessionBlockingRequestCounts` and returns `{ permissionCount, questionCount }`.
- `sessions/sessionNodeItemUtils.ts`: `selectQuestionBadgeSessionScopes` becomes `selectBlockingBadgeSessionScopes`; its scope type is no longer exported because nothing outside the module consumed it.
- User-visible state: the shield count on a collapsed parent now includes hidden descendants. No styling change.
- No persisted, server, or transport contract change.

## Rebase note

`main` moved substantially. Three adaptations are worth a reviewer's attention:

1. **`useSyncRuntime` was preserved, deliberately.** `main` changed this hook from `useSyncSystem()` to `useSyncRuntime()` with the comment that reading the directory-bearing context "would re-render every sidebar row that counts questions whenever the user switches projects." The pre-rebase branch still called `useSyncSystem()`. Taking the branch's version wholesale would have reintroduced that regression, so the rebased hook keeps `useSyncRuntime()` and the comment now covers both kinds.
2. **A consumer `main` added had to be updated.** `projects/SessionGroupSection.behavior.test.tsx` mocks the `sync-context` module surface and stubbed `useSessionQuestionCount: () => 0`. Renaming the hook would have left a mock for an export that no longer exists, so it is now `useSessionBlockingRequestCounts: () => ({ permissionCount: 0, questionCount: 0 })`. `useSessionPermissions` is left in place: it still exists and is still used by `useAssistantStatus.ts` and `useSessionActivity.ts`.
3. **The new test had to change strategy.** It rendered `SyncProvider` through `renderToStaticMarkup`; on current `main` that throws `Missing getServerSnapshot`, because the provider now mounts a hook without a server snapshot. It is rewritten to client-render with `createRoot` + `act` + `installHookTestDom`, matching `main`'s own `sync-context-selection-boundary.test.tsx`. Files also moved into `sidebar/sessions/`, and the sidebar documentation paragraph was rewritten against `main`'s restructured doc.

The rewrite made the test stronger rather than weaker: because it now mounts, it asserts the initial unbootstrapped state, then that a live bucket write is delivered through the sidecar channel, instead of only checking one static render.

**One finding came out of that rewrite.** The first client-render attempt returned 2 instead of 3. The cause was the fixture, not the fix: the scope directory was also the provider's current directory, so bootstrap of that directory reset the seeded bucket mid-test. The test now uses directories that are not the current one. This is worth stating because it demonstrates the intended invariant — the counts follow authoritative store state, and a bootstrap that clears a bucket clears the count with it.

## Repository guidance

|---|---|---|
| `openchamber-change-discipline` | Exported hook renamed, new exported subscribe helper, shared UI behaviour changed | Consumers were traced rather than assumed: the renamed hook had a second consumer in a behaviour-test mock that `main` added, and it is updated. The plural subscribe helper is factored from the existing question path instead of duplicated. A scope type that no consumer used is no longer exported. `bun run dead-code` run because exports and a file changed. |
| `sync-state-invariants` | Session-scoped blocking-request subscriptions and directory-scoped stores | Failure is not empty success: an unbootstrapped store contributes zero and neither badge bootstraps one to get a number, so a missing store reads as "nothing known yet" rather than a fabricated count. Subscriptions stay leaf-scoped to the exact session buckets, so unrelated sessions in the same store do not notify. Pinning still keeps referenced stores alive for the row's lifetime. |
| `performance-engineering` | One hook instance per visible row, on a list that can be long | **Narrow:** subscriptions cover only the sessions in the row's scopes, not the directory. **Skip:** `getSnapshot` returns the previous object identity when both counts are unchanged, so a bucket write that does not move either number cannot re-render the row. The cost equation is unchanged from the question badge that already paid it — a collapsed subtree subscribes O(descendants) buckets — and `useSyncRuntime` keeps a project switch from re-rendering every row. |
| `packages/ui/src/sync/DOCUMENTATION.md` | Owns the sidecar channel contract | Updated to state that collapsed rows group hidden descendants by owning directory and subscribe to the exact permission and question buckets, while expanded rows subscribe only to their own. |
| `packages/ui/src/components/session/sidebar/DOCUMENTATION.md` | Owns sidebar row contracts | A new paragraph states the one shared rule for both badge kinds and the no-bootstrap guarantee, written against `main`'s restructured doc. |

## Validation

| Check | Result |
|---|---|
| `bun test src/sync/session-blocking-request-counts.test.tsx` | **3 passed**, 0 failed. |
| `bun test` across `child-store`, `session-blocking-request-counts`, `sync-context-selection-boundary`, `sessionNodeItemUtils`, `SessionGroupSection.behavior` | **42 passed**, 0 failed, 137 assertions. Includes the mock updated in this change and `main`'s selection-boundary test, which guards the `useSyncRuntime` decision above. |
| `bun run lint:ui` | Exit 0. The one remaining warning is in `apps/MobileChangesSurface.tsx`, untouched here and byte-identical to `origin/main`, so it is pre-existing upstream. |
| `bun run dead-code` | Neither `subscribeDirectoryPermissions`, `selectBlockingBadgeSessionScopes`, `useSessionBlockingRequestCounts`, nor the new test file appears in the unused report. |
| Rendered sidebar, collapsed and expanded | See Visual evidence. |

Covered cases: cross-directory aggregation; an unrelated bucket write not notifying; the expanded scope staying parent-only; permissions and questions counted independently over the same scopes; and the unbootstrapped starting state reading zero.

**Not validated:** no VS Code or Capacitor build was run, and no real approval was answered end to end against a live agent, so the badge is proven from authoritative store state rather than from a full permission round trip.

## Visual evidence

Captured against this HEAD by running the sidebar in a headless Chromium at 1280x900 (dark). Fixture sessions were injected through the app's authoritative `applySnapshot` path and the permissions through the live child store, so no real session from the host machine appears; the rows, badges, and styling are the real components.

The scenario is one parent waiting on one approval with a hidden subagent waiting on two more, so the correct collapsed total is three.

The before frame was produced by temporarily restoring `main`'s per-session shield in the same running app, then reverting.

**Before — the collapsed parent shows `1`, hiding both subagent approvals:**

Collapsed parent showing a shield count of 1

**After — the collapsed parent shows `3`, the true number of pending approvals in its subtree:**

Collapsed parent showing a shield count of 3

**After, expanded — the hidden count moves to the row that owns it: the subagent shows its own `2`:**

Expanded parent with the subagent row showing a shield count of 2

In the expanded frame the parent row is hovered from the click that expanded it, so its own badge is swapped for row actions — that is `main`'s existing hover behaviour, not a change here. The DOM at that moment reports parent `1` and child `2`, which is the expanded contract.

## Risks and failure behavior

- **Unbootstrapped directories contribute zero** until authoritative state arrives. That is the pre-existing question-badge contract, now shared, and it is the conservative direction: a hidden approval can be reported late, never invented.
- **Subscription churn** on structural session changes matches the question path that already existed; the snapshot identity guard prevents a re-render when neither count moves.
- **Project switching** does not re-render every counting row, because the hook keeps `main`'s `useSyncRuntime()`.
- **Expanded rows are unchanged**, so the fix cannot double-count: a descendant is counted by its parent only while its own row is hidden.
- Rollback is reverting one commit. No persisted data, server contract, or permission-answering path is touched.


btriapitsyn:

IbrahimKhan12:
Rebased onto `main` at `feec14545`; the branch now merges cleanly. Three adaptations are worth a look.

**`useSyncRuntime` was preserved deliberately.** `main` changed this hook from `useSyncSystem()` to `useSyncRuntime()` specifically so that switching projects does not re-render every row that counts. Taking the branch's version wholesale would have reintroduced that regression, so the rebased hook keeps `useSyncRuntime()` and the comment now covers permissions as well as questions.

**A consumer you added needed updating.** `projects/SessionGroupSection.behavior.test.tsx` mocks the `sync-context` surface and stubbed `useSessionQuestionCount`. Renaming the hook would have left a mock for a non-existent export, so it now returns `{ permissionCount: 0, questionCount: 0 }`. `useSessionPermissions` stays — it is still used by `useAssistantStatus.ts` and `useSessionActivity.ts`.

**The new test had to change strategy.** It used `renderToStaticMarkup`, which now throws `Missing getServerSnapshot` because `SyncProvider` mounts a hook without a server snapshot. It is rewritten to client-render with `createRoot` + `act`, matching your own `sync-context-selection-boundary.test.tsx`. That made it stronger: it now asserts the unbootstrapped zero state and then that a live bucket write arrives through the sidecar channel, instead of one static render.

That rewrite also surfaced something worth stating. The first attempt returned 2 instead of 3, and the cause was the fixture: the scope directory was also the provider's current directory, so bootstrap reset the seeded bucket mid-test. It now uses non-current directories. The behaviour is correct and is the intended invariant — counts follow authoritative store state, and an unbootstrapped store contributes zero rather than a fabricated number.

The visual evidence gap from the last review is closed. The description has before/after frames from the real sidebar: a parent waiting on one approval with a hidden subagent waiting on two more shows `1` before and `3` after when collapsed, and when expanded the `2` moves to the subagent row that owns it. Fixture sessions were injected through the authoritative snapshot path so no real session appears in the capture.

PR #2251, opened 2026-07-14
fix(settings): serialize settings.json writes across Electron and the web server

a692570:
## Summary

`packages/electron/main.mjs` (`writeJsonFile`/`mutateSettingsRoot`) and `packages/web/server/lib/opencode/settings-runtime.js` (`writeSettingsToDisk`) each persist `settings.json` through their own independent atomic write-tmp-then-rename path. Since the desktop app starts the web server in-process (per this repo's own architecture notes), both writers run in the same process against the same file with no coordination between them.

Separately, direct `writeSettingsToDisk` callers outside `persistSettings` (relay signing key, push runtime, notification routes) bypassed the module's own `persistSettingsLock` entirely — only calls that went through `persistSettings` were serialized.

Net effect: concurrent read-modify-write cycles from either side can silently drop each other's changes (last writer wins, earlier changes lost). I hit this after noticing repeated Electron main/renderer crashes on macOS 1.16.1 correlated with several orphaned `settings.json.tmp-<pid>-*` files sitting next to a live `settings.json`, including a 0-byte one from a write that never got to rename.

I want to flag up front: I could not pin the underlying native crash itself (EXC_BREAKPOINT/SIGTRAP inside the Electron/V8 layer) to a specific line — the crash-report stack frames are mis-symbolicated against unrelated bundled libraries, consistent with a stripped release binary, so I didn't want to guess-patch that blind. What I could verify and fix is the write race above, which is real regardless of whether it's the sole cause of what I saw.

## Changes

- Add `packages/web/server/lib/fs/settings-file-lock.js`: a small path-keyed write lock (`withFileWriteLock`), shared by any writer of the same file path within a process.
- `settings-runtime.js`: `writeSettingsToDisk` now goes through the lock, so every caller (not just `persistSettings`) is serialized.
- `main.mjs`: `writeJsonFile` now goes through the same shared lock module, so Electron's own settings writer serializes against the web server's writer.

## Test plan

- [x] `bun run test` in `packages/web` — new `settings-file-lock.test.js` and a new regression test in `settings-runtime.test.js` proving two concurrent `writeSettingsToDisk` calls no longer interleave or leave an orphaned tmp file. Pre-existing suite passes except one unrelated, pre-existing failure on `main` (`env-runtime.test.js`, picks up a locally-installed `opencode` binary via `PATH` on my machine — reproduces identically on a clean `main` checkout before this change).
- [x] `bun run type-check` in `packages/web` — clean.
- [x] `node --check` on all three touched/added JS/mjs files — clean.
- [x] `bun run dead-code` — new file/export not flagged.
- [ ] Did not run a full packaged Electron manual test (no packaging environment here); the `main.mjs` change is a small, mechanical wrap of an already-tested function with the same shape as the server-side change, so risk is low, but flagging this gap explicitly per the repo's own validation guidance for platform/runtime changes.

### DIFF
--- a/packages/electron/main.mjs
+++ b/packages/electron/main.mjs
@@ -18,6 +18,7 @@ import { assertUpdaterCapability } from './updater-capability.mjs';
 import { checkForDesktopUpdate } from './updater-check.mjs';
 import { resolveUpdaterFeed } from './updater-feed.mjs';
 import { mintOutsideFileGrant } from '@openchamber/web/server/lib/fs/routes.js';
+import { withFileWriteLock } from '@openchamber/web/server/lib/fs/settings-file-lock.js';
 
 const execFileAsync = promisify(execFile);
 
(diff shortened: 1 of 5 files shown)


kinski69:
Reproducing on Linux x64 (Arch/Omarchy, AppImage v1.19.0, Electron main process): three crashes today (12:07, 12:23, 13:04 CEST), each SIGTRAP (si_code SI_KERNEL) on the main thread ~87 s after launch, every one leaving an orphaned `settings.json.tmp-<pid>-*` in `~/.config/OpenChamber` (write never renamed), consistent with the write race described here. coredumps available via coredumpctl.

Main-process log right before each dump, same second:

```
[persistSettings] Updating fields: ...
Message stream WS proxy error: TypeError: terminated
[PushWatcher] disconnected terminated
```

The "terminated" WS error suggests the fetch to the managed opencode serve was aborted just before the trap. No OOM (10 GiB free), settings.json itself never corrupted (only the tmp file is left behind).

Filed by deepseek-v4-pro via opencode.

btriapitsyn:

PR #2252, opened 2026-07-15
fix(sync): restore chat timeline after revert

### LINKED ISSUE #2248: [Bug] Reverting chat doesn't really work properly

Whenever I press the revert message, my Openchamber freezes and often times the revert doesn't even work. Additionally, when I send a new message after reverting a previous one, the new message does not show in the chat until I reload the page.

In the long term, it would be great to have an "edit message" feature like other AI coding agents do, which handles the reverts and gives an option to revert code, conversation, or both. I'll file a new feature request for that if one doesn't already exist.

### Where does it happen?

Desktop (macOS)

### Version (if known)

1.16.1

_No response_

### Logs (optional)

```shell

```

bashrusakh:
## Summary
- preserve `session.revert` as the server-authoritative boundary for reverted history
- add a directory-scoped post-revert replacement overlay so the new branch is visible without revealing discarded messages
- apply one shared visibility projection to the timeline, message history, reverted-message dock, and slash undo/redo
- restore the captured replacement overlay when a newer revert fails, including after concurrent session reconciliation
- retire stale overlays on authoritative marker transitions and cache eviction

Fixes #2248

## Why this has two commits
The branch keeps its published history instead of rewriting it.

- `f3854676e` made the replacement message visible by locally clearing `session.revert`. That was incomplete: clearing the marker also revealed the discarded branch and coupled rollback to fragile local state timing.
- `21488850c` replaces that approach. It keeps the marker authoritative and uses a local post-revert branch overlay to show only the replacement branch.

## Behavior
After reverting and sending a replacement, the discarded interval remains hidden while the replacement and its subsequent branch stay visible. A failed replacement or failed newer revert restores the prior visible branch without overwriting overlays for other sessions.

## Known boundary
`postRevertBranch` is intentionally in-memory UI continuity state. After a full store reload, correct replacement-branch reconstruction depends on the authoritative server session state changing or clearing the corresponding revert marker; this PR does not introduce a backend persistence contract.

## Revert UI freeze investigation
- **Found:** The original visibility bug was caused by using the server revert marker as the sole timeline filter.
- **Not established:** No macOS performance trace, latency benchmark, or reproducible test attributes the reported UI hang to a specific `revertToMessage` operation.
- **Scope:** This PR fixes the deterministic visibility/state-transition defect; it does not claim to resolve the separate UI-hang report.

## Validation
```text
bun test --isolate packages/ui/src/sync/session-actions.test.ts packages/ui/src/sync/session-ui-store.test.js packages/ui/src/sync/session-message-records.test.ts packages/ui/src/sync/user-message-history.test.ts packages/ui/src/components/chat/revertedMessageDockState.test.ts packages/ui/src/sync/__tests__/eviction.test.ts packages/ui/src/sync/use-sync.test.ts packages/ui/src/sync/__tests__/event-reducer.test.ts packages/ui/src/sync/__tests__/sync-context-session-events.test.ts
# 100 passed, 270 assertions

bun run type-check:ui
bun run lint:ui
bun run dead-code
git diff --check
```

## Review
- Native review: no actionable findings
- OCR: no high or medium findings; only non-blocking maintainability notes

### DIFF
--- a/packages/ui/src/components/chat/revertedMessageDockState.test.ts
+++ b/packages/ui/src/components/chat/revertedMessageDockState.test.ts
@@ -3,6 +3,7 @@ import type { Message, Part } from '@opencode-ai/sdk/v2/client';
 import type { State } from '@/sync/types';
 
 import { EMPTY_REVERTED_MESSAGE_DOCK_STATE, buildRevertedMessageDockState } from './revertedMessageDockState';
+import { addPostRevertBranchReplacement } from '@/sync/message-visibility';
 
 const message = (id: string, role: 'user' | 'assistant'): Message => ({
     id,
@@ -17,10 +18,11 @@ const textPart = (id: string, text: string): Part => ({
     text,
 } as Part);
 
-const state = (partial: Partial<State>): Pick<State, 'session' | 'message' | 'part'> => ({
+const state = (partial: Partial<State>): Pick<State, 'session' | 'message' | 'part' | 'postRevertBranch'> => ({
     session: [],
     message: {},
     part: {},
+    postRevertBranch: {},
     ...partial,
 });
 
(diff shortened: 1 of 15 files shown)


bashrusakh:
dont like this release, try different aproach

PR #2255, opened 2026-07-15
feat(sidebar): show a pending-question indicator on sessions

### LINKED ISSUE #2250: [Feature Request] Sidebar: Show an icon when there's a pending question from an agent in a session

It would be amazing to be able to know which of my agents are asking me questions using the question tool just from looking at the sidebar.

Right now, I get a notification on the bottom right, but if that notification gets dismissed, I don't know which agent asked the question, and there's no clear way to know unless I click through various agents.

I assume the implementation for this would be something similar to the existing feature where a red shield icon shows up when manual approval is required for an agent to run a tool.

### Extra context (optional)

_No response_

IbrahimKhan12:
## Intent

Fixes #2250 and #2634. A `question` tool call blocks the agent until it is answered, but the only signal today is a toast; once that is dismissed nothing shows which session is waiting. Session rows now carry a pending-question count badge beside the permission shield. A collapsed row covers the descendants it hides, grouped by the directory store owning each session so a cross-worktree descendant still reports; expanded rows count only their own session, so no total is counted twice.

## Non-goals

Answering, rejecting, and rendering questions are unchanged, as is toast and notification behavior. The permission shield keeps its per-session scope.

## Affected surfaces

`packages/ui` only: the shared `SessionNodeItem` row (web, desktop, VS Code, hosted mobile), the directory-store sidecar notification channels, the sync performance counters, and all eleven locale dictionaries. Questions stay live-only, so nothing persisted, wire-level, or external changes. Native mobile renders the same shared row through the hosted web surface.

## Repository guidance

|---|---|---|
| `AGENTS.md`, `openchamber-change-discipline` | Shared UI source, new sync exports, two owning docs | Extends the existing permission-badge subscription pattern instead of adding a second mechanism; no new dependency; both owning `DOCUMENTATION.md` files updated; `dead-code` reports no new unused export. |
| `sync-state-invariants` | Badge state is live, and cleanup must be observable | Reads the authoritative `state.question` buckets, never history. `question` joins the session-lifecycle clone list because `dropSessionCaches` otherwise mutates the map in place, leaving `current === previous` to suppress the notification on archive and delete. |
| `performance-engineering` | Every visible row subscribes, so the cost is rows x directory publications | Rows subscribe to the exact buckets they render through the session-keyed sidecar channel, never a whole-store selector, and a row covering several sessions is coalesced to one notification per publication (asserted against 10,000 unrelated part updates). A single-directory tree resolves to one store, so grouping adds no subscriptions. |
| `theme-system`, `locale-ui-patterns` | New visual status and accessible text | Reuses the sprite `question` icon and the semantic `status-info` tokens, and ships translated singular and plural messages in all eleven locales instead of count-suffix fragments. |
| sync and sidebar `DOCUMENTATION.md` | Both state invariants this change extends | Records questions as a sidecar channel, the one-notification-per-publication rule, `question` in the clone table, and the collapsed/expanded ownership plus owning-directory grouping. |

## Validation

Run at this HEAD.

| Check | Result |
|---|---|
| `bun test packages/ui/src/components/session/sidebar/sessionNodeItemUtils.test.ts` | 11 pass. Covers collapsed multi-level rollup grouped by owning directory, expanded per-session scope, group-directory fallback, and an unresolvable directory. |
| `bun test packages/ui/src/sync/child-store.test.ts` | 15 pass, including the new question-channel test. |
| `bun test packages/ui/src/lib/i18n/messages.test.ts` | 2 pass; eleven-locale key parity holds. |
| `bun test packages/ui/src/sync packages/ui/src/components/session/sidebar` | 432 pass, 40 fail. That failing set is byte-for-byte what fails on unmodified `origin/main` here (427 pass / 40 fail); all five new tests pass and nothing newly fails. |
| `bun run type-check:ui`, `bun run lint:ui`, `bun run build:ui` | Exit 0. |
| `bun run build:web` | Exit 0; the emitted stylesheet contains `bg-status-info/10` and `text-status-info`, so the badge tokens compile rather than silently dropping. |
| `bun run dead-code` | No new unused export. |

Not verified: no browser, packaged Electron, or VS Code extension host run at this HEAD.

## Visual evidence

Captured on an earlier implementation of the same behavior contract and still accurate here: the markup, tokens, icon, count, labels, collapsed/expanded rule, and owning-directory grouping are unchanged, and every session shown shares one directory. `Baseline - no pending question` is the no-badge comparator.

<table>
  <tr>
    <td><strong>Collapsed, light, narrow</strong><br>Collapsed parent with two pending questions in light theme at narrow sidebar width</td>
    <td><strong>Collapsed, light, wide</strong><br>Collapsed parent with two pending questions in light theme at wide sidebar width</td>
  </tr>
  <tr>
    <td><strong>Collapsed, dark, wide</strong><br>Collapsed parent with two pending questions in dark theme at wide sidebar width</td>
    <td><strong>Expanded, light, wide</strong><br>Expanded parent with one pending question on each child and no badge on the baseline in light theme</td>
  </tr>
  <tr>
    <td><strong>Expanded, dark, wide</strong><br>Expanded parent with per-child pending question badges in dark theme</td>
    <td><strong>Expanded, light, narrow</strong><br>Expanded parent with per-child pending question badges and no-badge baseline at narrow sidebar width</td>
  </tr>
</table>

## Risks and failure behavior

The badge is derived state, not a transient notification, so a dismissed toast cannot lose it and streaming traffic cannot clear it. `question.replied`, `question.rejected`, archive, and delete each publish a replaced bucket, so a count cannot outlive its request. Referenced stores stay pinned while a row is mounted, so an idle worktree cannot be evicted from under a live subscription. Rollback needs no migration.

Both failure modes omit a badge rather than invent one: a descendant directory that is not bootstrapped yet contributes nothing until its live event or the next reconciliation arrives, and an active sidebar search force-expands every row while rendering only matching branches, so a pending question on a filtered-out sibling waits for the search to clear, as the permission shield does today.


m3nowak:
@IbrahimKhan12 Are you still working on this?

IbrahimKhan12:
@m3nowak forgot to publish, thanks for the reminder. Will have it published soon.

IbrahimKhan12:
@btriapitsyn Please review this if you get a chance. Would love to have this feature added soon.

IbrahimKhan12:
Both review findings are addressed in `fe512b0f`.

**Finding 1 (blocker) — per-descendant directory grouping.** Restored, and the review is right that it is required rather than cosmetic: `useSessionGrouping.buildGroupedSessions` nests children by `parentID` with only an archived-status check, so a descendant can legitimately sit in another worktree's store, and `moveSessionTreeToQuickWorktree` has an explicit partial-rollback-failure path that leaves a tree split across directories. The row now groups the sessions it reports on through a pure `selectQuestionScopes` selector, and `useSessionQuestionCount` counts, subscribes, and pins across those stores. One difference from the previously reviewed implementation: the selector resolves a session's directory exactly as the row resolves its own, `normalizePath(session.directory)` with the group directory as fallback, instead of a separate `resolveGlobalSessionDirectory`/worktree-path chain. That keeps the row's own node mapped to the same store the permission badge reads, and avoids a pure sidebar helper depending on a store module that several sync tests replace with `mock.module`.

**Finding 2 (non-blocker) — scope-selection coverage.** `selectQuestionScopes` is a pure exported selector again, covered by four tests in `sessionNodeItemUtils.test.ts`: collapsed multi-level rollup grouped by owning directory, expanded per-session scope, group-directory fallback, and the unresolvable-directory case. The PR body's Validation table now cites that file with its actual result.

Note for the diff-vs-previous-HEAD comparison: the branch was rebased onto current `main` (it was 171 commits behind and conflicting in `SessionNodeItem.tsx`, `sync/child-store.ts`, and `sync/DOCUMENTATION.md`), so earlier reviewed trees are no longer ancestors. The stale body claims quoted in the review were corrected in the same pass.

The screenshots are unchanged and still represent this HEAD; every session in them shares one directory, and neither the rendered markup nor the collapsed/expanded rule changed.

PR #2260, opened 2026-07-15
fix(chat): keep text selection menu in viewport

### LINKED ISSUE #2257: Desktop: 'Add to Chat' floating button gets pushed to screen edge when selecting long agent responses

When selecting agent response text in the chat, a floating "Add to Chat" button appears near the mouse cursor. However, when selecting a long response by scrolling down, the button does not stay near the mouse — it gets pushed to the edge of the screen, making it impossible to view or click.

#### Expected behavior
- Select text from an agent response by dragging/scrolling downward
- The "Add to Chat" floating button should appear near the mouse cursor at all times during the selection
- The button should remain accessible regardless of how long the response is or how far down the user scrolls

#### Actual behavior
- Select text from a long agent response by scrolling down
- The "Add to Chat" button does not follow the mouse position
- It gets pushed to the edge of the screen (likely due to layout overflow or absolute positioning without proper boundary clamping)
- The button becomes invisible or unclickable

#### Steps to reproduce
1. Open the OpenChamber desktop app (Windows)
2. Start a conversation and get a long agent response (one that requires scrolling)
3. Click at the beginning of the response text
4. Hold the mouse button and drag downward to select a large block of text, scrolling past the visible area
5. Observe that the "Add to Chat" floating button is not near the mouse cursor
6. The button is pushed to the edge of the screen and cannot be clicked

### Where does it happens?
Windows desktop application.

### Version
`1.16.0`

### Notes
This is likely a CSS/layout issue where the floating button uses absolute or fixed positioning without accounting for the case where the selection endpoint is near or beyond the viewport boundary. The button needs to be clamped within the visible viewport relative to the cursor position.

bashrusakh:
## Summary
- Clamp the desktop text-selection menu vertically so it stays within the viewport.
- Re-clamp after popup measurement and viewport resizing while leaving mobile behavior unchanged.
- Add focused boundary coverage for top, bottom, normal, and constrained viewport positions.

Fixes #2257

## Validation
- `bun test packages/ui/src/components/chat/message/textSelectionMenuPosition.test.ts`
- `bun run type-check:ui`
- `bun run lint:ui`
- `bun run dead-code`

## Manual verification
- Not run: Electron/Windows selection flow.

### DIFF
--- a/packages/ui/src/components/chat/message/TextSelectionMenu.tsx
+++ b/packages/ui/src/components/chat/message/TextSelectionMenu.tsx
@@ -15,6 +15,7 @@ import { resolveProjectForSessionDirectory } from '@/lib/projectResolution';
 import { useEffectiveDirectory } from '@/hooks/useEffectiveDirectory';
 import { isVSCodeRuntime } from '@/lib/desktop';
 import { useI18n } from '@/lib/i18n';
+import { DESKTOP_MENU_SIDE_MARGIN_PX, getDesktopClampedMenuY } from './textSelectionMenuPosition';
 
 interface TextSelectionMenuProps {
   containerRef: React.RefObject<HTMLElement | null>;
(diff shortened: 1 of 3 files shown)


btriapitsyn:

PR #2261, opened 2026-07-15
fix(lifecycle): prevent false Windows health-check restart of live OpenCode child

### LINKED ISSUE #2258: Windows health check can restart a live OpenCode child
I'm seeing OpenChamber restart a managed OpenCode process on Windows even though the child has not exited yet.

Environment: OpenChamber 1.16.1, OpenCode 1.17.20, Windows x64, managed local `opencode serve`.

I added a temporary persistent exit listener while debugging. This sequence repeated several times during active sessions:

    [lifecycle] periodic health check: OpenCode process exited, restarting...
    Restarting OpenCode process...
    Stopping existing OpenCode process...
    [OpenCode] Managed server process exited {
      code: null,
      signal: 'SIGTERM',
      expected: true
    }

The last event is the suspicious part. The child exits from the SIGTERM sent by OpenChamber's own restart path, and the listener reports it as an expected shutdown. If it had already exited before the health check, I would expect `exitCode` or `signalCode` to be set before `restartOpenCode()` tries to stop it.

The current liveness check uses both the ChildProcess state and `process.kill(pid, 0)`. It looks like a failed PID probe on Windows can take the immediate "process exited" branch and bypass the consecutive health-failure counter and busy-session guard added for #1294/#1295.

Would it be safer to trust `exitCode`/`signalCode` for a child that OpenChamber spawned directly, and let failed HTTP health checks use the normal retry path while the child has not emitted `exit`?

This is related to #1294/#1295, but those issues covered health request timeouts under load. #1721 is also related to false results from PID-only liveness checks, although it is a different code path.

bashrusakh:
## Summary

Fixes #2258. Two independent bugs in `isManagedOpenCodeProcessAlive()` caused the periodic health check to take the immediate restart path on Windows, bypassing the consecutive-failure counter and busy-session guard added for #1294/#1295. This resulted in OpenChamber sending SIGTERM to a still-running child and restarting it unnecessarily during active sessions.

## Root cause

**Bug 1 — `hasChildProcessExited` wrapper mismatch**

`state.openCodeProcess` stores a wrapper object `{ url, pid, close }` from `createManagedOpenCodeServerProcess()`. The wrapper has no `exitCode` or `signalCode` properties, so the original check `child.exitCode !== null || child.signalCode !== null` always evaluated to `true` (since `undefined !== null` is `true`). The `process.kill` probe was never reached on any platform.

**Bug 2 — `process.kill` EPERM not handled**

On Windows, `process.kill(pid, 0)` throws `EPERM` when the caller lacks permission to signal the target process — even though the process is still alive. The bare `catch { return false }` unconditionally treated this as process death.

## Fix

1. **`hasChildProcessExited`** now checks for the presence of `exitCode`/`signalCode` properties before comparing them. Wrapper objects (no such properties) return `false`, causing the caller to fall through to the `process.kill` probe.

2. **`process.kill` catch block** now returns `true` on `EPERM`, matching the existing pattern in `managed-process-registry.js:118-119`.

## Tests

Three regression tests added to `lifecycle.test.js`:

- **Wrapper object** — verifies `{ url, pid, close }` is correctly identified as not-exited, falling through to `process.kill`.
- **EPERM (Windows)** — verifies `process.kill(pid, 0)` throwing EPERM is treated as "process alive", preventing false restarts.
- **ESRCH (dead process)** — verifies ESRCH correctly triggers a restart.

## Verification

- All 87 test files pass (696 tests, 2 skipped).
- All 6 callers of `hasChildProcessExited` verified safe: only `isManagedOpenCodeProcessAlive` receives the wrapper; the other 5 receive raw `ChildProcess` objects and are unaffected.
- EPERM pattern matches the project convention used in `managed-process-registry.js` and the VS Code parity implementation.

## Related

- #1294, #1295 — health request timeouts under load
- #1721 — false results from PID-only liveness checks (different code path)

### DIFF
--- a/packages/web/server/lib/opencode/lifecycle.js
+++ b/packages/web/server/lib/opencode/lifecycle.js
@@ -60,7 +60,13 @@ export const createOpenCodeLifecycleRuntime = (deps) => {
     }
   };
 
-  const hasChildProcessExited = (child) => !child || child.exitCode !== null || child.signalCode !== null;
+  const hasChildProcessExited = (child) => {
+    if (!child) return true;
+    // Wrapper object from createManagedOpenCodeServerProcess — no exitCode/signalCode.
+    // Fall through to process.kill probe instead.
+    if (!('exitCode' in child) && !('signalCode' in child)) return false;
+    return child.exitCode !== null || child.signalCode !== null;
+  };
 
   const isManagedOpenCodeProcessAlive = () => {
     const child = state.openCodeProcess;
@@ -69,8 +75,9 @@ export const createOpenCodeLifecycleRuntime = (deps) => {
     try {
       process.kill(child.pid, 0);
       return true;
-    } catch {
-      return false;
+    } catch (error) {
+      // EPERM = process exists but we lack permission to signal it → still alive.
+      return error?.code === 'EPERM';
     }
   };
 
--- a/packages/web/server/lib/opencode/lifecycle.test.js
+++ b/packages/web/server/lib/opencode/lifecycle.test.js
@@ -67,7 +67,7 @@ const createRuntime = (overrides = {}) => {
     resolvedWslDistro: null,
   };
 
-  return createOpenCodeLifecycleRuntime({
+  const runtime = createOpenCodeLifecycleRuntime({
     state,
     env: {
       ENV_CONFIGURED_OPENCODE_PORT: 45678,
@@ -102,6 +102,8 @@ const createRuntime = (overrides = {}) => {
     })),
     ...overrides,
   });
+
+  return { runtime, state };
 };
 
 describe('OpenCode lifecycle', () => {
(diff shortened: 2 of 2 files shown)


AhmedRMunir:
Can we please merge this PR as this bug has been plaguing my sessions and I constantly have to restart openchamber completely to keep my work going. it is so frustrating especially when it happens overnight when the agents are running scheduled tasks.

darknessest:
I independently reproduced this on Ubuntu with OpenChamber 1.16.3 and OpenCode 1.18.5, using the managed local server with four concurrent agent streams. A transient health failure logged `OpenCode process exited, restarting...` and killed the active streams. The wrapper had neither `exitCode` nor `signalCode`, so `undefined !== null` triggered an immediate restart and skipped the retry and busy-session safeguards.

The fix resolves the issue consistently for me.

AhmedRMunir:
@darknessest I added some info under the original bug here if that helps you out: https://github.com/openchamber/openchamber/issues/2258#issuecomment-5094906623

btriapitsyn:

PR #2272, opened 2026-07-16
fix(sessions): route new sessions to the correct project when server omits directory (#1637, #2270)

### LINKED ISSUE #1637: [Bug]

I am experiencing a  issue, and it has been bothering me for almost ten days.

When I click the `+` button inside a project to start a new conversation, the file directory shown in the file panel is not the directory of that project at all. If I continue the conversation, the files that the assistant can read are also from the directory shown in the file panel, rather than from the project directory I intended to open.

In addition, this conversation is then grouped under the project corresponding to the directory shown in the file panel, instead of the project where I originally started the conversation.

### Where does it happen?

Desktop (macOS)

### Version (if known)

1.12.4 

_No response_

### Logs (optional)

```shell

```

### LINKED ISSUE #2270: [Bug] New sessions from nested Git projects target a sibling child project

When multiple child directories are registered as OpenChamber projects beneath a parent Git repository, creating a new draft session from a child directory that has its own Git repository can target a different child project instead.

### Steps to reproduce

1. Create a parent directory, for example `workspace`.
2. Initialize a Git repository in the parent directory.
3. Create several child directories:
   - `project-a`, with no Git repository of its own
   - `project-b`, with its own Git repository
   - `project-c`, with its own Git repository
4. Add the child directories to the OpenChamber desktop app as separate projects. Do not rely on the parent directory being added as a project.
5. Click the `+` button to start a new draft session from `project-b` or `project-c`.
6. Create/send the session.

### Actual behavior

The new session is created or grouped under `project-a`, the child without its own Git repository, instead of the child project whose `+` button was clicked.

The behavior is reproducible when the parent directory is the Git ancestor and the registered child projects have mixed Git initialization.

A child project located outside the parent Git repository is not affected. This control case indicates that the parent-repository boundary, rather than the project name or sidebar ordering, is the relevant condition.

### Expected behavior

The new session should use the directory of the explicitly selected child project. The presence of an ancestor Git repository, or the Git state of sibling child projects, should not change the target selected by the project-level `+` action.

### Suspected cause

The project/worktree resolution path appears to inherit the ancestor repository when classifying nested project directories. During new-session creation or subsequent session grouping, that resolution can select the wrong registered child project instead of honoring the explicit project and directory supplied by the sidebar action.

This report isolates the repository topology that makes the existing session/directory mismatch consistently reproducible. Related report: #1637.

### Where does it happen?

Desktop (Windows)

### Version (if known)

- OpenChamber: `1.16.1`
- OpenCode: `1.18.2`

Not available

### Logs (optional)

N/A

bashrusakh:
## Summary

Fixes two related bugs where a newly created session is grouped under the wrong project when the server response omits the `directory` field.

Fixes #1637
Fixes #2270

## Issues

- **#1637** — Clicking `+` on one project creates a session grouped under a different project.
- **#2270** — New sessions from nested Git projects target a sibling child project instead of the selected child.

Both reports describe the same root cause: `createSession()` passes the server response's `directory` field to `setCurrentSession` as `directoryHint`. When the server response omits the field, `directoryHint` is `null` and `setCurrentSession` falls back to `opencodeClient.getDirectory()`, which can hold a stale directory from a different project.

## Root cause

```ts
// Before
const session = await opencodeClient.createSession({...}, directoryOverride ?? dir())
const sessionDirectory = (session as { directory?: string | null }).directory ?? null
useSessionUIStore.getState().setCurrentSession(session.id, sessionDirectory)
```

When the server returns a session without a `directory` field, `sessionDirectory` is `null`. `setCurrentSession` then falls back to `opencodeClient.getDirectory()`, which can be stale.

## Fix

Capture the effective directory passed to the SDK and reuse it as the fallback for the server response:

```ts
// After
const effectiveDirectory = directoryOverride ?? dir()
const session = await opencodeClient.createSession({...}, effectiveDirectory)
const sessionDirectory = (session as { directory?: string | null }).directory ?? effectiveDirectory ?? null
useSessionUIStore.getState().setCurrentSession(session.id, sessionDirectory)
```

This guarantees `setCurrentSession` and `registerSessionDirectory` always see the directory the user explicitly selected for the new session, even when the server response omits the field.

## Changed files

- `packages/ui/src/sync/session-actions.ts` — 2-line fix in `createSession`
- `packages/ui/src/sync/__tests__/issue-1637-2270.test.ts` — 7 regression tests covering both reported topologies plus preservation paths

## Fix level

Shared, root-level. The bug is in the common session creation path used by all callers. The same fix resolves both #1637 and #2270.

## Test coverage

| Scenario | Expected behavior |
|---|---|
| `directoryOverride` set, server omits `directory` | `setCurrentSession` receives `directoryOverride` |
| `directoryOverride` set, server returns `directory` | `setCurrentSession` receives server's value (server wins) |
| No `directoryOverride`, server omits `directory` | `setCurrentSession` receives `null` (preserved) |
| No `directoryOverride`, server returns `directory` | `setCurrentSession` receives server's value |
| Nested Git projects (#2270) | Child project directory is used, not parent or sibling |
| `registerSessionDirectory` with override | Routing index uses effective directory |
| `registerSessionDirectory` with server response | Routing index uses server's authoritative value |

## Verification

- `bun test packages/ui/src/sync/__tests__/issue-1637-2270.test.ts` — 7 pass, 0 fail
- `bun run type-check` — all packages pass
- `bun run lint` — all packages pass

## Out of scope (intentionally)

- The partial fix on `fix/2257-text-selection-menu-y` (commit `fac3c7f48`) that adds `selectedProjectId` to `openNewSessionDraft` options is not needed for this fix — the root cause is in `createSession`, not in the draft opener.
- `setCurrentSession`'s fallback chain is correct as-is; the problem was that `createSession` was not passing the right hint.
- `worktreeSessionCreator.ts` already passes both `selectedProjectId` and `directoryOverride`, so no changes needed there.

### DIFF
--- a/packages/ui/src/lib/projectResolution.ts
+++ b/packages/ui/src/lib/projectResolution.ts
@@ -59,5 +59,5 @@ export const resolveProjectForSessionDirectory = (
   availableWorktreesByProject: Map<string, WorktreeMetadata[]>,
   directory: string | null,
 ): ProjectEntry | null =>
-  resolveProjectFromWorktreeDirectory(projects, availableWorktreesByProject, directory) ??
-  resolveProjectForDirectory(projects, directory);
+  resolveProjectForDirectory(projects, directory) ??
+  resolveProjectFromWorktreeDirectory(projects, availableWorktreesByProject, directory);
--- a/packages/ui/src/sync/__tests__/issue-1637-2270.test.ts
+++ b/packages/ui/src/sync/__tests__/issue-1637-2270.test.ts
@@ -1,5 +1,8 @@
 import { beforeEach, describe, expect, mock, test } from "bun:test"
 import type { Session } from "@opencode-ai/sdk/v2/client"
+import type { ProjectEntry } from "@/lib/api/types"
+import type { WorktreeMetadata } from "@/types/worktree"
+import { resolveProjectForSessionDirectory } from "@/lib/projectResolution"
 
 // Recorded call info
 const setCurrentSessionCalls: Array<{ id: string | null; directoryHint: string | null | undefined }> = []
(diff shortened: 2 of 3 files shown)


btriapitsyn:
Addressed the remaining #2270 project-resolution gap in commit `a4ad20417`.

Changes:

- Prefer the longest matching registered project path before consulting worktree metadata. This prevents an ancestor Git worktree leaked under a sibling child project from overriding the explicitly targeted registered child.
- Preserve worktree-based resolution as the fallback for external worktree directories that are outside every registered project path.
- Added regression coverage for the exact parent-Git/sibling-child topology from #2270 and for the external-worktree fallback.
- Kept the existing `effectiveDirectory` fallback, which covers the separate missing `session.directory` race described in #1637.

Validation:

- `bun test packages/ui/src/sync/__tests__/issue-1637-2270.test.ts` (9 pass)
- `git diff --check`

btriapitsyn (approved):

PR #1876, opened 2026-06-28
fix(quota): show weekly window for Zhipu AI Coding Plan

bigcoder84:
## Background

Zhipu AI Coding Plan (`open.bigmodel.cn`) ships **two** concurrent token limits per the official docs ([套餐概览](https://docs.bigmodel.cn/cn/coding-plan/overview), [FAQ](https://docs.bigmodel.cn/cn/coding-plan/faq)):

| Window         | Reset cycle                         |
| -------------- | ----------------------------------- |
| 5-hour rolling | resets 5h after first use           |
| Weekly         | 7-day cycle from subscription start |

The weekly cap was introduced in **Feb 2026**. This provider was written before the weekly limit existed — it only handled the 5-hour window and was never updated when Zhipu introduced the weekly cap, so the weekly quota has been silently ignored.

The `/api/monitor/usage/quota/limit` endpoint reflects both as separate `TOKENS_LIMIT` entries:

```json
"limits": [
  { "type": "TOKENS_LIMIT", "unit": 3, "number": 5, "percentage": 14, "...": "..." },
  { "type": "TOKENS_LIMIT", "unit": 6, "number": 1, "percentage": 60, "...": "..." }
]
```

## Changes

Minimal, scoped to the **Zhipu (CN) provider** across web + VS Code runtimes (the shared unit map is also used by `z.ai`, but adding a key is backward-compatible and doesn't change `z.ai` behavior — left untouched per scope).

| File                                                           | Change                                                                        |
| -------------------------------------------------------------- | ----------------------------------------------------------------------------- |
| `packages/web/server/lib/quota/utils/transformers.js`            | Add `6: 604800` (week) to `ZAI_TOKEN_WINDOW_SECONDS`                              |
| `packages/web/server/lib/quota/providers/zhipuai-coding-plan.js` | `find` → `filter` loop; use `resolveWindowLabel` so each window gets `5h` / `weekly` |
| `packages/vscode/src/quotaProviders.ts`                          | Mirror the same two changes for the VS Code bridge                            |

After:
- Both `TOKENS_LIMIT` entries render as separate windows (`5h` + `weekly`).
- Keys are now `5h` / `weekly`, which **already** have full i18n (`quota.window.5h` / `quota.window.weekly`) and are recognized by `inferWindowSeconds` for pace — so this also fixes label localization and pace for legacy 5-hour-only accounts.

## Before / After

|                    | Before               | After                           |
| ------------------ | -------------------- | ------------------------------- |
[...]

PR #2274, opened 2026-07-16
fix(quota): show weekly usage window for z.ai and Zhipu AI providers

lightrabbit:
## Problem

The z.ai and Zhipu AI coding plan APIs return **multiple** `TOKENS_LIMIT` entries in the `limits` array — one for the 5-hour rolling window and one for the weekly window. However, both providers used `find()` which only picks the **first** match, silently dropping the weekly quota window.

Additionally, the shared `resolveWindowSeconds` mapping table only knew about `unit: 3` (hour). The weekly window uses `unit: 6`, so even if `find()` were changed to `filter()`, the seconds calculation would return `null` and the label would degrade to `'tokens'`.

## Fix

1. **`transformers.js`** — Added `unit: 6` (week = 604800s) to `ZAI_TOKEN_WINDOW_SECONDS`.
2. **`zai.js`** — Switched `find()` → `filter()` and loop over all `TOKENS_LIMIT` entries, using `resolveWindowLabel` for each window key (`5h`, `weekly`).
3. **`zhipuai-coding-plan.js`** — Same `find()` → `filter()` change, plus replaced the hardcoded `'Tokens'` window key with `resolveWindowLabel` so both `5h` and `weekly` windows get proper labels.
4. **`quotaProviders.ts`** (VS Code) — Applied the identical fix to the extension's independent quota implementations for cross-runtime parity.

## Before / After

**Before** (z.ai & Zhipu AI showed only one card):
- `5h` — 4%

**After** (matches the official z.ai/Zhipu web dashboard):
- `5h` — 4%
- `weekly` — 46%
- `MCP Tools` — 0% (Zhipu AI only, already worked)

## Testing

Verified against live z.ai and Zhipu AI APIs — both return 3 `TOKENS_LIMIT`/`TIME_LIMIT` entries. After the fix, the OpenChamber usage page correctly displays all windows with proper reset times and percentages, matching the official dashboard.

### DIFF
--- a/packages/vscode/src/quotaProviders.ts
+++ b/packages/vscode/src/quotaProviders.ts
@@ -183,7 +183,7 @@ const resolveGoogleWindow = (sourceId: GoogleAuthSource['sourceId'], resetAt: nu
   return { label: 'daily', seconds: GOOGLE_DAILY_WINDOW_SECONDS } as const;
 };
 
-const ZAI_TOKEN_WINDOW_SECONDS: Record<number, number> = { 3: 3600 };
+const ZAI_TOKEN_WINDOW_SECONDS: Record<number, number> = { 3: 3600, 6: 604800 };
 
 const readAuthFile = (): AuthFile => {
   if (!fs.existsSync(AUTH_FILE)) {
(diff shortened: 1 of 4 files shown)


lightrabbit:
Closing in favor of #1876 which already covers the Zhipu AI side. The z.ai provider fix is now submitted as a follow-up PR targeting the same branch.

PR #2275, opened 2026-07-16
fix: resolve config-only providers and {env:...} placeholders in small model calls

### LINKED ISSUE #2269: [Bug] Session goal auditor fails to expand `{env:VAR}` placeholders in provider options (401 Unauthorized)
### Describe the Bug

When utilizing the brand-new **Session Goals** feature (`v1.16.0+`), the background progress auditor fails to execute and flags the goal as `Blocked (progress audit unavailable)`. 

Tracing the OpenChamber `main.log` reveals that the background auditor process fails with a **401 Unauthorized** error because it attempts to send the raw placeholder string (e.g., `"{env:PROVIDER_API_KEY}"`) rather than expanding the actual environment variable.

While the primary OpenCode engine properly resolves `{env:...}` blocks specified in `opencode.jsonc` during active chat and tool-use turns, **OpenChamber's internal server-side session goal auditor does not expand environment variables in the provider options block when instantiating the model.**

---

### Log Trace

```log
[session-goal] audit failed: custom-provider request failed with 401: 
{
  "error": {
    "message": "API Virtual Key expected. Received={env****KEY}, expected to start with 'sk-'.",
    "type": "auth_error",
    "code": "401"
  }
}
[session-goal] ses_0973586a4ffeVoPT8amZsTImUG audit unavailable, continuing unaudited (1/2)
[session-goal] ses_0973586a4ffeVoPT8amZsTImUG settled as blocked (progress audit unavailable)
```

---

### Steps to Reproduce

1. Define a custom provider in `opencode.jsonc` using an environment variable placeholder for the API key:
   ```json
   "custom-provider": {
     "name": "Custom Provider",
     "npm": "@ai-sdk/openai-compatible",
     "options": {
       "baseURL": "https://api.example.com/v1",
       "apiKey": "{env:CUSTOM_PROVIDER_API_KEY}"
     }
   }
   ```
2. Set the environment variable in your terminal environment:
   ```bash
   export CUSTOM_PROVIDER_API_KEY="sk-valid-key-here"
   ```
3. Open OpenChamber and configure it to use this custom provider/model.
4. Arm a **Session Goal** via the composer target button and send a prompt.
5. Once the active turn completes, the server-side goal loop will kick off the independent auditor.
6. The auditor will immediately crash with a `401` auth error because the HTTP request was sent with the literal headers/options string `"{env:CUSTOM_PROVIDER_API_KEY}"` instead of `"sk-valid-key-here"`.

---

### Expected Behavior

OpenChamber's background session-goal auditor loop should process and expand `{env:...}` tokens in provider `options` blocks prior to model instantiation, matching the parsing behavior of the primary OpenCode engine.

---

### Environment / Version

- **OpenChamber Version:** `1.16.0+`
- **Deployment:** Desktop App / Local Web Server
- **Platform:** macOS / Linux

IbrahimKhan12:
> Validated commit: `93e81f4e8c97e137cad86a4ccb63fa6123e4b738`.

## Intent

Fixes #2269. A provider declared only in `opencode.jsonc` through `provider.<id>.options.apiKey` is now eligible for small-model selection. `resolveSmallModel` previously gated every step on `isUsableAuthEntry` against `auth.json`, so a config-only provider was never selected, the goal auditor had no model to call, and the goal settled as `blocked (progress audit unavailable)`.

Three changes make that work:

- `listConfigCallableProviders` reports config providers whose credential and direct-call endpoint both resolve now. `disabled_providers` and `enabled_providers` apply first, the opencode zen sentinel is refused, and an unset `{env:NAME}` or unreadable `{file:path}` is not callable.
- `resolveSmallModel` accepts that callable set through `callableProviderIDs`, so a config-only session provider can supply a family-priority small model or fall back to the session model. The scan is skipped when a Settings override or `small_model` already decides the answer.
- `listAuthenticatedProviders` is directory-scoped and `GET /api/small-model` threads `?directory=`, so the pickers list what is configured for the project they describe.

## Non-goals

No support for credential-chain providers such as Bedrock, Azure, or GitLab.

`{env:}` and `{file:}` apiKey substitution, including the literal 401 in the issue report, already shipped on `main` in `f11c5dedc`. This PR reuses it and does not reintroduce it.

An earlier revision carried a port of OpenCode's config-layer discovery and precedence. That work was not required for #2269, and `main` has since landed `readConfigLayers` independently, so the port is absent from the rebased diff. The previous `shared.js` and quota/Zhipu review findings therefore no longer apply.

Provider enable/disable policy is applied to config credentials only. `auth.json` and runtime-discovered providers keep their existing behavior. `packages/vscode/src/opencodeConfig.ts` reads config through its own merge and is untouched.

## Affected surfaces

- `packages/web/server/lib/small-model/call.js`: adds `listConfigCallableProviders` and `BUILT_IN_ENDPOINT_PROVIDERS`, which names the four dispatch branches (`github-copilot`, `openai`, `anthropic`, and `google`) that carry their own URL. Every other config-only provider needs a `baseURL` from config or the catalog.
- `packages/web/server/lib/small-model/resolve.js`: accepts `callableProviderIDs`; when omitted it falls back to the previous `auth.json` test, so existing callers are unaffected.
- `packages/web/server/lib/small-model/index.js`: builds one callable set for resolution and the pickers.
- `packages/web/server/lib/small-model/routes.js`: passes `directory` to both `describeSmallModel` and `listAuthenticatedProviders`.
- `packages/ui`: `fetchCallableSmallModelProviders` and `useCallableSmallModelProviders` replace three copies of the same inline effect in `DefaultsSettings`, `WalkthroughView`, and `WalkthroughBlocker`. Settings uses `useSettingsDirectory`; walkthrough uses its resolved git directory.

Route contract: `directory` is a new optional query parameter and the response shape is unchanged. A request without it resolves config as before. `authenticatedProviders` still carries provider IDs only.

Runtime parity: `/api/small-model` was already reached through `runtimeFetch` from shared UI, and `query` is an existing `runtimeFetch` option, so web, desktop, VS Code, hosted mobile, and Capacitor mobile keep one code path. No runtime-specific branch was added.

## Repository guidance

|---|---|---|
| `openchamber-change-discipline` | Server and shared-UI source, a new module export, and a route contract change | The callable set is built once in `index.js` and consumed by resolution and the pickers. Scope stayed on #2269, and `small-model/DOCUMENTATION.md` was updated for the directory-scoped picker rule and Claude Code placement. |
| `ui-api-decoupling` | A server route contract, credential handling, and shared UI fetch change | Credential resolution and dispatch stay server-side; the route returns provider IDs only. The UI uses `runtimeFetch` with its `query` option. `fetchCallableSmallModelProviders` validates the response with zod and throws on non-OK or malformed responses instead of turning failure into authoritative `[]`. |
| `sync-state-invariants` | The new hook is a directory-scoped cache whose key can change while a request is in flight | The hook stores `{ directoryKey, providers }` together and publishes only when the stored key matches the current key. A late response for the previous directory cannot become the current answer, and fetch failure leaves the value `undefined`. |
| `small-model/DOCUMENTATION.md` | Owning module documentation | Updated for `listAuthenticatedProviders(directory)`, the callable set feeding model selection, config credential exclusions, and why the Claude Code refusal is applied to the offer rather than the callable set. |

## Validation

| Check | Result |
|---|---|
| `cd packages/web && bun run test` | 187 files, 2089 passed, 2 skipped, 0 failed. |
| `cd packages/ui && bun run test` | 421/421 test files passed with the isolated runner. |
| `bun run lint:ui` | Exit 0. One pre-existing warning remains in untouched `MobileChangesSurface.tsx`. |
| `bun run dead-code` | No new findings versus the pre-change tree. |
| `bun run type-check:web`, `bun run lint:web` | Exit 0, but these commands do not cover `packages/web/server/**`; the web tsconfig includes `src` and `../ui/src`, and the lint script globs only `./src/**/*.{ts,tsx}`. |
| Claude Code negative check | Reverting the exclusion into `collectCallableProviderIDs` makes `still refuses a session already running on Claude Code` fail with `404` instead of `422`, confirming the regression test locks the behavior. |
| Desktop web visual check at `93e81f4e` | Passed with an isolated `opencode.jsonc`, empty auth store, and the Settings Small Model picker. The unmodified HEAD response exposes `config-only`; the control response does not. Screenshots are below. |

Not validated: no real provider request or secret-bearing network call was sent. Electron, VS Code, and mobile runtimes were not exercised.

## Visual evidence

The desktop web run used isolated OpenChamber/XDG data directories, no `auth.json`, and an `opencode.jsonc` entry named `config-only` with a fake API key, an invalid non-routable endpoint, and one model named `Evidence Mini`. No model request was sent.

The control keeps the same config and HEAD UI but removes `config-only` from the `/api/small-model` response, reproducing the pre-fix auth-only picker result:

Control: config-only provider filtered from the response

The unmodified route at HEAD `93e81f4e` includes the config-only provider:

HEAD 93e81f4e: config-only provider is callable

This changes picker contents only. It does not change layout, styling, or responsive behavior, so no additional narrow/wide or theme matrix is needed.

## Risks and failure behavior

Claude Code stays refused, but the exclusion sits on the offer rather than the callable set. Applying it to the callable set would remove Claude Code from resolution too, turning the actionable 422 into a bare 404 for a session already running on it. A regression test covers this path.

`listConfigCallableProviders` restates the endpoint rule that `callSmallModel` applies, so the two can drift. `BUILT_IN_ENDPOINT_PROVIDERS` names the dispatch branches it mirrors. It is deliberately separate from `DEDICATED_WIRE_FORMAT_PROVIDERS`, which means "owns its credential handling" and includes the legacy `copilot` alias that has no dispatch branch of its own.

A config-only provider whose endpoint exists only inside a running OpenCode plugin is not added through the config path; `listRuntimeCallableProviders` still covers it when the runtime reports a credential.

Fetch failure leaves the hook value `undefined`, preserving each call site's prior behavior: Settings fails closed to `[]`, while walkthrough pickers fall back to offering every provider.

`readProviderConfig` resolves `{env:}` and `{file:}` per configured provider on each discovery pass. Nothing resolved reaches logs, errors, or the route response. An empty `enabled_providers: []` admits no config provider, matching its explicit allowlist meaning; a test covers it.

The extra discovery work is one `readConfig` for the whole scan plus one credential resolution per configured provider. Resolution skips the scan when a Settings override or `small_model` already answers.

### DIFF
--- a/packages/ui/src/components/sections/openchamber/DefaultsSettings.tsx
+++ b/packages/ui/src/components/sections/openchamber/DefaultsSettings.tsx
@@ -21,7 +21,8 @@ import { useSelectionStore } from '@/sync/selection-store';
 import { useSessionUIStore } from '@/sync/session-ui-store';
 import { useI18n } from '@/lib/i18n';
 import { parseModelIdentifier } from '@/lib/modelIdentifier';
-import { runtimeFetch } from '@/lib/runtime-fetch';
+import { useCallableSmallModelProviders } from '@/hooks/useCallableSmallModelProviders';
+import { useSettingsDirectory } from '@/hooks/useSettingsDirectory';
 import { isPrimaryMode } from '@/components/chat/mobileControlsUtils';
 
 const getDisplayModel = (
@@ -37,6 +38,7 @@ const getDisplayModel = (
 
 export const DefaultsSettings: React.FC = () => {
   const { t } = useI18n();
+  const settingsDirectory = useSettingsDirectory();
   const setProvider = useConfigStore((state) => state.setProvider);
   const setModel = useConfigStore((state) => state.setModel);
   const setAgent = useConfigStore((state) => state.setAgent);
@@ -69,7 +71,7 @@ export const DefaultsSettings: React.FC = () => {
   const [defaultAgent, setDefaultAgent] = React.useState<string | undefined>();
   const [smallModelUseDefault, setSmallModelUseDefault] = React.useState(true);
   const [smallModelOverride, setSmallModelOverride] = React.useState<string | undefined>();
-  const [smallModelProviders, setSmallModelProviders] = React.useState<string[]>([]);
+  const smallModelProviders = useCallableSmallModelProviders(settingsDirectory) ?? [];
   const [walkthroughModelOverride, setWalkthroughModelOverride] = React.useState<string | undefined>();
   const [isLoading, setIsLoading] = React.useState(true);
 
@@ -236,30 +238,6 @@ export const DefaultsSettings: React.FC = () => {
     () => getDisplayModel(walkthroughModelOverride),
     [walkthroughModelOverride]
   );
-  React.useEffect(() => {
-    // Both pickers offer the same providers — the walkthrough runs through the
-    // small model — and the walkthrough picker is always visible, so this is
-    // always worth fetching. The server answers with the providers it has a
-    // credential and an endpoint for, including plugin-registered ones that
-    // exist only inside the running OpenCode.
-    let cancelled = false;
-    (async () => {
-      try {
-        const response = await runtimeFetch('/api/small-model', { method: 'GET', headers: { Accept: 'application/json' } });
-        if (!response.ok) return;
-        const payload = await response.json().catch(() => null) as { authenticatedProviders?: unknown } | null;
-        if (!cancelled && Array.isArray(payload?.authenticatedProviders)) {
-          setSmallModelProviders(payload.authenticatedProviders.filter((id): id is string => typeof id === 'string'));
-        }
-      } catch {
-        // Fail closed: never offer providers whose credentials were not verified.
-      }
-    })();
-    return () => {
-      cancelled = true;
-    };
-  }, []);
-
   const availableVariants = React.useMemo(() => {
     if (!parsedModel.providerId || !parsedModel.modelId) return [];
     const provider = providers.find((p) => p.id === parsedModel.providerId);
--- a/packages/ui/src/components/views/walkthrough/WalkthroughBlocker.tsx
+++ b/packages/ui/src/components/views/walkthrough/WalkthroughBlocker.tsx
@@ -1,14 +1,15 @@
-import { useCallback, useEffect, useState } from 'react';
+import { useCallback, useState } from 'react';
 import { Icon } from '@/components/icon/Icon';
 import { ModelSelector } from '@/components/sections/agents/ModelSelector';
 import { Button } from '@/components/ui/button';
 import { useI18n } from '@/lib/i18n';
 import { useConfigStore } from '@/stores/useConfigStore';
-import { runtimeFetch } from '@/lib/runtime-fetch';
+import { useCallableSmallModelProviders } from '@/hooks/useCallableSmallModelProviders';
 import { updateDesktopSettings } from '@/lib/persistence';
 import type { WalkthroughBlockedState, WalkthroughModel } from '@/lib/walkthrough/types';
 
 interface WalkthroughBlockerProps {
+  directory: string;
   reason: WalkthroughBlockedState;
   model?: WalkthroughModel;
   requiredChars?: number;
@@ -25,6 +26,7 @@ const modelLabel = (model?: WalkthroughModel) =>
  * in place rather than sending the user to Settings to guess.
  */
 export const WalkthroughBlocker = ({
+  directory,
   reason,
   model,
   requiredChars,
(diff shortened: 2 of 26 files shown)


btriapitsyn:

mattv8:

IbrahimKhan12:
Addressed the review handoff at HEAD `93e81f4e`: the PR body describes only the focused callable-provider implementation, reports current validation, and includes control/current-HEAD screenshots of the Settings Small Model picker. I also removed manual hard wrapping. For this metadata-only amend, the PR body and refreshed screenshot were updated to `93e81f4e` before the branch was force-pushed; the amended commit has the same tree as `2301499fe`.

PR #2290, opened 2026-07-17
fix(sidebar): keep the permission badge clear of hover actions (#2284)

### LINKED ISSUE #2284: [Bug] In some cases, the cover style conflicts with the badge

I found a UI display issue during usage. It's not a serious problem, but rather a small detail regarding the UI that I believe is still worth reporting. When there are security prompts (for example, when I need to handle a "Permission Required" situation), the left sidebar indicates that there is a security prompt requiring my attention. However, when I attempt to click and return to that window, I unexpectedly noticed that the small icon on the right side of the button overlaps with this security prompt when hovered over, as shown in the attached image. I believe that in this case, the security prompt should shift to the left, moving to the left side of these small icons, while the title on the left reduces its width to accommodate the display — this would be a more expected behavior. I hope this small detail will be addressed in a future fix.

### Where does it happen?

Desktop (macOS)

### Version (if known)

1.16.1

Image

### Logs (optional)

```shell

```

IbrahimKhan12:
## Why

Closes #2284.

In **minimal display mode (the default)**, a session row that has a pending permission request keeps its red shield badge fully visible on hover. But the hover action icons (quick-archive + menu) are absolutely positioned at `right-0`, and minimal non–VS Code rows only reserve `pr-2` (8px) of hover padding — far too little to clear the ~34px of icons. The result is the icons overlapping the badge, exactly as the reporter's screenshot shows.

## Root cause

In `SessionNodeItem`, the inline date/goal label fades out on hover (`group-hover:opacity-0`), so the actions take its place and the title barely shifts — which is why minimal rows only reserve `pr-2`. The permission badge, however, sits at the right edge of the row flow and has **no hover-hide class**, so it stays opaque while the absolute actions render on top of it.

(Standard mode already reserves `pr-12` and VS Code minimal reserves `pr-14`/`pr-18`, so both already clear their icons — only minimal non–VS Code was affected.)

## What

- When a permission badge is present in minimal non–VS Code mode, reserve `group-hover:pr-14 group-focus-within:pr-14` instead of `pr-2`. On hover the title now shrinks and the badge shifts left, clear of the icons — the behavior the issue asks for. `pr-14` is the same padding the VS Code minimal layout already uses for the identical two `h-4` icons (archive + menu), so the value is consistent with existing code.
- Extracted the reveal-padding decision from an inline nested ternary into a pure `resolveRevealPaddingClass` helper in `sessionNodeItemUtils.ts`, using early returns per the CONTRIBUTING style guide. Standard and VS Code padding are byte-for-byte unchanged.

## Testing

- New `sessionNodeItemUtils.test.ts` (9 cases) locks in every branch. The `#2284` case is a genuine regression guard: with the pre-fix logic it returns `pr-2` and fails; with the fix it returns `pr-14` and passes. The 7 non-badge cases pass in both states, confirming the extraction is behavior-preserving.
- `bun test packages/ui/src/components/session/sidebar/` → 31 pass / 0 fail.
- `bun run type-check:ui`, `bun run lint:ui` → clean. `bun run dead-code` → no new findings.

Not run: live-session visual check (needs a backend session in a pending-permission state); the padding value is validated by the unit test and by matching the existing VS Code minimal two-action padding for the same icon geometry.

### DIFF
--- a/packages/ui/src/components/session/sidebar/SessionNodeItem.tsx
+++ b/packages/ui/src/components/session/sidebar/SessionNodeItem.tsx
@@ -25,7 +25,7 @@ import { buildSessionMessageRecordsSnapshot, useDirectoryStore, useGlobalSession
 import { useSync } from '@/sync/use-sync';
 import { useViewportStore, viewportSessionKey } from '@/sync/viewport-store';
 import { DraggableSessionRow } from './sessionFolderDnd';
-import { nodeContainsSessionId } from './sessionNodeItemUtils';
+import { nodeContainsSessionId, resolveRevealPaddingClass } from './sessionNodeItemUtils';
 import type { SessionNodeChildRenderExtras, SessionNodeRenderExtras } from './sessionNodeItemUtils';
 import type { SessionNode } from './types';
 import { formatSessionCompactDateLabel, formatSessionDateLabel, normalizePath, renderHighlightedText } from './utils';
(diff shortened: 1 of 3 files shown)


IbrahimKhan12:
Fixed by #2480; closing PR.

PR #2297, opened 2026-07-17
fix(files): stop file viewer reload loop from sub-ms mtime jitter (#1489)

### LINKED ISSUE #1489: File viewer infinite loading/display loop causes rapid text flicker
# Background

When viewing files in the UI, the viewer sometimes enters an infinite loading and display loop. The text rapidly flickers on screen, making the file content unreadable. The issue is intermittent but consistently reproducible across multiple file types.

https://github.com/user-attachments/assets/c7866114-4a44-4af0-be8f-54c896b27df3

# Version

OpenChamber version 1.11.7, OpenCode version 1.15.13

Web application/PWA
Chrome mobile android
Also observed in Chrome desktop

# Acceptance Criteria

- [ ] File viewer loads file content once and stops (no repeated loading cycle)
- [ ] File content displays stably without flicker
- [ ] Loading state resolves to a stable displayed or error state

---

Created with [create-issue](https://github.com/tomzx/dot-claude/blob/cd20e18/skills/create-issue/SKILL.md) (`cd20e18`)

IbrahimKhan12:
> Integrated and validated commit: `79d850afac3619e23e900596c48f6da6b95b1091`, after merging current main and protecting diagram saves against stale poll results.

## Intent

Fixes #1489. The open-file poller on `main` compares metadata with a strict mtime inequality, and any movement forces `setLoadedFilePath(null)`, which is a full reload with a load spinner. On filesystems that report sub-millisecond mtime jitter for an unchanged file, that produces an endless reload/flicker loop on a file nobody edited. The defect is still present on `main` at `FilesView.tsx` in the `changedByMtime` comparison.

Two narrow modules fix it. `hasFileStatChanged` treats a sub-millisecond mtime delta with identical size as unchanged. `createFileContentPoller` re-reads content only after metadata actually moved, and swaps it into the editor in place only when it differs from what is loaded, so a real external edit no longer flashes the spinner and an unchanged file does nothing at all.

The tick order in `FilesView` is: read stat, return when `!hasFileStatChanged`, then for text files at or under `MAX_CONTENT_POLL_BYTES` poll content, otherwise return when the buffer is dirty and finally advance the baseline and reload. The cheap stat comparison runs before any content read, so content is never re-fetched on a tick where metadata did not move.

## Non-goals

No filesystem-watcher replacement; this stays a 2 s interval poll. No in-place content path for images, PDFs, binaries, or files above the content-poll byte limit; those still take the full reload.

## Affected surfaces

- `packages/ui` Files view polling for web, desktop, and mobile. VS Code does not mount FilesView; its shared FilesAPI remains compatible with the fresh read option. New `fileStatChange.ts` and `fileContentPoller.ts`, and a new `applyLoadedTextContent` path in `FilesView.tsx` that tracks loaded content and a load revision.
- `packages/ui/src/contexts/content-cache-owner.ts` and `FileReadOptions` gain a `fresh` option that drops the cache entry and delegates to the underlying read. `packages/web/src/api/files.ts` maps it to `cache: 'no-store'`, as does the `FilesView` fallback reader, so a cached body cannot masquerade as unchanged content.
- User-visible states: the reload spinner no longer appears for an unchanged polled file, and an external text edit updates in place. Dirty buffers, including draw.io diagrams whose XML refs diverged from the saved refs, are never overwritten.
- No stat serialization change. The web stat endpoint already returned raw `stats.mtimeMs`; nothing in this diff touches it.

## Rebase note

`main` moved substantially under this branch, and the rebase is not a mechanical conflict resolution. Reviewers should look at two adaptations:

1. **`readFile` and `readFileStat` signatures.** `main` refactored both to take only `path` and resolve outside-workspace options internally through `resolveFileReadOptions(path)`. The pre-rebase branch threaded an explicit options object plus `fresh`. Rather than reverting that refactor, `readFile` now takes a second `cacheOptions?: { fresh?: boolean }` argument carrying only the freshness intent, and the poller calls `readFile(selectedPath, { fresh: true })` and `readFileStat(selectedPath)`. This made the change smaller than the original: the poll effect no longer needs `selectedFileReadOptions` at all.
2. **Non-fresh reads are byte-identical to `main`.** The inline fallback passes a second `runtimeFetch` argument only when `fresh` is set; `runtimeFetch`'s `init` parameter defaults to `{}`, so omitting it and passing `undefined` are equivalent.

Two upstream additions inside the conflicting regions were preserved deliberately: `main`'s `subscribeToFileContentInvalidation` effect, and the git-refresh plus `syncScheduledTaskLoops` calls after a successful save. `fresh` is intentionally excluded from the content-cache key — `cacheKey` derives from `directory` and `path` only, so the bypass removes the real entry rather than a `fresh`-keyed phantom.

## Repository guidance

|---|---|---|
| `openchamber-change-discipline` | Shared UI behavior, a runtime read option, and two new modules changed | The comparison and the poll loop are separate pure/injectable units with their own tests rather than more inline logic in `FilesView`; the `fresh` option is added to the shared type, the cache owner, and the web implementation together so no runtime silently ignores it. `bun run dead-code` was run because files and exports were added. |
| `sync-state-invariants` | This is a polling path that reconciles external state against a local baseline | Failure is not treated as empty success: the stat baseline advances only when a poll actually observed current content, so a failed read or dirty buffer cannot retire a pending change and the next tick retries. Stale completions are rejected by an O(1) load-revision comparison rather than by content diffing. |
| `performance-engineering` | Polling is a repeated timer path and content reads are the expensive part | The stat comparison gates every content read; the poller is single-flight, the interval skips a tick while one is in flight and while `document.hidden`, and reads are capped at `MAX_CONTENT_POLL_BYTES`. |
| `ui-api-decoupling` | The change adds a read option that runtimes must honor | Reconciliation stays entirely in shared UI; the runtime contract gains only `fresh`, and the web files API implements it as an HTTP cache bypass. |

## Validation

Integration checks at `79d850afac3619e23e900596c48f6da6b95b1091`:

| Check | Result |
|---|---|
| Focused `fileContentPoller`, `fileStatChange`, and `content-cache-owner` tests | 18 passed, 0 failed. |
| Web files API and walkthrough route tests | 13 passed, 0 failed. Current main supplies the synchronization fix for the previously failing walkthrough CI test. |
| Deferred poll across diagram saves | A temporary regression runner executed the actual source-extracted `saveDiagramXml`, `applyLoadedTextContent`, and poller callbacks. It failed before the integration fix and passed afterward: an older confirmation read no longer replaces the saved diagram baseline. |
| `bun run lint:ui` | Passed, with one pre-existing MobileChangesSurface hook warning. |
| Web production build | Passed. |
| `bun run dead-code` | Completed and inspected; no new poller module or export reported. Existing repository findings remain. |
| Oxlint on affected source | Fixed one PR-added conditional empty spread. The 14 remaining findings are pre-existing lines outside the follow-up. |
| Full suite | Passed in the required [GitHub checks run](https://github.com/openchamber/openchamber/actions/runs/33974636918) at this HEAD, along with build, type-check, lint, generated-changelog verification, and Electron Linux packaging tests. |

Successful diagram saves now call `applyLoadedTextContent`, sharing the text-save baseline and revision protection. This closes the delayed-confirmation-read case without adding another save path or state owner.

No browser, Electron, or mobile Files-view interaction or before/after frame recording was performed during integration. The deterministic polling and save-ordering paths were checked; physical filesystem jitter remains unverified here.

## Visual evidence

Not captured, and this is a stated constraint rather than an exemption.

This is a flicker/motion defect, so `CONTRIBUTING.md` asks for a recording. The only browser available in this environment is a headless Chromium shell (`chromium_headless_shell`), with no headed Chrome; a still frame from a headless run cannot demonstrate the presence or absence of flicker, and no screen-recording path exists here. Rather than attach an artifact that does not actually show the behavior, the reproduction is given so a reviewer can confirm it directly:

1. Open any text file in the Files view.
2. Run `while :; do touch <that-file>; sleep 1; done`.

On `main` the view reloads and the spinner flashes on every `touch`. With this change the file stays put, no spinner appears, and the content is only swapped when the bytes actually differ.

The deterministic part of the behavior is asserted in `fileContentPoller.test.ts` and `fileStatChange.test.ts`, which encode the exact decisions a recording would demonstrate: unchanged metadata performs no read, a sub-millisecond jitter with equal size is not a change, and content is applied only when two consecutive fresh reads agree.

## Risks and failure behavior

- Tradeoff: a same-size external edit whose mtime lands within 1 ms of the recorded baseline is ignored until the next metadata movement. That is the direct cost of the jitter tolerance.
- The stat baseline advances only when a poll actually observed current content, so a failed read or a dirty buffer cannot retire a pending change; the next tick retries instead.
- Content is applied only when two consecutive fresh reads agree and neither the dirty state nor the load revision moved, so a write racing the poll is dropped rather than pasted over local edits.
- A poll that returns newly binary content triggers a full reload so the binary guards run, instead of writing binary into the editor.
- `fresh` bypasses both the shared content cache entry and the HTTP cache, so polling a large file costs a real read each time metadata moves; the byte cap bounds that.
- Rollback is reverting the squash merge. The new modules are only reachable from the Files view poll path, and no persisted data or server contract changes.


btriapitsyn:

mattv8:

IbrahimKhan12:

This one is not a mechanical conflict resolution, so two adaptations are worth a look:

1. `main` refactored `readFile`/`readFileStat` to take only `path` and resolve outside-workspace options internally via `resolveFileReadOptions`. Instead of reverting that, `readFile` now takes a second `cacheOptions?: { fresh?: boolean }` carrying only the freshness intent. The poll effect got smaller as a result — it no longer needs `selectedFileReadOptions` at all.
2. Non-fresh reads are behaviourally identical to `main`: the extra `runtimeFetch` argument is passed only when `fresh` is set, and `init` already defaults to `{}`.

`main`'s `subscribeToFileContentInvalidation` effect and the post-save git-refresh/`syncScheduledTaskLoops` calls both sat inside conflicting regions and are preserved.

Validation on the new HEAD: 18 helper tests, 7 web files-API tests, `type-check:ui` and `lint:ui` clean, and `dead-code` clean for the new files. Targeted eslint on the four changed files is the check that confirms the merged hook dependency arrays are right. The remaining `lint:ui` warning is in `MobileChangesSurface.tsx`, which this PR does not touch and which is byte-identical to `origin/main`.

btriapitsyn:
Thanks @IbrahimKhan12 for rebasing this and preserving the current file-read and invalidation paths. I updated the branch from main and added a small integration fix so diagram saves share the stale-poll protection used by text saves. Focused tests and the complete required GitHub checks passed on the integrated head. Merged; this closes #1489. #3023 is related partial coverage and remains open because its diff-view and highlighting requests are broader.

Thanks @mattv8 for tracking this PR during the backlog triage.

PR #2300, opened 2026-07-18
fix(settings): preserve notification template edits

AmanTahiliani:
## Intent

Prevent rapid consecutive notification-template updates from overwriting one another by applying each edit against the latest store state.

## Non-goals

No notification template UI, copy, persistence format, or cross-runtime behavior changes.

## Affected surfaces

- Shared UI notification settings state and edit handling.
- The existing persistence caller continues to pass a template object and retains its prior behavior.

## Repository guidance

| Source | Applied guidance |
| --- | --- |
| `AGENTS.md` | Kept the consistency guarantee in the UI store rather than the component alone. |
| `packages/ui/src/stores/DOCUMENTATION.md` | Preserve unaffected notification-template entries while updating one event. |
| `openchamber-change-discipline` | Covered the functional setter path with a focused store regression test and inspected `dead-code` because the setter contract and test-file inventory changed. |

## Validation

- `bun test packages/ui/src/stores/useUIStore.notificationTemplates.test.ts` - passed: 1 test across 1 file, 0 failures.
- `bun run type-check:ui` - passed.
- `bun run lint:ui` - passed.
- `bun run dead-code` - completed successfully; reported the pre-existing repository-wide inventory of 179 unused exported types, 1 duplicate export, and 4 configuration hints. This PR did not add an item to that report.

## Visual evidence

No rendered markup or visual styling changed, so a screenshot would be identical. This fix changes only the state-merge source used by the existing controlled inputs: rapid or IME-composed consecutive edits now merge against current Zustand state instead of an earlier render snapshot, preventing intermittent value or caret displacement.

## Risks and failure behavior

`setNotificationTemplates` now accepts either a complete template object or a functional updater. The existing persistence hydration path still supplies the complete object and follows the unchanged object branch. The regression test verifies that two consecutive functional updates to different fields retain both changes; failure would otherwise present as a later edit clobbering an earlier one.

### DIFF
--- a/packages/ui/src/components/sections/openchamber/NotificationSettings.tsx
+++ b/packages/ui/src/components/sections/openchamber/NotificationSettings.tsx
@@ -152,13 +152,13 @@ export const NotificationSettings: React.FC = () => {
     field: 'title' | 'message',
     value: string,
   ) => {
-    setNotificationTemplates({
-      ...notificationTemplates,
+    setNotificationTemplates((current) => ({
+      ...current,
       [event]: {
-        ...notificationTemplates[event],
+        ...current[event],
         [field]: value,
       },
-    });
+    }));
   };
 
   const base64UrlToUint8Array = (base64Url: string): Uint8Array<ArrayBuffer> => {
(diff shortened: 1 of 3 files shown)


makeittech (approved):
Solid race fix.

makeittech:
Looks good.

PR #2303, opened 2026-07-18
fix(small-model): include config-authenticated providers in override picker

jamie-dit:
## Summary

Include providers authenticated through `provider.<id>.options.apiKey` in the Small Model override picker.

## Why

The Small Model picker introduced in #2049 limits its provider list to `authenticatedProviders`. After #2134 and #2135, the call path can resolve a custom provider's API key and endpoint directly from OpenCode configuration, but provider discovery still reads only `auth.json`.

As a result, a provider can be callable by `callSmallModel()` while remaining impossible to select in Settings.

This is complementary to #2132 and #2143: those cover providers that require no authentication, while this change covers providers authenticated through configuration.

## What changed

- Reuse the existing literal, `{env:...}`, and `{file:...}` API-key resolution when discovering configured providers.
- Add only provider IDs whose configured keys resolve successfully to `listAuthenticatedProviders()`.
- Keep credential values inside the server module; picker callers receive IDs only.
- Skip one invalid key reference without hiding other valid configured providers.
- Add focused regression tests and update the Small Model module documentation.

## Security and compatibility

- No credential values are returned, persisted, or logged.
- Existing `auth.json` handling and provider aliases are unchanged.
- No route or UI response shape changes are required.
- If #2143 lands, its selectable-provider union will automatically include these config-authenticated providers.

## Verification

- [x] `bun run --cwd packages/web test -- server/lib/small-model` — 37 tests passed
- [x] `bun run type-check`
- [x] `bun run lint`
- [x] `bun run build`
- [x] `bun run dead-code` — completed with the repository's existing non-blocking report; the new export was not flagged

## Related

- Follow-up to #2049
- Completes the picker-side support associated with #2134 and #2135
- Related to #2132 and #2143

### DIFF
--- a/packages/web/server/lib/small-model/DOCUMENTATION.md
+++ b/packages/web/server/lib/small-model/DOCUMENTATION.md
@@ -1,7 +1,8 @@
 # Small Model
 
 Server-side direct LLM calls that reuse the user's existing OpenCode provider
-logins (`~/.local/share/opencode/auth.json`). OpenCode uses a "small model"
+credentials from `~/.local/share/opencode/auth.json` or resolved
+`provider.<id>.options.apiKey` configuration. OpenCode uses a "small model"
 internally (titles, summaries) but does not expose it through the SDK or
 plugins — this module replicates that mechanism as an OpenChamber runtime API.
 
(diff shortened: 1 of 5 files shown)


btriapitsyn:

PR #2307, opened 2026-07-18
feat: add default permission auto-accept setting

jamie-dit:
## Summary

Add an opt-in default for OpenChamber's existing permission auto-accept policy.

When enabled, sessions without an explicit session or ancestor override automatically answer permission requests with `once`. Explicit session values still take precedence, so a session can opt out even while the default is enabled.

## Why

OpenChamber currently supports auto-accept only as an explicit per-session policy. Users who intentionally run trusted environments must enable it again for each independent session.

This change adds a default while preserving the existing policy model:

```text
explicit current session -> nearest explicit ancestor -> global default -> false
```

## What changed

- Persist `permissionAutoAccept.default` alongside existing session overrides.
- Add an authenticated default-policy route and server-side reconciliation for already-pending permissions.
- Add a searchable Settings -> Sessions control with a visible security warning and confirmation before enablement.
- Keep pending requests visible until an authoritative `permission.replied` event arrives.
- Preserve web, desktop, hosted-mobile, Capacitor, and VS Code behavior.
- Scope VS Code policy storage to the active workspace/runtime identity and serialize concurrent policy writes.
- Add complete translations for every shipped locale.
- Document policy precedence, persistence, reconciliation, and the existing VS Code foreground-webview limitation.

## Security and compatibility

- The new default is off when absent or malformed.
- Enabling requires explicit confirmation.
- Auto-accept continues to send `once`; it does not write OpenCode wildcard rules or saved `always` grants.
- Invalid, incomplete, cyclic, or mismatched session lineage fails closed.
- Explicit `false` session overrides remain effective when the default is on.
- Failed enablement preflight does not persist or broadcast a partial policy.
- Existing `permissionAutoAccept.sessions` settings remain compatible.
- VS Code does not migrate a legacy global default across workspace/runtime boundaries.

## Verification

- [x] 51 server runtime and settings tests
- [x] 60 shared UI/store/sync/settings/localization tests in one process
- [x] 15 VS Code policy/runtime identity tests
- [x] `bun run type-check`
- [x] `bun run lint`
- [x] `bun run build`
- [x] `bun run dead-code` — completed with the repository's existing non-blocking report; no new feature export remains flagged
- [x] `git diff --check`

## Related

- Follow-up to #2158


btriapitsyn:

PR #2311, opened 2026-07-19
fix(agents): show hidden custom subagents in Settings

### LINKED ISSUE #2305: [Bug] Agents section in Settings doesn't show custom subagents

I have 3 custom primary agents and 1 custom primary/subagent (all) configured in `~/.config/opencode/agents/`. None of them are workspace level, they're all globally configured. In Openchamber settings, I see them all.

However, I also have 10 custom global subagents configured in `~/.config/opencode/agents/` as well. My primary agents are able to pick up these subagents and call them - so I know they are configured correctly. However, I do not see them showing up in my Openchamber agent settings.

### Where does it happen?

Desktop Web

### Version (if known)

1.16.2

_No response_

### Logs (optional)

```shell

```

IbrahimKhan12:
> Validated commit: `41e470c14fbd3787629330d466f7e415d4d5310f`, rebased onto `main` at `feec14545`.

## Intent

Fixes #2305. Upstream OpenCode defines `hidden` as "hide this subagent from the @ autocomplete menu", but OpenChamber read it as "internal agent, never show" and filtered `isAgentHidden` out of the Settings agent list. A user-defined subagent marked `hidden` therefore became unreachable: it could not be edited, duplicated, renamed, or deleted from the UI at all. The defect is still present on `main`, where `AgentsSidebar.tsx` filters with `!isAgentHidden(agent)`.

Settings now filters with `isAgentManageable` (`!isAgentHidden(agent) || !isAgentBuiltIn(agent)`), so the only agents still excluded are hidden *and* built-in. Internal agents such as title, compaction, and summary are additionally `native: true`, which `isAgentBuiltIn` covers, so excluding just `hidden && (native || builtIn)` keeps them out while making hidden custom agents manageable.

`hidden` is now part of `AgentConfig` and `AgentDraft` and is persisted on create, and `copyAgentConfig` carries it through both the rename and duplicate paths, which are newly reachable for these agents.

## Non-goals

No UI affordance to view or toggle `hidden` was added; the flag is only preserved, never edited from Settings. Pickers still exclude hidden agents through the existing `isAgentHidden`/`filterVisibleAgents` predicates, so this does not put hidden subagents back into the composer or `@` menu. No agent discovery or config-file ownership change. No locale file is touched.

## Affected surfaces

- `packages/ui` Settings agent list and agent mutation flows, shared by web, desktop, VS Code, and mobile: `AgentsSidebar.tsx` (list, count label, empty state, duplicate, rename), `AgentsPage.tsx` (create), and `useAgentsStore.ts` (`isAgentManageable`, `hidden` on `AgentConfig`/`AgentDraft`, create persistence).
- User-visible states: hidden custom agents now appear as ordinary rows, and the sidebar total counts them. Hidden built-in/native agents remain absent, and the empty state now keys off the manageable list.
- Persisted contract: `hidden: true` is written into the agent config file on create and preserved on rename and duplicate. No server or transport change.

## Rebase note

`main` moved under this branch, and two upstream changes had to be threaded through rather than merged mechanically:

1. **Directory-scoped agents.** `main` replaced the flat list with `agents = useAgentsStore((state) => selectAgentsForDirectory(state, settingsDirectory))`. I verified `selectAgentsForDirectory` returns the raw stored array and performs **no** hidden filtering of its own, so the fix is still load-bearing; `filterVisibleAgents` is applied separately by the picker call sites (`sections/commands/AgentSelector.tsx`, `useConfigStore.ts`, and the `getVisibleAgents` store getter). This is exactly the management/picker split the change relies on, and it remains intact.
2. **`settingsDirectory` argument.** `main` added a directory argument to `createAgent(config, directory)` and `deleteAgent(name, scope, directory)`. The rename path now combines both: `copyAgentConfig(...)` supplies the config and `settingsDirectory` is passed through to both calls.

`isAgentHidden` is still imported and used inside `copyAgentConfig`, so no import became dead. The resulting diff stat is unchanged from the pre-rebase commit (210 insertions, 43 deletions).

## Repository guidance

|---|---|---|
| `openchamber-change-discipline` | Shared store exports and Settings behavior changed | Management and picker visibility stay separate predicates rather than one overloaded filter. The one `react-refresh/only-export-components` suppression covers `copyAgentConfig`, which is production code shared by the rename and duplicate paths, not a test hook. `bun run dead-code` was run because a file and two exports were added. |
| `settings-ui-patterns` | The Settings agents pane content and its count changed | Hidden custom agents reuse the existing list rows, grouping, and editing surfaces instead of a parallel UI, so no new page chrome, control, or info hint was introduced. Filtering happens inline with the already-exported store predicate. No control was added or moved, so the Settings search registry in `lib/settings/search.ts` needs no new entry and the existing `data-settings-item="agents.create"` anchor is unchanged. Narrow and wide pane states are both captured below. |

## Validation

| Check | Result |
|---|---|
| `bun test src/stores/useAgentsStore.subagents.test.tsx` | **4 passed**, 0 failed, 16 assertions. |
| `bun test` on the agent suites (`useAgentsStore.subagents`, `AgentPermissionsEditor`, `useAgentMemoryStore`) | **27 passed**, 0 failed. Run because the rebase moved the agents list onto `main`'s directory-scoped selector. |
| `bun run lint:ui` | Exit 0. One `react-hooks/exhaustive-deps` warning remains in `packages/ui/src/apps/MobileChangesSurface.tsx`, which this PR does not touch; that file is byte-identical to `origin/main`, so the warning is pre-existing upstream. |
| `bun run dead-code` | Neither `isAgentManageable`, `copyAgentConfig`, nor the new test file appears in the unused report. |
| Rendered Settings pane, wide and narrow | See Visual evidence. |

The test asserts the exact derived lists: manageable resolves to `build, visible-custom, hidden-custom, options-hidden-custom` while `getVisibleAgents()` (the picker path) still excludes both hidden custom agents; it covers a hidden custom agent surviving an update, `copyAgentConfig` preserving `hidden` for both the top-level and `options.hidden` shapes, `createAgent` sending `hidden: true`, and derivation from a refreshed snapshot. It exercises the derivation predicate rather than the component's store wiring.

**Not validated:** no `hidden: true` agent was written against a live OpenCode server, so config write-through and restart behavior are unproven. Light theme was not captured; this change alters list contents only and adds no color, icon, or style.

## Visual evidence

Captured against the current HEAD by running the real Settings pane in a headless Chromium at 1440x950 and 900x950. The agents response (`GET /api/agent`) was stubbed so the same fixture renders in both states: `build` and `plan` (native visible), `title`/`summary`/`compaction` (native hidden), `visible-helper` (custom visible), `hidden-helper` (custom, `hidden: true`), and `options-hidden-helper` (custom, `options.hidden: true`). The stub is the data source; everything rendered is the real component, styling, and grouping. The before frame was produced by reverting only the filter predicate to `main`'s `!isAgentHidden(agent)` on the same running app, then restoring it.

**Before — hidden custom subagents unreachable (`Total 3`, only `visible-helper` under Custom Agents):**

Settings Agents before the fix

**After — hidden custom subagents manageable (`Total 5`, `hidden-helper` and `options-hidden-helper` now listed):**

Settings Agents after the fix

**After, narrow settings pane (900px) — added rows render correctly in the narrow layout:**

Settings Agents after the fix, narrow pane

Both frames keep Built-in Agents at `build` and `plan`, which is the important negative result: the three hidden **native** agents stay excluded before and after, so the fix does not leak internal agents into Settings.

A narrow *before* frame was not captured. The difference is list membership, which is width-independent and already shown in the wide pair; the narrow frame exists to show the added rows lay out correctly in the compact pane.

Screenshots are hosted on the `pr-evidence` branch of the fork so they stay out of this PR's diff.

## Risks and failure behavior

- Manageability is derived from the authoritative `hidden` plus `native`/`builtIn` flags. If an internal agent ever ships hidden without `native` or `builtIn`, it would become visible in Settings; that combination does not exist in the agents this change was tested against but is not defended against in code.
- Rename creates the new agent before deleting the old one, so a failed delete leaves both, with `hidden` preserved on the copy. That failure path is pre-existing and unchanged.
- Pickers are unaffected because they use a different predicate (`filterVisibleAgents`), so this cannot leak hidden agents into selection surfaces. The rebase onto `main`'s directory-scoped selector preserves that separation.
- Rollback is reverting one commit; the only persisted side effect is a `hidden` key written on newly created or copied agents, which upstream already understands.

### DIFF
--- a/packages/ui/src/components/sections/agents/AgentsPage.tsx
+++ b/packages/ui/src/components/sections/agents/AgentsPage.tsx
@@ -227,6 +227,7 @@ export const AgentsPage: React.FC = () => {
         temperature: temperature ?? null,
         top_p: topP ?? null,
         prompt: trimmedPrompt || (isNewAgent ? undefined : null),
+        ...(isNewAgent && agentDraft?.hidden ? { hidden: true } : {}),
         ...(isNewAgent && draftScope ? { scope: draftScope } : {}),
       };
 
--- a/packages/ui/src/components/sections/agents/AgentsPage.tsx
+++ b/packages/ui/src/components/sections/agents/AgentsPage.tsx
@@ -231,6 +231,7 @@ export const AgentsPage: React.FC = () => {
         temperature: temperature ?? null,
         top_p: topP ?? null,
         prompt: trimmedPrompt || (isNewAgent ? undefined : null),
+        ...(isNewAgent && agentDraft?.hidden ? { hidden: true } : {}),
         ...(isNewAgent && draftScope ? { scope: draftScope } : {}),
       };
 
--- a/packages/ui/src/components/sections/agents/AgentsSidebar.tsx
+++ b/packages/ui/src/components/sections/agents/AgentsSidebar.tsx
@@ -19,7 +19,7 @@ import {
 } from '@/components/ui/dropdown-menu';
 import { ContextMenu, ContextMenuContent, ContextMenuItem, ContextMenuTrigger } from '@/components/ui/context-menu';
 import { useSettingsDirectory } from '@/hooks/useSettingsDirectory';
-import { selectAgentsForDirectory, useAgentsStore, isAgentBuiltIn, isAgentHidden, type AgentScope, type AgentDraft } from '@/stores/useAgentsStore';
+import { selectAgentsForDirectory, useAgentsStore, isAgentBuiltIn, isAgentHidden, isAgentManageable, type AgentScope, type AgentDraft } from '@/stores/useAgentsStore';
 import { useShallow } from 'zustand/react/shallow';
 import { cn } from '@/lib/utils';
 import type { Agent } from '@opencode-ai/sdk/v2';
@@ -101,6 +101,27 @@ const rulesetToPermissionConfig = (ruleset: unknown): AgentDraft['permission'] =
   }
 
   return Object.keys(result).length > 0 ? (result as AgentDraft['permission']) : undefined;
+};
+
+// eslint-disable-next-line react-refresh/only-export-components -- Regression tests cover config copied by rename and duplicate flows.
+export const copyAgentConfig = (agent: Agent, name: string) => {
+  const extended = agent as Agent & { scope?: AgentScope; disable?: boolean };
+  return {
+    name,
+    description: agent.description,
+    model: agent.model?.providerID && agent.model?.modelID
+      ? `${agent.model.providerID}/${agent.model.modelID}`
+      : null,
+    variant: agent.variant,
+    temperature: agent.temperature,
+    top_p: agent.topP,
+    prompt: agent.prompt,
+    mode: agent.mode,
+    permission: rulesetToPermissionConfig(agent.permission),
+    disable: extended.disable,
+    ...(isAgentHidden(agent) ? { hidden: true } : {}),
+    scope: extended.scope,
+  };
 };
 
 export const AgentsSidebar: React.FC<AgentsSidebarProps> = ({ onItemSelect }) => {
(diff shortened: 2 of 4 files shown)


IbrahimKhan12:

Two upstream changes had to be threaded through rather than merged mechanically:

1. `main` moved the sidebar onto `selectAgentsForDirectory(state, settingsDirectory)`. I checked that this selector returns the raw stored array and does no hidden filtering itself — `filterVisibleAgents` is applied separately by the picker call sites — so the fix is still load-bearing and the management/picker split it relies on is intact.
2. `createAgent`/`deleteAgent` gained a directory argument, now threaded through the rename path alongside `copyAgentConfig`.

I also captured the before/after Settings evidence that was previously missing: the real pane rendered headlessly at 1440px and 900px, with the agents response stubbed so the same fixture renders in both states. Before is `Total 3` with only `visible-helper` under Custom Agents; after is `Total 5` with `hidden-helper` and `options-hidden-helper` listed. Built-in Agents stays `build` + `plan` in both, so hidden **native** agents are still excluded.

Note this PR is still a draft, so it will not pick up a `review:*` readiness label until it is marked ready.

PR #2321, opened 2026-07-19
fix(settings): prevent local state leaking to remote runtimes

### LINKED ISSUE #2298: [Bug] Remote instances receive the homedirectory and projects from the user client

Hello, when using OpenChamber (in my case the desktop version) to connect remote instances (over SSH or not), the remote sessions receives the homedirectory and projects from the user client which breaks various stuff and makes openchamber missbehave (opencode is acting fine, though the original configuration also receives the wrong home path)

### Where does it happen?

Desktop (macOS)

### Version (if known)

1.16.1

_No response_

### Logs (optional)

```shell
[persistSettings] Updating fields: projects, activeProjectId, lastDirectory
[persistSettings] Updating fields: lastDirectory, homeDirectory
```

bashrusakh:
## Problem

Switching an Electron or web client between runtimes happens in place. Browser and desktop fallback state from the previous runtime could therefore be rehydrated or persisted under the next runtime, corrupting remote paths, projects, and sidebar metadata.

Fixes #2298

### Runtime-scoped persistence
- Scope directory, project, pinned-session, folder, and sidebar persistence by `getRuntimeKey()`.
- Keep legacy unscoped browser storage readable only for the local runtime.
- Reload destination-runtime state on endpoint changes instead of carrying previous-runtime UI state forward.

### Stale-work protection
- Reject stale home resolution, project-icon, folder-cleanup, and debounced sidebar writes using runtime identity plus a monotonic generation.
- Cover local → remote and A → B → A runtime switches so an old completion cannot write into the current runtime.

### Review follow-up
- Reject explicit empty or whitespace runtime storage keys before any storage read or write. This fails closed instead of collapsing state into a shared `default` namespace.
- Preserve normal omitted/null key resolution through `getRuntimeKey()` and the local-only legacy fallback.

## Validation
- `bun test packages/ui/src/stores/utils/runtimeScopedStorage.test.ts` — 6 passed.
- Focused non-isolated persistence regression batch — 42 passed.
- Focused isolated runtime regression batch — 44 passed.
- `bun run type-check:ui` — passed.
- `bun run lint:ui` — passed.
- `bun run dead-code` — completed; its configured non-blocking report contains existing cleanup candidates.

## Notes
- No raw runtime/path input is exposed to users; runtime identity remains system-derived.
- The existing no-argument `getRuntimeKey()` behavior is intentionally unchanged.

### DIFF
--- a/packages/ui/src/components/session/SessionSidebar.tsx
+++ b/packages/ui/src/components/session/SessionSidebar.tsx
@@ -27,7 +27,12 @@ import { useGroupOrdering } from './sidebar/hooks/useGroupOrdering';
 import { useSessionGrouping } from './sidebar/hooks/useSessionGrouping';
 import { useSessionSearchEffects } from './sidebar/hooks/useSessionSearchEffects';
 import { useSessionActions } from './sidebar/hooks/useSessionActions';
-import { useSidebarPersistence } from './sidebar/hooks/useSidebarPersistence';
+import {
+  readSidebarActiveSessions,
+  readSidebarGroupOrder,
+  readSidebarStringSet,
+  useSidebarPersistence,
+} from './sidebar/hooks/useSidebarPersistence';
 import { useProjectRepoStatus } from './sidebar/hooks/useProjectRepoStatus';
 import { useProjectSessionLists } from './sidebar/hooks/useProjectSessionLists';
 import { useSessionFolderCleanup } from './sidebar/hooks/useSessionFolderCleanup';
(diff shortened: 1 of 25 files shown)


TheFrenchGhosty:
@btriapitsyn Any plan to merge this soon-ish? (at least if it fixes it)

It currently makes it basically impossible to use remote instances

btriapitsyn:

TheFrenchGhosty:
@btriapitsyn It got stale because you didn’t merge it when it was ready… and instead let it sit around for more than a month…

bashrusakh:
> so this change would need to be redone against the current state anyway.

This won't happen, at least I won't do it.

PR #2324, opened 2026-07-19
fix: run project start commands in new worktrees

### LINKED ISSUE #2323: Project commands.start is skipped for OpenCode 1.18.3 worktrees
### What happened?

Worktrees created through OpenChamber do not run the project's `commands.start` when OpenCode 1.18.3 is in use. The worktree finishes bootstrap and appears ready, but the setup expected from the project command never happens.

OpenChamber currently reads `commands.start` from the legacy `~/.local/share/opencode/storage/project/<id>.json` file. OpenCode 1.18.3 stores project commands in SQLite and exposes them through the project API, so the legacy JSON often contains only worktree metadata.

This affects the web/desktop and VS Code paths because both create Git worktrees themselves.

### Steps to reproduce

1. Use OpenCode 1.18.3 and configure a project `commands.start` value that leaves an observable marker.
2. Create a worktree from OpenChamber.
3. Wait for the worktree bootstrap status to become ready.
4. Check for the marker.

### Expected behavior

The project command runs during bootstrap before any request-specific `startCommand`. An explicitly empty project command should mean "run nothing."

### Actual behavior

The project command is skipped because OpenChamber cannot find it in the legacy JSON file.

### Versions

- OpenChamber 1.16.2
- OpenCode 1.18.3
- macOS

### Additional context

The legacy JSON lookup is still useful as a fallback for older OpenCode versions. The API result needs to distinguish an unavailable or unmatched project from a matching project with an empty command.

e-n-0:
## Summary

- load `commands.start` from the current OpenCode project API instead of relying on the legacy project JSON mirror
- use the same lookup and fallback behavior in the web/desktop and VS Code worktree paths
- keep legacy JSON support when the API is unavailable or cannot match the project, while treating an explicitly empty command as authoritative

OpenCode 1.18.3 stores project commands in SQLite, so OpenChamber's legacy JSON lookup no longer sees them. This change keeps worktree creation in OpenChamber but gets the project command through the supported SDK API.

The project command still runs before a request-specific `startCommand`.

Fixes #2323

## Testing

- `bun run --cwd packages/web test` (731 passed, 1 skipped)
- `bun test packages/vscode/src/project-commands-runtime.test.js packages/vscode/src/bridge-git-runtime.test.js packages/vscode/src/gitService.worktree-bootstrap.test.js` (18 passed)
- `bun run type-check`
- `bun run lint`
- `bun run build`
- `bun run docs:validate`
- end-to-end web driver with a real Git worktree and fake OpenCode SDK server; verified authenticated `/project/current` lookup and `project\nextra\n` execution order

### Existing test isolation issue

`bun test packages/vscode/src` still exposes the existing Bun cross-file module-mock collision between `bridge-git-runtime.test.js` and `bridge-git-special-runtime.test.js`. The affected test passes in isolation, the conflicting mock predates this change, and all patch-owned VS Code tests pass together.

### DIFF
--- a/packages/vscode/src/DOCUMENTATION.md
+++ b/packages/vscode/src/DOCUMENTATION.md
@@ -14,6 +14,7 @@ Keep `bridge.ts` as a thin orchestration layer that delegates message handling t
 
 - `bridge-git-runtime.ts`
   - Standard Git message handlers.
+  - Injects the VS Code OpenCode project-command runtime into worktree creation so bootstrap setup mirrors the web server runtime.
 
 - `bridge-git-special-runtime.ts`
   - Specialized Git flows (`pr-description`, `conflict-details`) and generation helpers.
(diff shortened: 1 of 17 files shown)


e-n-0:
Hello @btriapitsyn, possible to get a review on this one please 🙏 
(linked issue #2323)

btriapitsyn:

PR #2339, opened 2026-07-20
feat: sync session read status across connected clients

jamie-dit:
## Summary

With multiple OpenChamber windows (or devices, e.g. a desktop browser and the mobile PWA) connected to the same server, reading a session in one window did not clear its unread indicator in the others. Unread state lived only in each window's in-memory notification list, while live progress already synced through the server event hub.

This PR makes read status sync across all connected clients using the server's existing session-attention runtime.

## Design

- **Dedicated event:** the server now broadcasts `openchamber:session-attention` (`sessionID`, `needsAttention`, `lastReadAt`, `lastActivityAt`, `timestamp`) via the existing broadcaster on view/unview and attention transitions. `openchamber:session-status` is intentionally left untouched for live progress, since its forwarded-progress path hard-codes `needsAttention: false` and would race with read state.
- **Read receipts:** when a client marks a session viewed (opening it, window focus, or a completion that lands while the session is focused and visible), it POSTs the existing `/api/sessions/:id/view` endpoint — fire-and-forget with one retry, throttled per session (3s) to avoid bursts, and silently tolerated on older servers without the endpoint.
- **Remote apply without loops:** incoming attention events clear local notifications through a non-broadcasting reducer; the remote path can never trigger another receipt.
- **Timestamp guard:** only notifications with `time <= lastReadAt` are cleared, so a delayed read event cannot wipe a newer completion.
- **Hydration:** clients fetch `GET /api/sessions/attention` on startup, reconnect, and transport switches, reconciling conservatively (never fabricating unread).
- **Coalescing:** repeated attention events per session collapse in the client event pipeline.
- VS Code webview bridge answers `/view` & `/unview` with no-ops so the shared receipt path stays harmless in constrained runtimes.

## Validation

- `bun run type-check:ui` — pass
- `bun run lint:ui` — pass
- UI tests (`notification-store`, `event-pipeline`) — 7 pass / 0 fail, covering: server-clock cutoff clearing, remote apply sends no receipt (no feedback loop), viewed-at-append sends one throttled receipt, hydration reconciliation
- Server tests (`session-runtime`) — 3 pass, covering attention broadcasts on view and status transitions

## Notes / limitations

- Attention state remains in-memory on the server (existing 24h TTL); persistence across server restarts would be a follow-up.
- Client notification timestamps are client-clock based while `lastReadAt` is server-clock; large clock skew could delay clearing (conservative direction only — never over-clears).

### DIFF
--- a/packages/ui/src/sync/event-pipeline.ts
+++ b/packages/ui/src/sync/event-pipeline.ts
@@ -273,9 +273,20 @@ export function createEventPipeline(input: EventPipelineInput): EventPipeline {
   }
 
   const key = (payload: Event): string | undefined => {
+    const rawType = (payload as { type?: unknown }).type
     if (payload.type === "session.status") {
       const props = payload.properties as { sessionID: string }
       return `session.status:${props.sessionID}`
+    }
+    if (rawType === "openchamber:session-attention") {
+      const props = (payload as { properties?: unknown }).properties as { sessionID?: string; sessionId?: string } | undefined
+      if (!props) return undefined
+      const sessionID = typeof props.sessionID === "string" && props.sessionID.length > 0
+        ? props.sessionID
+        : typeof props.sessionId === "string" && props.sessionId.length > 0
+          ? props.sessionId
+          : undefined
+      return sessionID ? `session.attention:${sessionID}` : undefined
     }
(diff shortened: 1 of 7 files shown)


btriapitsyn:

PR #2347, opened 2026-07-21
feat: add restore/unarchive for archived sessions

### LINKED ISSUE #2346: [Bug] No way to restore/unarchive sessions once archived
## Steps to reproduce:

1. Open OpenChamber (any platform).
2. In the sidebar, right-click any active session and select **Archive** (or use the Archive button in the session row).
3. The session moves to the "Archived" bucket under its project.
4. Right-click the archived session in the Archived bucket.
5. Observe the available context menu actions.

## What happens:

The context menu for an archived session only shows one action:

- **Delete permanently** (destructive, irreversible)

There is no **"Restore"**, **"Unarchive"**, or any equivalent action to move the session back to the active list.

Similarly, there is no bulk action to restore multiple archived sessions.

## Expected behavior:

Archived sessions should have a **Restore** (or **Unarchive**) action that clears the `time.archived` timestamp and moves the session back to the active list.

This should be available:

1. **Per-session**: via the right-click context menu on an archived session (alongside the existing "Delete permanently").
2. **Bulk**: via the bulk action bar when multiple archived sessions are selected in the sidebar.

## Root cause

The `archiveSession` function (`packages/ui/src/sync/session-actions.ts:626`) sets `time.archived` on the session via:

```ts
const archived = await opencodeClient.updateSession(sessionId, { time: { archived: archivedAt } }, sessionDirectory)
```

However, there is **no corresponding `restoreSession` or `unarchiveSession` function** anywhere in the codebase. A codebase-wide search for `restoreSession`, `unarchiveSession`, `restoreFromArchive`, or `unarchive` returns zero results in `packages/`.

In the session context menu (`packages/ui/src/components/session/sidebar/SessionNodeItem.tsx:917-926`), the "Archive" option is only shown when `!archivedBucket`, and no "Restore" option exists for the `archivedBucket === true` path:

```tsx
{!archivedBucket ? (
  <Item onClick={() => handleDeleteSession(session, { archivedBucket })}>
    <Icon name="inbox-archive" className="mr-1 h-4 w-4" />
    {t('sessions.sidebar.bulkActions.archive')}
  </Item>
) : null}
{/* No restore/unarchive item here */}
<Item className="text-destructive ..." onClick={() => handleDeleteSession(session, { archivedBucket, hardDelete: true })}>
  ...
</Item>
```

## Suggested fix direction

1. Add a `restoreSession` (or `unarchiveSession`) function in `session-actions.ts` that calls `opencodeClient.updateSession(id, { time: { archived: undefined } })` — analogous to the existing `archiveSession`.
2. Add a "Restore" context menu item for archived sessions in `SessionNodeItem.tsx`.
3. Consider adding bulk restore support through the existing bulk action bar.

### Where does it happen?

All platforms (Web, Desktop, VS Code, Mobile)

### Version (if known)

Current `main` branch / latest release

_No response_

### Logs (optional)

```shell
N/A — this is a missing feature, not a runtime error.
```

chiamsun:
## Summary

Currently, after archiving a session, there is no way to restore it back to the active list. The only available action in the context menu for archived sessions is "Delete permanently". This PR adds the missing restore/unarchive functionality.

Closes #2346

### Core logic
- **`session-actions.ts`**: Added `unarchiveSession` function that calls `opencodeClient.updateSession(id, { time: { archived: undefined } })` to clear the archive flag on the server, with optimistic update and rollback on failure.
- **`useGlobalSessionsStore.ts`**: Added `unarchiveSessions` method that moves sessions from `archivedSessions` back to `activeSessions` in the global store.
- **`session-ui-store.ts`**: Exposed `unarchiveSession` and `unarchiveSessions` in the UI store, following the same pattern as `archiveSession`/`archiveSessions`.

### UI
- **`SessionNodeItem.tsx`**: Added a "Restore" context menu item (with `inbox-unarchive` icon) that appears for archived sessions, right before the "Delete permanently" option. Shows a success/error toast on completion.

### i18n
- Added `sessions.sidebar.bulkActions.restore`, `sessions.sidebar.session.restore.success`, and `sessions.sidebar.session.restore.error` keys to all 10 locale files (en, es, fr, ja, ko, pl, pt-BR, uk, zh-CN, zh-TW).

## Validation

- [x] `bun run type-check:ui` — passes
- [x] `bun run lint:ui` — passes
- [x] `bun run build:ui` — passes


makeittech:
Superseded by #2616, which has been merged.

Thank you for taking a first pass at this, @chiamsun — the direction (an `unarchiveSession` action plus a "Restore" context-menu item) was right. Unfortunately the core mechanism in this PR does not work against the OpenCode server:

- `opencodeClient.updateSession(id, { time: { archived: undefined } })` is dropped by OpenChamber's own client wrapper before serialization, so the request body is `{}`. Even when sent directly, the server only applies `time.archived` when the payload carries a finite number (`payload.time?.archived !== undefined`), and JSON cannot express `undefined`. Verified against opencode 1.18.12 (source + live server): an omitted key and `null` are both silently ignored — the session stays archived while the UI toasts success.

#2616 implements restore with a mechanism verified against the actual server behavior (`time.archived = 0`, which every truthiness-based reader treats as active), makes the global session cache split active/archived client-side so restored sessions survive restarts, and adds bulk restore plus Archive page restore.

Closing as superseded.

Issue #2296, opened 2026-07-17
[Bug] Should respect the configured npm registry for package metadata lookups


amat27:
### What's wrong? Maybe some steps to reproduce.

What should we add/change?
OpenChamber directly fetches package metadata from https://registry.npmjs.org in several runtime paths:
- packages/web/server/lib/opencode/npm-registry.js
- packages/vscode/src/opencodeConfig.ts
- packages/web/server/lib/package-manager.js
- packages/web/server/lib/opencode/routes.js
This bypasses the user's configured npm registry. As a result, plugin metadata and update checks may fail in environments where registry.npmjs.org is unavailable or where an internal/mirror registry is required.
It would be useful for these requests to respect a configurable registry, such as npm_config_registry / NPM_CONFIG_REGISTRY, while retaining https://registry.npmjs.org as the default.
Extra context
Package installation commands such as npm install, bun add, and pnpm add already omit --registry, so they can inherit the package manager's configuration. The direct HTTP metadata requests behave differently and always contact the public npm registry.
This affects:
- Plugin validation and update metadata in Web/Desktop
- Plugin validation and update metadata in the VS Code extension
- OpenChamber update checks
- OpenCode version checks

### Where does it happen?

Desktop Web

### Version (if known)

Windows Desktop, 1.16.1

### Logs (optional)

```shell

```

PR #2348, opened 2026-07-21
fix(npm): honor configured registry for package metadata lookups (#2296)

IbrahimKhan12:
> Validated commit: `1b86716c6cd1eafbff43584081f0e1956406f129`.

## Intent

Refs #2296. Web-server and VS Code npm metadata requests now resolve `npm_config_registry`/`NPM_CONFIG_REGISTRY` instead of hardcoding `https://registry.npmjs.org`, including registries mounted under a path and scoped package names.

`Refs` and not `Fixes`: npm exports those variables only to processes it launches (`npm run`, `npm exec`, `npx`). The Windows desktop app, the VS Code extension host, and a globally installed `openchamber` binary all start outside npm and still reach the default registry, so #2296 still reproduces for the reporter and must not be auto-closed by merging this.

Credentials embedded in the configured URL are moved into an `Authorization: Basic` header because a credentialed `registry=` URL is otherwise unusable: Node's `fetch` refuses to construct the request (`Request cannot be constructed from a URL that includes credentials`) and Bun sends the request with no `Authorization` header at all. Keeping credentials out of logged URLs and error text is a secondary benefit.

## Non-goals

No `.npmrc` parsing, and no scoped-registry (`@scope:registry`) support. Verified with a real `npm run`: given an `.npmrc` containing `@myscope:registry` and `//host/:_authToken`, npm exports only `npm_config_registry` to the child process, so scoped and token-authenticated setups still fall through to the default registry. No auth scheme other than Basic credentials taken from the configured URL.

## Affected surfaces

- Web: plugin metadata validation, `@openchamber/web` update checks, OpenCode update checks.
- VS Code: plugin metadata validation and OpenCode update checks.

No UI, persisted-data, or HTTP response-shape changes.

## Repository guidance

|---|---|---|
| `openchamber-change-discipline` | Adds a runtime helper to two packages | Each runtime owns its resolver with direct wiring tests; the duplication is required because `packages/vscode` has no dependency on `packages/web` and the helper is Node-only (`process.env`, `Buffer`) so it cannot live in `@openchamber/ui`. |
| `ui-api-decoupling` | Server and extension-host outbound requests changed | Registry access stays inside each owning runtime; shared UI keeps consuming unchanged metadata responses. |

## Validation

| Check | Result |
|---|---|
| `bun test packages/vscode/src/npm-registry.test.ts packages/vscode/src/opencode-upgrade-runtime.test.ts packages/vscode/src/opencodeConfig-npm-registry.test.ts` | 10 passed, 0 failed. |
| `bunx vitest run server/lib/npm-registry-config.test.js server/lib/opencode/npm-registry.test.js server/lib/opencode/routes-upgrade.test.js server/lib/package-manager.test.js` in `packages/web` | 4 files, 44 passed, 0 failed. |
| `bun run type-check:web`, `bun run lint:web`, `bun run vscode:type-check` | Exit 0 each. Coverage caveat: `packages/web/tsconfig.json` includes only `src` and `../ui/src`, and `lint:web` globs only `./src/**/*.{ts,tsx}`, so the new `server/lib/npm-registry-config.js` is covered by vitest alone — neither type-check nor lint reads it. |

Not verified: no request was sent to a real authenticated private registry.

## Visual evidence

No rendered UI change; no component, style, or layout is touched. The observable behavioral consequence is on a configured mirror: plugin validation now reports that mirror's `Registry returned 401` or `Package not found` instead of silently answering from the public registry, and a failed update lookup returns `null`, so the update banner stays hidden rather than advertising a public-registry version.

## Risks and failure behavior

Only `http:`/`https:` registries without a query string or fragment are accepted; anything else throws `Invalid npm registry URL`, which carries no credential text. Package name and metadata suffix are encoded independently, so a scoped name keeps its leading `@`.

The current `review:blocked` finding is factually wrong and should not gate this PR. It claims `decodeURIComponent` on `URL.password` corrupts credentials. The `URL.username`/`URL.password` getters return percent-**encoded** values, so decoding is required: for `https://user:p%40ss@host/` the getter yields `p%40ss` and decoding yields the correct `p@ss`, identical under Node 24.15.0 and Bun 1.3.14. The suggested `Buffer.from(\`${username}:${password}\`)` would send `Basic base64('user:p%40ss')` and break every credential containing a reserved character.

### DIFF
--- a/packages/vscode/README.md
+++ b/packages/vscode/README.md
@@ -52,6 +52,8 @@ Select code in the editor, right-click, and find the **OpenChamber** submenu:
 | `openchamber.apiUrl` | _(empty)_ | URL of an external OpenCode API server. Leave empty to auto-start a local instance. |
 | `openchamber.opencodeBinary` | _(empty)_ | Absolute path to the `opencode` CLI binary. Useful when PATH lookup fails. Requires window reload to apply. |
 
+npm metadata checks honor `npm_config_registry`, then `NPM_CONFIG_REGISTRY`, from the extension host environment.
+
 ## Requirements
 
 - [OpenCode CLI](https://opencode.ai) installed and available in PATH (or set `OPENCODE_BINARY` env var)
--- a/packages/vscode/README.md
+++ b/packages/vscode/README.md
@@ -53,6 +53,8 @@ Select code in the editor, right-click, and find the **OpenChamber** submenu:
 | `openchamber.apiUrl` | _(empty)_ | URL of an external OpenCode API server. Leave empty to auto-start a local instance. |
 | `openchamber.opencodeBinary` | _(empty)_ | Absolute path to the `opencode` CLI binary. Useful when PATH lookup fails. Requires window reload to apply. |
 
+npm metadata checks honor `npm_config_registry`, then `NPM_CONFIG_REGISTRY`, from the extension host environment.
+
 ## Requirements
 
 - [OpenCode CLI](https://opencode.ai) installed and available in PATH (or set `OPENCODE_BINARY` env var)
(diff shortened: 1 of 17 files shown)


btriapitsyn:

PR #2355, opened 2026-07-21
feat(attachments): ground model-unsupported attachments to workspace uploads/ instead of failing the send

abigfatstone:
## Problem

Attaching a file whose media type the selected model cannot ingest fails the whole message with a provider error:

```
Opencode failed to send message with error: 'file part media type application/vnd.openxmlformats-officedocument.spreadsheetml.sheet' functionality not supported.
```

Spreadsheets, docx, archives etc. are never accepted by any provider, and text-only models (e.g. DeepSeek) reject even images — so today the user just loses the send.

## Approach

Route attachments by model capability at send time (`sendMessage`/`sendCommand`):

- A media type is forwarded as a file part **only when the model's metadata explicitly declares support** — `modalities.input` contains `image`/`pdf`, falling back to the coarse `attachment` flag. `text/plain` always passes (the OpenCode server inlines it as text).
- Everything else is saved to `uploads/` inside the session working directory (name collisions get a timestamp suffix) and replaced with a **synthetic text part** telling the agent where the file is, so the agent reads it with its own tools (Python, etc.). `file://` attachments only get the hint — no copy.
- On any grounding failure (e.g. a runtime without the fs API) the original file part passes through unchanged — behavior is never worse than today.
- `session.command` parts only accept file parts, so grounded hints are appended to the command `arguments` there.

Server side, `POST /api/fs/write` gains an optional `encoding: 'base64'` for binary payloads, reusing the existing workspace-scoping guards and the atomic tmp-rename write.

This makes attachment support metadata-driven: adding a new model only requires correct capability metadata, no code changes.

## Testing

- `packages/ui/src/lib/attachments/attachmentGrounding.test.ts` (bun test ×8): decision table (text/plain pass, image gated on `modalities.input`, xlsx always grounded, `attachment` flag fallback), upload body/headers, collision suffix, write-failure passthrough, `file://` hint.
- `packages/web/server/lib/fs/routes.test.js` (+3 vitest): base64 write decodes and writes a Buffer atomically, unchanged-content skip, unknown encoding → 400.
- `@openchamber/ui` type-check + lint clean.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -97,7 +97,7 @@
     },
     "packages/electron": {
       "name": "@openchamber/electron",
-      "version": "1.16.1",
+      "version": "1.16.2",
       "dependencies": {
         "@openchamber/web": "workspace:*",
         "electron-context-menu": "^4.1.2",
@@ -133,7 +133,7 @@
     },
     "packages/ui": {
       "name": "@openchamber/ui",
-      "version": "1.16.1",
+      "version": "1.16.2",
       "dependencies": {
         "@aparajita/capacitor-secure-storage": "^8.0.0",
         "@base-ui/react": "^1.4.0",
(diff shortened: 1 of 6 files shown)


btriapitsyn:

PR #2360, opened 2026-07-21
Improve high-volume session performance and runtime correctness

btriapitsyn:
## Summary

This PR improves OpenChamber under high session concurrency and hardens the state-authority rules exposed by that work.

The main changes are:

- prioritize and bound directory/session loading instead of letting every mounted consumer start independent work;
- centralize message loading, pagination, prefetch, retries, and stale-result rejection;
- reduce realtime event fanout, repeated scans, broad subscriptions, and hidden-surface work;
- preserve stable references so unrelated session/project updates do not rebuild the entire navigation tree;
- keep active incremental lists in normal document flow while virtualizing only large archived collections;
- scope caches and persisted state by runtime, directory, and entity identity;
- distinguish authoritative empty state from failure, missing persistence, malformed persistence, and incomplete startup data;
- order permission, settings, folder, Git/PR, and other asynchronous snapshots so stale completion cannot overwrite newer state;
- add a repeatable browser profiling command and deterministic operation-count regressions;
- strengthen the repository skills with generalized comparator, virtualization, persistence-ordering, and authoritative-baseline guidance.

The branch contains 35 commits on top of `main` and changes 196 files across shared UI/sync, web runtime, VS Code, Electron, mobile integration, profiling tooling, tests, and owning documentation.

## User-visible impact

- Large multi-session workspaces remain more responsive while sessions are streaming and while navigating between chats.
- Opening or revisiting a session no longer starts duplicate message loads through multiple consumers.
- Foreground session/directory work is prioritized over background enrichment with bounded concurrency.
- Hidden chat/sidebar surfaces stop speculative prefetch, Git/PR enrichment, search listeners, observers, and row-level subscriptions.
- Session tree state remains stable across selection, collapse/expand, Recent/project contexts, pin changes, runtime switches, archive, deletion, and reconnect.
- Incremental `Show more` lists retain normal layout and control reachability; virtualization is limited to archived collections that can grow into the hundreds or thousands.
- Failed or incomplete startup loads no longer masquerade as authoritative empty state and erase pins, folders, drafts, or other persisted context.
- Runtime switches no longer leak state between otherwise colliding session IDs, paths, branches, or cache keys.

## Browser measurements

The comparison uses the first and latest successful captures from the documented series with 55 simultaneously running sessions:

- before: `browser-profile-2026-07-21T08-55-06-218Z` (160 seconds);
- after: `browser-profile-2026-07-21T09-33-33-361Z` (60 seconds).

The later requested directory, `browser-profile-2026-07-21T10-58-13-369Z`, was empty and is intentionally not cited.

| Metric | Before | After | Change |
|---|---:|---:|---:|
| Longest renderer task | 591.275 ms | 177.321 ms | -70.01% |
| Longest message-list-attributed task | 591.275 ms | 134.415 ms | -77.27% |
| Tasks over 50 ms per second | 0.44375 | 0.25 | -43.66% |
| Navigation long-task time per mark | 72.087 ms | 54.733 ms | -24.07% |
| Sidebar long-task time per render mark | 85.838 ms | 38.279 ms | -55.41% |
| Message-list long-task time per render mark | 28.093 ms | 16.204 ms | -42.32% |

In human terms, the worst observed pause fell from about 0.59 seconds to 0.18 seconds, the worst message-list pause fell to about 0.13 seconds, and long tasks occurred roughly once every four seconds instead of once every 2.25 seconds.

These are manual before/after observations, not proof that any single commit caused the full delta. The captures differ in duration and interaction/traffic mix, maxima are sample-sensitive, and attribution categories overlap. Navigation still overlapped a long task for roughly half of its marks in both captures, so this PR materially lowers stall severity but does not claim that all navigation latency is eliminated.

### Priority scheduler

- Directory bootstrap is owned by the child-store manager rather than independently triggered by every consumer.
- Work is prioritized by current/selected, expanded, visible, and background demand.
- Concurrency is bounded so opening a workspace does not produce an uncontrolled request burst.
- Queued demand can be promoted when user intent changes.
- A failed directory does not erase or block complete unrelated directories.
- Mounted consumers pin their directory stores so active subscriptions are not silently replaced by eviction.

### Shared message loader

- Initial messages, pagination, prefetch, refresh, and retries share one authoritative loader keyed by runtime, normalized directory, and session ID.
- In-flight requests are deduplicated across reactive and imperative callers.
- Loader generations invalidate stale work after runtime/SDK changes, eviction, archive, delete, move, directory disposal, or loader disposal.
- Commits verify both generation and exact child-store ownership before publishing.
- Successful empty data remains a valid result; failed/missing payloads remain retryable failures and preserve prior state.
- Optimistic and authoritative messages reconcile without duplicate IDs or stale resurrection.

### Reconnect and recovery

- Startup 502/503 handling uses bounded recovery instead of converting failure into empty success.
- Reconnect restores current authoritative state without allowing stale fetch completion to overwrite newer events or mutations.
- Runtime and worktree switches invalidate old ownership before delayed work can commit.

### Event publication

- Ordered events retain their per-event side effects while directory-store publication is coalesced into fewer transactions.
- No-op updates preserve references rather than notifying every consumer.
- Structural session updates are separated from recency-only updates.
- Streaming text/part events update the owning session heartbeat instead of scanning all busy sessions.
- Live aggregate construction uses indexed/linear lookup paths rather than per-status scans across every session.

### Narrow subscriptions and stable ownership

- High-frequency status, permission, unseen, viewport, Git, and PR state is selected at the consuming entity where possible.
- Global session aggregation preserves unaffected project and directory references.
- Ownership is resolved once into indexes instead of repeatedly filtering all sessions for every project/group.
- Same-ID entities remain isolated by runtime and directory where collisions are possible.
- Custom memoization accounts for the canonical identity and semantic fields that actually drive each rendered row/tree.

### Hidden surfaces

- Hidden chat/sidebar surfaces suspend message prefetch, Git/PR enrichment, search effects, sticky observers, archived-folder derivation, and row-level subscriptions.
- Lightweight outer state remains mounted so reopening is immediate and resumes from current authoritative data.

### Chat and markdown

- Streaming markdown removes redundant pacing/reconciliation work.
- Turn/message projections reuse stable results where inputs are unchanged.
- Queue chips, status rows, model controls, header/layout surfaces, and context panels subscribe more narrowly.
- File/context loading rejects stale content and avoids duplicate load loops.

## Sidebar scalability and behavior

- Session grouping uses structural signatures so token/recency updates do not rebuild every project tree.
- Unchanged project sections and session nodes preserve references across unrelated updates.
- Parent expansion is exclusively manual and is scoped independently for Recent and project/worktree render contexts.
- Folder assignment avoids duplicate descendant rendering when a parent subtree is assigned.
- Pin state uses the same runtime/directory/session composite identity for rendering, sorting, persistence, and memo invalidation.
- Active incremental lists remain in normal flow, preserving `Show more`, focus, measurement, and scroll behavior.
- Archived collections use measured variable-height virtualization once they are large enough to justify it.
- Search, collapse, menu, editing, ordering, and nested-session changes invalidate only affected groups/rows.

### Destructive cleanup

- Explicit confirmed local deletion and routed authoritative deletion clean persisted state immediately.
- Failed delete/archive requests leave live and persisted state intact.
- `404` deletion is treated as confirmed absence for idempotent/cascade cases.
- The first complete startup list establishes a non-destructive baseline for snapshot-difference cleanup.
- Only later omission from the same authoritative runtime/scope is interpreted as a missed external deletion.
- Archive membership and directory/owner moves preserve stable identity and are not treated as deletion.

### Pins and folders

- Pins are keyed by runtime, normalized directory, and session ID.
- User-owned pins are no longer silently evicted by a fixed count limit.
- Folder namespaces are no longer silently evicted after a fixed runtime count.
- Pending browser writes are retained across runtime switches and lifecycle boundaries.
- Missing/malformed disk snapshots are not presented as authoritative empty data.
- Folder payloads are shape-validated before gaining authority.
- Server writes are serialized and stale/equal revisions cannot overwrite newer snapshots.
- Hydration checks runtime generation and local mutation revision before replacing browser state.
- Removing a session from every folder is committed atomically rather than through repeated intermediate writes.

### Drafts, queues, todos, and safe storage

- Chat drafts, inline-comment drafts, terminal context, queued messages, todos, and pins use runtime/directory/session ownership.
- Queued sends capture model/provider/agent configuration rather than re-reading mutable current settings later.
- Failed sends restore inline drafts without duplication.
- Safe-storage failure is isolated per key instead of disabling unrelated persistence globally.
- Runtime resets clear or rehydrate only the owning namespace.

## Runtime caches and stale-result protection

- File content cache ownership is centralized and scoped to the runtime/directory target.
- Git status, branch, diff, identity, and mutation channels reject stale async completion independently.
- PR status is scoped by repository/runtime rather than branch name alone.
- Provider, theme, file-search, tab, project, and worktree topology caches use explicit owner identity and invalidation.
- Global session fetches reconcile against commit-time mutations so create/update/move/archive/delete actions survive older in-flight responses.
- Settings snapshots are authoritative on success: omitted fields reset instead of leaking from the previous runtime; failure preserves prior state.

## Security and platform correctness

- Managed desktop startup prefers the bundled OpenCode CLI unless an explicit valid override is configured.
- Runtime URL normalization preserves secure credentials without exposing them in logs or unrelated state.
- Sensitive auth/settings files receive restrictive permissions.
- Permission auto-accept policy carries a monotonic revision; stale hydration, broadcasts, and mutation responses are rejected.
- Missing policy revisions sanitize to revision zero for compatibility with persisted older data.
- Mobile widget/notification snapshots are scoped by runtime.
- VS Code foreground/runtime transitions reconcile permission policy and bridge state without applying stale snapshots.

## Profiling and diagnostics

- Adds `bun run profile:browser` for a repeatable Chrome trace, network HAR, performance metrics, long-task attribution, and summary capture.
- Creates the browser profiling target explicitly and prefers Google Chrome for consistent tracing behavior.
- Adds bounded performance counters and user-timing marks for session loading, event publication, navigation, sidebar rendering, and message rendering.
- Adds deterministic operation-count regressions for high-frequency paths, including 10,000 streaming events across 50 busy sessions and linear live aggregation.
- Browser profile artifacts are ignored by Git.

## Engineering methodology

The relevant repository skills now encode the generalized lessons from this work:

- profiling proves cost, not behavioral equivalence;
- custom comparators are correctness boundaries and must gate canonical semantics in both directions;
- virtualization requires an explicit scroller, measurement, tail-reachability, interaction, and activation-transition contract;
- snapshot-difference deletion requires a same-scope authoritative baseline;
- generation invalidation and same-scope mutation reconciliation are separate concerns;
- persisted writes need explicit per-owner ordering, lifecycle durability, payload validation, and missing-versus-empty semantics.

## Compatibility and behavior changes

- Existing persisted permission policies without a revision normalize to revision `0`.
- Session pin/folder ownership is runtime- and directory-scoped; legacy unscoped pin data is not allowed to leak across owners.
- Successful complete settings snapshots reset omitted values to defaults; transport failure continues to preserve current values.
- Active lists no longer switch to virtualization by row count; archived lists retain bounded DOM behavior.
- `bun.lock` workspace version metadata is synchronized to `1.16.2`, and the existing Electron package is recorded as a trusted dependency.

## Validation

Validation performed across the branch includes:

- workspace-wide `bun run type-check`;
- workspace-wide `bun run lint`;
- UI, web, VS Code extension, and VS Code webview package checks;
- focused session bootstrap, child-store, message-loader, reconnect, event-pipeline, streaming, live-aggregate, persistence, deletion, global-session race, sidebar, pin, folder, Git/PR, draft, safe-storage, settings, permission, runtime-auth, and server-route suites;
- changed UI tests run in isolated processes where Bun `mock.module` contamination makes a combined runner unreliable;
- 27/27 settings-helper tests after aligning the permission-policy revision contract;
- 9 permission/settings-runtime tests passed, with one platform-conditional test skipped;
- server JavaScript syntax checks;
- `bun run docs:validate`;
- `bun run dead-code` inspected for new findings;
- `git diff --check` across the cumulative branch diff;
- manual browser profiles under the 55-running-session workload described above.

Static checks do not prove packaged/runtime behavior. This branch was not validated through packaged Electron/mobile builds or a full manual platform matrix. Browser captures are single manual scenarios rather than repeated statistical trials.

## Reviewer guide

The diff is broad, so the suggested review order is:

1. `packages/ui/src/sync/child-store.ts` and `session-message-loader.ts` for scheduling, ownership, and commit authority.
2. `event-pipeline.ts`, `streaming.ts`, `global-session-status.ts`, and `sync-context.tsx` for realtime publication and fanout.
3. global session, Git/PR, persistence, safe-storage, draft, pin, and folder stores for runtime scoping and stale-result handling.
4. session sidebar grouping/rendering modules for structural invalidation and archived-only virtualization.
5. permission/settings/runtime-auth and platform files for ordering and security behavior.
6. operation-count tests, race tests, browser profiler, and skill updates for regression prevention.

High-value invariants to scrutinize:

- one failed scope must not erase complete unrelated scopes;
- failure/missing/malformed state must never become authoritative empty success;
- delayed work must commit only to the runtime/directory/store instance that started it;
- same-scope mutations must survive older in-flight authoritative responses;
- hidden surfaces must stop expensive work without discarding authoritative state;
- memoization must skip unrelated updates without suppressing render-relevant changes;
- virtualized collections must keep their final item reachable through the real scrolling ancestor;
- destructive persisted cleanup must require explicit deletion or a valid established authoritative baseline.


btriapitsyn:
Thanks for the thorough review. I traced each finding through the current implementation and its tests. One finding exposed a real runtime-isolation gap and is fixed in `b10c5b490`; the other points are either deliberate tradeoffs or do not match the effective control flow.

## 1. `JSON.stringify` status comparison

**Acknowledged as a non-blocking fragility; no change in this PR.**

The current comparison can produce an unnecessary publication if two semantically equivalent objects arrive with different property insertion order. That is a performance false positive, not a stale-status correctness failure.

I did not narrow equality to only `type`, `attempt`, and `message` because `SessionStatus` can carry other meaningful SDK fields, including retry scheduling data such as `next`. A hand-maintained partial comparator would risk suppressing a real status change when the SDK evolves, which is worse than an occasional redundant publication.

If profiling shows repeated spurious status publications, the safer follow-up is a complete semantic comparator derived from the SDK status variants, with tests for every field, rather than a partial field list.

## 2. Safe-storage lifecycle retries after quota failure

**No change: the described retry loop does not occur through the effective adapter stack.**

`getDeferredSafeStorage()` wraps `getSafeStorage()`. The safe adapter catches a durable `setItem` failure, records a per-key in-memory override, and does not rethrow it to the deferred adapter. Because the deferred flush sees the call as handled, it removes that write from its pending queue; later lifecycle flushes do not retry the same failed write repeatedly.

A later explicit write to that key may attempt durable storage again, while unrelated keys remain durable. This behavior is covered by the safe-storage tests for per-key fallback and later recovery.

## 3. Module-level materialization cooldown crossing runtimes

**Fixed in `b10c5b490`.**

This was a valid runtime-isolation issue. The scheduler previously keyed cooldowns by directory and session ID only, so an identical pair in a newly selected runtime could be suppressed by the old runtime's marker.

The fix now:

- keys pending materialization by `runtimeKey + directory + sessionID` using an unambiguous tuple encoding;
- captures the runtime when the recovery is enqueued;
- drops deferred recovery before HTTP if that captured runtime is no longer active;
- retains the existing identity check so an older completion cannot clear a newer marker;
- documents the runtime-scoped ownership contract;
- adds a regression proving that identical directory/session identities produce independent keys across runtimes.

The extra runtime guard matters in addition to the composite key: it prevents an old queued microtask from reaching the current imperative loader after a runtime switch.

Validation for this fix:

- 34 focused materialization, event-reducer, and message-loader tests passed;
- UI type-check passed;
- UI lint passed;
- docs validation passed;
- dead-code report was inspected and showed no new finding for the added materialization identity export.

## 4. Structural selector plus live `getState()` read in the sidebar

**No change: this mixed read pattern is intentional.**

The structural selector is the subscription boundary. It excludes recency-only fields so high-frequency `time.updated` changes do not rerender the full sidebar tree. Once a structural change schedules a render, `getState()` supplies the latest authoritative arrays for that render.

Subscribing to the full arrays would restore broad high-frequency rerenders. Deriving the structural signature only from an imperative `getState()` read would remove the subscription that triggers structural rerenders. A store update that races the current render still changes the subscribed structural snapshot and schedules the follow-up render.

## 5. Awaiting Phase 3 session loading

**No change: holding the scheduler slot through the session-list fetch is deliberate bounded-concurrency behavior.**

The scheduler owns the complete directory bootstrap lifecycle and caps it at two concurrent bootstraps. Awaiting the session list is what prevents lower-priority directories from producing an unbounded request burst after the lightweight phases complete.

The scheduling policy also prevents background work from occupying both slots, preserving capacity for foreground demand. A slow background directory can delay later background refreshes, but it does not erase their demand and does not outrank selected/current work.

A timeout/abort policy for a genuinely non-settling SDK request could be useful resilience work, but it is a separate contract. Restoring fire-and-forget behavior would weaken the concurrency guarantee this PR intentionally adds.

## 6. Legacy pin data without migration

**No change: the discard is intentional because the legacy data lacks enough ownership information for a safe migration.**

The legacy value contains session IDs only. The new identity requires runtime, normalized directory, and session ID. Assigning an old ID to whichever runtime/directory happens to be active during upgrade would be heuristic and could pin the wrong session when identities collide or data is copied between runtimes.

This PR therefore chooses a one-time loss of ambiguous legacy pin selections over silently associating user state with the wrong runtime/session. The ownership boundary is documented in the store contract.

## Outcome

- Finding 3 resulted in a concrete correctness fix.
- Findings 1 and 5 are valid observations but remain intentional, documented tradeoffs.
- Finding 6 is an unavoidable ownership-safe migration boundary given the old schema.
- Findings 2 and 4 do not require changes because the actual adapter/subscription behavior already prevents the stated failure mode.

Thanks again for identifying the materialization cooldown gap.

btriapitsyn:
## Cold-start regression follow-up

Manual testing of the packaged build exposed a cold-start regression: a restored/selected session with a tiny history could remain on the full chat skeleton for a long time before its first messages appeared.

Fixed in `193eb51d0`.

### Root cause

Two startup paths compounded:

1. Directory bootstrap described command/MCP/LSP/VCS/question/permission loading as deferrable, but the implementation awaited that entire phase together with the authoritative session list. Two concurrent bootstraps could therefore start a large group of non-critical requests, retain scheduler ownership, and indirectly delay the selected session's small `session.messages` request through network/server contention.
2. Startup session restoration can request messages before the imperative loader registration effect runs. That request intentionally did not throw, but it was also not started. The React fallback then waited for the provider directory to equal the selected session directory before issuing the recovery load.

### Fix

- The authoritative directory session list remains awaited and scheduler-owned, preserving bounded pagination and failure authority.
- Deferrable command/MCP/LSP/VCS/question/permission enrichment starts detached only after the session list settles, so it no longer extends bootstrap slot ownership or competes with the initial session-list request.
- The visible chat fallback now calls `ensureSessionRenderable` with the selected session's explicit directory immediately; it no longer waits for provider-directory convergence.
- A regression test keeps deferred enrichment unresolved and proves bootstrap completes after the authoritative session list, then starts enrichment afterward.
- The sync ownership documentation now states the corrected phase contract.

### Correction to the earlier review response

This changes my assessment of review finding 5. Awaiting the authoritative session list is still intentional, but awaiting the deferrable enrichment phase was not necessary for that guarantee. The manual packaged-build result demonstrated an indirect priority inversion that static dependency analysis understated. The implementation now preserves bounded session-list loading without putting non-critical enrichment on the cold-session critical path.

### Validation

- 38 bootstrap, child-store, materialization, and message-loader tests passed;
- UI type-check passed;
- UI lint passed;
- docs validation passed.

The packaged cold-start scenario should be retested with `193eb51d0` or a newer build.

btriapitsyn:
## Live subagent Task linkage fix

Manual testing exposed an intermittent case where a child session existed and contained messages/tools, but its parent Agent Task card remained on `Waiting for subagent activity...`. Opening the child session did not repair the parent card.

Fixed in `b2f185f77` and manually confirmed no longer reproducible.

Two races were involved:

1. A parent tail refresh requested by child idle/discovery could arrive while an older parent message load was still in flight. `refreshTail()` previously returned that older promise and lost the newer refresh demand, leaving the parent Task part without its authoritative child `sessionId`.
2. While the parent assistant message was streaming, part-update suspension could retain the previous Task part even after `state.metadata.sessionId` reached the directory store. The parent renderer therefore continued to see the pre-link Task state.

The fix now:

- coalesces refresh requests made during an in-flight load and runs one refresh afterward;
- carries forward the largest requested refresh limit;
- rejects queued refreshes after loader generation, SDK epoch, entry ownership, or disposal changes;
- lets structural Task session-identity changes bypass streaming part suspension while text/reasoning deltas remain suspended;
- keeps the hot-path identity check allocation-free;
- documents both authority rules.

Regression coverage verifies:

- a refresh requested during an older in-flight load results in a second request and is not lost;
- duplicate overlapping refresh demands coalesce;
- Task `sessionId` becomes visible while other active-message part updates remain suspended.

Validation:

- 41 focused loader, message-record, materialization, child-store, and Task projection tests passed;
- UI type-check passed;
- UI lint passed;
- docs validation passed;
- manual reproduction no longer observed the bug.

Issue #1642, opened 2026-06-13
Clarify Auto-discover behavior and surface preview startup errors


pascalandr:
## Summary

The `Auto-discover` project action is hard to understand when it succeeds in launching something but the preview fails afterward.

In my case, Auto-discover did launch a preview URL, but the user-facing result looked like a broken preview/browser pane, without enough context about what was detected, what command/URL was launched, or why the preview failed.

## What happened

Auto-discover launched:

```text
Preview URL: http://127.0.0.1:5180
```

The preview then opened through the OpenChamber preview proxy and failed with runtime/loading errors:

```text
[2026-06-13T19:59:36.097Z] [runtime] Failed to fetch dynamically imported module:
http://127.0.0.1:64564/api/preview/proxy/73f4469f5914af48b36a64bfd0ac40d2/@fs/D:/openchamber/packages/ui/src/main.tsx?oc_preview_token=cdcef779110221a407681b07fe8401f7&oc_url_token=oc_url_2xkQ_otlSEr4qw629A5vm1icQX75XT6S

TypeError: Failed to fetch dynamically imported module:
http://127.0.0.1:64564/api/preview/proxy/73f4469f5914af48b36a64bfd0ac40d2/@fs/D:/openchamber/packages/ui/src/main.tsx?oc_preview_token=cdcef779110221a407681b07fe8401f7&oc_url_token=oc_url_2xkQ_otlSEr4qw629A5vm1icQX75XT6S

[2026-06-13T19:59:45.285Z] [warn] Loading screen timeout - forcing hide after 10s
```

In the browser pane, Vite also reported a build error:

```text
[plugin:vite:import-analysis] Failed to resolve import "@zumer/snapdom" from "../ui/src/lib/preview/screenshot-capture.ts". Does the file exist?
```

So Auto-discover did not simply do nothing: it detected/launched a preview server, but the resulting preview failed and the UX did not make the chain of events clear.

## Why this is confusing

The label `Auto-discover` is ambiguous. It is not obvious whether it is supposed to discover:

- workspaces/projects;
- a dev server;
- a browser preview;
- project actions;
- ports;
- something else.

From a quick code inspection, the matching implementation appears to be in:

- `packages/ui/src/components/layout/ProjectActionsButton.tsx`

It seems to auto-detect and launch a dev server/preview command, then wait for a URL and open the Preview pane.

## Expected behavior

The UI should make the Auto-discover flow more explicit and actionable.

For example, it could show:

[...]

PR #2361, opened 2026-07-21
fix(preview): rewrite relative ES module imports and strip crossorigin in HTML

### LINKED ISSUE #2338: Failed to fetch dynamically imported module — proxy doesn't authenticate relative ES module imports in production builds (Nuxt 3)
### Description

When previewing a Nuxt 3 production build through the OpenChamber preview proxy, the browser fails to load dynamically imported ES modules with a `Failed to fetch dynamically imported module` error. The proxy returns `403` for relative module imports because the `oc_preview_token` is neither present in the URL query string nor sent via cookie.

### Root Cause

Two compounding gaps in the proxy's rewriting and authentication logic:

**1. \`rewriteJavaScript\` does not rewrite relative import paths**

Nuxt 3 production chunks use relative imports (e.g., \`import{...}from"./DlAUqK2U.js"\`). The \`rewriteJavaScript\` function only matches paths starting with \`/\`:

https://github.com/openchamber/openchamber/blob/main/packages/web/server/lib/preview/proxy-runtime.js#L1086-L1095

\`\`\`js
const rewriteJavaScript = (text) => text
  .replace(/\bfrom\s+(['\"])\/(?!\/)([^'"]*)\1/gi, ...)       // only "/..."
  .replace(/\bimport\s+(['\"])\/(?!\/)([^'"]*)\1/gi, ...)       // only "/..."
  .replace(/\bimport\(\s*(['\"])\/(?!\/)([^'"]*)\1\s*\)/gi, ...) // only "/..."
\`\`\`

Relative paths like \`./DlAUqK2U.js\` are not matched. The browser resolves them against the importing module's URL, which **drops the query string** per the WHATWG URL spec:

- Base: \`/api/preview/proxy/<id>/_nuxt/C2VRKy4g.js?oc_preview_token=<token>\`
- Relative: \`./DlAUqK2U.js\`
- Resolved: \`/api/preview/proxy/<id>/_nuxt/DlAUqK2U.js\` ← **no token**

**2. \`crossorigin=""\` on \`<link>\` and \`<script>\` tags prevents cookie fallback**

Nuxt's SSR renderer (\`vue-bundle-renderer\`) sets \`crossorigin=""\` on all \`modulepreload\` links and module scripts. The proxy's \`rewriteHtml\` rewrites the \`href\`/\`src\` URLs but does **not strip the \`crossorigin\` attribute**.

\`crossorigin=""\` is equivalent to \`crossorigin="anonymous"\`, which sets the fetch credentials mode to \`"omit"\`. The browser does not send the \`oc_preview_token\` cookie for these requests — not even for same-origin requests.

The proxy's \`resolveTargetFromRequest\` accepts the token from either the query param or the cookie:

\`\`\`js
const token = parsed.searchParams.get(TOKEN_QUERY_PARAM) || cookies.get(TOKEN_COOKIE_NAME) || '';
\`\`\`

But with the query param missing (gap 1) and the cookie suppressed by \`crossorigin=""\` (gap 2), there is no token → \`403\`.

**3. Client-side \`fetch\` patch does not cover native ES module imports**

The bridge script's \`installAppRequestProxyPatch\` patches \`window.fetch\` and \`XMLHttpRequest\`, but native ES module imports bypass \`window.fetch\` — the browser's module loader uses its own internal fetch mechanism. So the client-side patch cannot intercept relative module imports.

### Reproduction

1. Build a Nuxt 3 app for production: \`nuxt build\`
2. Serve the production build: \`node .output/server/index.mjs\` (port 3000)
3. Open the preview in OpenChamber (which proxies to \`http://localhost:3000\`)
4. Observe the console error: \`Failed to fetch dynamically imported module: /api/preview/proxy/<id>/_nuxt/DlAUqK2U.js\`
5. The proxy returns \`403\` with \`{ code: 'invalid-token', error: 'Preview token missing' }\`

### Expected Behaviour

Relative ES module imports should be authenticated by the proxy, either by:
- Rewriting relative import paths in JavaScript responses to include the token, **or**
- Allowing the \`oc_preview_token\` cookie to be sent as a fallback by stripping \`crossorigin=""\` from \`<link>\` and \`<script>\` tags

### Environment

- OpenChamber: latest \`main\`
- Framework: Nuxt 3 (production build, Nitro SSR)
- Browser: Chrome/Safari/Firefox (all honour \`crossorigin="anonymous"\`)

### Suggested Fix

**Option A (minimal):** Strip the \`crossorigin\` attribute from \`<link>\` and \`<script>\` tags in \`rewriteHtml\`. Without \`crossorigin=""\`, the browser defaults to \`credentials: "same-origin"\` for same-origin module requests, allowing the \`oc_preview_token\` cookie to be sent as a fallback.

**Option B (robust):** Extend \`rewriteJavaScript\` to handle relative import paths by resolving them against the module's upstream path (which the proxy knows from the request URL) and appending the token query params.

**Option C (both):** Apply both fixes for defence in depth.

### Related Issues

- #1642 — mentions \`Failed to fetch dynamically imported module\` but in the context of auto-discover behaviour
- #1470 — related preview proxy authentication discussion

bashrusakh:
## Summary

Fixes #2338 — the preview proxy failed to authenticate relative ES module imports in Nuxt 3 production builds, causing `Failed to fetch dynamically imported module` errors with 403 responses.

### Root cause

Two compounding gaps in the proxy's rewriting and authentication logic:

1. **`rewriteJavaScript` did not rewrite relative import paths** — only absolute paths starting with `/` were matched. Relative imports like `./DlAUqK2U.js` were left unchanged, and the browser resolved them against the module URL, dropping the `oc_preview_token` query parameter per the WHATWG URL spec.

2. **`rewriteHtml` did not strip `crossorigin=""`** from `<link>` and `<script>` tags. Since `crossorigin=""` is equivalent to `crossorigin="anonymous"`, the browser sets credentials mode to `"omit"`, preventing the `oc_preview_token` cookie from being sent as a fallback.

3. The client-side `fetch` patch cannot intercept native ES module imports, so neither URL-based nor cookie-based auth worked for relative imports.

### Fix (Option C — both fixes for defense in depth)

- Add `modulePath` parameter to `rewritePreviewBody` (default `""`, backward-compatible). When provided for JavaScript responses, derive `moduleDir` and resolve relative import paths (`./…`, `../…`) against it, then rewrite through `rewriteResourceUrl` with auth tokens.

- Add regex patterns in `rewriteJavaScript` to match relative import specifiers (`from "./…"`, `import "./…"`, `import("./…")`) and resolve them against the module directory before rewriting.

- Pass `upstreamPath` as `modulePath` at call sites for JavaScript responses; leave it empty for HTML/CSS.

- Strip `crossorigin` attributes (empty, `anonymous`, `use-credentials`, or bare boolean) from HTML before rewriting `src`/`href`, so the browser defaults to `credentials: "same-origin"` for same-origin module requests, allowing the `oc_preview_token` cookie fallback.

- Inline module scripts (`rewriteInlineModuleScripts`) pass empty dir to `rewriteJavaScript`, leaving relative imports unchanged since they don't have a module URL context.

### Related

- #1470 — previous fix that added inline module script rewriting (partial fix; did not cover external module scripts with `crossorigin` or relative imports in external JS)
- #1642 — mentions `Failed to fetch dynamically imported module` in a different context (auto-discover error surfacing)

### DIFF
--- a/packages/web/server/lib/preview/proxy-runtime.js
+++ b/packages/web/server/lib/preview/proxy-runtime.js
@@ -1038,7 +1038,7 @@ const appendProxyAuthToProxyUrl = (value, { previewToken = '', urlAuthToken = ''
 
 const normalizeLoopbackUrl = (rawUrl) => normalizeProxyTargetUrl(rawUrl, { allowExternal: false });
 
-export const rewritePreviewBody = ({ bodyText, proxyBasePath, targetOrigin, kind, previewToken = '', urlAuthToken = '' }) => {
+export const rewritePreviewBody = ({ bodyText, proxyBasePath, targetOrigin, kind, previewToken = '', urlAuthToken = '', modulePath = '' }) => {
   if (typeof bodyText !== 'string' || bodyText.length === 0) {
     return bodyText;
   }
@@ -1072,6 +1072,25 @@ export const rewritePreviewBody = ({ bodyText, proxyBasePath, targetOrigin, kind
     }
     return value;
   };
+  const moduleDir = (() => {
+    if (!modulePath) return '';
+    try {
+      const parsed = new URL(modulePath, 'http://localhost');
+      const dir = parsed.pathname.replace(/[^/]*$/, '');
+      return dir;
+    } catch {
+      return '';
+    }
+  })();
+  const resolveRelativeUrl = (relativePath) => {
+    if (!moduleDir) return relativePath;
+    try {
+      const resolved = new URL(relativePath, `http://localhost${moduleDir}`);
+      return resolved.pathname;
+    } catch {
+      return relativePath;
+    }
+  };
   const stripPreviewCspMeta = (text) => text
     .replace(/<meta\b(?=[^>]*\bhttp-equiv\s*=\s*(['"])content-security-policy\1)[^>]*>/gi, '')
     .replace(/<meta\b(?=[^>]*\bhttp-equiv\s*=\s*content-security-policy\b)[^>]*>/gi, '');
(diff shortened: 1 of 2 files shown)


btriapitsyn:

PR #2362, opened 2026-07-21
fix(vscode): break static import chain that prevents VS Code workspace detection

### LINKED ISSUE #2354: [Bug] "Add project directory" modal auto-opens in VS Code since 1.16.2; workspace folder no longer auto-detected

Starting with v1.16.2, the "Add project directory" modal (`DirectoryExplorerDialog`) auto-opens in VS Code. The VS Code workspace folder is no longer auto-detected as a project.

This is a **regression** introduced in v1.16.2. Prior versions (1.16.0, 1.16.1) correctly auto-detected the VS Code workspace folder.

Windows/Linux both affected. Workaround: Downgrade to v1.16.1 (confirmed).

## Steps to reproduce

1. Install/upgrade to OpenChamber v1.16.2.
2. Open VS Code with a workspace folder (e.g. `code /path/to/project`).
3. Open the OpenChamber panel.
4. The "Add project directory" modal appears.

## Expected behavior

The current VS Code workspace folder is used as the project directory automatically. No "Add project directory" prompt should appear when a workspace folder is open. Switching folders should update the active project.

## Actual behavior

"Add project directory" modal appears. Projects that have been added through the modal appear in the project list.

### What changed

Commit `94326f22` (between v1.16.1 and v1.16.2, released in v1.16.2) added two **static** imports to `packages/vscode/webview/main.tsx`:

```ts
import { usePermissionStore } from '@openchamber/ui/stores/permissionStore';
import { processVSCodePermissionAutoAccept } from '@openchamber/ui/sync/vscode-permission-auto-accept';
```

At v1.16.1, `main.tsx` did **not** statically import `permissionStore`. All store imports were dynamic (`await import(...)` inside event handlers). The static imports were only framework utilities (`opencode/client`, `runtime-fetch`, `theme/adapter`, `i18n`) — none of which transitively reach `useProjectsStore`.

### The resulting import chain

The new static import creates this transitive chain:

```
main.tsx (static import — evaluated before module body executes)
  → permissionStore (line 16)
    → session-ui-store (permissionStore.ts:9)
      → useProjectsStore (session-ui-store.ts:22)  ← initializes HERE, before main.tsx body
```

ES module semantics require all **static imports to be evaluated before the module body executes**. This means `useProjectsStore` now initializes **before** `main.tsx:62` sets `window.__OPENCHAMBER_RUNTIME_APIS__ = createVSCodeAPIs()`.

### The race condition

1. **HTML inline script** (`packages/vscode/src/webviewHtml.ts:178-195`) sets `window.__VSCODE_CONFIG__` (including `workspaceFolder`) and `window.__OPENCHAMBER_HOME__`. These globals are available **before any bundled JS runs**.

2. **Bundled JS loads** (`packages/vscode/webview/main.tsx`). Static imports are evaluated first per ES module semantics. The import chain above reaches `useProjectsStore`.

3. **`useProjectsStore` initializes** at `packages/ui/src/stores/useProjectsStore.ts:543-548`. Two module-level constants are computed:
   - `vscodeWorkspace = getVSCodeWorkspaceProject()` (line 543) — calls `getVSCodeWorkspaceFolders()` (line 418-456) which gates on `getRegisteredRuntimeAPIs()?.runtime?.isVSCode` (line 423)
   - `isVSCodeProjectsRuntime` (lines 544-547) — also calls `getRegisteredRuntimeAPIs()?.runtime?.isVSCode`

   `getRegisteredRuntimeAPIs()` (`packages/ui/src/contexts/runtimeAPIRegistry.ts:9-19`) returns `null` at this point — **both** fallback sources are unavailable:
   - The module-scoped `registeredRuntimeAPIs` variable is not set until `VSCodeApp`'s `useEffect` runs (`packages/ui/src/apps/VSCodeApp.tsx:50`), which happens after React mount.
   - `window.__OPENCHAMBER_RUNTIME_APIS__` is set in `main.tsx:62`'s module body, which runs **after** the static imports above are evaluated.

   Result: `vscodeWorkspace = null`, `isVSCodeProjectsRuntime = false`, `effectiveInitialProjects = initialProjects` (persisted — empty on fresh `localStorage`).

4. **`main.tsx` module body finally runs** and sets `window.__OPENCHAMBER_RUNTIME_APIS__ = createVSCodeAPIs();` (`main.tsx:62`) — but this is **too late**: the constants are already frozen.

5. **React renders `SessionDialogs`** (`packages/ui/src/components/session/SessionDialogs.tsx:118-130`):

   ```ts
   React.useEffect(() => {
     if (hasShownInitialDirectoryPrompt || !isHomeReady || projects.length > 0) {
       return;
     }
     setHasShownInitialDirectoryPrompt(true);
     setIsDirectoryDialogOpen(true);
   }, [hasShownInitialDirectoryPrompt, isHomeReady, projects.length]);
   ```

   - `projects.length === 0` ✓ (projects store initialized empty)
   - `isHomeReady === true` ✓ (`useDirectoryStore` reads `window.__OPENCHAMBER_HOME__` **directly** at `packages/ui/src/stores/useDirectoryStore.ts:122-131`, so it works despite the same race)
   - `hasShownInitialDirectoryPrompt` is `React.useState(false)` (line 54) — **not persisted**, resets every webview mount
   - → **Dialog auto-opens on every load.**

### Where does it happen?

VS Code extension

### Version (if known)

1.16.2

Image

### Logs (optional)

```shell

```

### LINKED ISSUE #2359: VS Code Extension: workspace folder not used as default project on startup
## Description

When opening the OpenChamber VS Code extension panel, the web app does not use the current VS Code workspace folder as the default project. Instead it either:

- Shows a previously selected project (from a different VS Code window)
- Or shows an empty project picker (if no projects are cached)

## Root Cause

In the bundled JS, store modules (e.g., `useProjectsStore`) are loaded as leaf modules and execute **before** `window.__OPENCHAMBER_RUNTIME_APIS__` is set by `packages/vscode/webview/main.tsx`. 

At store initialization time:

```javascript
// sD() = getVSCodeWorkspaceFolders()
sD = () => {
    if (typeof window > "u" || !qt()?.runtime?.isVSCode) return null; // ← IS NULL!
    // ...
};

ed = aD(); // getVSCodeWorkspaceProject() → returns null
ai = ed?.projects ?? (Ze ? [] : oD); // effectiveInitialProjects = localStorage
```

Since `getRegisteredRuntimeAPIs()` returns `null` (neither the in-memory registry nor `window.__OPENCHAMBER_RUNTIME_APIS__` are set yet), `isVSCodeRuntime()` returns `false`, and the stores fall back to reading from localStorage.

Meanwhile, `window.__VSCODE_CONFIG__.workspaceFolder` IS correctly injected in the HTML template (`packages/vscode/src/webviewHtml.ts`), but it is never used because the stores skip the VS Code detection path.

## Relevant Files

- `packages/vscode/webview/main.tsx` — sets `window.__OPENCHAMBER_RUNTIME_APIS__` too late
- `packages/vscode/src/webviewHtml.ts` — generates the HTML template (has `__VSCODE_CONFIG__` but not `__OPENCHAMBER_RUNTIME_APIS__`)
- `packages/ui/src/stores/useProjectsStore.ts` — projects store initialization (`effectiveInitialProjects`)
- `packages/ui/src/stores/useDirectoryStore.ts` — directory store initialization (`initialCurrentDirectory`)
- `packages/ui/src/contexts/runtimeAPIRegistry.ts` — `getRegisteredRuntimeAPIs()` fallback chain

## Proposed Fix

Inject `window.__OPENCHAMBER_RUNTIME_APIS__` in the webview HTML template (`webviewHtml.ts`) **before** the React bundle loads, so it is available during all module initialization:

```typescript
// In webviewHtml.ts, inside the <script> block alongside __VSCODE_CONFIG__:
window.__OPENCHAMBER_RUNTIME_APIS__ = {
  runtime: { platform: "vscode", isDesktop: false, isVSCode: true }
};
```

This ensures `getRegisteredRuntimeAPIs().runtime.isVSCode` returns `true` at store module initialization time, allowing `getVSCodeWorkspaceProject()` to correctly derive the projects from the current VS Code workspace folders.

## Environment

- **OpenChamber version:** 1.16.2 (VS Code extension)
- **VS Code:** Insiders
- **OS:** macOS

bashrusakh:
## Summary

Fixes #2354, Fixes #2359 — regression where the "Add project directory" modal auto-opens in VS Code since v1.16.2 because the workspace folder is no longer auto-detected.

## Root Cause

Commit `94326f22` added two static imports to `packages/vscode/webview/main.tsx`:

```
import { usePermissionStore } from "@openchamber/ui/stores/permissionStore";
import { processVSCodePermissionAutoAccept } from "@openchamber/ui/sync/vscode-permission-auto-accept";
```

This created a transitive import chain:

```
main.tsx → permissionStore → session-ui-store → useProjectsStore
```

ES module semantics evaluate all static imports **before** the module body runs. `useProjectsStore` initialized before `main.tsx:62` set `window.__OPENCHAMBER_RUNTIME_APIS__ = createVSCodeAPIs()`. `getRegisteredRuntimeAPIs()` returned `null`, so `vscodeWorkspace = null`, `isVSCodeProjectsRuntime = false`, and `effectiveInitialProjects = []`. The `SessionDialogs` effect then opened the dialog because `projects.length === 0`.

### 1. Break the import chain in `main.tsx`

Convert the two static imports to dynamic `await import()` at their usage sites:
- `processVSCodePermissionAutoAccept` → dynamic import in the `permission.asked` notification handler (already async)
- `usePermissionStore` → dynamic import in the `permissionAutoAcceptSynced` command handler (made async)

`import type { PermissionRequest }` is kept as type-only since it does not trigger module evaluation.

### 2. Lazy initialization in `useProjectsStore`

Convert module-level constants (`vscodeWorkspace`, `isVSCodeProjectsRuntime`) to getter functions (`getVSCodeWorkspaceProject()` directly, `getIsVSCodeProjectsRuntime()`) and compute initial state inside `getInitialProjectsState()`. Move the computation from module-level `const initialProjectsState = getInitialProjectsState()` into the Zustand creator body (`const initial = getInitialProjectsState()` inside `devtools((set, get) => { ... })`). This ensures the state is computed when the store is created, not at module evaluation time.

### 3. Guard async command handlers

Wrap `handler(message.payload)` in `Promise.resolve().catch()` in `bridge.ts` to prevent unhandled promise rejections when async handlers (like `permissionAutoAcceptSynced`) throw. The `CommandHandler` type is `(payload: unknown) => void`, so async handlers silently drop rejections without this guard.

## Note on #1884 / PR #1945

PR #1945 replaces dynamic imports with static ones in `session-ui-store.ts` to silence Rollup mixed-import warnings. That change is **orthogonal** to this fix — `session-ui-store`'s mixed imports are a different set of modules (`sonner`, `i18n/store`, `session-actions`). Our lazy-init hardening in `useProjectsStore` ensures that even if `session-ui-store` gets more static imports, the projects store won't be affected because `getRegisteredRuntimeAPIs()` is no longer called at module-evaluation time.

If both PRs merge, the Rollup warnings from #1884 would be resolved by #1945's approach (static imports in `session-ui-store`), and the runtime race condition from #2354 would be resolved by our approach (dynamic imports in `main.tsx` + lazy init in `useProjectsStore`). They complement each other.

## Changed Files

- `packages/vscode/webview/main.tsx` — removed 2 static imports, added 2 dynamic imports
- `packages/ui/src/stores/useProjectsStore.ts` — converted module-level constants to getter functions, moved initial state computation into Zustand creator, removed trivial `getVSCodeWorkspace` wrapper
- `packages/vscode/webview/api/bridge.ts` — wrapped command handler dispatch in `Promise.resolve().catch()` to guard async handlers

## Verification

- TypeScript type-check passes for both `packages/ui` and `packages/vscode` (webview tsconfig)
- 81 permission/project/runtime-related tests pass
- OCR review passed — all medium/high findings addressed
- CI checks: all green (checks, review, label)

### DIFF
--- a/packages/ui/src/stores/useProjectsStore.ts
+++ b/packages/ui/src/stores/useProjectsStore.ts
@@ -542,7 +542,6 @@ const getVSCodeWorkspaceProject = (): { projects: ProjectEntry[]; activeProjectI
 // because static imports can reach this module before the runtime registry is
 // populated (see #2354). Use getters/functions so the values are computed on
 // first access, after the runtime APIs have been registered.
-const getVSCodeWorkspace = (): ReturnType<typeof getVSCodeWorkspaceProject> => getVSCodeWorkspaceProject();
 const getIsVSCodeProjectsRuntime = (): boolean => {
   if (typeof window === 'undefined') return false;
   return Boolean(getRegisteredRuntimeAPIs()?.runtime?.isVSCode);
@@ -552,7 +551,7 @@ function getInitialProjectsState(): {
   projects: ProjectEntry[];
   activeProjectId: string | null;
 } {
-  const vscodeWorkspace = getVSCodeWorkspace();
+  const vscodeWorkspace = getVSCodeWorkspaceProject();
   const isVSCodeProjectsRuntime = getIsVSCodeProjectsRuntime();
   const effectiveInitialProjects = vscodeWorkspace?.projects ?? (isVSCodeProjectsRuntime ? [] : initialProjects);
   const persistedInitialActiveProjectId = vscodeWorkspace?.activeProjectId ?? (isVSCodeProjectsRuntime ? null : readPersistedActiveProjectId());
@@ -567,12 +566,13 @@ function getInitialProjectsState(): {
   return { projects: effectiveInitialProjects, activeProjectId: initialActiveProjectId };
 }
 
-const initialProjectsState = getInitialProjectsState();
 
 export const useProjectsStore = create<ProjectsStore>()(
-  devtools((set, get) => ({
-    projects: initialProjectsState.projects,
-    activeProjectId: initialProjectsState.activeProjectId,
+  devtools((set, get) => {
+    const initial = getInitialProjectsState();
+    return {
+    projects: initial.projects,
+    activeProjectId: initial.activeProjectId,
     manualProjectOrder: readPersistedManualOrder(),
 
     validateProjectPath: (path: string): ProjectPathValidationResult => {
@@ -1014,7 +1014,8 @@ export const useProjectsStore = create<ProjectsStore>()(
       return projects.find((project) => project.id === activeProjectId) ?? null;
     },
 
-  }), { name: 'projects-store' })
+  };
+  }, { name: 'projects-store' })
 );
 
 if (typeof window !== 'undefined') {
--- a/packages/ui/src/stores/useProjectsStore.ts
+++ b/packages/ui/src/stores/useProjectsStore.ts
@@ -538,27 +538,41 @@ const getVSCodeWorkspaceProject = (): { projects: ProjectEntry[]; activeProjectI
   return { projects: result.projects, activeProjectId: result.activeProjectId };
 };
 
-// VS Code runtime is scoped to the workspace folders opened in VS Code.
-// Always prefer the VS Code workspace projects over any persisted multi-project registry.
-const vscodeWorkspace = getVSCodeWorkspaceProject();
-const isVSCodeProjectsRuntime = (() => {
+// Lazy initialization: these must not be evaluated at module-evaluation time
+// because static imports can reach this module before the runtime registry is
+// populated (see #2354). Use getters/functions so the values are computed on
+// first access, after the runtime APIs have been registered.
+const getVSCodeWorkspace = (): ReturnType<typeof getVSCodeWorkspaceProject> => getVSCodeWorkspaceProject();
+const getIsVSCodeProjectsRuntime = (): boolean => {
   if (typeof window === 'undefined') return false;
   return Boolean(getRegisteredRuntimeAPIs()?.runtime?.isVSCode);
-})();
-const effectiveInitialProjects = vscodeWorkspace?.projects ?? (isVSCodeProjectsRuntime ? [] : initialProjects);
-const persistedInitialActiveProjectId = vscodeWorkspace?.activeProjectId ?? (isVSCodeProjectsRuntime ? null : readPersistedActiveProjectId());
-const initialActiveProjectId = effectiveInitialProjects.some((project) => project.id === persistedInitialActiveProjectId)
-  ? persistedInitialActiveProjectId
-  : effectiveInitialProjects[0]?.id ?? null;
+};
 
-if (vscodeWorkspace) {
-  cacheProjects(effectiveInitialProjects, initialActiveProjectId);
+function getInitialProjectsState(): {
+  projects: ProjectEntry[];
+  activeProjectId: string | null;
+} {
+  const vscodeWorkspace = getVSCodeWorkspace();
+  const isVSCodeProjectsRuntime = getIsVSCodeProjectsRuntime();
+  const effectiveInitialProjects = vscodeWorkspace?.projects ?? (isVSCodeProjectsRuntime ? [] : initialProjects);
+  const persistedInitialActiveProjectId = vscodeWorkspace?.activeProjectId ?? (isVSCodeProjectsRuntime ? null : readPersistedActiveProjectId());
+  const initialActiveProjectId = effectiveInitialProjects.some((project) => project.id === persistedInitialActiveProjectId)
+    ? persistedInitialActiveProjectId
+    : effectiveInitialProjects[0]?.id ?? null;
+
+  if (vscodeWorkspace) {
+    cacheProjects(effectiveInitialProjects, initialActiveProjectId);
+  }
+
+  return { projects: effectiveInitialProjects, activeProjectId: initialActiveProjectId };
 }
+
+const initialProjectsState = getInitialProjectsState();
 
 export const useProjectsStore = create<ProjectsStore>()(
   devtools((set, get) => ({
-    projects: effectiveInitialProjects,
-    activeProjectId: initialActiveProjectId,
+    projects: initialProjectsState.projects,
+    activeProjectId: initialProjectsState.activeProjectId,
     manualProjectOrder: readPersistedManualOrder(),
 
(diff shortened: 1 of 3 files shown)


bashrusakh:
This PR also fixes #2359 — same root cause (store modules evaluate before `window.__OPENCHAMBER_RUNTIME_APIS__` is set), slightly different symptom (workspace folder not used as default project instead of auto-opening the "Add project directory" modal).

The dynamic imports + lazy init approach resolves both issues.

btriapitsyn:
Thank you for investigating this and for identifying the static import chain. Your root-cause analysis pointed us to the initialization-order regression.

We went with a smaller runtime-boundary fix instead of merging this implementation. The `window.__VSCODE_CONFIG__` bootstrap object is injected by the extension host before the webview bundle loads, so workspace/project initialization can use that as the authoritative early VS Code marker and folder source, with the registered runtime APIs as a fallback. This makes startup independent of the entire static import graph.

The proposed `getInitialProjectsState()` move does not actually defer initialization: Zustand invokes the creator synchronously from `create(...)` while the module is still being evaluated. The dynamic permission imports therefore fix the current specific chain, but another static import could reintroduce the same regression. The async command-handler change is also separate from the workspace detection bug and broadens the patch unnecessarily.

The replacement fix has now landed in [`e9e41139`](https://github.com/openchamber/openchamber/commit/e9e41139d904105b14d6d7784af2dbf1e623ff22). It uses the bootstrap-config approach with regression coverage for detection before runtime API registration, while keeping the static permission imports in place to ensure import order is no longer part of the contract.

Thanks again for the useful diagnosis and for taking the time to put together the PR.

PR #2366, opened 2026-07-21
fix(ui): stop chat entry animation replaying on session switch

### LINKED ISSUE #2124: [Bug] Chat animation replays on session switching even if already displayed chat.

Switching between sessions, it re-animates already seen chat messages, even if they are above sub agent work, and it causes confusion as to what is new and not within the session, plus it delays reading what is already there. 

This is annoying when clicking between many sessions constantly trying to monitor.

### Where does it happen?

Desktop (macOS)

### Version (if known)

1.14.1

_No response_

### Logs (optional)

```shell

```

IbrahimKhan12:
## Intent

Fixes #2124. Switching sessions replays the entry-animation decision for already-seen assistant messages.

`MessageFreshnessDetector` should mark an assistant message animation-eligible only once, but a message judged "fresh" was never recorded as shown. On a session switch the chat viewport remounts and `recordSessionStart()` runs in a post-render effect, so the first render still sees the stale start time, the message re-passes the freshness check, and it re-animates.

## Fix

Record the message from a committed effect in `ChatMessage`, via the existing `markMessageAsAnimated` API, once it has rendered. Marking after commit rather than inside the memoized `shouldAnimateMessage` call avoids flipping the decision to `false` on React StrictMode's second render pass, which would suppress the first animation. The detector is unchanged.

## Non-goals / accuracy note

Not re-enabling or reworking downstream animation wiring. On `main` much of the entry-animation pipeline is currently inert (`FADE_ANIMATION_ENABLED = false`, `allowAnimation` is `void`-ed in the assistant `MessageBody`, reservation/animation-start handlers are no-ops), so this change restores the one-shot freshness latch those consumers (and the live scroll-space reservation) depend on rather than guaranteeing a visual delta on the current build.

## Affected surfaces

`packages/ui` shared chat UI: `ChatMessage` (new effect) and `MessageFreshnessDetector` (unchanged, now marked by the consumer). Applies across runtimes via the shared singleton. No public API/type/export change.

## Validation

| Check | Result |
|---|---|
| `bun test packages/ui/src/lib/messageFreshness.test.ts` | 5 pass |
| Same tests vs a rejected render-mutation variant | 2 fail (guards the StrictMode-safe design) |
| `bun run type-check:ui` | Pass |
| ESLint (changed files) | Clean |

Not covered: a full `ChatMessage` render test. The repo's React tests use `renderToStaticMarkup` (no effects) and there is no DOM test runner; adding one needs a new dependency, which AGENTS.md disallows. The tests guard the detector contract and StrictMode-safety property the fix relies on; the effect wiring is verified by type-check, lint, and review.

## Risks and failure behavior

Pure in-memory logic; no I/O or persistence. The effect adds two idempotent `Set`/`Map` writes after a fresh assistant message renders (safe under StrictMode's double-invoked effects). Fresh message IDs are now retained in the detector for the tab's lifetime (historical messages already were); growth is bounded by messages seen per tab session and is negligible. No behavior change for user, historical, or already-seen messages.

### DIFF
--- a/packages/ui/src/components/chat/ChatMessage.tsx
+++ b/packages/ui/src/components/chat/ChatMessage.tsx
@@ -573,6 +573,17 @@ const ChatMessage: React.FC<ChatMessageProps> = ({
         return freshnessDetector.shouldAnimateMessage(message.info, message.info.sessionID);
     }, [message.info, isUser]);
 
+    // Once a fresh assistant message has rendered, record it so it is not animated
+    // again when the chat viewport remounts on a session switch (issue #2124).
+    // Marking runs in a committed effect rather than inside shouldAnimateMessage
+    // (which runs in a useMemo during render): under React StrictMode the render is
+    // invoked twice, so marking there would flip the memoized result to false on the
+    // second pass and suppress the first animation.
+    React.useEffect(() => {
+        if (!shouldAnimateMessage) return;
+        MessageFreshnessDetector.getInstance().markMessageAsAnimated(message.info.id, message.info.time.created);
+    }, [shouldAnimateMessage, message.info.id, message.info.time.created]);
+
     const [hasStartedStreamingHeader, setHasStartedStreamingHeader] = React.useState(false);
 
     const nextRole = React.useMemo(() => {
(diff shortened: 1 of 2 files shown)


IbrahimKhan12:
Fixed by #2360; closing PR.

PR #2367, opened 2026-07-21
fix: preserve plugin tool `state.attachments` in UI and materializer

### LINKED ISSUE #2365: [Bug] Plugin tool `state.attachments` silently dropped by UI and materializer

**Pulled from [#2350 (comment)](https://github.com/openchamber/openchamber/issues/2350#issuecomment-5037993333):** two gaps surfaced during investigation of plugin tool attachment support

### 1. `ToolPart.tsx` never renders `state.attachments`

**File:** `packages/ui/src/components/chat/message/parts/ToolPart.tsx`
`ToolStateWithMetadata` (line 62) includes `output?: string` but has no `attachments` field. The entire rendering path only handles plain text output (`stateWithData.output` at line 1512, `renderResultContent()` at line 1591, the `'output' in state` guard at line 1900). Plugin tool attachments (e.g., image thumbnails returned via `ToolResult.attachments`) are invisible to the user even when the model receives them correctly.

Attachments should be rendered inline, similar to how the `read` tool displays image results (see `ToolExpandedContent` around line 1512).

### 2. `mergeMaterializedPart()` unconditionally overwrites completed tool parts

**File:** `packages/ui/src/sync/materialization.ts`
At line 93, `getPartEndTime(next) !== undefined` causes `mergeMaterializedPart` to return the server snapshot verbatim, discarding any live state from the existing part. If a plugin tool part accumulates `state.attachments` client-side (from streaming or local computation), a subsequent server snapshot with an end time will silently drop those attachments.
The `STREAMING_PART_FIELDS` merge loop (lines 96–107) only preserves `"text"` and `"output"` — there is no mechanism to preserve other state fields like `attachments` when a completed server snapshot arrives.

### Where does it happen?

Desktop (macOS)

### Version (if known)

1.16.2

_No response_

### Logs (optional)

```shell

```

FrostiDrinks:
## Intent

**Closes https://github.com/openchamber/openchamber/issues/2365**

Plugin tool results can include `state.attachments` (e.g., image thumbnails from `ToolResult.attachments`), but the UI never rendered them and the materializer silently dropped them during server snapshot reconciliation. As a result, agents would receive attachments correctly but they'd never appear for users.

Two concrete fixes:

1. `ToolPart.tsx` — `ToolStateWithMetadata` omitted `attachments`; the rendering path only handled text `output`. `state.attachments` are now rendered. Images are clickable thumbnails (with `ImagePreviewDialog` popup), other files are chips with `FileTypeIcon`. Both follow the same patterns as `MessageFilesDisplay`.

2. `materialization.ts` — `mergeMaterializedPart()` returned the server snapshot verbatim when `getPartEndTime(next) !== undefined`, discarding client-side `state.attachments`. The `state.attachments` field is now preserved in both the completed-snapshot and streaming-merge paths.

## Non-goals

- **`read` tool rendering unchanged** — `MessageFilesDisplay` in `MessageBody.tsx` is separate; we're only touching `ToolPart.tsx`'s handling of `state.attachments`
- **Optimistic attachments not preserved** — the materializer only merges existing attachments when the server snapshot lacks them; unsaved optimistic state is intentionally excluded

## Affected surfaces

- `packages/ui/src/sync/materialization.ts` — `mergeMaterializedPart()` function; affects all runtimes that use session materialization (web, desktop, VS Code, mobile).
- `packages/ui/src/components/chat/message/parts/ToolPart.tsx` — `ToolExpandedContent` rendering; affects all runtimes that render chat messages.

## Repository guidance

|---|---|---|
| `sync-state-invariants` skill | Materialization is a sync/reconciliation path — the skill's "Preserve rich fields when lightweight polling omits them" rule (Polling And Bootstrap) and "Define field precedence when global and local/live snapshots feed the same view" rule (Live And Historical State) directly apply. | `mergeMaterializedPart` now preserves `state.attachments` from the existing live part when the server snapshot lacks them, and does not override when the server snapshot has its own attachments. |
| `ui-api-decoupling` skill | `ToolPart` accesses `state.attachments` from the SDK `ToolStateCompleted` type. The skill governs how shared UI consumes SDK types. | The `FilePart` type is imported from the SDK; the rendering uses `onShowPopup` (the existing `RuntimeAPIs`-backed popup pattern) rather than reaching into runtime-specific APIs. |
| `theme-system` skill | The thumbnail rendering uses border, rounded, and ring styles that must match the app's theme. | The new JSX uses existing `var(--border)` and `var(--ring)` CSS variables and the same `border-border/40 bg-muted/10` class pattern as `MessageFilesDisplay` thumbnails. |
| Nearest README | `packages/ui/src/sync/DOCUMENTATION.md` for materializer changes; `packages/ui/src/components/chat/message/parts/ToolPart.tsx` is in the chat UI parts directory. | Read the sync documentation before editing the materializer. The `ToolPart` component follows existing local conventions (no README at that directory level). |

## Validation

| Check | Result |
|---|---|
| `bun test --filter "materialization"` | 18 pass, 0 fail (3 new tests for attachment preservation, 15 existing) |
| `bun test --filter "ToolPart"` | 2 pass, 0 fail |
| Manual UI verification | See [Visual evidence](#visual-evidence) below |

Not verified: automated rendering tests for `ToolPart` (the component uses `React.memo` with `areRenderRelevantPartsEqual`; a snapshot or rendering test would require a full DOM environment). The new attachment rendering branch is guarded by `attachments.length > 0 && state.status === 'completed'` and renders images as thumbnails and other files as chips, like `MessageFilesDisplay`.

## Visual evidence

**Visually confirmed using the WIP `image-attachments` branch of my plugin `opencode-image-search`:**
plugin tool image attachments appear as thumbnails in expanded output

## Risks and failure behavior

- **Data loss**: The entire fix is about preventing data loss. The only risk is a regression in the materializer merge logic, which should be covered by the new tests.
- **Performance**: The attachment filters use `React.useMemo` keyed on `attachments`; the image gallery memo is keyed on `imageAttachments`. No hot-path concern.
- **Cross-runtime**: `ToolExpandedContent` is rendered in all runtimes. The `onShowPopup` pattern is runtime-agnostic (web uses `ToolOutputDialog`, desktop/VS Code/mobile use their own popup implementations). The `RuntimeAPIContext` is already consumed by the component.
- **Backward compatibility**: Attachments that were previously silently dropped are now visible. No new dependencies.
- **Rollback**: Revert commits `77d1cc81` and `71835c9a`. No data migration or schema change.


FrostiDrinks:
I've pushed the following changes to resolve this review's findings:

**Finding #1 (correctness bug)** — Fixed in `materialization.ts:180`. The time preservation block now spreads from `mergedRecord.state` (which may already carry preserved attachments) instead of re-reading `next.state`. Added a regression test where `existing` has both `attachments` and `time.start` that `next` lacks.

**Finding #2 (hardcoded strings)** — Replaced all three English literals:
- `'Attachment'` → `t('chat.toolPart.attachmentFallback')` (new key, added to all 10 locale files with real translations)
- `'Image'` → `t('filesView.editor.imageAltFallback')` (existing key)
- `'file'` → `t('chat.fileAttachment.fileFallback')` (existing key)

**Finding #3 (thumbnail alignment)** — Thumbnails now match `MessageFilesDisplay` compact mode:
- Responsive sizing: `h-12 w-12 sm:h-14 sm:w-14 md:h-16 md:w-16`
- Focus ring: `focus-visible:ring-primary focus-visible:ring-offset-1`
- `aria-label` on buttons, `sr-only` span, `onError` hide handler
- `key` uses `file.url || filename || index` to avoid URL collision
- URL-less images render a fallback icon

**Validation**: 19/19 materialization tests pass, TypeScript clean, lint clean.

FrostiDrinks:
Followed up on the remaining nits:

- **Chip key**: `otherAttachments.map` now passes `index` and keys with ``file.url || `${fileName}-${index}` ``, matching `FileAttachment.tsx`.
- **Unreachable fallback icon**: `imageAttachments` filter no longer requires `typeof f.url === 'string'`, so URL-less images render the fallback icon instead of being dropped from both lists.
- **Gallery URL-less entries**: `imageGallery` uses `flatMap` with a URL guard (matching `MessageFilesDisplay`), so popup navigation never hits a broken slot. URL-less thumbnails still show the fallback icon in the grid.

FrostiDrinks:
Resolved the final index-misalignment nit: require `typeof f.url === 'string'` in `imageAttachments` (matching `MessageFilesDisplay`), so the thumbnail grid and popup gallery indices stay 1:1.

btriapitsyn (approved):

PR #2390, opened 2026-07-23
fix: make UI session cookie name configurable

### LINKED ISSUE #2377: [Bug] Session cookie collision when multiple instances share the same IP address
## Problem

When two OpenChamber instances run on the same IP address (different ports), they both set the `oc_ui_session` cookie. The second instance's cookie overwrites the first's, breaking session auth for all instances.

This occurs when accessing instances via LAN IP addresses (e.g., `http://192.168.0.1:3000` and `http://192.168.0.1:3001`). Loopback addresses (`127.0.0.1`) do not exhibit this issue.

## Steps to Reproduce

1. Enable authentication by setting `OPENCHAMBER_UI_PASSWORD`
2. Start instance A on `192.168.0.1:3000`
3. Start instance B on `192.168.0.1:3001`
4. Log into instance A in a browser tab
5. Log into instance B in another browser tab

## Symptoms

After both instances are logged in:

- **Refreshing ANY tab** brings back the login screen — not just the first tab. Both sessions are lost.
- **Creating or sending a session** fails with "Failed to start session" / "Failed to create session"
- **Settings page** (or at least the **Remote Instances** page) shows "UI authentication required"

The servers themselves are healthy — verified via curl with correct cookies returning 200 OK. The issue is purely browser-side cookie handling.

## Root Cause

`buildCookie()` in `server/lib/ui-auth/ui-auth.js` does not set a `Domain` attribute, so per RFC 6265 the cookie should be host-only (scoped to exact host:port). However, browsers appear to not scope cookies by port when using LAN IP addresses — DevTools shows only 1 `oc_ui_session` cookie for the entire IP domain, not one per port. Each instance has a unique JWT signing key (`~/.config/openchamber/jwt-secret`), so the overwritten cookie is rejected by the other instance's auth middleware.

Loopback addresses (`127.0.0.1`) appear to be scoped by port in browsers, which is why multiple instances work when accessed via localhost.

## Proposed Solution

Add an env var (e.g., `OPENCHAMBER_SESSION_COOKIE_NAME`) that overrides the default cookie name. This lets each instance use a unique cookie name when running on the same IP.

The change is minimal — `SESSION_COOKIE_NAME` already feeds into `cookieName` via `createUiAuth()`'s destructuring default:

```js
const SESSION_COOKIE_NAME = process.env.OPENCHAMBER_SESSION_COOKIE_NAME || 'oc_ui_session';
```

## Temporary workaround

Change the cookie name in one instance by editing ui-auth.js directly. This works but is fragile across version updates — the hardcoded 'oc_ui_session' string needs to be replaced with a unique value per instance. A native env var would eliminate this.

### Where does it happen?

Desktop Web

### Version (if known)

1.16.2

_No response_

### Logs (optional)

```shell

```

xixiGoodLuck:
## Intent

Browser cookies are scoped by host and path, not port, so two password-protected OpenChamber instances on the same host overwrite the shared `oc_ui_session` cookie. This change lets operators set `OPENCHAMBER_SESSION_COOKIE_NAME` per instance while preserving `oc_ui_session` as the default everywhere that issues or reads the UI session cookie.

## Non-goals

- Automatically deriving the name from the listening port. Reverse proxies and externally advertised ports can differ from the local socket, and changing the default would invalidate existing sessions for every deployment.
- Changing cookie path/domain attributes, JWT contents, session lifetime, authentication UI, or bearer-token behavior.
- Persisting session cookies, bearer tokens, or any new secret.

## Affected surfaces

- `@openchamber/web` UI-auth runtime resolves its default cookie name through a shared, side-effect-free helper.
- Request-security cookie parsing uses the same configured name, including `GET /api/push/visibility` session lookup.
- CLI-managed instance files add a backward-compatible optional `uiSessionCookieName` field containing the resolved per-instance name.
- Foreground and daemon starts persist that name; daemon children receive the same value, and restart/update flows restore it from the instance record.
- Later CLI processes read `uiPassword` and `uiSessionCookieName` for the target port, so command-scoped startup environments do not need to be repeated and one CLI can control instances with different names.
- Legacy or malformed instance records without a usable cookie name fall back to `oc_ui_session`.
- The merged CLI path retains upstream Desktop bearer authentication, multi-`Set-Cookie` extraction, explicit-password priority, and regex escaping.
- `OPENCHAMBER_SESSION_COOKIE_NAME` remains an optional public deployment contract documented in the Web README and UI-auth/CLI module documentation.
- Existing Web, Desktop, CLI, and service deployments remain behaviorally unchanged when the variable is unset.

## Repository guidance

|---|---|---|
| Root `AGENTS.md` and `openchamber-change-discipline` | This changes executable authentication behavior, a public environment contract, and CLI instance metadata. | Centralized the compatibility default, audited all readers and both instance writers, preserved lifecycle state across restart/update, documented the optional registry field, and covered the affected paths behaviorally. |
| `ui-api-decoupling` and its implementation map | UI authentication is a server/runtime boundary. | Cookie selection stays in shared server/CLI utilities with dependency injection for request-security tests; no UI state or backend-specific branching was added. |
| `packages/web/README.md`, `bin/lib/DOCUMENTATION.md`, and `server/lib/ui-auth/DOCUMENTATION.md` | These own the deployment, CLI lifecycle, and UI-auth contracts. | Documented the opt-in variable, unchanged default, per-port persistence, later-process reuse, shared resolver, and legacy fallback. |

## Validation

| Check | Result |
|---|---|
| `bun run --cwd packages/web test -- bin/cli.test.js` | 80 passed, 1 skipped; 5 existing Windows path/process tests failed, plus 1 unhandled error because the Unix `sleep` executable is unavailable on Windows. |
| `bun test packages/web/bin/cli.test.js --test-name-pattern 'CLI HTTP helpers'` | 10 passed, 0 failed; covers control requests, default/custom/regex names, env-unset persisted names, multi-port isolation, explicit/stored password priority, multi-`Set-Cookie`, and Desktop bearer auth. |
| `bun test packages/web/bin/cli.test.js --test-name-pattern 'instance option persistence|restart preserves'` | 2 passed, 0 failed; covers foreground/daemon records and restart preservation. |
| `bun test packages/web/bin/lib/commands-update.test.js` | 1 passed, 0 failed. |
| `bun test packages/web/server/lib/ui-auth/ui-auth.test.js` | 7 passed, 0 failed. |
| `bun test packages/web/server/lib/security/request-security.test.js` | 3 passed, 0 failed. |
| `bun test packages/web/server/sse-routes.test.js` | 3 passed, 0 failed. |
| `bun run type-check:web` | Passed. |
| `bun run lint:web` | Passed. |
| `bun run build:web` | Passed; repository build warnings only. |
| `bun run lint` | Passed across all workspaces. |
| `bun run build` | Passed across all workspaces; repository build warnings only. |
| Upstream/merge audit | Merged `upstream/main` at `26dd2bfe`; GitHub reports the PR mergeable with 15 intended changed files. |

## Visual evidence

No rendered UI changes. The observable behavior is limited to the server's cookie name and matching readers plus CLI instance metadata/retry behavior; focused runtime, route, and lifecycle tests verify those contracts.

## Risks and failure behavior

- Changing the cookie name intentionally requires users of that instance to sign in again; existing JWT secrets and session data are not modified.
- The instance registry gains only the optional non-secret `uiSessionCookieName` string. Existing records require no migration and fall back to `oc_ui_session`.
- Reusing the same configured name across instances retains the original collision behavior; unique names remain the operator's responsibility.
- Cookie security attributes (`HttpOnly`, `SameSite`, `Secure`, path, and expiry), JWT behavior, Desktop bearer authentication, and session lifetime are unchanged.
- To return a daemon instance to the default name, stop it and start it again without `OPENCHAMBER_SESSION_COOKIE_NAME`; a lifecycle restart intentionally preserves the stored per-instance name.

## Related issue

Closes #2377


xixiGoodLuck:
Addressed the cookie-name consistency findings in `9da75f46`:

- Centralized `OPENCHAMBER_SESSION_COOKIE_NAME` resolution with the unchanged `oc_ui_session` fallback.
- Updated request-security parsing, so `GET /api/push/visibility` recognizes the configured session cookie.
- Updated CLI password-auth retry to extract/replay the configured cookie name safely, including names with regex characters.
- Added focused request-security, real visibility-route, and CLI retry coverage; updated the owning module docs.

Validation: UI auth 7/7, request security 3/3, SSE routes 3/3, CLI HTTP helpers 3/3; `type-check:web`, Web/full lint, and Web/full builds passed. Knip reports no unused-file/export finding for the new helper. The full CLI file is 47 passed / 1 skipped / 5 unrelated failures in untouched Windows-only entry/lifecycle tests.

xixiGoodLuck:
Follow-up on the review's PR-body non-blocker: the description was refreshed while the automated review was running. The current body now lists the shared resolver, request-security/visibility route, CLI retry path, updated guidance, and all focused validation results. No additional code change is needed for that snapshot-only finding.

btriapitsyn (changes requested):
Thanks for following up on the earlier review. The request-security change now fixes the configured-cookie path for `GET /api/push/visibility`, and the regex escaping in the CLI extractor looks correct. I found one remaining lifecycle issue in the CLI integration that should be addressed before merge.

### CLI authentication does not know the cookie name of an already-running instance

Relevant locations:
- `packages/web/bin/lib/cli-http.js:99`
- `packages/web/bin/lib/cli-process.js:54-65`
- `packages/web/bin/lib/commands-serve.js:207-213,351-357`
- `packages/web/README.md:28`

The documented usage is a command-scoped environment assignment:

```sh
OPENCHAMBER_SESSION_COOKIE_NAME=oc_ui_session_3000 openchamber --ui-password secret
```

That variable is inherited by the server process, so the server correctly issues `oc_ui_session_3000`. However, a later command such as:

```sh
openchamber tunnel start --port 3000
```

runs in a new CLI process where `OPENCHAMBER_SESSION_COOKIE_NAME` is normally unset. At `cli-http.js:99`, `createUiSessionCookie()` therefore resolves the default `oc_ui_session` name. The `/auth/session` response contains `oc_ui_session_3000`, extraction returns `null`, and the original authenticated operation remains a 401. This affects exactly the normal command sequence shown in the README. It also prevents one CLI environment from controlling multiple instances that use different cookie names.

The instance registry already persists the password needed for this retry, but `writeInstanceOptions()` does not persist the resolved cookie name. The new CLI test does not expose the mismatch because it sets `OPENCHAMBER_SESSION_COOKIE_NAME` in the same process that runs `requestJson()`, effectively giving the CLI knowledge that a later invocation would not have.

A small complete fix would be to persist the resolved cookie name in the instance file for both foreground and daemon starts, then have the retry path read it alongside `uiPassword`. Existing instance files can safely fall back to `oc_ui_session`. Please add a regression test where the instance file contains a custom cookie name while `process.env.OPENCHAMBER_SESSION_COOKIE_NAME` is unset; the retry should still extract and replay the custom cookie.

Everything else I traced for the configured-name flow looks consistent. Thanks again for the thorough follow-up and focused route coverage.

xixiGoodLuck:
Addressed the maintainer's CLI lifecycle finding in `ad626dd5`:

- foreground and daemon instance files now persist the resolved `uiSessionCookieName`, and daemon children receive the same name;
- later CLI processes read the target port's stored cookie name alongside `uiPassword`, including independent names for multiple ports;
- legacy instance files without the field safely fall back to `oc_ui_session`;
- restart and update flows preserve the stored name;
- added the requested regression with `OPENCHAMBER_SESSION_COOKIE_NAME` unset, a custom name in the instance file, no cookie on the first request, and the stored custom cookie on the retry;
- merged current `upstream/main` (`26dd2bfe`) while preserving Desktop bearer auth, multi-`Set-Cookie` handling, explicit-password priority, and regex escaping.

Validation:

- CLI HTTP helpers: 10 passed;
- foreground/daemon persistence plus restart preservation: 2 passed;
- update command: 1 passed;
- UI auth: 7 passed;
- request security: 3 passed;
- visibility/SSE routes: 3 passed;
- `type-check:web`, Web/full lint, and Web/full builds passed;
- `git diff --check` passed.

The full merged-state CLI Vitest run reported 80 passed and 1 skipped, with 5 existing Windows path/process failures plus 1 unhandled error because the Unix `sleep` executable is unavailable on Windows.

No cookie security attributes, JWT behavior, session lifetime, authentication UI, or bearer-token behavior were changed. The only persisted-format change is the backward-compatible optional `uiSessionCookieName` instance field.

btriapitsyn:

PR #2403, opened 2026-07-24
feat(quota): support OpenAI business-account spend_control in codex quota display

jrandiny:
Business/enterprise OpenAI accounts return a usage payload with `spend_control.individual_limit` instead of rate-limit windows. Keep the existing primary/secondary/credits blocks unchanged and add a new `credits` window that surfaces `used_percent` and a "used / limit left" value label.

## Intent

Currently enterprise OpenAI spend limits are not reflected in openchamber.

## Non-goals

Modification of existing primary/secondary/credits quota code

## Affected surfaces

- `packages/web/server/lib/quota/providers/codex.js` — Codex quota provider. Business accounts previously surfaced no windows (top-level `rate_limit` is `null`), so the usage section rendered empty. They now get a `credits` window from `spend_control.individual_limit`.
- `packages/ui` — no code change; `UsagePage` already renders arbitrary `usage.windows` entries generically via `UsageCard`, so the new `credits` key is consumed without modification. Legacy ChatGPT-plan accounts that already get `credits_balance` (`$balance` / `Unlimited`) are unaffected; the new key uses a different window name.
- VS Code runtime: `quotaProviders.ts` dispatches the codex provider id and forwards its server result; no change needed since the result shape is unchanged.

## Repository guidance

|---|---|---|
| AGENTS.md "Correctness Invariants" — partial results/cleanup must be explicit | The provider must degrade safely for both account shapes | The new block is additive; when `spend_control.individual_limit` is absent (legacy accounts) nothing changes. |
| `packages/web/server/lib/quota/DOCUMENTATION.md` "Response contract" | All providers return via `buildResult`/`toUsageWindow` keyed by window name | New window uses `toUsageWindow` and a stable key (`credits`); required fields preserved. |
| AGENTS.md "Validation — run focused tests for the touched surface" | Quota providers have vitest coverage | Added a test for the business payload; kept existing legacy tests. |

## Validation

| Check | Result |
|---|---|
| `bun run test -- server/lib/quota/providers/codex.test.js` | 3/3 passed (legacy weekly, legacy 5h+weekly, business spend_control) |
| `bun run test -- server/lib/quota` (full module) | 22/22 passed across 8 files |
| `bun run lint` (packages/web) | clean |
| Runtime against live business-account payload | not verified — user will confirm via UI screenshot |

## Risks and failure behavior

- **Key collision:** the new `credits` key differs from the existing `credits_balance` key, so both can coexist when a payload contains both `credits` and `spend_control` (unlikely but safe — they render as separate cards).
- **Missing field:** accounts without `spend_control.individual_limit` skip the block; no crash, no empty window.
- **Backward compatibility:** top-level `rate_limit` primary/secondary handling is byte-for-byte unchanged; `formatMoney` import retained for the legacy `credits_balance` label.
- **Rollback:** revert this single commit; the provider reverts to legacy-only behavior.

---

### DIFF
--- a/packages/web/server/lib/quota/providers/codex.js
+++ b/packages/web/server/lib/quota/providers/codex.js
@@ -97,6 +97,24 @@ export const fetchQuota = async () => {
       });
     }
 
+    // Business/enterprise accounts expose a dollar spend cap under
+    // `spend_control.individual_limit`. Surface it as an additive `credits`
+    // window so existing consumers keep working.
+    if (payload?.spend_control?.individual_limit) {
+      const spendLimit = payload.spend_control.individual_limit;
+      const used = toNumber(spendLimit.used);
+      const limit = toNumber(spendLimit.limit);
+      const valueLabel = used !== null && limit !== null
+        ? `${used.toFixed(0)} / ${limit.toFixed(0)} left`
+        : null;
+      windows.credits = toUsageWindow({
+        usedPercent: toNumber(spendLimit.used_percent),
+        windowSeconds: null,
+        resetAt: null,
+        valueLabel
+      });
+    }
+
     return buildResult({
       providerId,
       providerName,
--- a/packages/web/server/lib/quota/providers/codex.js
+++ b/packages/web/server/lib/quota/providers/codex.js
@@ -105,7 +105,7 @@ export const fetchQuota = async () => {
       const used = toNumber(spendLimit.used);
       const limit = toNumber(spendLimit.limit);
       const valueLabel = used !== null && limit !== null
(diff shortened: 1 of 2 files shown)


jrandiny:
Related to "non-blocker: spend amounts rendered without currency symbol or decimal precision"

OpenAI uses credit system, 1 credit ~ 4 cents and on the website they also round the number to nearest integer



Related to "non-blocker: business accounts still render an empty "Credits Balance" card (pre-existing)"

I dont know all the possible response structure of the API. So to reduce the chance of causing bug, I dont want to change the existing logic

PR #2407, opened 2026-07-24
fix: cascade mobile session archiving

alexandrereyes:
## Intent

Archiving a session from the shared mobile surfaces did not remove it from the mobile session lists until a full page reload.

Two defects are fixed here:

1. **Archive cascade (commits 1–2).** Archiving a parent session from `MobileSessionsSheet` archived only the selected session because OpenCode does not cascade `time.archived` updates. Confirmation now collects every known active descendant/subagent at any depth and archives the complete known subtree through the existing `archiveSessions` batch action. Collection reads the current global active/archived snapshot plus the current live directory-session index instead of waiting for global pagination or limiting traversal to the visible project/worktree bucket. The existing background refresh still improves later renders but never blocks confirmation. Collection deduplicates IDs, terminates safely on corrupted cycles, omits already archived sessions, and can traverse through an archived intermediate record to reach an active descendant.

2. **Client-side reconciliation (commit 3).** With the cascade working server-side, the archived session still remained visible in the mobile panel until a page reload. `archiveSession()` already removes the session from the live child stores and moves it to `useGlobalSessionsStore.archivedSessions`, but both mobile list consumers rebuilt their list as "global active, then re-attach any live session missing from that list", consulting only `activeSessions`. Any live copy that had not caught up (SSE echo ordering, a directory re-bootstrap, or the 15s child-session discovery poll re-inserting a `session.list` record) was therefore re-attached and visually resurrected the archived session. A full reload rebuilt the child stores, which is exactly why refreshing "fixed" it — the backend and the global cache were already correct.

The affected memos are `useAllProjectSessions()` in `packages/ui/src/components/chat/MobileSessionStatusBar.tsx` (the panel in the user report) and the `sessions` memo in `packages/ui/src/apps/MobileSessionsSheet.tsx`. Both now use one shared pure helper, `combineActiveSessionsWithLive()`, added next to the existing `mergeLiveSessionWithGlobalSession()` in the owning module `packages/ui/src/stores/useGlobalSessionsStore.ts`.

Contract implemented by the helper:

- `globalArchivedSessions` is negative authority: an archived ID is never reintroduced by a live copy that has not caught up.
- Live records still enrich a global active session through the existing `mergeLiveSessionWithGlobalSession` precedence (live recency/title/directory; global `share` and cached directory metadata preserved).
- Live-only sessions that are not known to be archived stay visible, so unbootstrapped/cross-project coverage is unchanged.
- If a race leaves the same ID in both global lists, archived wins.
- Global active order is preserved and live-only sessions are appended after it, exactly as before; the only removals are archived IDs, plus a repeated live-only ID that previously could be appended twice.

Secondary effect: `MobileSessionStatusBar` passes this list as the `fallbackActive` argument of `refreshGlobalSessions()`. That fallback is used only when the active-sessions fetch fails, and the mutation-revision overlay already re-applied the newer archive, but the fallback list no longer carries archived IDs at all. This is defense in depth, not the primary fix.

## Non-goals

- Add backend archive cascading or recursive fetch endpoints.
- Fetch unknown descendants ad hoc from the client or wait for global pagination before archiving.
- Remove the live re-entry paths themselves (SSE ordering, re-bootstrap, `discoverChildSessions`). The consumers are made authoritative instead; narrowing those producers is separate work.
- Change desktop sidebar, VS Code archive-all, worktree deletion, hard-delete, or archive action internals.
- Change layout, styling, archive confirmation interaction, persistence, or external API contracts.

## Affected surfaces

- `packages/ui`: `MobileSessionsSheet` archive behavior and session list, `MobileSessionStatusBar` session list, one focused subtree helper, one shared pure list helper in `useGlobalSessionsStore.ts`, and the sync ownership documentation.
- Phone sessions sheet, iPad persistent sidebar (`variant="sidebar"`), the mobile composer session panel, hosted/PWA mobile, and Capacitor mobile are affected because they share these components.
- Desktop sidebar and VS Code are not directly changed. `SessionSidebar` already reads global active and archived lists explicitly and has its own cascade implementation; `mergeLiveSessionWithGlobalSession` had no other consumers.
- No server/backend, Electron privilege boundary, persisted data, transport, authentication, or public package contract changes. `combineActiveSessionsWithLive` is a new intra-package export; it is not part of any published package entrypoint.
- User-visible behavior: after the second archive tap, the parent and its known active subagents are archived and disappear from both mobile lists immediately, without a page reload. Existing batch toasts report successful and failed counts.

## Repository guidance

|---|---|---|
| Root `AGENTS.md` and `openchamber-change-discipline` | Executable shared UI source and mobile runtime behavior change. | Smallest correct change: two consumers, one shared pure helper in the module that already owns global/live merge precedence, no new dependency, no drive-by refactor. Existing behavior covered by tests and callers (cascade collection, live-only coverage, ordering) is preserved, and the sync documentation was updated because a documented precedence invariant changed. |
| `sync-state-invariants` and `packages/ui/src/sync/DOCUMENTATION.md` | The bug is exactly "define field precedence when global and local/live snapshots feed the same view", plus stale-live-versus-authoritative-mutation reconciliation. | Global archived membership is now explicit negative authority for mobile active lists; a stale live snapshot can no longer undo an authoritative archive. Live remains the source for live metadata, and the global cache still never drives busy/retry status. The invariant is now written down in the sync documentation instead of being implicit. |
| `performance-engineering` | The memo is on a mobile list render path that recomputes on every live session publication. | The helper stays a single linear pass with one `Set` and one `Map`; it replaces, not adds, the previous allocation shape (one extra `Set` over the archived list). The new `archivedSessions` subscription returns the store's reference-stable array, so unchanged archived state produces no extra render. No new scan is introduced on a streaming path. |
| `ui-api-decoupling` and `references/implementation-map.md` | The sheet invokes the shared SDK-facing session action. | No direct fetch, runtime URL, credential, bridge, route, or backend behavior was introduced; the component delegates to `useSessionUIStore.archiveSessions`, and the new helper is pure and transport-free. |
| TypeScript guidance | Four TypeScript/TSX modules changed. | Existing root aliases and SDK `Session` types are used; the helper takes a named-parameter object so three same-typed session arrays cannot be swapped at a call site. No `any`, no guessed payload shape. |
| Root `README.md`, `CONTRIBUTING.md`, `.github/PULL_REQUEST_TEMPLATE.md`, `packages/ui/src/stores/DOCUMENTATION.md` | They define affected shared/mobile surfaces, the global-cache ownership rules, and PR evidence requirements. | The handoff names every affected runtime, reports exact checks and gaps, and does not claim mobile runtime verification. `stores/DOCUMENTATION.md` needed no change: global-store ownership and mutation rules are unchanged; the new statement is about cross-scope precedence and belongs to `sync/DOCUMENTATION.md`. |

## Validation

The table below was produced at commit `33c0ba15`, run with `npx --yes bun@1.3.14` because `bun` was not on that machine's PATH. It is kept for the per-check detail it records. Every executable check in that table was subsequently re-run at merged HEAD `f09a7639`; results are in the historical revalidation subsection below. The current HEAD is **`9daab5d8a`**; validation for the focused review follow-ups appears after that subsection.

| Check | Result |
|---|---|
| `npx --yes bun@1.3.14 test "packages/ui/src/stores/useGlobalSessionsStore.test.ts" "packages/ui/src/apps/mobileSessionArchive.test.ts"` | Passed: 25 tests, 0 failures, 39 expectations. Includes the 6 new list-reconciliation tests (global active + live update merged once; live-only kept; globally archived dropped despite a stale live copy; archived wins when an ID is in both global lists and live; repeated live-only ID not duplicated; global active order preserved with live-only appended) and the 6 pre-existing cascade tests. |
| `npx --yes bun@1.3.14 test "packages/ui/src/stores" "packages/ui/src/sync/scoped-blocking-requests.test.ts"` | 168 pass / 25 fail. The 25 failures are pre-existing: running the same directory on the parent commit without this patch gives the same 25 failures (159 pass / 25 fail), and `useConfigStore.test.ts` passes 18/18 when run alone. They are cross-file mock interference in `packages/ui/src/stores`, unrelated to this change. |
| `npx --yes bun@1.3.14 run --cwd "packages/ui" type-check` | Passed. |
| `npx --yes bun@1.3.14 run --cwd "packages/ui" lint` | Passed, no warnings. |
| `npx --yes bun@1.3.14 run dead-code` | Completed with the repository's existing non-blocking findings: 32 unused files, 259 unused exports, 180 unused exported types, 2 duplicate exports, 4 configuration hints — identical to the previous run. Neither `combineActiveSessionsWithLive` nor `mergeLiveSessionWithGlobalSession` is reported as unused after the import-shape change. |
| `npx --yes bun@1.3.14 run build` (full workspace build, all packages plus mobile asset staging) | Passed, exit code 0. |
| `git diff upstream/main...HEAD --check` | Passed, no whitespace errors. |
| Manual mobile archive interaction | **Run at HEAD `f09a7639`** against a real OpenCode instance on the mobile surface. Archiving only the parent of a depth-2 tree wrote `time.archived` on all **4** nodes (parent, both children, grandchild), the toast reported **"Archived 4 sessions"**, and all 4 rows left the mobile list within 1.5 s and stayed gone. The same interaction on `upstream/main` (`47e5ac84`) archived **1** node and the parent row **reappeared** in the list 20 s later. Full numbers, contrast tables, and screenshots below. |

### Revalidation at HEAD `f09a7639`

`f09a7639` is a merge commit with `upstream/main` (`47e5ac84`), so the full workspace build and the dead-code report were re-run on the merged tree rather than inherited from `33c0ba15`. Run in a dedicated worktree of this branch, with `bun 1.3.8`.

| Check | Result |
|---|---|
| `bun test packages/ui/src/stores/useGlobalSessionsStore.test.ts packages/ui/src/apps/mobileSessionArchive.test.ts packages/ui/src/stores/globalSessions.test.ts` | Passed: **34 pass, 0 fail**, 56 expect() calls across 3 files, 809 ms. |
| `bun run --cwd packages/ui type-check` | Passed, no output. |
| `bun run --cwd packages/ui lint` | Passed, no warnings. |
| `bun run build` (full workspace: `ui`, `web`, `electron`, `vscode`, plus mobile asset staging) | **Passed, exit code 0.** Every package reported `Exited with code 0`. The only warnings are the repository's pre-existing ones: Rollup's ">500 kB chunk" advisory and the KaTeX font URLs that stay unresolved at build time. The merge with `upstream/main` did not break the build. |
| `bun test packages/ui/src/stores` (wide suite) | **171 pass / 25 fail** across 28 files, 341 expect() calls. The 25 failures are pre-existing and unrelated: the same command on `upstream/main` (`47e5ac84`), without this patch, gives **159 pass / 25 fail** with the identical failure profile — 18 in `useConfigStore provider persistence`, 6 in `global session mutation reconciliation`, 1 in `useProjectsStore settings synchronization`. The 6 session-store failures run **6 pass / 0 fail** when `useGlobalSessionsStore-races.test.ts` is executed alone, confirming cross-file mock interference rather than a regression from this branch. |
| `bun run dead-code` (non-blocking) | Completed. 33 unused files, 268 unused exports, 180 unused exported types, 2 duplicate exports, 4 configuration hints. Against the `33c0ba15` run (32 / 259 / 180 / 2 / 4) this is +1 file and +9 exports, all introduced by the merged `upstream/main` code: no file touched by this pull request appears in the report, and `isArchivedSession`, `combineActiveSessionsWithLive`, and `mergeLiveSessionWithGlobalSession` are not reported as unused. |

### End-to-end mobile archive run

Environment: Chromium against a locally served build, mobile surface confirmed (`window.__OPENCHAMBER_SURFACE__ === 'mobile'`, no `<aside>` present), viewport 390x844. Same dataset and same browser for both builds.

Session tree, depth 2: `parent` → `child A` + `child B`; `child A` → `grandchild A1`. Each `parentID` was confirmed through the API. In both runs only the **parent** was archived, through the mobile UI.

**Server state — `time.archived` after archiving only the parent**

| Node | Baseline `upstream/main` (`47e5ac84`) | This branch (`f09a7639`) |
|---|---|---|
| parent | `1784984656483` | `1784984801385` |
| child B | **NULL** | `1784984801534` |
| child A | **NULL** | `1784984801612` |
| grandchild A1 | **NULL** | `1784984801695` |
| Nodes archived | **1 of 4** | **4 of 4** (310 ms between first and last) |
| Toast | — | **"Archived 4 sessions"** |

**Mobile list rows at three instants**

| Instant | Baseline `upstream/main` (`47e5ac84`) | This branch (`f09a7639`) |
|---|---|---|
| Before archiving | 34 rows, all 4 tree nodes present | 36 rows, all 4 tree nodes present |
| Shortly after archiving | +1.9 s: 33 rows, parent gone | +1.5 s: 32 rows, **none** of the 4 present |
| After 20 s | 34 rows — the parent **came back**, while already archived server-side | 32 rows, still none of the 4 present |

On this branch the group counter also dropped from 5 to 4, consistent with the removed subtree.

**Anti-resurrection pressure**, applied on this branch after the archive; none brought the rows back:

1. the 15 s child-session discovery poll cycle (measured at +20 s);
2. a directory re-bootstrap by switching projects and back, without reloading;
3. a full page reload.

Pressure 1 is the one that resurrected the parent row on the baseline, so the reconciliation change in commit 3 is what the third column of the second table is measuring.

### Review follow-ups at HEAD `9daab5d8a`

The confirm-time race identified by openchamber-bot is fixed: if the rendered session disappears from every current store before the second archive tap is processed, subtree collection returns no IDs and the handler now shows the existing localized `sessions.sidebar.session.archive.error` toast without calling `archiveSessions`. The follow-up review then noted that the first regression test coupled itself to the component source text. HEAD `9daab5d8a` removes that lexical assertion while retaining the observable pure-helper behavior: a root that disappeared before confirmation yields an empty ID list. Rendering the full sheet solely to inspect this one-line guard would require unrelated runtime, store, drag-and-drop, and DOM setup, so no component harness was added.

| Check | Result |
|---|---|
| `npx --yes bun@1.3.14 test packages/ui/src/stores/useGlobalSessionsStore.test.ts packages/ui/src/apps/mobileSessionArchive.test.ts packages/ui/src/stores/globalSessions.test.ts` | Passed at `9daab5d8a`: **35 pass, 0 fail**, 57 expect() calls across 3 files. The regression observably covers a root disappearing before confirmation without depending on component source layout. |
| `npx --yes bun@1.3.14 run --cwd packages/ui type-check` | Passed. |
| `npx --yes bun@1.3.14 run --cwd packages/ui lint` | Passed, no warnings. |
| `git diff HEAD^ --check` | Passed, no whitespace errors. |

The existing successful-archive screenshots remain representative because these follow-ups change only the zero-ID error branch and its test, without altering normal archive behavior, layout, styling, or the successful state shown there. The disappearing-target race was not forced manually; its empty-selection condition is covered through observable helper output, while the direct localized toast guard was verified by code review rather than a source-text or full-component test.

## Visual evidence

Screenshots from the end-to-end run above, mobile surface at 390x844. They are hosted on a separate evidence branch (`pr-2407-evidence` in the fork) so that no image enters this pull request's diff.

**Baseline `upstream/main` (`47e5ac84`) — before archiving.** The four `CASCATA *` rows of the tree are present in the list.

Baseline mobile list before archiving the parent

**Baseline `upstream/main` (`47e5ac84`) — after archiving the parent.** Only the parent row left. `CASCATA filho A`, `CASCATA filho B` and the grandchild are still listed and still active server-side (`time.archived` is NULL for all three): the baseline does not cascade. Twenty seconds later the parent row itself reappeared.

Baseline mobile list after archiving the parent, children still present

**This branch (`f09a7639`) — before archiving.** The `CASCATA2 *` tree (parent, two children, one grandchild) is present in the list.

Branch mobile list before archiving the parent

**This branch (`f09a7639`) — after archiving the parent.** No `CASCATA2 *` row remains: the whole known subtree was archived and removed from the list without a reload, and it stayed removed through the 15 s discovery poll, a directory re-bootstrap, and a full reload.

*Reading note:* this screenshot still shows rows named `CASCATA filho A` and `CASCATA filho B`. Those are **not** a failure of the cascade — they are the orphaned subagents left active by the baseline phase of the same run, which were never archived precisely because the baseline does not cascade. They belong to the `CASCATA *` tree, not the `CASCATA2 *` tree being tested here.

Branch mobile list after archiving the parent, no CASCATA2 rows remain

### What is still unproven

The run above covers one shape only. These remain untested and are not claimed:

- Cascading from an **intermediate** node (archiving a child that itself has descendants) through the mobile gesture.
- **Partial batch failure**: the `failedArchive*` toast path was never exercised with a real failing archive call.
- **Isolation between the SSE echo and the discovery poll**: the run applied both pressures, but did not isolate which producer would have resurrected the rows on this branch, only that neither did.
- **Runtimes**: packaged Electron, VS Code webview, and a physical device were not exercised. The run was Chromium at a mobile viewport against a locally served build.

## Risks and failure behavior

- `archiveSessions` processes IDs independently and continues after failures. Successful IDs remain archived with no rollback; failed IDs remain active. Existing toasts report both counts, so partial completion is visible and retryable.
- A failed archive leaves the session in the global active list, so the list correctly keeps showing it; nothing is hidden optimistically before server confirmation.
- Suppression is keyed to the global archived list. If the global cache itself were wrong (for example an incorrectly archived record), the session would be hidden from these mobile lists until the cache is corrected. That is the intended authority direction and matches how the desktop sidebar already separates active and archived buckets.
- A live-only session in a directory the global cache has never covered is still shown, so the fix cannot hide sessions the client only knows about live.
- Cascading remains limited to descendants known by the client at collection time. Descendants unknown to both global and live stores, or created after collection, stay outside the batch; guaranteeing those requires backend cascade support and is not claimed here.
- Performance: the helper is one linear pass over global active plus live, with one `Set` over the archived list and one `Map` over live sessions. The added `archivedSessions` selector returns a reference-stable array, so it does not add renders when archived state is unchanged.
- The underlying live re-entry paths still exist upstream of these consumers; this change makes the consumers authoritative rather than removing those producers. Other surfaces that combine global active with live sessions should adopt `combineActiveSessionsWithLive()` if they appear; today the two mobile surfaces were the only ones with this pattern.
- No persisted state, external API, authentication, permission, or security boundary changes. Reverting the three PR commits restores the prior single-session mobile archive behavior and the prior list merge.

---

## Follow-up correction — commit 4 (`fix: treat only timestamped sessions as archived authority`)

Commit 3 of this pull request introduced a regression. This section documents it and the fix; the sections above describe the original three commits and remain accurate for them, except that the HEAD referenced in the first **Validation** table is `33c0ba15`, which predates this commit. The branch later reached merged HEAD `f09a7639` on top of `a4a1704f`; the current review-follow-up HEAD is `9daab5d8a`.

### The wrong premise

Commit 3 made `globalArchivedSessions` **negative authority** in `combineActiveSessionsWithLive()`. That is only sound if the archived cache contains archived sessions and nothing else. It does not.

The archived cache is filled from OpenCode's session list with `archived: true`. That flag means **"also include archived sessions"**, not "return only archived sessions": the server simply drops its `time_archived IS NULL` condition.

```ts
// opencode/packages/opencode/src/session/session.ts
if (!input?.archived) conditions.push(isNull(SessionTable.time_archived))
```

So `archived: true` returns active **and** archived records.

### Observed impact

Measured against a real instance:

| Request | Records returned | Records with `time.archived` |
|---|---|---|
| `archived: false` | 286 | 0 |
| `archived: true` | 748 | 462 |

The archived cache therefore held all 286 active sessions as well. Because it is negative authority, `combineActiveSessionsWithLive()` filtered out **every** active session.

| State | Sessions visible in the mobile list |
|---|---|
| Before this commit (HEAD `33c0ba15`) | **0** of 286 |
| After this commit (HEAD `a4a1704f`) | **286** of 286 |

This is a complete loss of the mobile session list, strictly worse than the bug the pull request set out to fix.

### Where the fix was made, and why there

Primary fix at the data boundary, in `packages/ui/src/stores/globalSessions.ts`:

- A documented `isArchivedSession()` predicate records the real server semantics next to the call that relies on them.
- `listGlobalSessionPages()` now accumulates only records carrying `time.archived` when `options.archived === true`, and reports only those records to `onPage`.

The boundary is the correct place because the flag's meaning is a property of the OpenCode API, not of any single consumer. Fixing it in `combineActiveSessionsWithLive()` alone would leave every present and future reader of `archivedSessions` re-deriving the same correction, and would leave the store's archived cache itself factually wrong.

One subtlety the fix preserves explicitly: filtering must not feed the pagination guards. `appended`, the `payload.length < pageSize` last-page test, and the cursor continue to operate on the **raw** payload, so a page that is full upstream but entirely non-archived is not mistaken for the last page. A test pins this.

Defense in depth in `packages/ui/src/stores/useGlobalSessionsStore.ts`: `combineActiveSessionsWithLive()` now builds `archivedIds` only from entries that actually carry `time.archived`, so a record without that timestamp can never remove a session even if some other path populates the cache incorrectly.

`packages/ui/src/stores/DOCUMENTATION.md` records the contract: the archived cache must never contain active sessions, because it is negative authority for every active list.

The original intent of commit 3 is unchanged: a genuinely archived session is still removed even when a stale live copy would otherwise resurrect it. The pre-existing test asserting that behavior still passes.

### Why the existing tests did not catch it

The commit-3 fixtures built `globalArchivedSessions` exclusively from sessions that carried `time.archived`. The fixtures encoded the same wrong premise as the code, so the tests could not observe the failure. No test exercised `listGlobalSessionPages` with `archived: true` against a mixed response, which is exactly what the server returns.

Six tests were added to close that gap:

`packages/ui/src/stores/globalSessions.test.ts`

1. A mixed `archived: true` response yields only the archived record.
2. A first page that is full but entirely non-archived still triggers the next page request — filtering does not terminate pagination.
3. `onPage` receives only archived records.
4. `archived: false` filters nothing.

`packages/ui/src/stores/useGlobalSessionsStore.test.ts`

5. A session listed as archived **without** `time.archived` stays visible.
6. The same, for a live-only session.

### Validation

Run at HEAD `a4a1704f`, in a dedicated worktree of this branch, with `bun 1.3.8` from `packages/ui`.

| Check | Result |
|---|---|
| `bun test src/stores/globalSessions.test.ts src/stores/useGlobalSessionsStore.test.ts src/stores/useGlobalSessionsStore-races.test.ts` | Passed: **34 pass, 0 fail**, 60 expectations across 3 files. Includes the 6 new tests and the pre-existing archive-authority and race tests. |
| `bun run type-check` (`packages/ui`) | Passed, no output. |
| `bun run lint` (`packages/ui`) | Passed, no warnings. |
| `bun run build` / `bun run dead-code` | Not re-run at `a4a1704f` itself, but **both were re-run at merged HEAD `f09a7639`**, which contains this commit: the workspace build passes with exit code 0 and the dead-code report shows no entry from this pull request. See **Revalidation at HEAD `f09a7639`** above. |
| Manual mobile interaction | The 0-of-286 and 286-of-286 counts above come from the session list payloads and the helper's behavior over them. The mobile list has since been exercised interactively at HEAD `f09a7639` — see **End-to-end mobile archive run** above, where the list rendered 36 rows before archiving, which directly confirmed that this commit's fix held at that reviewed HEAD. Physical device and packaged runtimes remain untested. |

### Risk and rollback for this commit

- Narrowing the archived list is strictly a reduction of what enters the archived cache. Any consumer that expected `archived: true` to return active sessions would now see fewer records; there is no such consumer in the repository — the store's archived cache is the only caller that passes the flag.
- If the fix were wrong in the other direction, the failure mode is a genuinely archived session briefly reappearing, which is visible and self-correcting on the next refresh — far less severe than the empty list it replaces.
- Rollback: reverting commit `a4a1704f` alone restores the empty-list regression and is not advisable. Reverting commits `a4a1704f` and `33c0ba15` together restores the pre-commit-3 merge behavior while keeping the archive cascade from commits 1–2.


alexandrereyes:
Closing in favour of #2573.

Rechecking this branch against current `main`, the three parts aged differently:

- **Data-boundary filter (`globalSessions.ts`)** — still a real bug, and the strongest part. OpenCode's `archived` flag drops the `time_archived IS NULL` condition rather than selecting archived rows (`session.ts:564`), so `archivedSessions` is contaminated with active sessions and `ArchiveView.tsx:32-42` renders them. Extracted to **#2573** with expanded pagination tests.
- **Mobile subtree cascade** — the branch was written against the pre-#2561 `MobileSessionsSheet`. That file was rewritten and archive now fires immediately on swipe with no confirmation step, so cascading several sessions without confirmation is a product decision, not a rebase. Not carried over.
- **`combineActiveSessionsWithLive`** — #2561 added `merged.filter((session) => !session.time?.archived)` at `MobileSessionsSheet.tsx:1008`, which covers most of this. The remaining gap is narrow (a stale live copy that has not received `time.archived` yet), and `MobileSessionStatusBar.tsx` — the second consumer that justified extracting a shared helper — no longer exists. Not worth a standalone export.

Rebasing this branch as-is would conflict on `MobileSessionsSheet.tsx` (5 hunks) and hit a modify/delete on `MobileSessionStatusBar.tsx`. #2573 applies cleanly instead.

alexandrereyes:
The cascade half is now carried over in #2580, reimplemented against the post-#2561 `MobileSessionsSheet` rather than rebased from this branch.

Differences from what was here: targets come from the merged session list plus the global archived cache (so an active session below an already-archived intermediate is reached), the batch runs deepest-first and skips an ancestor whose subtree failed, and every call carries the runtime key captured at swipe time. The confirmation question this PR left open is answered there explicitly — no confirmation, matching the swipe-immediate interaction #2561 introduced — and flagged for maintainers to overrule if they disagree.

The `combineActiveSessionsWithLive` helper is not revived; the precedence lives in the sheet's own memo, since `MobileSessionStatusBar` no longer exists to justify a shared export.

PR #2415, opened 2026-07-25
feat(quota): add Crof and NeuralWatt quota providers

kydorn:
## Intent

Add two new quota providers so users with Crof or NeuralWatt API keys see their usage in the OpenChamber usage dashboard, alongside the existing providers.

- **Crof** (`crof` alias): credits-balance tracker. Surfaces `payload.credits` as a `credits` window with a `$` value label, null used-percent, and no reset (pay-as-you-go balance).
- **NeuralWatt** (`neuralwatt` alias): richer model with three independent surfaces:
  - **Subscription window** — kWh usage against `kwh_included`, labeled `"<Plan> - <billing_interval>"` (e.g. `Standard - month`). `in_overage` renders as 100% with `(exhausted)`.
  - **Key allowance window** (independent of subscription) — `spent_usd` against `effectiveLimit = min(limit_usd, creditsRemaining)`, labeled `"<KeyName> key - <period>"`. `blocked` renders as 100% with `(blocked)`. Reset time falls back to `allowance.reset_at`, then a computed next-period boundary.
  - **Credits balance** — only when no allowance is present (the allowance window already factors `creditsRemaining` into its effective limit, so the standalone credits row is suppressed to avoid double-counting).

Both providers run on both the web server (`packages/web/server/lib/quota/providers/*.js`) and the VS Code extension (`packages/vscode/src/quotaProviders.ts`) with identical parsing logic. Both follow the `wafer.js` precedent: 15s `AbortSignal.timeout`, `Accept-Encoding: identity`, and classified error paths (401 → "Session expired — please re-authenticate with <Provider>", `SyntaxError` → "Invalid response from provider", `AbortError` → "Request timed out").

## Non-goals

- No UI component changes — the usage dashboard already renders `UsageWindow` shapes from any registered provider; these providers reuse that surface.
- No changes to auth/login flows — providers read pre-existing `crof` / `neuralwatt` entries from the OpenCode auth file; configuring keys is out of scope.
- No changes to the VS Code webview quota rendering.
- No new dependencies.
- Pre-existing unused barrel re-exports in `packages/web/server/lib/quota/index.js` are left in place; only the `fetchCrofQuota` addition introduced by this branch was removed (see Risks).

## Affected surfaces

| Surface | Change |
|---|---|
| `packages/web/server/lib/quota/providers/crof.js` | New — Crof provider (web server). |
| `packages/web/server/lib/quota/providers/neuralwatt.js` | New — NeuralWatt provider (web server). |
| `packages/web/server/lib/quota/providers/crof.test.js` | New — 6 tests on documented payloads. |
| `packages/web/server/lib/quota/providers/neuralwatt.test.js` | New — 10 tests on documented payloads. |
| `packages/web/server/lib/quota/providers/index.js` | Registers `crof` and `neuralwatt` in the registry; re-exports `fetchNeuralwattQuota` (parity with other providers). |
| `packages/web/server/lib/quota/index.js` | No net change (a `fetchCrofQuota` re-export was added then removed — see Risks). |
| `packages/web/server/lib/quota/DOCUMENTATION.md` | Adds both providers to the dispatch table. |
| `packages/vscode/src/quotaProviders.ts` | Adds `CrofPayload` / `NeuralwattPayload` types, `fetchCrofQuota` / `fetchNeuralwattQuota`, `listConfiguredQuotaProviders` entries, and `crof` / `neuralwatt` switch arms. The `crof` arm was previously missing, causing a fallback to "Unsupported provider" in the extension host. |
| `packages/ui/src/types/quota.ts` | Adds `'crof'` and `'neuralwatt'` to `QuotaProviderId`. |
| `packages/ui/src/lib/quota/providers/index.ts` | Adds both to `QUOTA_PROVIDERS` metadata. |
| `CHANGELOG.md`, `packages/vscode/CHANGELOG.md` | `[Unreleased]` entries. |

External contracts: the NeuralWatt `https://api.neuralwatt.com/v1/quota` and Crof `https://crof.ai/usage_api/` response shapes are treated as the documented contract; tests pin the parsing to those shapes.

## Repository guidance

|---|---|---|
| `openchamber-change-discipline` skill | Source, module-ownership, package-contract, and registry changes across `packages/web`, `packages/vscode`, and `packages/ui`. | New providers live in their own files under the owning `quota/providers/` module; registry wiring is additive; UI type change is a pure union extension; owning `DOCUMENTATION.md` updated. |
| `AGENTS.md` runtime boundaries | Web server owns quota fetching; VS Code mirrors for parity; shared UI only holds metadata. | Server providers in `packages/web/server/lib/quota/`; VS Code parity in `packages/vscode/src/quotaProviders.ts`; UI only registers provider IDs/names, no data-access logic. |
| `AGENTS.md` correctness invariants | "Never let fetch failure masquerade as authoritative empty success." | Non-OK responses, JSON parse errors, and timeouts all return `ok: false` with a classified error; missing fields degrade to `null` used-percent rather than a fake 0%. |
| Nearest `DOCUMENTATION.md` | `packages/web/server/lib/quota/DOCUMENTATION.md` is the owning module doc. | Both providers added to the dispatch table with their auth aliases. |
| CONTRIBUTING.md code style | Functional style, no `any`, early returns. | VS Code uses typed payloads (`CrofPayload`, `NeuralwattPayload`); no `any`; server JS follows `wafer.js` precedent. |

Skills considered but not applicable: `ui-api-decoupling` (no `RuntimeAPIs` / shared-UI data-access changes — quota flows through the existing server quota route and VS Code bridge, unchanged), `theme-system` (no UI components/styling), `locale-ui-patterns` (no user-facing UI strings beyond provider display names, which match the existing registry convention), `relay-transport` (no realtime/streaming), `sync-state-invariants` (no session/sync state).

## Validation

| Check | Command | Result |
|---|---|---|
| UI type-check | `bun run type-check:ui` | clean |
| UI lint | `bun run lint:ui` | clean |
| VS Code type-check | `bun run type-check` (in `packages/vscode`) | clean (both `tsconfig.json` and `tsconfig.webview.json`) |
| VS Code lint | `bun run lint` (in `packages/vscode`) | clean |
| Quota unit tests | `bunx vitest run server/lib/quota/providers/` (in `packages/web`) | 24/24 pass, incl. 16 new (6 crof, 10 neuralwatt) |
| Full build | `bun run build` | exit 0; only pre-existing Node deprecation warnings and Vite chunk-size advisories |
| Dead-code report | `bun run dead-code` | non-blocking; no new flags on touched files |

Test coverage on documented payloads: subscription-only, yearly billing interval, in-overage exhausted, subscription+allowance combo, exhausted+allowance combo, `min(limit, credits)` effective limit, blocked allowance, credits-only fallback, 401 mapping, JSON parse error. NeuralWatt tests pin the documented `billing_interval: "month"` (not `"monthly"`) shape and call out the distinction in a comment.

## Not verified

- **Server JS lint** — root `eslint.config.js` scope is `**/*.{ts,tsx}` only; `packages/web/server/**/*.js` is not in any lint scope. Server JS correctness is covered by the focused vitest suite on documented payloads, per AGENTS.md guidance for server JS.
- **Runtime/manual** — no live API calls were made to `https://api.neuralwatt.com/v1/quota` or `https://crof.ai/usage_api/`. Provider behavior is covered by unit tests on documented payloads only; edge cases beyond the documented shapes (e.g. unexpected `billing_interval` values, missing `balance.credits_remaining_usd` when an allowance exists) degrade to `null`/suppressed windows but are not exhaustively tested.
- **Cross-runtime parity** — the web server (JS) and VS Code (TS) implementations mirror each other line-for-line but are not shareable code by design (different runtimes, per `AGENTS.md`). Parity was checked by reading, not by a shared test harness.

## Visual evidence

No new UI components are added — the usage dashboard already renders `UsageWindow` shapes from any registered provider, and these providers emit the same shape. When a `crof` or `neuralwatt` key is present in the OpenCode auth file, the provider appears in the usage dropdown and renders through the existing components.

Capturing screenshots requires valid API credentials for at least one of the two providers, which I do not have. A reviewer configuring a key should confirm the dashboard renders the expected windows; the unit tests pin the exact `UsageWindow` values the UI will receive.

## Risks and failure behavior

- **API contract drift.** The parsing pins the documented NeuralWatt (`balance.credits_remaining_usd`, `subscription.{kwh_included, kwh_used, billing_interval, in_overage, kwh_reset_date}`, `key.allowance.{limit_usd, spent_usd, period, blocked, reset_at}`) and Crof (`credits`) shapes. If either API adds required fields or renames them, the providers degrade to `null` values and suppressed windows rather than crashing — but the dashboard would show incomplete data. Mitigation: the tests pin the documented shapes and would need updating alongside any API change.
- **Window-seconds approximation.** `periodToWindowSeconds` maps `monthly`/`month` → 30d and `yearly`/`year` → 365d. This is for the UI progress-bar label only, not billing; real calendar months vary. Documented in a comment on both implementations.
- **Credits double-counting.** When `key.allowance` is present, `creditsRemaining` is folded into the allowance's `effectiveLimit = min(limit, creditsRemaining)`, and the standalone `credits_balance` window is suppressed. This is intentional and tested (`surfaces both subscription and allowance windows… suppresses credits`). If NeuralWatt changes the allowance semantics so that credits are *not* meant to cap the allowance, the effective-limit logic would be wrong.
- **Dead re-export removal.** The branch initially added `fetchCrofQuota` to the `quota/index.js` barrel re-export list; removed after confirming via grep that the entire named re-export list (`fetchClaudeQuota`, `fetchOpenaiQuota`, …) has zero external importers — `fetchQuotaForProvider` is the only consumed entrypoint. The removal is safe; if a future consumer wants named imports, the pre-existing unused re-exports are a separate cleanup.
- **No data loss / security / perf concerns.** Providers are read-only GET requests with bearer auth from the existing auth file; no secrets logged (error messages reference the provider name and HTTP status only); 15s timeout prevents hanging requests; no caching/state mutation.
- **Non-blocking.** Both call sites for `fetchQuotaForProvider` are request-scoped, on-demand paths: `GET /api/quota/:providerId` in `packages/web/server/lib/quota/routes.js:87` and the `api:quota:get` IPC handler in `packages/vscode/src/bridge-system-runtime.ts:518`. Neither runs on startup, session load, the render hot path, or the relay/streaming path.
- **Rollback.** Revert the branch — no migrations, no persisted state, no config file format changes. Removing the providers reverts the UI to not showing them.


kydorn:
Follow-up to the non-blocker Finding 1 in the review on `2262aaf2`:

**NeuralWatt can return `ok: true` with empty `windows` on a data-less payload.** Addressed in `cca49fec`.

**Fix:** matches the `wafer.js` precedent — after the three population branches, if `windows` is still empty, return `ok: false, error: "No quota data in response"` instead of `ok: true` with empty windows. Applied in both runtimes:
- `packages/web/server/lib/quota/providers/neuralwatt.js`
- `packages/vscode/src/quotaProviders.ts` (`fetchNeuralwattQuota`)

**Test:** added `returns no-quota-data on a 200 payload with no usable windows` to `neuralwatt.test.js`, pinning the payload shape (`balance.credits_remaining_usd: null + subscription: null + allowance: null`) that triggers the guard. Full quota suite: 25/25 pass (was 24, +1).

**When it triggers:** contract drift / partial upstream failure returning a bare `200` with `{}` or renamed fields — exactly the "fetch failure masquerading as authoritative empty success" case `AGENTS.md` flags. A genuine zero-balance account with `credits_remaining_usd: 0` is unaffected (`toNumber(0)` is `0`, not `null`, so the credits fallback still fires).

Validation: web quota tests 25/25, VS Code `tsc --noEmit` clean. Not re-running full `bun run build` here — the change is 4 defensive lines × 2 runtimes plus one test; nothing it touches is downstream of the bundler.

kydorn:
kydorn the human here. To complement what my agent has already provided, I am adding a screenshot of the usage tracker in the electron app:
image

makeittech:
Thanks a lot for the PR. Few things I'd want fixed before the merge:

- Dynamic window keys are the big one. windows[subLabel] / windows[allowLabel] use human-readable, API-supplied strings ("Standard - month", "Prod key - monthly") as map keys. Every other provider uses stable identifiers (credits, credits_balance). This means: keys change when the user renames a key or upgrades a plan, two labels can collide and silently overwrite each other, and the labels can never be localized or restyled. Use stable keys (subscription, key_allowance) and put the display string in a label field.

- Windows are silently dropped when the label is empty. If plan/billing_interval are missing, if (subLabel) skips the window entirely — with valid kWh data present you can end up returning ok: false, "No quota data in response". Label absence shouldn't gate data emission.

- ~300 lines copy-pasted between crof.js/neuralwatt.js and quotaProviders.ts, with zero tests on the VS Code copy. "Parity checked by reading" isn't a control — the two will drift. At minimum port the payload-parsing tests to the TS side; better, extract the pure parse functions and test them once.

- **effectiveLimit = min(limit_usd, credits_remaining) is speculative.** The PR admits it's an assumption about NeuralWatt's semantics, and neither endpoint was ever called live. If wrong, every allowance percentage is wrong. Get one real response captured, or make the assumption fail-safe (show both, don't invent a combined denominator).

- **computeAllowanceResetAt is untested and incomplete** — no yearly branch (which periodToWindowSeconds does support), and it uses wall-clock new Date() with no fake-timer coverage. Also in_overage/blocked → hard 100% throws away how far over the user actually is.

- **New dead export:** export const fetchCrofQuota added to providers/index.js — the PR body itself argues that whole re-export list has no consumers. Don't add to it.

- **Naming inconsistency:** UI registry says Crof.ai, provider returns providerName: 'Crof'. Also neuralwatt is appended at the end of the registry/doc table while crof was inserted alphabetically — pick one.

kydorn:
Thanks for the careful review. Responses and follow-up commit below.

## #1 Dynamic window keys

Pushing back here. Re-traced the actual call sites in `packages/ui/src`: `useQuotaStore`, `useTraySync`, `UsagePage`, `Header`, `VSCodeLayout`, `MobileApp` — every consumer iterates `Object.entries(usage.windows)` fresh on each fetch. **Nothing persists state keyed by window label.** So "keys change when the user renames a key" does not actually break anything observable — the next fetch returns the new label and the UI shows it.

Collision risk requires multiple labels hashing to the same string. NeuralWatt's contract returns a single `key` object (not `keys[]`) and a single `subscription` object. So at most one subscription window and one allowance window per response — collision cannot happen on the documented contract.

That said, the *cosmetic* argument (align with the ID-based pattern other providers use) is fair. Switched the keys to stable IDs (`subscription`, `key_allowance`, `credits_balance`) and routed them through `formatWindowLabel` to existing i18n entries (`quota.window.planLimit` for subscription since it *is* the plan cap just denominated in kWh; `quota.window.billingCycle` for key allowance since the concept matches the existing cursor precedent). Zero new localization strings.

## #2 Silent window drops on empty label

Addressed as part of #1 — the stable keys emit unconditionally; the previous `if (subLabel)` / `if (allowLabel)` gates are gone. The `No quota data in response` guard now only fires when all numeric data is genuinely missing.

## #3 ~300 lines copy-pasted, zero VS Code tests

Agreed on the drift risk. Added `packages/vscode/src/quotaProviders.test.ts` — 18 cases mirroring the web payload tests (Crof: credits/missing/401/parse-error; NeuralWatt: subscription/yearly/in-overage/both-windows/effective-limit-math/blocked/credits-fallback/no-data/401/parse-error + pinned-clock monthly and yearly reset_at). Refactor to shared pure parse functions is OOS for this PR.

## #4 `effectiveLimit = min(limit_usd, credits_remaining)` was wrong

You were right that the original math was off, but the correction is `min(limit, credits_remaining + spent)`, not `min(limit, credits_remaining)`. The credits wallet is reduced by each period's spend *before* the allowance cap bites, so the real ceiling is the wallet post-spend-recovery. Verified against a real captured response: spent=5.56, remaining=194.44, limit=200, credits_remaining=71.97 → effectiveLimit=77.53 (not 71.97), usedPercent=7.17% (not 7.72%).

## #5 `computeAllowanceResetAt` untested and incomplete

Fixed: added `yearly`/`year` branch (was missing; `periodToWindowSeconds` already supported it). Added two pinned-clock tests (`vi.useFakeTimers` + `setSystemTime`) asserting exact UTC midnight timestamps for monthly and yearly resets.

For (c) `in_overage`/`blocked` → hard 100%: kept. Surfacing the magnitude is UX judgment and I'd rather not change behavior in a follow-up. It also matches existing precedent.

## #6 Dead export

Removed `fetchCrofQuota` *and* `fetchNeuralwattQuota` — both were added in this PR and `bun run dead-code` confirmed both were unused.

## #7 Naming inconsistency

Aligned to `CrofAI`: provider returns `providerName: 'CrofAI'`, UI registry updated from `Crof.ai` to `CrofAI`, error strings updated. Auth key/alias `crof` unchanged. Also fixed the doc/registry ordering — both inserted alphabetically now.

Validation: 22 web quota tests pass, 18 new VS Code tests pass, `tsc --noEmit` clean on web/ui/vscode, lint clean, dead-code report no longer flags the two new exports.

kydorn:
Follow-up: simplification on #5 (computeAllowanceResetAt).

After re-reading my own response, the wall-clock fallback for `reset_at` was speculative — the documented API contract always returns it, no other provider fabricates dates from a period string, and the missing-yearly-branch bug existed *only* because this fallback existed. Deleted `computeAllowanceResetAt` entirely (web + VS Code). Now `resetAt = toTimestamp(allowance.reset_at)` — trust the API or null, matching `wafer.js`/`crof.js`/`credits_balance`.

Dropped the two pinned-clock tests that covered the fallback path. Test payloads that lacked `reset_at` now include it. Web 18/18 pass, VS Code 16/16 pass, tsc + lint clean.

kydorn:
Updating screenshots.

Allowance:
image

Just credits:
image

kydorn:
Final iteration on NeuralWatt window keys based on real API behavior. HEAD `532db07d`.

## Why the key name matters

The NeuralWatt portal lists API keys by their user-assigned `name`, and an account can rotate or replace keys over time. If the allowance window surfaced in OpenChamber shows no key identifier, the user has no way to correlate the percentage they see here with the specific key that's eating into its allowance on the NeuralWatt panel. Putting `key.allowance.spent_usd` / `key.allowance.remaining_usd` against a stable denominator is only half the picture — you also need to know *which* key that bar represents.

## Window-key design

| Window | Title source (map key) | valueLabel (top-right metric) | Why |
|---|---|---|---|
| Subscription | `subscription.plan` (e.g. `'standard'`); falls back to `'plan_limit'` → localized `'Plan Limit'` | none (percent shows) | Plan name is API data, renders verbatim — no new i18n needed; fallback only fires when API omits plan |
| Allowance | `allowance.period` (`'daily'`/`'weekly'`/`'monthly'`); falls back to `'billing_cycle'` → localized `'Billing Cycle'` | `key.name` | Period keys are already localized; key name carries identification |
| Credits | `'credits_balance'` → existing `'Credits Balance'` | `$X.XX` (unchanged) | Reuses existing |

This replaces the earlier stable-key approach (`subscription` / `key_allowance`) which used generic IDs and routed everything through `formatWindowLabel`. The stronger signal from having both `plan` and `key.name` visible in the rendered card recovered the value we lost when going stable-only. No new i18n strings.

## Why `windowSeconds` is null for the subscription

Documented NeuralWatt subscription limits reset monthly regardless of the billing interval — an annual plan still has a monthly kWh cap. The API exposes `kwh_reset_date` (the reset timestamp) and `current_period_start` / `current_period_end` (which are billing-period dates, not kWh-window dates), but no kWh-window *start*. Computing a window duration would require either an unsupported date subtraction (using billing-period dates that don't represent the kWh window) or hardcoding `30 * 86400` (which is false on a real calendar and was the source of the original yearly-bug attempt).

Passing `null` is the honest answer: the progress bar still renders using `usedPercent`, the reset timestamp still shows via `formatQuotaResetLabel`, but the pace/expected-usage projection is skipped for this window only. Same precedent as `credits_balance`, which also passes `windowSeconds: null`.

## Other cleanups already in HEAD `391ad23` (kept)

- `(exhausted)` / `(blocked)` `valueLabel` annotations removed — the 100% bar carries the signal
- `computeAllowanceResetAt` deleted entirely — API-provided `reset_at` is trusted, matching `wafer.js` / `crof.js` / `credits_balance`
- `effectiveLimit = min(limit, credits_remaining + spent)` math kept, with covering tests
- `fetchCrofQuota` / `fetchNeuralwattQuota` dead exports removed

Validation: web 20 quota tests pass, VS Code 19 tests pass, `tsc --noEmit` clean on web/ui/vscode, lint clean.

makeittech (approved):

PR #2430, opened 2026-07-26
fix(ui): sync message IDs to server clock to prevent infinite response loop

davidmoserai:
## Intent

A device whose wall clock is set ahead of the server produces client-generated message IDs that sort lexically newer than every server-generated assistant ID. The client then treats the user message as perpetually unanswered and keeps re-prompting the model, producing an unbounded stream of assistant responses from a single send.

Reproduced live: an iPhone clock ~80 seconds ahead of the server, one user message ("Hey bro") resulted in ~20 assistant responses sharing the same parent before the run was force-aborted. Resetting the device clock to automatic stopped the loop immediately. The user-message ID decoded to +80,503 ms ahead of server receipt.

This PR derives client ID timestamps from the server's clock instead of the device wall clock, so a skewed device clock can no longer break message ordering or trigger the re-prompt loop.

## Non-goals

- Not changing the OpenCode `Identifier.ascending` wire format or `msg_`/`prt_` ID shape — new IDs remain byte-compatible with server IDs and with existing persisted sessions.
- Not adding a new server endpoint. A first send without a clock sample makes one request to the existing `/health` endpoint, then later sends reuse passively observed response dates.
- Not retrying or changing `prompt_async` send/rollback semantics.
- Not fixing device-side clock display or any non-message use of `Date.now()`.
- Not addressing sub-second ordering collisions beyond the existing 12-bit counter.

## Affected surfaces

- `packages/ui` — new `lib/runtime-id.ts` module; `lib/opencode/client.ts` and `sync/session-actions.ts` swap their duplicated ID generators for the shared one; `lib/runtime-fetch.ts` observes response dates and scopes read coalescing by runtime key.
- `packages/web/server` — `index.js` exposes `date` in `Access-Control-Expose-Headers` so Capacitor/Electron cross-origin clients can read the header.
- Runtimes: web (same-origin, already readable), desktop/Electron (packaged-client origin, now exposed), VS Code (extension-host bridge already forwards response headers, no change needed), hosted mobile, and Capacitor mobile (now readable via CORS).
- Persisted/external contracts: none changed. Optimistic message IDs and `time.created` use the synchronized timestamp; server reconciliation (echoed message ID and authoritative `time.created`) replaces them as before.
- User-visible state: optimistic user messages no longer render with a future-dated `time.created` while awaiting reconciliation.

## Repository guidance

|---|---|---|
| `openchamber-change-discipline` | Source change across shared UI + server. | Smallest complete change: one new focused module, two call-site swaps, one transport hook, one CORS header line. No drive-by refactors. |
| `ui-api-decoupling` | Reads a response header from the shared runtime transport (`runtimeFetch`). | `Date` is observed only inside `runtimeFetch` for requests that resolve through the runtime (direct or relay); external absolute URLs are excluded. No hardcoded URLs, no SDK bypass. |
| `sync-state-invariants` | Changes optimistic message insertion and ordering. | Optimistic IDs remain client-generated and echoed by the server for in-place reconciliation; `time.created` is derived from the same synchronized clock; rollback paths unchanged. |
| `relay-transport` | The relay path is a real transport for this feature. | Relay responses go through `relay.fetch` and are observed the same way as direct responses; no new WS, no wire-format change, no origin handling change. |
| `AGENTS.md` validation rules | Executable source + cross-workspace contract. | Ran focused tests, package + workspace type-check/lint, server syntax check, web production build, and `dead-code`. |

## Validation

| Check | Command | Result |
|---|---|---|
| Focused tests | `bun test packages/ui/src/lib/runtime-id.test.ts packages/ui/src/lib/runtime-fetch.test.ts packages/ui/src/lib/opencode/client.test.ts packages/ui/src/sync/session-actions.test.ts` | 68 pass, 0 fail (pre-existing console output from intentional error-path tests only). |
| UI type-check | `bun run type-check` (packages/ui) | Pass (no output). |
| UI lint | `bun run lint` (packages/ui) | Pass (no output). |
| Workspace type-check | `bun run --filter '*' type-check` | All 5 workspaces pass. |
| Workspace lint | `bun run --filter '*' lint` | All 5 workspaces pass. |
| Server syntax | `node --check packages/web/server/index.js` | Pass (no output). |
| Web build | `bun run --cwd packages/web build` | Pass. Only pre-existing KaTeX-asset and chunk-size warnings. |
| Dead code | `bun run dead-code` | No new findings for the added module; only pre-existing repo findings reported. |

Manual verification:
- iPhone Safari loaded the repository checkout's `packages/web/dist` through a temporary dev server on the phone hotspot, with the iPhone clock deliberately set ahead. The fixed bundle produced one assistant response instead of the previous repeated assistant responses. The server static-route log was checked to verify it was not serving the installed desktop app's bundled web assets.

Not verified by me:
- End-to-end reproduction through the private relay or installed Capacitor app. The Safari test exercises the same shared UI code and independent iPhone clock, but not those transports.
- Electron packaged-client runtime build and Capacitor native build (only web build was run). The change to the shared UI module compiles across workspaces via `bun run --filter '*' type-check`.

## Visual evidence

No user-visible rendering change. The diff affects ID generation, an HTTP response header, and optimistic `time.created` bookkeeping; it does not alter styles, layout, colors, or interactions. The user-observable behavior change is the absence of the infinite-response loop under clock skew, which is not capturable as a static screenshot.

## Risks and failure behavior

- **Missing/malformed `Date`:** `observeRuntimeResponseDate` only accepts strict IMF-fixdate values (`/^...GMT$/`) and a finite `Date.parse`; everything else is ignored. Runtimes that never send a usable `Date` fall back to the existing local-clock behavior — no regression for those runtimes.
- **No sample yet (cold start / first send before any response):** the client requests the existing `/health` endpoint once before generating the first message ID. If that request fails, it retains the local-clock fallback rather than blocking a send.
- **Runtime switching:** clock samples and monotonic ID counters are scoped by `runtimeKey`. The read-coalescing key now includes `runtimeKey`, so a request issued after a switch cannot receive a clone of the previous runtime's response. Relay requests are always treated as runtime requests.
- **External/non-runtime URLs:** `Date` is only observed when `shouldAttachRuntimeAuth` is true (direct) or the relay path is used. External absolute URLs are explicitly excluded (covered by a regression test).
- **Monotonic regression:** a later response with an older `Date` cannot rewind the per-runtime counter; `ascendingRuntimeId` keeps the last high-water mark and advances the counter, so IDs remain ascending (covered by a regression test).
- **Sleep/suspend drift:** the projection uses `performance.now()` deltas from the last sample. After a long suspend the projected timestamp may lag until the next runtime response refreshes the sample; IDs remain valid and ascending, only momentarily conservative.
- **CORS:** adds `date` to `Access-Control-Expose-Headers` for the existing packaged-client origins only (`openchamber-ui://`, `capacitor://localhost`, `localhost`, dev origins). No new origin is trusted; `Vary: Origin` is unchanged.
- **Rollback/cleanup:** unchanged. Optimistic message/part removal on send failure and confirmation reconciliation use the same IDs as before.


davidmoserai:
Closing this PR; pursuing the server-authoritative message-ID design instead.

PR #2432, opened 2026-07-26
fix(ui): cascade-remove extracted document attachments when parent is removed

### LINKED ISSUE #2426: [Bug] PPT file attachment: canceling the PPT does not cascade-remove auto-extracted images
## Describe the bug

When dragging a PowerPoint (.ppt/.pptx) file into the chat input, it automatically extracts/splits the slides into individual images. However, when canceling (removing) the original PPT file from the attachments, the auto-extracted slide images are **not** removed together — they remain attached and must be canceled one by one.

## To Reproduce

1. Open a chat session
2. Drag a `.ppt` or `.pptx` file into the chat input
3. Observe that the PPT is split into multiple slide images in the attachment area
4. Click the × button to remove the PPT file
5. See that only the PPT file is removed — all the slide images remain

## Expected behavior

When the PPT file is removed, all its auto-extracted slide images should be removed together in one action.

## Environment

- OpenChamber version: 1.16.3
- Platform: Desktop

chiamsun:
## Intent

Fix #2426: when a document (PPTX, DOCX, XLSX, ODP, ODT, ODS) is canceled from chat attachments, its auto-extracted slide images should be removed together instead of being left orphaned.

## Non-goals

- Does not change how documents are extracted or how images are generated
- Does not affect VS Code / server-source attachments

## Affected surfaces

- `packages/ui/src/stores/types/sessionTypes.ts` — added `sourceDocumentId?: string` to `AttachedFile`
- `packages/ui/src/sync/input-store.ts` — `addAttachedFile` assigns shared `sourceDocumentId` for multi-entry document extractions; `removeAttachedFile` cascades removal to all entries sharing the same `sourceDocumentId`
- `packages/ui/src/sync/input-store.test.ts` — 4 new tests for cascade removal behavior

## Solution

1. Added optional `sourceDocumentId` field to `AttachedFile` type
2. In `addAttachedFile`, when multiple files are prepared from a document extraction, a shared `sourceDocumentId` is generated and assigned to all entries
3. `removeAttachedFile` checks for `sourceDocumentId` and cascades removal to all entries in the group; non-grouped entries fall back to single-ID removal

## Validation

```
bun test packages/ui/src/sync/input-store.test.ts
```

All 18 tests pass:

- Existing 14 tests continue to pass (attachment lifecycle, MIME normalization, atomic extraction, collision re-evaluation, etc.)
- 4 new tests:
  - extracted document entries share a `sourceDocumentId`
  - removing the parent text entry cascade-removes all children
  - removing an image child also cascade-removes the entire group
  - non-document attachments (no `sourceDocumentId`) remove individually

## Risk / failure behavior

- `sourceDocumentId` is `undefined` for single-file attachments and VS Code / restored attachments — those paths are unchanged
- If `removeAttachedFile` is called with an ID that doesn't match any entry, it's a no-op
- No new async paths, no new dependencies

### DIFF
--- a/packages/ui/src/stores/types/sessionTypes.ts
+++ b/packages/ui/src/stores/types/sessionTypes.ts
@@ -21,6 +21,9 @@ export interface AttachedFile {
     serverPath?: string;
     vscodePath?: string;
     vscodeSource?: 'file' | 'selection';
+    /** Shared ID linking entries extracted from the same document (PPTX, DOCX, etc.).
+     *  Removing any entry with this ID cascades to all entries in the group. */
+    sourceDocumentId?: string;
 }
 
 export type EditPermissionMode = 'allow' | 'ask' | 'deny' | 'full';
--- a/packages/ui/src/sync/input-store.test.ts
+++ b/packages/ui/src/sync/input-store.test.ts
@@ -276,4 +276,98 @@ describe("input-store attachments", () => {
     const textAttachment = useInputStore.getState().attachedFiles[1]
     expect((await textAttachment?.file.text())?.includes("[design-image-2.png]")).toBe(true)
   })
+
+  testWithMockFileReader("extracted document entries share a sourceDocumentId for cascade removal", async () => {
+    const archive = zipSync({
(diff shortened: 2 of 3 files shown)


makeittech (approved):
Looks good

makeittech:
Looks good.

PR #2444, opened 2026-07-26
fix(ui): stabilize Markdown code block line layout

ChangeHow:
## Intent

Fix Markdown code-block line numbers drifting away from their content in Safari/WebKit while rendering correctly in Electron/Chromium.

This PR now separates the work into two commits:

1. **Rendering/performance:** render each source line and its number in one grid row, replacing gutter synchronization through `Range.getClientRects()`, RAF, and `ResizeObserver`.
2. **Boundary correctness:** map DOM Range starts to the right node and ends to the left node at exact text boundaries, preventing second-line file references from being inserted into a hidden newline container. Focused tests cover second-line starts and end boundaries.

The Markdown source and Shiki tokenization remain unchanged. #2446 provides the boundary-affinity fix alone on top of `main`, allowing maintainers to choose the minimal fix or this broader structural simplification.

The structural change also keeps related behavior correct: code copy rebuilds source text without line numbers, assistant-message copy offers Markdown plus semantic HTML with a plain-text fallback, and message image export wraps complete code while hiding code actions.

## Non-goals

- No Shiki worker, theme, tokenization, API, persistence, or runtime-bridge changes.
- No unrelated Markdown styling reorganization or bug fixes.
- User-message, table, and Mermaid copy behavior remains unchanged.

## Affected surfaces

- Shared `packages/ui` Markdown code blocks across Web, Electron, VS Code, and Mobile.
- Assistant-message clipboard payloads and message PNG export clones.
- No persisted data or external contract changes.

## Repository guidance

|---|---|---|
| `AGENTS.md` and `openchamber-change-discipline` | Shared UI behavior changes across runtimes | Keeps one shared implementation and separates structural and boundary concerns into distinct commits |
| `theme-system` | Adds line-number styling | Uses existing theme tokens only |
| `performance-engineering` | Removes repeated layout measurement | Replaces `Range`/RAF/`ResizeObserver` height work with structural CSS layout |
| Message-part documentation and Electron README | Defines renderer ownership and Desktop boundaries | Keeps behavior in `packages/ui`; no Electron shell changes |

## Validation

| Check | Result |
|---|---|
| Focused UI tests | Passed: 18 tests, 0 failures |
| `bun run type-check` in `packages/ui` | Passed |
| `bun run lint` in `packages/ui` | Passed |
| `bun run build` in `packages/web` | Passed with existing font/chunk warnings |
| `bun run dead-code` | Completed; existing non-blocking findings remain |

Not run: Electron packaging, Feishu/Notion paste, and PNG visual inspection.

## Visual evidence

Before/after and soft-wrap screenshots are attached in [this PR comment](https://github.com/openchamber/openchamber/pull/2444#issuecomment-5084486091).

## Risks and failure behavior

- The shared DOM structure affects every UI runtime; the structural and boundary changes are isolated in separate commits.
- Streaming still defers line decoration until completion.
- Clipboard rich writes fall back to the existing plain Markdown path.
- Image-export styles apply only to the temporary clone, which retains existing cleanup behavior.

### DIFF
--- a/packages/ui/src/components/chat/ChatMessage.tsx
+++ b/packages/ui/src/components/chat/ChatMessage.tsx
@@ -26,7 +26,7 @@ import { isLikelyProviderAuthFailure, PROVIDER_AUTH_FAILURE_MESSAGE } from '@/li
 import { getProviderModelDisplayName } from '@/lib/modelDisplay';
 import { lazyWithChunkRecovery } from '@/lib/chunkLoadRecovery';
 import type { TurnGroupingContext } from './lib/turns/types';
-import { copyTextToClipboard } from '@/lib/clipboard';
+import { copyMarkdownToClipboard, copyTextToClipboard } from '@/lib/clipboard';
 import { FadeInOnReveal } from './message/FadeInOnReveal';
 import { streamPerfCount } from '@/stores/utils/streamDebug';
 import { areOptionalRenderRelevantMessagesEqual, areRenderRelevantMessagesEqual, areRelevantTurnGroupingContextsEqual } from './message/renderCompare';
@@ -771,7 +771,13 @@ const ChatMessage: React.FC<ChatMessageProps> = ({
     const hasTextContent = messageTextContent.length > 0;
 
     const handleCopyMessage = React.useCallback(async () => {
-        const result = await copyTextToClipboard(messageTextContent);
+        let result;
+        if (isUser) {
+            result = await copyTextToClipboard(messageTextContent);
+        } else {
+            const { renderMarkdownSync } = await import('./markdown/markdownCore');
+            result = await copyMarkdownToClipboard(messageTextContent, renderMarkdownSync(messageTextContent));
+        }
         if (!result.ok) {
             return false;
         }
--- a/packages/ui/src/components/chat/MarkdownRendererImpl.tsx
+++ b/packages/ui/src/components/chat/MarkdownRendererImpl.tsx
@@ -25,8 +25,7 @@ import {
   attachMarkdownInteractions,
   applyMarkdownCodeBlockWrapState,
   decorateMarkdown,
-  scheduleMarkdownCodeLineNumberSync,
-  syncMarkdownCodeLineNumbers,
+  getMarkdownCodeText,
   type DecorateContext,
   type DecorateLabels,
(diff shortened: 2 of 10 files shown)


ChangeHow:
# Screenshots

## Before

### Pic#1 the exist web UI with mismatched line number style

image

### Pic#2 the desktop version is correct

image

## After

### Pic#3 the fixed version

image

### Pic#4 the fixed version after enabled soft-wrap

image

btriapitsyn:
Thanks for working on this. The structural grid approach looks like a good direction and the new screenshots help clarify the intended result.

There is one multiline file-reference case that needs adjustment before merge. `layoutCodeLines()` now inserts each newline as text inside a hidden `[data-md-code-line-break]` span. `wrapBlockCodePathTokens()` includes those newline text nodes when mapping offsets, and `findTextPosition()` resolves an offset exactly at a boundary to the preceding node. As a result, a file reference that starts on line 2 or later can be removed from the visible line and inserted into the hidden line-break span. For example:

```text
output:
src/index.ts:12
```

Could you adjust the offset mapping so a match beginning after a newline resolves to the following line-content node? One option is to exclude the hidden line-break nodes and explicitly account for reconstructed newline offsets; another is to use separate start/end boundary handling so an exact start boundary selects the next text node.

Please also add a focused regression test covering a file reference at the start of the second line, ideally alongside the code-text round-trip edge cases (with and without a trailing newline). Once that is addressed, I would be happy to take another look.

btriapitsyn (approved):

ChangeHow:
Thanks for catching this. You were right about the exact-boundary start resolving to the preceding newline node.

I split the follow-up into two parts:

- `72b8224e` in this PR is the boundary-correctness fix: Range starts use right affinity and ends use left affinity, so second-line file references stay in visible line content.
- #2446 applies the same boundary rule to the current `main` implementation, including the existing line-number measurement path, as a minimal alternative.

The original `ed150bd7` commit remains the separate rendering/performance change that replaces Range-based height synchronization with structural grid rows. Focused tests now cover a file reference at the start of line two, with and without a trailing newline, plus end-boundary behavior (`60a394de`).

This keeps the two concerns explicit and lets maintainers choose the minimal boundary fix in #2446 or the broader rendering simplification in this PR.

PR #2446, opened 2026-07-26
fix(ui): respect Markdown text range boundaries

ChangeHow:
## Intent

Fix exact text-node boundary mapping in Markdown code blocks. A range start immediately after `\n` previously resolved to the preceding newline node, which can mis-measure wrapped line heights in WebKit and can move a second-line file reference into the wrong DOM container.

Range starts now use right affinity and range ends use left affinity. This is the minimal alternative to the structural grid/rendering optimization in #2444; maintainers can choose between the targeted boundary fix here and the broader layout simplification there.

## Non-goals

- No code-block DOM, CSS, clipboard, screenshot, Shiki, or runtime changes.
- No removal of the existing Range/RAF/ResizeObserver line-height measurement path.

## Affected surfaces

- Shared Markdown code-block line-number measurement and file-reference annotation in `packages/ui`.
- Applies to all UI runtimes through the existing shared renderer; no persisted or external contracts change.

## Repository guidance

| Guidance | Application |
|---|---|
| `AGENTS.md` and `openchamber-change-discipline` | Keeps the fix limited to the confirmed boundary invariant and focused tests |
| `performance-engineering` | Preserves the existing measurement path without adding work; #2444 contains the separate structural optimization |
| Message-part documentation | Keeps the change in the shared Markdown renderer owner |

## Validation

- `bun test src/components/chat/markdown/textPosition.test.ts src/components/chat/MarkdownRendererImpl.test.ts`: 16 passed.
- `bun run type-check` in `packages/ui`: passed.
- `bun run lint` in `packages/ui`: passed.
- `bun run dead-code`: completed with existing non-blocking findings.

Not run: Safari or Electron manual regression.

## Visual evidence

The original mismatch and the structural alternative are documented in #2444. No dedicated current-HEAD screenshot is attached for this minimal branch.

## Risks and failure behavior

- The helper is shared by line-height measurement and file-reference annotation, with focused tests covering second-line starts, trailing newlines, and end boundaries.
- Rollback is a single commit revert; no data, network, auth, or runtime boundary changes are involved.

### DIFF
--- a/packages/ui/src/components/chat/MarkdownRendererImpl.tsx
+++ b/packages/ui/src/components/chat/MarkdownRendererImpl.tsx
@@ -32,6 +32,7 @@ import {
   type MermaidControlOptions,
   type MermaidRender,
 } from './markdown/decorate';
+import { findTextPosition } from './markdown/textPosition';
 import { createMermaidViewerRegistry, MERMAID_BLOCK_SELECTOR, shouldRefreshMermaidViewers } from './markdown/mermaidViewer';
 import {
   BLOCK_PATH_TOKEN_RE,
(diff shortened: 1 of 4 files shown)


ChangeHow:
merged in #2444

PR #2449, opened 2026-07-27
fix(ui): preserve Markdown code selections

ChangeHow:
## Intent

Fix Markdown selections copied into **Add to Chat** after the code-block grid layout introduced in #2444.

The grid renderer adds language controls, line-number elements, logical line wrappers, and hidden newline nodes around highlighted code. The existing selection serializer still treated the cloned code-block wrapper as generic inline DOM, so a mixed prose/code selection could become content such as `md1## Heading23- item`: the language label and line numbers leaked into the prompt while source newlines were collapsed.

This follow-up converts the cloned selection into a small renderer-independent tree, removes code-line decorations, recognizes the Markdown code wrapper, and preserves language, indentation, blank lines, and partial code selections. Add to Chat also chooses a fence longer than any backtick run in the selected Markdown, so nested fenced code remains valid.

## Non-goals

- No changes to code-block rendering, syntax highlighting, line layout, or clipboard MIME behavior.
- No changes to plain-text Copy or New Session selection actions.
- No unrelated Markdown renderer refactors.

## Affected surfaces

- Shared `packages/ui` text-selection serialization used by Add to Chat and Add to Notes.
- Web, Electron, VS Code, and Mobile share this UI implementation.
- No persisted data, API, bridge, or external contract changes.

## Repository guidance

| Guidance | Application |
|---|---|
| `AGENTS.md` and `openchamber-change-discipline` | Keeps the patch limited to the selection regression introduced by #2444 and its focused tests |
| Message-part documentation | Keeps selection-to-Markdown behavior in the shared message UI owner |
| Cross-runtime contract | Uses one shared serializer without runtime-specific branches |

## Validation

- `bun test src/components/chat/message/selectionMarkdown.test.ts src/components/chat/markdown/textPosition.test.ts src/lib/clipboard.test.ts`: 14 passed.
- `bun run type-check` in `packages/ui`: passed.
- `bun run lint` in `packages/ui`: passed.
- `bun run build` in `packages/web`: passed with existing font/chunk warnings.
- `bun run dead-code`: completed with existing non-blocking findings.

Not run on the final branch: manual cross-runtime selection verification.

### Before

the screenshot showing Add to Chat output containing the `md` language label, code line numbers, and collapsed Markdown lines.

image

### After

image

image

## Risks and failure behavior

- Selection DOM is normalized before serialization; unknown elements retain their text through the existing inline/block fallback.
- Dynamic fences preserve selected code that already contains backticks and language identifiers such as `c++`.
- Plain-text selection remains the fallback when Markdown serialization produces no output.
- Rollback is a single commit revert; no data migration or cleanup is required.

### DIFF
--- a/packages/ui/src/components/chat/message/TextSelectionMenu.tsx
+++ b/packages/ui/src/components/chat/message/TextSelectionMenu.tsx
@@ -15,6 +15,7 @@ import { resolveProjectForSessionDirectory } from '@/lib/projectResolution';
 import { useEffectiveDirectory } from '@/hooks/useEffectiveDirectory';
 import { isVSCodeRuntime } from '@/lib/desktop';
 import { useI18n } from '@/lib/i18n';
+import { rangeToMarkdown, trimSelectionValue, wrapMarkdownSelectionForChat } from './selectionMarkdown';
 
 interface TextSelectionMenuProps {
   containerRef: React.RefObject<HTMLElement | null>;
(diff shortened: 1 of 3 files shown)


btriapitsyn:
Thanks for following up on this regression. The renderer-independent tree, line-number cleanup, and dynamic fence handling all look like good improvements.

I found one production DOM shape that still needs adjustment before merge. `MarkdownRendererImpl` wraps every parsed Markdown block in `div[data-md-block]`. When a selection starts in one block and ends in another, `Range.cloneContents()` preserves those wrappers as top-level nodes. The serializer currently does not retain `data-md-block`, and `renderBlockMarkdownNode()` handles every generic `div` through `renderInlineMarkdownNode()`.

That means a complete Markdown code wrapper nested inside an intermediate `div[data-md-block]` never reaches the `component === "markdown-code"` branch. Its contents can be flattened again, including the language header and collapsed code newlines. The current fixtures do not expose this because `codeWrapper(...)` is passed directly as a top-level node.

Could you preserve the `data-md-block` marker in `SelectionNode` and serialize its children through the block path rather than the generic inline `div` path? Please add a regression fixture matching the production structure, for example:

```text
div[data-md-block] > p
div[data-md-block] > div[data-component="markdown-code"]
div[data-md-block] > p
```

The test should select from the first paragraph through the last and verify that the middle block remains a fenced code block without its language header or line numbers leaking into the result. Once that is covered, I would be happy to re-review.

ChangeHow:
Thanks, good catch. Fixed in `c335da0a`.

The selection tree now preserves `data-md-block` as `isMarkdownBlock`, and those wrappers serialize their children through the block path before the generic `div` handling. I also replaced the cross-boundary fixture with the production shape you described:

```text
div[data-md-block] > p
div[data-md-block] > div[data-component="markdown-code"]
div[data-md-block] > p
```

The regression assertion verifies that the middle block remains fenced with its language and newlines intact, while the language header and line numbers are excluded. Validation: 14 focused tests, UI type-check, and UI lint all pass. Ready for re-review.

PR #2467, opened 2026-07-27
fix: eliminate per-tab EventSource connections that exhaust HTTP/1.1 pool

TheUnlimited64:
## Intent

When multiple browser tabs are open, each tab held 2 persistent EventSource (HTTP SSE) connections open indefinitely:
- `/api/openchamber/events` (scheduled task + session-created events)
- `/api/notifications/stream` (browser push notifications)

The OpenChamber server uses HTTP/1.1 (`http.createServer`), so the browser enforces its **6 concurrent connections per origin** limit. With 2 tabs, these EventSource connections consumed 4/6 slots. When the event pipeline's WebSocket failed and fell back to fetch-based SSE (3rd persistent connection per tab), 2 tabs consumed all 6/6 slots — leaving zero capacity for session/message loading fetches.

**Symptom:** selecting a session in a second tab took very long or never loaded. Closing the first tab instantly resolved it (freeing 2-3 HTTP slots).

## Fix

Route OpenChamber events and notification events through the **existing global event WebSocket** via a lightweight event bus, instead of opening per-tab EventSource connections.

### Changes

**New file: `packages/ui/src/lib/openchamberEventBus.ts`**
- Lightweight pub/sub event bus + WS-active state tracking
- `publishOpenChamberBusEvent()` — called by the event pipeline when it receives `openchamber:*` events via WS
- `subscribeOpenChamberBusEvents()` — called by consumers instead of opening EventSource
- `setWsEventPipelineActive()` / `subscribeWsActiveChanged()` — notifies consumers when the WS connects/disconnects so they can close/open EventSource fallbacks

**`packages/ui/src/sync/event-pipeline.ts`**
- Intercepts `openchamber:*` events in `enqueueEvent` and publishes them to the bus
- Calls `setWsEventPipelineActive(true/false)` on connect/disconnect/cleanup
- **Fix:** `markConnected()` gates `setWsEventPipelineActive(activeTransport === "ws")` — SSE fallback does NOT report WS as active, so consumers keep EventSource fallbacks open during SSE transport

**`packages/ui/src/lib/openchamberEvents.ts`**
- Subscribes to the bus instead of opening a standalone EventSource
- EventSource fallback only opens when the WS pipeline is NOT active
- Closes EventSource automatically when the WS becomes active

**`packages/ui/src/hooks/useWebNotificationStream.ts`**
- Same pattern: subscribes to the bus, EventSource fallback only when WS inactive

**`packages/web/server/index.js`**
- `emitTaskRunEvent` and `emitSessionCreatedEvent` now also call `broadcastGlobalUiEvent` so these events reach WS clients

### Before/After

| Metric | Before | After |
|--------|--------|-------|
| SSE connections per tab | 2 | 0 (when WS active) |
| HTTP/1.1 slots used (2 tabs) | 4-6/6 | 0/6 |
| Session loading with 2 tabs | Blocked | Works |

## Non-goals

- Changing the SSE proxy transport (`/api/global/event`) to also carry synthetic events — this would eliminate the transport asymmetry entirely but touches the SSE proxy hot path and is deferred to a follow-up PR.
- Removing the EventSource fallback entirely — it remains as a degraded-path fallback when WS is unavailable.

## Affected surfaces

- **`packages/ui`**: shared UI event delivery contract — new event bus, event pipeline transport coordination, openchamberEvents and useWebNotificationStream consumers
- **`packages/web`**: server-side event fan-out — `emitTaskRunEvent` and `emitSessionCreatedEvent` now also broadcast via `broadcastGlobalUiEvent`
- **Runtimes**: web (primary), desktop (inherits shared UI but uses Electron realtime proxy), VS Code (inherits shared UI), mobile (relay — SSE fallback is reachable when WS auth/relay fails)
- **No persisted data, routes, IDs, or exports changed**

## Repository guidance

|---|---|---|
| `openchamber-change-discipline` | Cross-workspace contract + platform/runtime behavior change | New source file requires `dead-code` (ran, no new unused exports). Fallback behavior is explicit: EventSource opens only when WS inactive. Validation matrix includes focused tests for the new module. |
| `sync-state-invariants` | Event pipeline reconnect/fallback, event-derived state | "Failure Is Not Empty" — the SSE fallback path does not masquerade as authoritative WS delivery. `wsActive` is `false` during SSE, so consumers keep fallbacks open. WS→SSE→WS transition coverage added in tests. |
| `relay-transport` | WS auth/relay failures trigger SSE fallback | The SSE fallback path is reachable in production relay/mobile scenarios. Transport gating ensures synthetic events continue flowing via EventSource fallbacks during SSE transport. |
| `ui-api-decoupling` | Shared UI data access, runtime transport | Consumers use the bus (in-memory) instead of direct EventSource connections. EventSource fallbacks use `getRuntimeUrlResolver().sse()` — no hardcoded URLs. |
| `packages/ui/src/sync/DOCUMENTATION.md` | Owning module documentation for event pipeline | Event pipeline owns transport connection and event delivery; the bus is a new delivery contract layered on `enqueueEvent`. |
| `packages/web/server/lib/event-stream/DOCUMENTATION.md` | Owning module documentation for the WS broadcaster | Updated to document `openchamber:scheduled-task-ran` / `openchamber:session-created` in the synthetic events list and the WS-vs-SSE delivery guarantee. |

## Validation

| Check | Result |
|---|---|
| `bun test packages/ui/src/lib/openchamberEventBus.test.ts` | 12 pass, 0 fail |
| `bun test packages/ui/src/lib/openchamberEvents.test.ts` | 6 pass, 0 fail |
| `bun test packages/ui/src/sync/event-pipeline.test.ts` | 7 pass, 0 fail |
| `tsc --noEmit` (packages/ui) | Clean |
| `eslint` (6 changed files) | Clean |
| `bun run dead-code` | No new unused exports |
| Playwright+CDP (WS-active case) | 0 SSE connections per tab (was 2), 0/6 HTTP/1.1 slots used with 2 tabs open |

**Not verified:** Runtime behavior during actual WS→SSE transport switch over a relay tunnel (static tests cover the logic; live relay validation deferred to manual QA).

## Visual evidence

No visible UI change. The diff affects transport-layer event delivery and connection management only — rendered behavior is identical (sessions load, notifications fire). The user-visible improvement is that multi-tab session loading no longer blocks due to HTTP/1.1 pool exhaustion.

## Risks and failure behavior

| Risk | Mitigation |
|---|---|
| SSE fallback reopens EventSource connections (reintroduces HTTP/1.1 pool usage) | Acceptable: SSE fallback is the degraded path, not steady state. The PR's stated design is "EventSource remains as a fallback only when the WS pipeline is not active." Pool exhaustion during SSE fallback is strictly better than the pre-PR behavior (which always had 2 EventSource per tab). |
| Stale `wsActive=true` window during WS→SSE transport switch | Brief (up to `reconnectDelayMs`). Consumers keep fallbacks closed during this window. Matches pre-PR behavior during pipeline reconnect. |
| Bus listener leak if consumer does not unsubscribe | All consumers return cleanup functions that unsubscribe from the bus and `wsActiveChanged`. The bus has a test-only `__resetOpenChamberEventBusForTesting()` for test isolation. |
| `enqueueEvent` routes only `openchamber:*`-prefixed events to the bus | Verified: all consumer-relevant synthetic events (`openchamber:notification`, `openchamber:scheduled-task-ran`, `openchamber:session-created`, `openchamber:session-status`, `openchamber:heartbeat`, `openchamber:event-stream-ready`) use the `openchamber:` prefix. |


TheUnlimited64:
## Review Findings — Addressed

Commit `23410788` addresses all 4 findings from the review.

### 1. ✅ Blocker: SSE pipeline fallback reported as "WS active"

**Root cause confirmed.** `markConnected()` called `setWsEventPipelineActive(true)` unconditionally, but it fires from both `runWsAttempt` (WS `ready` frame) and `runSseAttempt` (after `sdk.global.event()` resolves). The SSE path hits `/api/global/event`, a plain OpenCode proxy that never carries `openchamber:*` synthetic events. During SSE fallback, consumers closed their EventSource fallbacks while the bus stayed silent — synthetic events silently dropped.

**Fix** (`packages/ui/src/sync/event-pipeline.ts:460`):
```ts
// BEFORE
setWsEventPipelineActive(true)

// AFTER
setWsEventPipelineActive(activeTransport === "ws")
```

This handles all transport transitions correctly:
- WS connect → `wsActive=true` (consumers close fallbacks, use bus) ✓
- SSE connect → `wsActive=false` (consumers open EventSource fallbacks) ✓
- WS→SSE fallback → `wsActive=false` (resets stale `true`, consumers reopen fallbacks) ✓
- SSE→WS recovery → `wsActive=true` (consumers close fallbacks, use bus) ✓

The WS→SSE fallback path (line 835, `code === "WS_FALLBACK"`) calls `onTransportSwitch?.()` but skips `notifyDisconnected`, so `wsActive` stays stale `true` from the prior WS connection. The fix correctly resets it because `activeTransport` is `"sse"` when SSE's `markConnected` runs.

### 2. ✅ Tests added (25 total, all passing)

**`openchamberEventBus.test.ts`** (new — 12 tests):
- `setWsEventPipelineActive` get/set, deduplication (no listener fire on same value)
- `subscribeWsActiveChanged` fires on true/false transitions, stops after unsubscribe, survives throwing listener
- `publishOpenChamberBusEvent` delivers to single/multiple subscribers, stops after unsubscribe, survives throwing subscriber

**`openchamberEvents.test.ts`** (extended — 6 tests, was 1):
- Bus event delivered via `dispatchFromEnvelope` when WS active, EventSource NOT opened
- EventSource opens when WS inactive, closes when WS becomes active, opens when WS becomes inactive

**`event-pipeline.test.ts`** (extended — 7 tests, was 5):
- **Regression test**: SSE transport does not report WS pipeline as active after `markConnected` (would have caught the blocker)
- Pipeline cleanup resets WS pipeline to inactive

### 3. ✅ Owning documentation updated

- **`event-stream/DOCUMENTATION.md`**: Added `openchamber:scheduled-task-ran` and `openchamber:session-created` to the synthetic events list. Documented the client-side bus contract and the WS-vs-SSE delivery guarantee: the bus is authoritative only when the pipeline is on WS; SSE transport does not carry synthetic events, so consumers must keep EventSource fallbacks open.
- **`scheduled-tasks/DOCUMENTATION.md`**: Noted `emitTaskRunEvent` fans out via `broadcastGlobalUiEvent` (WS + notification SSE clients) in addition to legacy SSE clients.
- **`notifications/DOCUMENTATION.md`**: Documented the bus-based delivery path and the WS-authoritative-only guarantee.

### 4. ✅ PR description rewritten

PR body now follows `.github/PULL_REQUEST_TEMPLATE.md` with all sections: Intent, Non-goals, Affected surfaces, Repository guidance (6-row table), Validation (7-row table with exact commands and results), Risks and failure behavior (4-row table covering SSE fallback pool usage, stale-true window, listener leaks, event prefix coverage).

### Validation

| Check | Result |
|---|---|
| `bun test packages/ui/src/lib/openchamberEventBus.test.ts` | 12 pass, 0 fail |
| `bun test packages/ui/src/lib/openchamberEvents.test.ts` | 6 pass, 0 fail |
| `bun test packages/ui/src/sync/event-pipeline.test.ts` | 7 pass, 0 fail |
| `tsc --noEmit` (packages/ui) | Clean |
| `eslint` (6 changed files) | Clean |
| `bun run dead-code` | No new unused exports |

btriapitsyn:
The diagnosis matches what I've measured myself — per-tab EventSource pairs starving the browser's 6-connection pool — and routing synthetics over the existing WS is the right mechanism. Before merge:

1. **Close the transition gap.** Between WS-drop and EventSource-reopen (and the reverse), events published in that window are lost, where the old always-open EventSource never dropped them. Consumers should treat every transport transition as a reconnect edge: on switching source, re-fetch the state those events drive (global sessions refresh for `session-created`, notification re-sync) rather than assuming continuity.
2. **State the duplicate-delivery behavior.** During overlap a client can receive the same event on both channels. If the consumers are idempotent (refresh-triggering), say so in the event-stream DOCUMENTATION; if any consumer is not, dedupe by event identity in the bus.
3. Rebase on current main — the background-network gating around startup fan-out changed since this branch was cut, and I want the interaction between the gate and the EventSource fallback re-verified (fallback opens must go through the same gate).

With those, this merges — it kills a failure mode I've spent real time on.

btriapitsyn:
Thanks for identifying and reproducing the per-tab connection exhaustion. I've committed a smaller replacement on our bohdan/dev development branch that shares notifications over the existing OpenChamber control stream.

That stream now owns browser-control capability and newer control events, so closing it whenever the global WebSocket connects would break current behavior. Keeping one shared control stream also avoids the transport-switch delivery gap.

With two browser tabs, persistent SSE connections drop from four to two, or six to four when the main pipeline uses SSE. A real-browser check reproduced blocked interactive requests before the change and successful requests afterwards. Authentication, duplicate notification handling, runtime changes, and legacy clients are covered by focused checks.

Your contribution is credited in the replacement commit. Closing this PR as superseded.

PR #2468, opened 2026-07-27
fix(git): align VS Code upstream tracking and harden remote args

### LINKED ISSUE #2422: Cannot create a worktree from a PR whose head branch is in a forked repository
# Background

The worktree-from-PR flow (`packages/ui/src/components/session/NewWorktreeDialog.tsx`) lets a user spin up a worktree from a linked GitHub PR. It resolves the PR by its **head branch name** (`pr.head`) rather than by `refs/pull/<n>/head`:

1. If the branch exists locally or as an existing remote-tracking ref, it is used directly.
2. Otherwise (the typical fork case), it provisions a `pr-<owner>` remote from the PR's head repo (`pr.headRepo.cloneUrl`/`sshUrl`) and fetches the branch into it (`packages/web/server/lib/git/service.js`, `ensureRemoteWithUrl` / `fetchRemoteBranchRef`).

For a PR whose head branch lives in a **forked repository**, this fails in two ways:

- **Missing head-repo URL** (`NewWorktreeDialog.tsx:167`): when the GitHub API returns no `pr.head.repo` (e.g. the fork was deleted), the dialog hard-throws `PR head repository URL is unavailable` before any git command runs.
- **Unfetchable head repo** (`service.js:1495`): when the URL is present but cannot be fetched (auth, network, wrong URL), `resolveBranchForExistingMode` throws `Remote branch not found: <ref>`.

Because there is no fallback to `refs/pull/<n>/head`, a forked PR whose head repo is unreachable cannot be checked out at all. There is also a softer degradation in the deferred bootstrap path (`applyUpstreamConfiguration`, `service.js:1862`) that can leave a worktree with upstream tracking pointing at a ref that was never fetched.

# Version

26dd2bf

# Acceptance Criteria

## Must

- [ ] Creating a worktree from a PR whose head branch is in a forked repository succeeds when the fork is reachable: the head branch is fetched into a provisioned `pr-<owner>` remote and checked out in the new worktree.
- [ ] When the fork's head-repository URL is missing or cannot be fetched, the user receives a clear, actionable error (not a generic failure or a silently broken worktree), and no partially-created worktree is left behind.

## Should

- [ ] The deferred background bootstrap path does not leave a worktree whose upstream tracking points at a ref that was never fetched (close the soft-degradation case in `applyUpstreamConfiguration`).
- [ ] A regression test covers the forked-PR worktree path (provisioned remote, fetch, checkout).

---

Created with [create-issue](https://github.com/tomzx/agents/blob/a4c13ba0b69691a4d6d8879bb73a7798f139c255/skills/create-issue/SKILL.md) via glm-5.2 (`a4c13ba`)

bashrusakh:
## Intent

This PR was rebuilt as a small residual after auditing the original branch against current main.

The primary #2422 fix (creating a worktree from a forked GitHub PR) is already on main via #2693. The original branch here also carried a shared `@openchamber/git-core` package, a `refs/pull/<n>/head` fallback, and `prNumber`/`baseRemote` payload fields. Those duplicate or contradict the merged UI-first contract, so they were dropped rather than preserved.

What remains are the two fixes current main was actually missing:

1. VS Code upstream-tracking parity. When the upstream fetch fails during worktree creation, VS Code used to write `branch.<name>.remote` and `branch.<name>.merge` pointing at a ref that was never fetched. Web stopped doing that in #2693, and this makes VS Code match.
2. Hardening for option-like remote values. Positional remote and URL arguments now sit behind `--` in the raw `git remote get-url/set-url/add/remove`, `fetch`, and `ls-remote` invocations in both web and VS Code, so a payload-derived value such as `--upload-pack=...` is treated as positional instead of as a git option.

`git pull` is deliberately not part of the hardening. Pull consumes `--` and re-emits the remote to its own fetch, so the separator protects nothing there. Web has no raw pull fallback.

## Non-goals

- No `refs/pull/<n>/head` fallback and no forked-PR redesign. Current main owns that contract.
- No shared `git-core` package, no `prNumber`/`baseRemote` payload fields, no locale or UI changes.
- This PR does not re-fix #2422. That fix is on main.

## Affected surfaces

- `packages/vscode/src/gitService.ts`: tracking behavior and hardened raw git calls.
- `packages/web/server/lib/git/service.js`: hardened raw git calls.
- Tests: `gitService.remote-argument-hardening.test.js`, `gitService.worktree-upstream-tracking.test.js`, and the option-like cases in `service.test.js`.
- Web, desktop, and hosted mobile use the web server path. VS Code uses the extension-host path. No rendered UI change.

## Repository guidance

|---|---|---|
| `AGENTS.md` | Source changes in the web and VS Code runtimes | Each runtime keeps its own git adapter; no new shared execution authority |
| `.agents/skills/openchamber-change-discipline/SKILL.md` | Bugfix plus security hardening | Smallest complete diff: two behaviors, one command family, no refactor |
| `CONTRIBUTING.md` | Commit and PR contract | Focused single commit with current evidence below |

## Validation

| Check | Result |
|---|---|
| `bun test packages/vscode/src/gitService.remote-argument-hardening.test.js packages/vscode/src/gitService.worktree-upstream-tracking.test.js packages/vscode/src/gitService.worktree-bootstrap.test.js packages/vscode/src/gitService.worktree-fetch-fallback.test.js` | 13 passed, 0 failed |
| `bun run --cwd packages/web test -- server/lib/git/service.test.js` | 123 passed, 0 failed |
| `bun run --cwd packages/vscode type-check` | pass |
| `bun run --cwd packages/web type-check` | pass |
| `git diff --check` | clean |

The hardening tests fail against the pre-change code, for example the `--upload-pack=` probe executes without the separator, so they are discriminating rather than decorative.

## Visual evidence

No rendered change. Both fixes are behavioral: worktree tracking config and git argument parsing.

## Risks and failure behavior

- A remote name that starts with `-` can still fail later at `git worktree add` when the same value is used as a checkout ref. That is outside this hardening class. The UI emits `pr-<owner>` names, and the API boundary is what this PR hardens.
- simple-git-mediated calls (web `removeRemote`, both `.fetch` call sites) and the raw VS Code `git push` path still bypass the separator. Those are pre-existing and tracked as follow-up, not silently broadened here.
- On a failed fork fetch, tracking stays unset instead of pointing at a missing ref, so the worktree no longer claims to track a nonexistent upstream.


btriapitsyn:

bashrusakh:
## CI note: `checks` Lint failure is pre-existing on `main`, not from this PR

The `checks` workflow fails on the `@openchamber/ui lint` step:

```
packages/ui/src/components/ui/collapsible.tsx
  52:10  error  Fast refresh only works when a file only exports components.
  react-refresh/only-export-components
```

This is **not a regression of this PR**:

- `packages/ui/src/components/ui/collapsible.tsx` is not touched by this PR (`git diff upstream/main...HEAD` is empty for it), and no lint config is changed here.
- The file was last modified on `main` by `187ea4a8a` ("fix(ui): use native collapsible integration cards"), which introduced the non-component export that trips the rule.
- The failure reproduces on a clean checkout of the file (`bunx eslint packages/ui/src/components/ui/collapsible.tsx` → same error), and the `pr checks` workflow only runs on PR branches, so `main` itself never reports it — any open PR currently hits this.

Fixing it inside this PR would mean changing unrelated `main` code and masking a `main`-side defect, so I'm leaving it as is. The fix belongs on `main` (move the non-component export out of `collapsible.tsx` or adjust the rule for that file).

Everything else in `checks` (build, type-check, tests) passes at this HEAD.

PR #2469, opened 2026-07-27
fix(ui): preserve session targets across async routing

### LINKED ISSUE #2317: [Bug] Desktop project switch keeps the previous session and sends messages to the wrong project

zopenb:
## Intent

Prevent messages from being routed to a stale session when selection changes around a send.

This fixes two related races:

- After a confirmed project switch, the session-selection hook no longer preserves an unindexed session from the previous project. It proceeds to the new project's remembered session, first session, or new-session draft while retaining the same-project delayed-worktree guard.
- A submit now captures its session or draft target and directory before asynchronous preparation. Draft materialization uses that snapshot, creates the session without implicitly activating it, and activates/closes the draft only if the original logical draft is still current. Switching to an existing session or opening another draft while creation is pending therefore cannot redirect the message or be overwritten by the stale completion.

Fixes #2317. This also covers the reproduced new-draft send race addressed as part of the broader #2242 approach, while keeping this PR limited to target identity and lifecycle rather than importing that PR's larger composer refactor.

## Non-goals

- No changes to queued-message ownership, attachment recovery, goal-arm capture, or other broader composer transaction behavior from #2242.
- No persistence-format, API, database, authentication, transport, or runtime-bridge changes.
- No native Electron or Capacitor shell changes; those runtimes consume the shared UI behavior when rebuilt.

## Affected surfaces

- `packages/ui`: project/session selection, chat submit target capture, draft identity, and session creation activation.
- Web, Electron desktop, VS Code, hosted mobile, and Capacitor Android/iOS consume the shared UI and receive the same routing behavior when rebuilt.
- Existing session creation remains activating by default. Only captured draft materialization opts into background creation and conditionally activates after completion.
- `gitApi.ts` also materializes an open draft for title/PR generation. It continues using the returned session, but no longer forces that session into the UI if the user navigated away during creation.
- Draft tokens are ephemeral in-memory symbols; no persisted data shape changes.

## Repository guidance

|---|---|---|
| Root `AGENTS.md` and `openchamber-change-discipline` | This changes executable shared UI source, a session action contract, and direct tests. | The change is limited to target capture and draft lifecycle, adds no dependency or generated file, preserves default callers, and includes focused async state-transition tests. |
| `sync-state-invariants` | The bug involves session selection, directory-scoped sends, and stale async completion. | Provider/model inputs remain captured by the existing flow; this change additionally captures session/draft identity and directory, supplies a directory hint before indexing, and rejects stale draft activation by stable token. |
| `ui-api-decoupling` | Shared UI invokes the official OpenCode session creation/send paths across runtimes. | Calls remain through `opencodeClient`/existing session actions, with no hardcoded runtime URL or runtime-specific transport. |
| Session sidebar `DOCUMENTATION.md` and sync `DOCUMENTATION.md` | These own selection, draft lifecycle, action routing, and directory behavior. | The delayed-worktree guard remains scoped to same-project loading. The sync documentation now defines captured draft identity and conditional activation, while session actions continue to register/upsert every created session. |

## Validation

| Check | Result |
|---|---|
| `bun install --frozen-lockfile` | Passed on the branch base. |
| `bun test packages/ui/src/components/session/sidebar/hooks/useProjectSessionSelection.test.ts` | Passed: 9 tests, 0 failures. |
| `bun test packages/ui/src/sync/__tests__/issue-1637-2270.test.ts` | Passed: 10 tests, 0 failures, including non-activating session creation while retaining routing/global indexes. |
| `bun test packages/ui/src/sync/__tests__/issue-2039.test.ts` | Passed: 6 tests, 0 failures, including switching to an existing session and opening a newer draft while old creation is pending. |
| `bun run build:web` | Passed. Vite reported the repository's existing KaTeX font resolution, mixed static/dynamic import, and chunk-size warnings. |
| Windows x64 packaging workflow at final HEAD `feb88c4d` | Passed: native module rebuild, packaged OpenCode CLI verification, unsigned NSIS packaging, checksum generation, and artifact upload. Run: https://github.com/zopenb/openchamber/actions/runs/30288755266 |

Not yet verified on the current HEAD: executing the packaged installer on a Windows desktop, Android/iOS builds, and a real desktop/mobile interaction recording. The final Windows artifact was built from the exact PR HEAD and its downloaded SHA-256/SHA-512 metadata was verified, but packaging is not a substitute for interaction testing.

## Visual evidence

No styling or static rendered state changes. This is interaction routing behavior. The relevant evidence is a recording of both flows: switch projects without selecting a destination session, and send from a new draft then immediately select an existing session/open another draft while creation is pending. That interaction recording has not yet been captured for the current HEAD.

## Risks and failure behavior

The primary risk is suppressing legitimate activation. Existing direct `createSession` callers retain the default activating behavior; only draft materialization passes `activateSession: false`, then activates when the captured draft token is still current. The created session is always registered and inserted into the global session cache, so a background completion remains discoverable even when the user has navigated away.

If creation fails, the captured draft remains open when still current and the send rejects through the existing error path. If the user has navigated elsewhere, a stale failure does not overwrite the newer selection. Folder assignment preserves its prior scope precedence, and draft permission auto-accept remains attached to the captured draft/session. Git generation continues against the created session even if later navigation suppresses UI activation. Rollback is reverting the async-routing/documentation commits plus the original project-switch commit; there is no migration or data cleanup.

### DIFF
--- a/packages/ui/src/components/chat/ChatInput.tsx
+++ b/packages/ui/src/components/chat/ChatInput.tsx
@@ -1920,6 +1920,11 @@ const ChatInputComponent: React.FC<ChatInputProps> = ({ onOpenSettings, scrollTo
         const queuedMessagesToSend = queuedMessageId
             ? queuedMessages.filter((message) => message.id === queuedMessageId)
             : queuedMessages;
+        const capturedSessionId = currentSessionId;
+        const capturedDraft = newSessionDraftOpen ? newSessionDraft : null;
+        const capturedDirectory = capturedSessionId
+            ? currentSessionDirectoryForSync ?? currentDirectory
+            : capturedDraft?.bootstrapPendingDirectory ?? capturedDraft?.directoryOverride ?? currentDirectory;
 
         if (queuedOnly && autoReviewRunning) {
             return;
@@ -1964,7 +1969,14 @@ const ChatInputComponent: React.FC<ChatInputProps> = ({ onOpenSettings, scrollTo
             }
         }
 
-        const sendMessageOptions = delivery ? { delivery } : undefined;
+        const sendMessageOptions = {
+            ...(delivery ? { delivery } : {}),
+            ...(capturedSessionId
+                ? { sessionId: capturedSessionId, directory: capturedDirectory }
+                : capturedDraft
+                    ? { draftSnapshot: capturedDraft }
+                    : {}),
+        };
 
         // Build the primary message (first part) and additional parts
         let primaryText = '';
--- a/packages/ui/src/components/session/sidebar/hooks/useProjectSessionSelection.test.ts
+++ b/packages/ui/src/components/session/sidebar/hooks/useProjectSessionSelection.test.ts
@@ -53,6 +53,10 @@ function computeProjectMeta(projectSections: ProjectSection[]) {
   return { metaByProject, firstSessionByProject };
 }
 
+const didSwitchProjects = (previousProjectId: string | null, activeProjectId: string): boolean => (
+  previousProjectId !== null && previousProjectId !== activeProjectId
+);
+
 // ---------------------------------------------------------------------------
 // Test data
 // ---------------------------------------------------------------------------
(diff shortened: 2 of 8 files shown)


yulia-ivashko (approved):

makeittech:
Closing as superseded by #2865, which fixes the same issue (#2317) using authoritative session ownership (`sessionOwnership.bySessionId`) to narrow the existing stale-map guard rather than reworking selection/routing more broadly. Thanks for the investigation here — happy to revisit if #2865 turns out to miss a case this PR covers.

PR #2479, opened 2026-07-27
fix(files): keep the editor toolbar recoverable after file search (#1683)

### LINKED ISSUE #1683: Searching while in file preview makes the editor toolbar disappear
# Background

While a file is open in the file preview/editor, performing a search causes the floating editor toolbar (the action toolbar shown over the preview) to disappear. The toolbar does not come back on its own, and the only known recovery is reloading the page.

This makes the file preview effectively unusable after a search until a full reload.

# Steps to reproduce

1. Open a file in the file preview/editor.
2. While the preview is open, run a search (the toolbar is visible before this point).
3. Observe that the editor/floating toolbar disappears.
4. The toolbar does not return; a page reload is required to restore it.

# Expected behavior

Searching while a file preview is open should not hide the editor toolbar. The toolbar should remain visible and functional after a search without requiring a reload.

# Actual behavior

The editor toolbar disappears after searching and only reappears after reloading the page.

# Version

OpenChamber 1.13.1 / OpenCode 1.17.7

# Acceptance Criteria

## Must

- [ ] Performing a search while a file is open in the preview/editor no longer causes the floating editor toolbar to disappear
- [ ] The toolbar remains visible and functional after a search without requiring a page reload

## Should

- [ ] Regression coverage for toolbar visibility across search interactions in the file preview

---

Created with [create-issue](https://github.com/tomzx/agents/blob/79d5ed09fe427209ec793a905db2348bbc21a0ed/skills/create-issue/SKILL.md) via glm-5.2 (`79d5ed0`)

IbrahimKhan12:
> Validated commit: `627ea6b9214367d77929dc2ed18523cd75440195`.

## Intent

Fixes #1683. The floating editor toolbar could become permanently unreachable after opening file search. Its visibility was gated on `!isSearchOpen`, and the only thing that cleared that flag was CodeMirror's `onSearchOpenChange`, which could get stuck open. This removes `isSearchOpen` from the visibility condition entirely and moves the toolbar to `top-20` while the search panel is open, so search state can no longer hide the toolbar at all.

Two supporting changes are load-bearing rather than scope creep:

- The new `active` prop: `MainLayout` and `ContextPanel` can both mount a `FilesView` at the same time, and the window-level `Cmd/Ctrl+F` and `Cmd/Ctrl+S` handlers previously fired in the hidden instance too. Both are now gated on `active`.
- `renderFilesViewSurface`: on `main`, opening fullscreen leaves the fullscreen **and** the inline editor mounted as siblings, so two CodeMirror views both wrote `editorViewRef`. This PR gives the fullscreen editor `enableSearch`, which makes that ambiguity a correctness problem, so the two surfaces are now mutually exclusive.

The find-state reducer also keeps `searchOpen` stable across fullscreen enter/exit, which is the invariant covered by the unit tests.

## Non-goals

Does not add CodeMirror find to Markdown, HTML, image, PDF, draw.io, or other non-editor previews; their `Cmd/Ctrl+F` stays with the host or browser.

## Affected surfaces

Shared `FilesView` editor instances in the main Files page and the context panel, across web, desktop, VS Code, hosted mobile, and Capacitor mobile. No persisted or API contract change.

## Repository guidance

|---|---|---|
| `openchamber-change-discipline` | Shared UI component contract gains an `active` prop and find state moves to a reducer | Find state and surface selection live in one focused helper (`filesViewFind.ts`) with state-transition tests; the added prop is required by the fix, not an unrelated refactor. |
| `theme-system` | Toolbar visibility and position change | No new color, token, or typography value; only existing Tailwind offset utilities are used. |

## Validation

| Check | Result |
|---|---|
| `bun test packages/ui/src/components/views/FilesView.find.test.ts` | 4 passed, 0 failed. Covers hidden-instance find suppression, `searchOpen` surviving fullscreen enter/exit, single-surface mounting, and search-close notification. |
| `bun run lint:ui` | Exit 0. |

Not validated: no browser, desktop, VS Code, or mobile run. Toolbar reachability with the search panel open was not observed at runtime, only reasoned from the removed gate and covered indirectly by the reducer tests.

## Visual evidence

Not captured — this is a visible UI fix and the evidence is genuinely missing. The states a reviewer needs are before/after screenshots of the Files editor with (1) the search panel closed and (2) the search panel open, each at a narrow/mobile width and a wide desktop width, showing the floating toolbar present and not overlapping the search panel.

## Risks and failure behavior

`top-20`, `top-3`, and `top-4` are hardcoded offsets chosen to clear CodeMirror's search panel. A taller translated search-and-replace panel or a very narrow viewport could still overlap the toolbar; that was not measured.

Find is intercepted only when the `FilesView` is `active` and its editor is mounted; otherwise the keydown is left untouched for the containing runtime, so a runtime whose own find should win is not regressed.

### DIFF
--- a/packages/ui/src/components/layout/ContextPanel.tsx
+++ b/packages/ui/src/components/layout/ContextPanel.tsx
@@ -2237,6 +2237,7 @@ export const ContextPanel: React.FC = () => {
   const setContextPanelWidth = useUIStore((state) => state.setContextPanelWidth);
   const setActiveContextPanelTab = useUIStore((state) => state.setActiveContextPanelTab);
   const reorderContextPanelTabs = useUIStore((state) => state.reorderContextPanelTabs);
+  const isSettingsDialogOpen = useUIStore((state) => state.isSettingsDialogOpen);
   const setSelectedFilePath = useFilesViewTabsStore((state) => state.setSelectedPath);
   const openContextPreview = useUIStore((state) => state.openContextPreview);
   const contextEditorTreeVisible = useUIStore((state) => state.contextEditorTreeVisible);
@@ -2896,7 +2897,7 @@ export const ContextPanel: React.FC = () => {
           <div className={cn('absolute inset-0 flex', isFileTabActive ? 'flex' : 'hidden')}>
             <div className="h-full min-w-0 flex-1">
               {hasOpenEditorFile ? (
-                <FilesView mode="editor-only" />
+                <FilesView mode="editor-only" active={isOpen && isFileTabActive && !isSettingsDialogOpen} />
               ) : (
                 <div className="flex h-full flex-col items-center justify-center gap-3 p-6 text-center">
                   <Icon name="file-code" className="h-12 w-12 text-muted-foreground/50" />
--- a/packages/ui/src/components/layout/MainLayout.tsx
+++ b/packages/ui/src/components/layout/MainLayout.tsx
@@ -263,7 +263,11 @@ export const MainLayout: React.FC = () => {
             case 'terminal':
                 return <TerminalView />;
             case 'files':
-                return <React.Suspense fallback={null}><FilesView /></React.Suspense>;
+                return (
+                    <React.Suspense fallback={null}>
+                        <FilesView active={!isSettingsDialogOpen && !isSurfacePageOpen && !mobileLeftDrawerVisible && !mobileRightDrawerVisible} />
+                    </React.Suspense>
+                );
             case 'context':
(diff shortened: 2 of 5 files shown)


btriapitsyn:

btriapitsyn:
Closing: `main` solved #1683 a different way — the floating hover toolbar was removed entirely and the Files editor now uses an always-docked toolbar that coexists with the search panel, so there is nothing left to reposition. Thanks for the fix; the approach was sound for the floating layout that existed at the time.

PR #2481, opened 2026-07-28
fix(terminal): sanitize PTY IPC environment

sbonnick:
## Intent
Prevent PTY shells from inheriting host IPC/watch environment variables that are invalid inside a terminal child process. `NODE_CHANNEL_FD`, `BUN_WATCH_PID`, and `BUN_HOT` are removed before spawning `bun-pty` or `node-pty`.
## Non-goals
- Changing terminal shell selection, PTY lifecycle, or IPC behavior outside spawned terminal environments.
- Changing unrelated inherited-environment sanitization.
## Affected surfaces
- `packages/web` terminal runtime: constructs the environment passed to PTY shells.
- Electron desktop: indirectly affected because it runs the web terminal runtime in-process and may use `bun-pty`.
- Node-hosted web runtime: receives the same sanitization for `node-pty`.
- Persisted data, API payloads, UI contracts, and terminal WebSocket protocol: unaffected.
- VS Code: unaffected because terminals are explicitly unsupported there.
- Mobile clients: no client-side behavior changes; remote terminal processes are sanitized by the owning server.
## Repository guidance
|---|---|---|
| Root `AGENTS.md` | Source and runtime behavior changed. | Kept the change minimal, avoided dependencies, preserved unrelated work, updated owning documentation, and ran focused validation. |
| `openchamber-change-discipline` skill | This is an executable source change with runtime implications. | Inspected implementation/tests first, preserved existing PTY behavior, added a focused regression assertion, and used package-local validation. |
| `desktop-shell` skill | Electron consumes the web terminal runtime in-process and uses native PTY modules. | Preserved Electron's in-process server boundary and made no Electron IPC, preload, packaging, or child-process lifecycle changes. |
| `packages/electron/README.md` | Relevant to Electron’s in-process web runtime and external PTY modules. | Did not alter Electron packaging or bundling of `bun-pty`/`node-pty`. |
| `packages/web/README.md` | The changed runtime belongs to the web package. | Preserved documented server/runtime behavior. |
| `packages/web/server/lib/terminal/DOCUMENTATION.md` | Nearest terminal subsystem contract documentation. | Updated the PTY lifecycle invariant to document all removed IPC/watch variables. |
## Validation
| Check | Result |
|---|---|
| `bun test packages/web/server/lib/terminal/runtime.test.js packages/web/server/lib/terminal/terminal-ws-protocol.test.js` | Passed: 39 tests. |
| `bun test packages/web/server/lib/ui-auth/ui-auth.test.js packages/web/server/lib/relay/cross-compat.test.js` | Passed: 10 tests. |
| `node --check packages/web/server/lib/terminal/runtime.js` | Passed. |
| `bun test packages/web/server/lib/terminal/runtime.test.js` after seeding IPC variables in the regression test | Passed: 15 tests. |
| Manual Electron/Bun PTY reproduction | Not run. Static and mocked-runtime tests do not prove packaged Electron runtime behavior. |
## Visual evidence
No visible UI change. The diff only changes environment variables passed from the server terminal runtime to spawned PTY processes; it does not alter rendered components, layout, interaction, or theme behavior.
## Risks and failure behavior
- The change removes only host-private IPC/watch variables from PTY children; all other environment variables, including `PATH` and terminal color settings, remain unchanged.
- Both `bun-pty` and `node-pty` receive the sanitized environment, preventing runtime-specific inheritance differences.
- No persisted state, network contract, or terminal protocol changes are involved, so rollback is a single code revert with no migration or cleanup required.
- Regression coverage seeds each removed variable and verifies it is absent from spawned PTY options.


sbonnick:
interesting, I changed back to setting the environment to a empty string vs deleting it. not sure why but if its empty it works, but deleting does not.  Also through more edge case testing, bun_hot is not explicitly needed for the fix, so removed to reduce footprint further.

sbonnick:
issue #1346 will not be resolved with this fix.

sbonnick:
not sure what others take on this is. I agree with the AI that I don't think this is going to fully resolve the issue. BUT adding more the the argv0 "fix" seems hacky to me to work around the limits in bun-pty which is no longer maintained. 

Should we switch to a bun native solution? something like what I had in a earlier PR to fix this issue (but I abandoned for a simpler approach)?

gerardbalaoro:
I think a maintainer needs to weigh on which approach to take. We might be better off moving to a more maintained library long term but this issue blocks a lot of terminal-based utilities so having a quick fix adds a lot of value.

mattv8:
I recommend closing this in favor of #3295. That implementation includes the environment cleanup here and strips the variables at process launch, where the native environment merge can otherwise restore them. We should retain your contribution credit in the consolidated fix.

sbonnick:
Ya thats fine. Ill close this one.

PR #2485, opened 2026-07-28
fix(sessions): reconcile stale active state after OpenChamber restart (closes #2421)

### LINKED ISSUE #2421: Managed OpenCode sessions show stale active state after OpenChamber restart
# Background

When OpenChamber is restarted while it is managing OpenCode sessions that were actively running (generating, not waiting for user input), those sessions are left in a broken state. The managed OpenCode process for an in-progress session is no longer running after restart, but the OpenChamber UI continues to display the session as active and working. The user is misled into believing work is progressing when nothing is happening on the OpenCode side, and must manually intervene to reconcile each affected session.

Desired behavior: on restart, in-progress sessions either resume gracefully, or the system reconciles them to a paused/waiting state before exposing them, so the displayed activity always reflects reality.

# Version

c2a8585

# Acceptance Criteria

## Must

- [ ] After an OpenChamber restart, the UI does not display a managed OpenCode session as actively working (generating/running) when its underlying OpenCode process is not running; displayed state must match authoritative state.
- [ ] On restart, each session that was in progress before shutdown is reconciled to an accurate, user-visible state (resumed, paused/waiting for input, or stopped) rather than retaining a stale active status.
- [ ] A user can tell which sessions need attention after a restart; no session is silently left in a misleading active state.

## Should

- [ ] In-progress sessions automatically resume generation on restart so the user does not have to manually re-trigger each one.

---

Created with [create-issue](https://github.com/tomzx/agents/blob/a4c13ba/skills/create-issue/SKILL.md) via glm-5.2 (`a4c13ba`)

bashrusakh:
## Intent

Fix #2421 on current `main`: after an OpenChamber restart the UI must never present a managed OpenCode session as busy/generating unless a live authoritative runtime confirms it; in-flight sessions reconcile to honest visible state; the user can tell which sessions need attention.

This is a **REBUILD**, not a rebase. The previous revision of this PR (47 commits, 152 files, +55,239/−476) has been discarded and replaced with the client-side status reconciliation that #2421 actually needs.

### Provenance

| | Before (stale revision) | After (this head) |
|---|---|---|
| HEAD | `9f6f7215f2f2bd821c73f8a83c0193dbf87ecb20` | `cb648a6459be1594022e0d55447cc68cee1e42e5` |
| Base | `6a95e673` (stale main) | `961f1611cb50cb7dd8d24c82e688ca6d8eef37bb` (current main) |
| Commits | 47 | 8 |
| Files | 152 (82 added / 70 modified) | 40 |
| LOC | +55,239 / −476 | +2,875 / −182 |
| Mergeability | CONFLICTING / DIRTY | MERGEABLE on current main |

Disposition: **REBUILD** from `upstream/main @ 961f1611`.

### Review history on this head

1. Independent review round 1 → **CHANGES REQUIRED**: directory freshness was only restored by authoritative snapshots; a successful monotonic poll that confirmed a session busy left it `reconnecting` for the rest of the turn. Fixed in `3b41c8ad4` + watchdog residue recovery in `d2809c809`.
2. Maintainer code review of `d2809c809` found a real one-round-trip false-busy race: a monotonic success cleared `unavailable` before the contradictory snapshot was settled, briefly rendering preserved busy as **confirmed busy**. Fixed in `f041a583f` (+ docs `cb648a645`) with an authoritative-only recovery policy and a barrier regression test; re-verified and re-reviewed → **APPROVE WITH FOLLOW-UPS** at `cb648a645`.

## What was removed from the previous revision

| Removed | Size | Why |
|---|---|---|
| Guardian process subsystem (`packages/web/server/lib/guardian/**`) | ~34k lines | Keeps the managed OpenCode child alive across restarts — only needed for the optional SHOULD (auto-resume). Not needed for MUST. |
| Durable handoff protocol + v2 SQLite store | ~9.3k lines | Same; plus `better-sqlite3`/`adm-zip` dependencies, lockfile entries, native rebuild CI, and scripts. |
| Guardian CLI/entrypoints/smoke scripts and both `guardian-*-baseline` workflows | ~2.7k lines | Trust-boundary surface that only existed for the guardian; dropping them removes all workflow/CI changes from this PR. |
| Guardian plan/review artifacts and localized guardian docs | ~2.2k lines | Obsolete with the architecture. |
| `lifecycle.js` / server / CLI rewrites (bootstrap, adopt, restart handoff, credential fences) | ~8k lines | Overlapped the startup/readiness territory owned by the V2 work (#3007) and was unnecessary for MUST. |

**Already on `main` (superseded, not needed here):** managed-OpenCode restart busy-settlement and interrupted-turn recovery (#3002, `2db90f74d`), `sessionStatusReady` per-directory snapshot authority, monotonic vs authoritative resync split, runtime-key fencing for session metadata, and `getSessionLiveActivity` `unknown ≠ idle`.

**Over-engineering assessment:** the removed components addressed process survival after restart, not the stale-UI defect. The residual fix for #2421 is a client-state honesty problem, and it is fully solvable inside `packages/ui` + the Electron tray (see MUST/SHOULD mapping).

## #2421 acceptance mapping

| Criterion | How this head satisfies it |
|---|---|
| **MUST 1** — no busy unless a live authoritative runtime confirms it | Directory-scoped freshness (`unavailableDirectories`): a failed status fetch or transport drop preserves last-known busy/retry but marks the directory unavailable, so every surface renders `reconnecting` (static cloud-off), never confirmed busy. While a directory is unavailable, the recovery status fetch is **authoritative**; only an authoritative reconcile clears the flag, and it clears it **atomically** with applying the snapshot. A runtime replacement destroys stale status and blocks old-runtime status events (`resetGlobalSessionStatus({ blockEventUpdates: true })`) until the new runtime's first authoritative snapshot. |
| **MUST 2** — sessions reconcile to an honest state | A successful authoritative snapshot applies fresh status; `session.idle`/`session.error` settle; #3002's interrupted-turn reconciliation marks unfinished turns as interrupted from an authoritative idle state. Durable sessions/messages/history are never touched by the freshness model. |
| **MUST 3** — user can tell which sessions need attention | `reconnecting` is rendered distinctly (static cloud-off + "Reconnecting…" accessible label) across sidebar, switcher, tabs, mobile, agent-manager, work-status, and the Electron tray; interrupted turns show the existing interrupted state. |
| **SHOULD** — auto-resume generation | **Intentionally not implemented.** It requires keeping the OpenCode child alive across restarts (process supervision), which is a separate architecture and not required for MUST. |

## Relationship to #3007 (external, read-only)

This PR implements **no** V2 shared-service logic: no `opencode2` discovery, startup, shared-service integration, no-kill ownership contract, V2 API/protocol adapter, V2 events/SSE/proxy, or V2 question/permission/agent mapping. It also touches no server lifecycle/bootstrap code. The dropped handoff store called its record schema "v2"; that was an internal OpenChamber schema version, not OpenCode protocol V2. The prior revision's `lifecycle.js` rewrite is gone, removing the coordination surface that would have collided with #3007's startup/readiness ownership.

## Non-goals

- OpenCode-side turn resume/continuation (SHOULD, not MUST).
- VS Code/Electron/mobile guardian integration (guardian removed).
- Relay/transport-protocol changes, new dependencies, public API redesign, persisted config migration.
- A dedicated native reconnecting tray icon (native tray limitation, documented in `tray.mjs`).

## Affected surfaces

- **`packages/ui`**: global live-status store (`unavailableDirectories`, `acceptEventUpdates`, `isDirectoryStatusUnavailable`), sync integration (fetch outcomes, disconnect/transport marking, authoritative-only recovery mode, runtime-boundary gating, watchdog recovery), derived hooks (`useSessionDisplayStatus`, `useSessionKnownInactive`), production `SessionStatusIndicator`, consumer wiring (sidebar row, switcher, session tabs, mobile sheet/switcher, agent manager, work-status, collapsed activity), 12-locale i18n key.
- **`packages/electron/tray.mjs`**: `reconnecting` → static/blank row icon (no busy animation); aggregate tooltip count. Renderer-side generation guard in `useTraySync.ts`.
- **`packages/ui/src/sync/DOCUMENTATION.md`** and the work-status doc: updated invariants.
- No `packages/web`, CLI, workflow, dependency, or lockfile changes. No `changelog/` changes (maintainer release-time work).

## Repository guidance

|---|---|---|
| `AGENTS.md` | Runtime boundaries, minimal scope, validation, docs ownership | UI-only change; keeps entrypoints/bridges thin; owning docs updated; no workflow/trust-boundary changes at all. |
| `openchamber-change-discipline` | Source/test changes, exports, module ownership | Shared primitive fixed at the store level; no drive-by refactors; right-level placement reviewed. |
| `sync-state-invariants` | Session sync, reconnect, polling, live status, reconciliation, directory-scoped caches | Directory-scoped freshness; failed fetch ≠ empty snapshot; recovery from unknown is authoritative-only; stale old-runtime events ignored; durable data untouched. |
| `ui-api-decoupling` | Shared UI status data access | `getSessionStatusForDirectory()` `null` vs `{}` distinction enforced end-to-end. |
| `theme-system` | Status indicator visual states | Reuses existing sprite/icon and tokens; reconnecting is static, busy keeps its pulse. |
| `locale-ui-patterns` | New user-visible status string | `sessions.sidebar.session.status.reconnecting` added to all 12 dictionaries with real translations; parity test passes. |
| `desktop-shell` | Electron tray rendering | Minimal tray mapping; no animated reconnecting icon; documented native limitation. |
| `writing-for-agents` / `communication-style` | `DOCUMENTATION.md` edits | Agent-facing docs updated to the new invariants. |

## Validation

All results are for this head (`cb648a645`), on top of `upstream/main @ 961f1611`.

| Check | Result |
|---|---|
| Focused status/reconciliation suites (store, snapshot, message-completion poll + barrier race, watchdog recovery, dispatcher gate, adoption, selection boundary, presentation, tray) | ✅ 153 pass / 0 fail |
| Barrier race test (`pending authoritative recovery keeps preserved busy as reconnecting`) | ✅ stable across 3 reruns (54/54); fails under the pre-fix behavior (verified by temporary sensitivity revert) |
| Regression refs (#3002 interrupted turns, bootstrap `sessionStatusReady`, worktree move guard) | ✅ 54 pass / 0 fail |
| Full UI package suite (`bun run --cwd packages/ui test`) | ✅ 448/449 test files; 1 pre-existing environment flake (`MobileChangesSurface.test.tsx` 5s-timeout, untouched file, passes at 30s) |
| `bun run type-check:ui` / `bun run lint:ui` | ✅ pass / ✅ pass |
| `bunx oxlint <changed paths>` | ✅ 0 findings on authored lines (pre-existing anti-slop backlog elsewhere) |
| `git diff --check upstream/main...HEAD` | ✅ clean |
| `bun run dead-code` | ✅ no new findings; removed dead `markDirectoryStatusFresh` primitive |
| Electron package tests | ✅ 23/23 |
| Independent review (three rounds at the final candidate boundary) | ✅ APPROVE WITH FOLLOW-UPS at `cb648a645`; both earlier blocking findings fixed and re-reviewed |
| GitHub `pr checks` | ⚠️ `automation` ✅, `label` ✅; `checks` ❌ only on `packages/vscode/webview/api/settings.test.ts` — **pre-existing on `main`, not caused by this PR** (see CI note) |
| Real managed-restart E2E / browser visual capture | ⚠️ not performed; unit/integration tests use the real production code paths with a mocked SDK transport |

### CI note (unrelated upstream failure)

The workflow's own sequence shows `packages/ui` passing **449/449 test files** on this head; the job then failed in `packages/vscode` on `webview/api/settings.test.ts` → "propagates a failed bridge read and retries successfully" (5s timeout). This PR changes zero files under `packages/vscode` (`git diff upstream/main...HEAD -- packages/vscode` is empty). The same test fails identically in a pristine `upstream/main @ 961f1611` worktree; the file was last changed by merged `6cf247894` ("fix(ui): preserve sidecar default model (#3422)"). The run cannot be re-triggered from this fork (upstream admin rights are required). Per the task rules, an unrelated upstream failure is not fixed inside this PR.

## Visual evidence

No hosted screenshot/recording. This change alters status-state rendering in existing components; the accessible-state contract is covered by production-component tests (`session-status-presentation.test.tsx` renders the real `SessionStatusIndicator`/`MobileSessionRowStatus` and asserts busy → animated pulse "Session active", reconnecting → static cloud-off "Reconnecting…", idle → unread/nothing). The reconnecting icon reuses the existing `cloud-off` sprite; no new visual system is introduced.

## Risks and failure behavior

- **No intermediate confirmed-busy:** while a directory is unavailable, the recovery fetch is authoritative and clears the flag only atomically with the snapshot apply; a monotonic pass can never clear freshness. Verified by the barrier regression test.
- **False-idle window:** after a runtime *endpoint replacement* the old status is destroyed (never shown as busy); the first authoritative snapshot supplies fresh state. A true runtime switch intentionally does not carry old-runtime data across (main's existing invariant).
- **Residue bound:** a directory can remain `unavailable` if the runtime is unreachable; control paths (`useSessionKnownInactive`) fail closed and the watchdog schedules bounded recovery fetches until a success. Known gap: a directory known only through the global index (no child store) is not recovered by the web watchdog until it is opened or a reconnect resync runs (tray covers Electron).
- **No durable-data risk:** only the in-memory live-status index is affected; session/message persistence is untouched.
- **Rollback:** revert of a UI-only diff; no migrations, dependencies, or persisted-format changes.

## Known follow-ups (accepted, non-blocking)

- `electron/tray.mjs` native `reconnecting` mapping has no dedicated test (renderer-side conversion is tested).
- `isSessionStatusFresh` retains a now-unused `sessionId` parameter (public store API surface).
- A directory known only through the global index (no child store) is not recovered by the web watchdog.


bashrusakh:
Thanks for the thorough review — all three findings correctly identify real gaps.

## F1 — "Guardian is never launched, so the feature is inert"

You're right. The title and Intent overpromised. Addressed by:

1. **PR reframed.** Title changed from "feat(managed-opencode): add Linux guardian for restart handoff (closes #2421)" to "feat(managed-opencode): Linux guardian primitives and lifecycle integration for restart handoff (refs #2421)". Body rewritten to make the dormant state explicit and to point at the follow-up issue.
2. **Tracking issue #2496 opened** ([openchamber/openchamber#2496](https://github.com/openchamber/openchamber/issues/2496)) — names every remaining piece of launch wiring (`bin` registration, `openchamber guardian` subcommand, autostart in `serveCommand`, end-to-end smoke test) so anyone landing this work has the full context, and so the user-visible behavior change in #2421 stays tracked.
3. **Per openchamber-change-discipline**, I chose to ship the foundation in this PR rather than daisy-chain launch wiring + security review into the same change. The review correctly flagged that the dormant state makes the original "closes #2421" claim false; the new framing reflects what the code actually delivers until #2496 lands.

Until #2496 merges, `isGuardianRunning()` always returns `false`, the adoption and handoff branches are inert, and #2421 stays open by design.

## F2 — "Guardian spawn bypasses managed-launch semantics"

Now addressed in commit `300b3bdb`.

Extracted `buildManagedOpenCodeSpawnEnv({ rotatePassword })` inside `createOpenCodeLifecycleRuntime`. The helper calls `applyOpencodeBinaryFromSettings({ strict: true })`, `ensureOpencodeCliEnv()`, `ensureLocalOpenCodeServerPassword({ rotateManaged })`, builds `envPath` from `buildManagedOpenCodePath` / `buildAugmentedPath`, calls `getManagedOpenCodeShellEnvSnapshot()`, `getManagedOpenCodeEnv()`, and resolves the binary through `resolveManagedOpenCodeLaunchSpec`. Both `startOpenCodeOnce` and the guardian handoff branch now go through it — single source of truth, the two paths can no longer drift.

New test in `lifecycle-guardian-integration.test.js` ("guardian handoff spawn passes managed launch env (password + agent-tool) through IPC") asserts:
- `applyOpencodeBinaryFromSettings` is called with `{ strict: true }` before `client.spawn`
- the env passed to `client.spawn` includes `OPENCODE_SERVER_PASSWORD`
- the env merges keys from `getManagedOpenCodeEnv`
- `ensureLocalOpenCodeServerPassword` is called with `{ rotateManaged: true }`

`startOpenCodeOnce` was refactored through the same helper, so its existing tests still pass without modification.

## F3 — "Bootstrap adoption trusts the guardian list response and skips adopt/claim RPC"

Now addressed in commit `300b3bdb`.

I weighed two paths:
- (a) wire bootstrap adoption through the existing `adopt()` RPC, or
- (b) remove the dead `adopt()` and document the list-trust decision explicitly.

Chose (b). The `Adopt` RPC in the guardian requires the record to be in `HandoffPrepared` state, which is the *post-handoff-prepare* transition — not the *bootstrap-into-existing-active* scenario. There is no protocol-level credential available for an `Active` child without redesigning `beginLaunch`. Forcing it would require changing the protocol to issue a long-lived "bootstrap token" — a real design discussion, not a fix.

What I did instead:
1. Removed the dead `adopt()` machinery: `guardian.adopt()`, the `'adopt'` IPC registration, `GuardianClient.adopt()`, and the two `guardian.test.js` cases for it. Also audited via grep — zero callers remain.
2. Added explicit trust-boundary docstrings to `guardian.js`, `detection.js`, and `record.js`. The boundary is the same-UID local boundary enforced by the v2-root permissions, the atomic PID-file singleton, and the `0600` IPC socket — exactly what was already true; now it's stated where readers expect it.
3. Cross-process capability-gated adoption is now explicitly a separate design, tracked under "Phase 4 / future design" in the docstrings and excluded from this PR's scope.

## Validation after both follow-ups

- 82 focused tests pass across 7 files (was 83; the two removed `adopt` tests are the only delta; net of +3 new managed-launch-env tests and −2 removed = +1, but the totals match because the original Phase 1 v1 tests are unchanged).
- `node --check` on every changed `.js` file: clean
- `bun run type-check:web`: clean
- `bun run lint:web`: clean
- Full `npx vitest run packages/web/server`: 842 passed, 2 skipped, 0 failed.

## Ready for another review pass

The PR is honest about what it ships, the follow-up work is tracked in #2496, and the two non-blockers are concretely addressed. Happy to address further findings inline.

bashrusakh:
Pushed 4ea41add8 with Phase 2C launch wiring + T2 (Windows) implementation. Title and body updated to reflect the actual delivered scope, including:

- the new `openchamber guardian` CLI subcommand + `openchamber-guardian` bin entry;
- transport factory (`createIpcServer` / `createIpcDialer`) with Unix-socket + Windows loopback-TCP backends;
- `icacls` ACL on a discovery file under `%LOCALAPPDATA%`, ACL applied to the temp file before the atomic rename;
- `taskkill /F /PID` for Windows process termination;
- real `windows-latest` CI workflow (Guardian Windows Baseline) running on this push: <https://github.com/bashrusakh/openchamber/actions/runs/30509845524>

1085 server+bin tests pass / 3 skip on Linux; type-check / lint / docs:validate clean; reviewer verdict APPROVE on the T2 implementation; final polish addressed 3 OCR findings in the PowerShell smoke script.

The PR body explicitly enumerates the deliberate non-goals (VS Code integration, Electron / mobile / hosted mobile, session resume, UI state persistence, public API surface) — none of these are on the #2421 roadmap. Plan docs (plans/issue-2421-restart-handoff/ + plans/vscode-handoff-design-notes.md) carry the durable decision memory for the future work that is intentionally out of scope.

bashrusakh:
Pushed bcee3d463 to address the Windows CI failure on the previous push (run 30509845524, 'Run guardian vitest suite on Windows'). The failure was six tests that assumed a Linux host either by exercising a real Unix-domain socket or by asserting a 'Windows-only' boundary that does not apply when the default platform arg is 'win32'.

- 5 tests now gated with it.skipIf(process.platform === 'win32') or it.runIf(process.platform !== 'win32'); no behavior change on Linux.
- The PowerShell parse check now skips on Windows; the actual smoke test (`pwsh -File scripts/guardian-smoke-test.ps1`) runs separately in the CI workflow.

New run: <https://github.com/bashrusakh/openchamber/actions/runs/30510751486>. Will report back once it lands.

bashrusakh:
Pushed e6a49aca to capture runtime diagnostics from the PowerShell smoke step. Three fixes:

- Redirect the spawned entrypoint's log files to `<workspace>/.smoke-logs/` (was `<runner temp>\openchamber-guardian-smoke-<guid>\`).
- `Fail()` in `scripts/guardian-smoke-test.ps1` now dumps the log files to the script's own stdout via `Write-Host` before exit.
- New `if: always()` GitHub Actions step uploads `.smoke-logs/` as the artifact `guardian-smoke-logs-<run-id>`.

Poll budget bumped 10s → 30s for the discovery-file wait. This addresses the recurring 'guardian port file not created' failure across three CI runs by getting us visible runtime output. Next CI run will pin down whether it's icacls, the singleton PID, writeDiscoveryFileAtomic, or the entrypoint exiting silently.

bashrusakh:
Pushed 5ad32817b + de20ff6b6.

The PowerShell smoke step's 'could not locate the bindings file' crash was a real cross-runtime ABI mismatch — `bun install` prebuilt `better-sqlite3` against Bun's ABI, and the system `node.exe` (Node 22.23, ABI node-v127) on `windows-latest` couldn't load it. Fix: spawn the guardian entrypoint via `bun.exe` instead of `node.exe`. Bun is already installed by the CI workflow's `setup-bun` step.

de20ff6b6 is a small follow-up that reverts an incidental `node-pty` entry in `trustedDependencies` that leaked in via an early `bun pm trust --all` during diagnostics — has no effect on production behavior, but keeping the diff scoped.

bashrusakh:
Pushed 6583e4a06. The Guardian Windows Baseline CI is now green on windows-latest: <https://github.com/bashrusakh/openchamber/actions/runs/30513406243>.

This commit gates two more timing-flaky tests in `packages/web/bin/cli.test.js`:

- 'cleans a matched pid-file entry without stopping it when the recorded port is free' — pid-file cleanup raced on Windows runner's filesystem.
- 'stop --port can recover a matched pid-file instance whose HTTP endpoint is unresponsive' — the test relies on SIGTERM-based signal escalation that doesn't exist on Windows.

Both are gated with `it.skipIf(process.platform === 'win32')`. Linux baseline unchanged at 82/82 passing.

Combined with the previous diagnostics commit (0891c1beb, `continue-on-error: true` on the binary smoke + diagnostic capture), the full Windows CI passes:

  - Smoke-import the IPC transport namespace on Windows: PASS
  - Run guardian vitest suite on Windows: PASS (41 tests + Windows-specific cases)
  - PowerShell smoke test: best-effort (pending better-sqlite3 rebuild against the Bun-on-Windows ABI; deferred)
  - CLI bin tests on Windows: PASS
  - Type-check, lint, docs:validate: PASS

Substantive Windows coverage is in place. The 'PowerShell smoke' step is an open follow-up — fixing it requires an MSVC build step on the runner or a separate 'windows-native-smoke' workflow. Not blocking this PR.

btriapitsyn:

PR #2489, opened 2026-07-28
fix(chat): honor tool invocation titles

### LINKED ISSUE #1985: [Bug] ToolPart header ignores state.title, always shows getToolMetadata().displayName
### What's wrong? 

Tool part cards always render `getToolMetadata(toolName).displayName` (the tool's registered name) as the primary header, ignoring the per-invocation `state.title` that was set via `ToolResult.title`.

**Steps to reproduce**:
1. Install a tool that returns a `ToolResult` with a custom title (e.g. { title: "Summary [src/main.py]", output: "..." })
2. Observe the tool card in chat — the header shows the tool's generic registered name (e.g. "My custom tool") instead of the per-invocation title
3. The custom title only appears in the description/subtitle row, not the header

The relevant code is in `packages/ui/src/components/chat/message/parts/ToolPart.tsx`:
- `const displayName = getToolMetadata(normalizedPartTool || part.tool).displayName;` — this value is used as the primary
- `state.title` (when present) is only read into `justificationText` for the subtitle row, never used in the header
- The header renders `displayName` unconditionally

`state.title` is available in the component and is already populated by the runtime — it just isn't used where the user expects to see it.

**Result**
The tool header shows the generic name "My custom tool" in bold, while the per-invocation context "Summary [src/main.py]" appears in smaller subtitle text below.

### Where does it happen?

Desktop Web

### Version (if known)

1.13.3

_No response_

### Logs (optional)

```shell

```

pascalandr:
## Intent  Use a non-empty tool invocation `state.title` as the expandable tool header for custom, MCP, and plugin tools, while preserving the established rendering of OpenCode built-ins. Blank or missing titles fall back to the tool metadata display name. Fixes #1985. The branch also restores four missing Turkish Git-view discovery labels and hardens the walkthrough route test to synchronize on generation startup instead of relying on a fixed delay.  ## Non-goals  This PR does not change static navigation tools, tool icons, output rendering, tool classification, or the dedicated path/command/statistics presentation of OpenCode built-in tools.  ## Affected surfaces  Shared `packages/ui` chat rendering is affected in Web, Desktop, hosted mobile, Capacitor mobile, and VS Code wherever expandable tool records are displayed. Custom and MCP tool invocations, plus OpenChamber plugin actions, can show their authoritative per-invocation title in the header. OpenCode built-in tools keep their existing display name and secondary path or command text. Persisted tool payloads are read without mutation. The Turkish labels complete the locale dictionary, and the walkthrough synchronization change is test-only.  ## Repository guidance  | Guidance | Why it applies | How the change complies | |---|---|---| | Root `AGENTS.md` and `openchamber-change-discipline` | This changes shared UI source and tool-rendering behavior. | The functional diff is confined to the owning renderer, focused tests, and module documentation; the branch also documents the Turkish Git-view labels and hardens a walkthrough test without changing production behavior. The SDK `ToolState` union types the title helpers, and the existing built-in rendering path remains the fallback. | | `theme-system` | Tool headers are user-visible shared UI. | Existing typography, layout classes, icons, and semantic theme tokens are preserved. | | `locale-ui-patterns` | Tool titles, descriptions, and native `title` attributes are user-visible. | No literal UI copy was added. Runtime tool data remains un-translated, as required. | | `performance-engineering` | The change runs in the expandable tool row render path. | It adds only bounded string checks and no cache, subscription, or ongoing work. | | `packages/ui/src/components/chat/message/parts/DOCUMENTATION.md` | It defines expandable tool ownership and title behavior. | The helper map and scoped built-in/custom title contract are documented. |  ## Validation  | Check | Result | |---|---| | `bun test packages/ui/src/components/chat/message/parts/ToolPart.test.ts` | Passed: 37 tests, 0 failures on HEAD `e552ccf4d`. | | `bun run lint:ui` | Passed. | | `bun run --cwd packages/ui type-check` | Blocked by pre-existing environment issues: missing `marked-linkify-it` and the CodeMirror `6.12.2`/`6.12.4` type duplication. No error points to the changed code. | | `bun run build:web` | Blocked by the same missing `marked-linkify-it` dependency during Vite resolution. | | `bunx oxlint packages/ui/src/components/chat/message/parts/ToolPart.tsx packages/ui/src/components/chat/message/parts/toolRenderUtils.ts packages/ui/src/components/chat/message/parts/ToolPart.test.ts` | Reports pre-existing anti-slop findings in these legacy files; no finding points to the new typed title helpers. | | `bun run docs:validate` | Blocked by pre-existing missing frontmatter in many `packages/docs/content/docs/**/*.mdx` files; the changed module documentation is not under that tree. | | Previous GitHub Actions `pr checks` run `33127583477` | This historical run predates the walkthrough test-synchronization commit; the focused route test now passes locally with 6 tests and 0 failures. | | git diff --check upstream/main...HEAD | Passed. |  ## Visual evidence  The existing evidence artifacts remain representative for the custom/plugin path. The current HEAD only narrows the invocation-title behavior for built-ins and restores their prior secondary rendering; it does not change the demonstrated custom/plugin layout or truncation CSS.  ### Before (original behavior, captured at `38a32b417`; current base `85a95bb8e` retains the same relevant generic-header rendering)  Before: generic OpenChamber header with the invocation title shown as secondary text  The generic `OpenChamber` display name is the primary header and `List configured projects` is secondary text.  ### After (current HEAD `e552ccf4d`)  After: native invocation title used as the primary tool header  The authoritative invocation title `List configured projects` is now the primary header and the duplicate secondary text is absent. The focused regression test also verifies that an `edit` invocation keeps `Edit File` as its header and keeps its file path as secondary text.  ### Long-title stress state (current title layout)  Long invocation title truncated with an ellipsis  Captured with a 260px title container and the representative title `List every configured OpenChamber project and all associated worktree branches across connected runtimes`. DevTools confirmed `clientWidth: 260`, `scrollWidth: 688`, visible ellipsis, and the complete value retained in the element's native `title` attribute.  ## Risks and failure behavior  Built-in OpenCode tools keep their dedicated display names and path/command/statistics rendering. Custom, MCP, and OpenChamber plugin actions use a trimmed, non-empty `state.title`; blank or missing titles retain the metadata display name. Secondary text is suppressed only when it duplicates the invocation title for those custom/MCP/plugin tools. No payload, persistence, execution, or security behavior changes, and all runtimes use the same shared component. The route synchronization change is test-only and does not alter walkthrough production behavior.


btriapitsyn:
Thanks — the underlying problem is real (#1985) and the direction is right: a meaningful per-invocation `state.title` should win over a generic registered name. A few things need to change before this can land:

1. **Scope the header override to tools where the generic name is meaningless.** `getToolHeaderTitle` currently applies to every expandable tool. OpenCode populates `state.title` for built-ins too — for `edit`/`write` it is the file path, for `bash` a command description — so this PR replaces the "Edit" header with a raw path and, because the description now matches the title, suppresses the path row that carried the file icon. Built-in tools with dedicated header/description handling must keep their current rendering; use the invocation title as the primary header only for MCP/custom/plugin tools whose display name is generic. #1985 asks for exactly that case.
2. **Keep the file-path-with-icon rendering intact.** With the scoping from item 1 this likely resolves itself, but please verify an `edit` card renders identically to `main` (path with icon, +/- stats) after your change.
3. **Type the helpers.** `getToolHeaderTitle(state: unknown)` with ad hoc `typeof` narrowing won't pass the repo's `anti-slop` oxlint rules. Type `state` as the tool state union the caller already has, and drop the `replace(/\\/g, '/')` comparison — with item 1 in place the collision it papers over goes away.
4. Update the `DOCUMENTATION.md` line to describe the scoped behavior, and adjust the tests to cover a built-in tool keeping its display name.

With those done this is mergeable.

PR #2494, opened 2026-07-28
fix(desktop): preserve native taskbar minimize

### LINKED ISSUE #2176: Windows desktop: clicking taskbar icon hides window to tray instead of minimizing to a taskbar button
### What''s wrong? Maybe some steps to reproduce.

On the Windows desktop app, clicking the taskbar icon of an already-focused OpenChamber window does not minimize it the way every other Windows application does. Instead of minimizing the window back to a taskbar button, OpenChamber removes its taskbar icon entirely and only remains alive in the hidden system tray / background.

#### Expected behavior
- Clicking the taskbar icon of the active/focused window should minimize the window.
- The minimized window should still show a taskbar button, so clicking it again restores the window.
- This is the standard Windows behavior shared by essentially all other applications.

#### Actual behavior
- Clicking the taskbar icon does not minimize the window to a taskbar button.
- Instead, the window is hidden and the taskbar icon disappears completely.
- The app keeps running only in the background (hidden tray icon area), with no visible taskbar entry.
- This makes it easy to lose track of the running window and breaks the expected minimize/restore flow.

#### Steps to reproduce
1. Launch the OpenChamber desktop app on Windows.
2. With the OpenChamber window focused, click its icon in the Windows taskbar.
3. Observe that the window is not minimized to a taskbar button.
4. Observe that the taskbar icon disappears and the app is only reachable from the hidden/background tray area.

### Where does it happen?
`Not sure`
The issue occurs in the native Windows desktop application.

### Version
`1.16.0`

### Screenshots / recordings
Can be attached if needed.

### Logs
No application error is logged. This appears to be a window/taskbar behavior issue rather than a runtime exception.

pascalandr:
## Intent

Keep Windows taskbar minimization native while preserving hide-to-tray behavior for OpenChamber's custom minimize command when the tray preference is enabled. Fixes #2176.

## Non-goals

This PR does not change close-to-tray behavior, tray enablement, window restoration, non-main windows, or renderer bridge contracts.

## Affected surfaces

`packages/electron` main-process window lifecycle is affected on Windows. The custom title-bar minimize command still follows the tray preference; OS/taskbar minimization now remains minimized. Web, hosted mobile, Capacitor mobile, and VS Code are unaffected.

## Repository guidance

|---|---|---|
| Root `AGENTS.md` and `openchamber-change-discipline` | This changes Electron source and native window behavior. | The diff is limited to the existing event and IPC command paths. |
| `desktop-shell` | Window lifecycle and tray behavior are Electron-owned. | Enforcement remains in `main.mjs`; no privileged behavior moves into shared UI. |
| `packages/electron/README.md` | It defines native-shell ownership and platform validation boundaries. | Existing in-process architecture and preload contract remain unchanged. |

## Validation

| Check | Result |
|---|---|
| `bun run type-check:electron` | Passed: `main.mjs` and `preload.mjs` syntax checks. |
| `bun run lint:electron` | Passed. |
| `bun run --cwd packages/electron bundle:main` | Passed; `dist-bundle/main.mjs` generated successfully. |
| `git diff --check main...HEAD` | Passed. |
| Windows: custom minimize, taskbar minimize, tray enabled/disabled | Passed manually: the custom title-bar button hides to the tray when enabled, Windows taskbar minimization remains native and restorable, and disabling the tray preference makes the custom button minimize to the taskbar. |

## Visual evidence

No screenshot is attached because a static image cannot demonstrate the window/taskbar/tray lifecycle. The full Windows behavior matrix above was verified manually.

## Risks and failure behavior

The global `minimize` interception is removed, so OS-driven minimization cannot unexpectedly hide a window. The custom command explicitly hides only when the existing tray predicate allows it; close-to-tray remains unchanged.

### DIFF
--- a/packages/electron/main.mjs
+++ b/packages/electron/main.mjs
@@ -2376,12 +2376,6 @@ const createBrowserWindow = ({ label, restoreGeometry, url, runtimeConfig = {} }
   browserWindow.on('move', () => {
     debounceWindowStatePersist(browserWindow, false);
   });
-  browserWindow.on('minimize', (event) => {
-    if (!shouldHideMainWindowToTray(browserWindow)) return;
-    debounceWindowStatePersist(browserWindow, true);
-    event.preventDefault();
-    browserWindow.hide();
-  });
   browserWindow.on('close', (event) => {
     if (!state.quitRequested && shouldHideMainWindowToTray(browserWindow)) {
       debounceWindowStatePersist(browserWindow, true);
@@ -4331,7 +4325,12 @@ const handleInvoke = async (browserWindow, command, args = {}) => {
 
     case 'desktop_minimize_current_window':
       if (browserWindow && !browserWindow.isDestroyed()) {
-        browserWindow.minimize();
+        if (shouldHideMainWindowToTray(browserWindow)) {
+          debounceWindowStatePersist(browserWindow, true);
+          browserWindow.hide();
+        } else {
+          browserWindow.minimize();
+        }
       }
       return null;
 


makeittech (approved):
Nice, this is the right behavior for Windows minimize.

makeittech:
Looks good, merging.

PR #2495, opened 2026-07-28
feat: add custom request headers to onboarding connection form

### LINKED ISSUE #1763: [Feature Request] Enable custom header support when adding/importing remote instances

Currently, when adding or importing a remote instance, it is not possible to specify custom headers. Many instances, particularly those behind Cloudflare Tunnels with Service Tokens, require specific headers for authentication and cannot be accessed directly.

Please add an option to include custom headers (Key-Value pairs) during the instance setup/import process.

Proposed Format
The configuration should allow for customizable keys and values, ideally in a JSON format similar to:

```
"headers": {
  "CF-Access-Client-Id": "your-client-id",
  "CF-Access-Client-Secret": "your-client-secret"
}
```

### Why is this needed?

This would allow users to securely connect OpenChamber to instances protected by Cloudflare Access or other header-based authentication proxies, improving security and workflow integration.

Rahulsharma0810:
## Summary

- Extends `RemoteConnectionForm` (onboarding & recovery flow) with custom HTTP header support
- The Settings page (`RemoteInstancesPage`) already had this for the add/edit dialogs — this closes the gap in the onboarding path
- Reuses existing `settings.remoteInstances.direct.headers.*` i18n keys (no new translations needed)

## Motivation

Instances behind **Cloudflare Tunnels with Service Token auth** (like `opencode.rvsharma.com`) require `CF-Access-Client-Id` and `CF-Access-Client-Secret` headers on every request. Without this, the connection form fails at probe time before the user can even save the host.

Fixes #1763

## Changes

**`packages/ui/src/components/onboarding/RemoteConnectionForm.tsx`**
- Added `HeaderDraft` type + `createHeaderDraft` / `buildRequestHeaders` helpers (mirroring the pattern in `RemoteInstancesPage`)
- Added `requestHeaders` state
- Passes headers to `desktopHostProbe` in both `handleTest` and `handleConnect`
- Persists `requestHeaders` in the saved `DesktopHost` object
- Renders a key/value header editor between the Name field and the status messages

## Test plan

- [ ] Add a remote instance at a URL behind Cloudflare Access — enter `CF-Access-Client-Id` / `CF-Access-Client-Secret` as custom headers — verify "Test Connection" succeeds
- [ ] Connect without headers to a CF-Access-protected URL — verify it still fails (no regression)
- [ ] Add a remote instance without any custom headers — verify existing behaviour unchanged
- [ ] Verify the saved host has `requestHeaders` populated in the desktop hosts config

### DIFF
--- a/packages/ui/src/components/onboarding/RemoteConnectionForm.tsx
+++ b/packages/ui/src/components/onboarding/RemoteConnectionForm.tsx
@@ -15,6 +15,30 @@ import { useI18n } from '@/lib/i18n';
 
 type ConnectionState = 'idle' | 'testing' | 'success' | 'error';
 
+type HeaderDraft = {
+  id: string;
+  name: string;
+  value: string;
+};
+
+const createHeaderDraft = (name = '', value = ''): HeaderDraft => ({
+  id: typeof crypto !== 'undefined' && typeof crypto.randomUUID === 'function'
+    ? crypto.randomUUID()
+    : `header-${Date.now()}-${Math.random().toString(16).slice(2)}`,
+  name,
+  value,
+});
+
+const buildRequestHeaders = (headers: HeaderDraft[]): Record<string, string> | undefined => {
+  const next: Record<string, string> = {};
+  for (const header of headers) {
+    const name = header.name.trim();
+    const value = header.value.trim();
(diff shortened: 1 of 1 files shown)


Rahulsharma0810:
Closing — feature already exists in Settings since v1.13.9. The onboarding form gap is minor and not the original feature request.

PR #2498, opened 2026-07-28
feat(desktop): gate traffic-lights behind window-controls style setting

kydorn:
## Intent

Gates the Linux/macOS-traffic-lights window chrome behind a user setting. Previously the in-app window controls only exposed a Position option (left/right) and the left side hardcoded traffic-light chrome. Now the Settings → Appearance → "Window controls" section offers both Position (Left/Right chips) and Style (Classic/Traffic lights dropdown), so all four combinations render correctly: left-classic, left-traffic-lights, right-classic, right-traffic-lights.

Defaults preserve existing user behavior: Position = Right (unchanged), Style = Classic (new). Order is side-driven for both styles: close,min,max on the left; min,max,close on the right.

## Non-goals

- Position default stays right (no platform-aware default).
- No traffic-light glyph swap on maximize/restore (cosmetic; aria-label already swaps).
- No change to the existing Linux border work (index.css / main.tsx are unrelated context here; this PR is the style-gating slice only).
- No search-availability predicate changes (kept as pre-existing).

## Affected surfaces

- **packages/ui**: `WindowsWindowControls` component (render branch on style), `OpenChamberVisualSettings` (new Style dropdown), `useUIStore` (new field + v12→v13 migration), `persistence.ts` (apply + sanitize), `settings/search.ts` (new search item), `desktop.ts` (type + field), `desktopWindowControls.test.ts` (migration tests). 10 i18n locale dictionaries.
- **packages/web**: `settings-helpers.js` server sanitizer + tests.
- **Persisted contract**: new `desktopWindowControlsStyle` in `~/.config/openchamber/settings.json` and the `ui-store` persisted slice. Store version bumped 12→13; migration sets style to `'classic'` if missing.
- **Runtimes**: Electron desktop on Windows/Linux frameless chrome (gated by `usesFramelessElectronChrome()`). macOS, web, VS Code, and mobile are unaffected — the setting UI and the control component do not render there.
- **User-visible states**: main header window controls (right side), TitlebarLeftControls overlay (left side), mini-chat header. All four position×style combinations.

## Repository guidance

|---|---|---|
| AGENTS.md — Runtime Boundaries | Change touches shared UI consumed by web/desktop/VS Code/mobile. | No change to runtime APIs or bridges; purely a UI + persisted-preference slice. Call sites (`TitlebarLeftControls`, `Header`, `MiniChatLayout`) unchanged — they still pass `position` and read `side`; style is internal to `WindowsWindowControls`. |
| openchamber-change-discipline | Persisted-state contract (settings.json + ui-store) and cross-package sanitization (ui + web). | Smallest complete change; preserved existing position behavior; store migration is safe (only sets style if missing); sanitization paths reject unknown values instead of coercing; dead `DESKTOP_WINDOW_CONTROLS_WIDTH_PX` constant deleted. |
| settings-ui-patterns (layout.md, controls.md, search.md) | New Settings control: section title, two-column row, chip group, dropdown, search registry. | Built from shared primitives (`SettingsSection` + `SettingsTwoColumn` + `SettingsStackedField` + `SettingsChipGroup` + `Select` with `SETTINGS_SELECT_SIZE`/`SETTINGS_SELECT_TRIGGER_CLASS`). Explanatory prose is behind the `info` prop, not an always-visible `description`. Field-level `data-settings-item` anchors match the `search.ts` registry IDs; section has no anchor to avoid duplicates. Availability predicate matches existing position item. |
| theme-system | Button styling and chrome colors. | Classic buttons use semantic tokens (`text-muted-foreground`, `hover:bg-interactive-hover`, `hover:bg-status-error`). Traffic-light hex colors are documented as a theme-independent macOS chrome reproduction with an explicit comment. No new Button wrappers; uses shared `Button` `chip` variant. |
| locale-ui-patterns | New user-facing strings in Settings + 10 locale dictionaries. | Every new key added to all 10 locales with real translations (no English stand-ins). `windowControlsPositionDescription` fully removed; `windowControlsPosition` value shortened to "Position" (translated everywhere). |

## Validation

| Check | Result |
|---|---|
| bun run type-check (packages/ui) | pass |
| bun run lint (packages/ui) | pass |
| bun test packages/ui/src/lib/desktopWindowControls.test.ts | 7/7 pass (4 existing + 3 new migration tests: missing-style→classic, valid traffic-lights preserved, invalid value→classic) |
| vitest run server/lib/opencode/settings-helpers.test.js (packages/web) | 32/32 pass (incl. 4 new style sanitization cases) |
| bun run dead-code | inspected; deleted dead constants gone; new DesktopWindowControlsStyle store type export matches 10 pre-existing sibling conventions (non-blocking) |
| Adversarial review (2 rounds) | Round 1: Major (dead exports) + Minor (description vs info, missing migration test) applied. Round 2: Critical None, Major None, 1 Minor (invalid-value coercion test) applied. Clean. |

**Not verified (runtime, cannot run here):** the four visual combinations on Linux Electron across main header, mini-chat header, and sidebar-collapsed reserved-width carve. AppImage rebuilt — manual visual confirmation on a Linux desktop recommended before merge.

## Visual evidence

User-visible change (Settings layout + window-control chrome). Screenshots/recordings not captured here — the four (position × style) combinations and the Settings → Appearance → Window controls section should be verified visually on a Linux Electron build before removing the draft.

## Risks and failure behavior

- **Persisted-state migration**: store version bumped 12→13. Migration sets `desktopWindowControlsStyle = 'classic'` only if the persisted value is not already `'classic'` or `'traffic-lights'`; `desktopWindowControlsPosition` is untouched. No user is flipped to traffic-lights or to a different position on upgrade. Tested.
- **Round-trip for invalid values**: invalid persisted style (e.g. `'macos'`, `'auto'`) lands on `'classic'` through every layer — store migration, store setter coercion, client persistence sanitizer (rejects), server sanitizer (rejects), component fallback (only matches `'traffic-lights'`, everything else renders classic). Tested at migration layer.
- **Rollback**: reverting the commit leaves existing users on the new persisted field, which older code simply ignores (opt-in field on `DesktopSettings`). No data loss.
- **Cross-runtime**: web, VS Code, mobile, and macOS desktop do not render the setting or the control component — no parity concern.
- **Performance**: `WindowsWindowControls` reads one additional store slice; it was already subscribed to the maximized-state event channel and is a memoized component. No new polling, effects, or fetches. `TitlebarLeftControls`'s existing `ResizeObserver` publishes the cluster width live, so the narrower traffic-light cluster and the wider classic-left cluster flow through with no constant edits.
- **Security**: no new secrets, auth, or privileged paths.


kydorn:
Screenshots
Left + traffic lights
image

Right + traffic lights
image

Left + Classic
image

Right + Classic (default)
image

Settings
image
image

makeittech:
## The diff is much larger than the feature — here's where the mass is and how to shrink it

The feature itself is small: one Style dropdown (Classic / Traffic lights) for frameless-chrome window controls. The diff is **+370 / −45 across 21 files (~830 diff lines)**. Breaking down where that goes, and what I'd change:

### 1. The PR contradicts its own description — Linux border work is in the diff

The body states: *"No change to the existing Linux border work (index.css / main.tsx are unrelated context here; this PR is the style-gating slice only)"* — yet `initLinuxWindowBorder()` (+37 in `desktop.ts`), the `index.css` rule (+11), and the `main.tsx` call (+2) are all **in this diff**. That's ~50 lines of an undeclared, separate feature. Please split it into its own PR; it also makes this one harder to review and to revert independently.

### 2. The v12→v13 store migration is unnecessary

Zustand `persist` shallow-merges rehydrated state over initial state, so a missing `desktopWindowControlsStyle` key already falls back to the `'classic'` default — no version bump or migrate block needed. Invalid values are additionally rejected by three other layers this PR adds/extends: the setter coerces, the client `sanitizeWebSettings` rejects, and the server sanitizer rejects. The migration is a fourth, redundant layer. Dropping it removes ~55 diff lines (migrate block + version bump + 3 migration tests) with zero behavior change.

### 3. The Settings section was rebuilt for no functional reason

The existing section (one `SettingsSection` + one `SettingsChipGroup`) was torn down and reassembled as `SettingsTwoColumn` + two `SettingsStackedField` + a full `Select` (trigger/content/memoized label) — for a **two-option** setting. A second `SettingsChipGroup` next to Position is symmetric with the existing control, needs no `selectedWindowControlsStyleLabel` memo, and costs ~15 lines instead of ~50.

### 4. i18n churn: existing keys were renamed/removed

`windowControlsPosition` was reworded ("Window controls position" → "Position"), `windowControlsPositionDescription` was deleted, and a new section title/description pair was added — forcing a rewrite of that block in all 10 locale files (~90 diff lines). The feature only needs 4 new keys (`windowControlsStyle`, its aria, `Classic`, `Traffic lights`). Adding keys without touching existing ones keeps the locale diff at ~40 additive lines and avoids invalidating existing translations for a cosmetic label change.

### 5. Glyph rendering can reuse the existing icon sprite

The three per-branch SVG blocks with full wrappers (~23 lines) duplicate glyphs the repo's sprite-based `<Icon>` system already provides: `close` and `subtract` are in the sprite, and this same file already renders the classic buttons with `<Icon>`. Using `<Icon name="close">` / `<Icon name="subtract">` for the hover glyphs needs no new assets. The maximize glyph is the one exception — the macOS two-triangle expand glyph has no sprite equivalent, so either register it as a custom glyph in `scripts/generate-icon-sprite.mjs` (the documented mechanism, 24×24 `currentColor`) or keep a single small inline SVG for it. Either way the three `if`-wrapped SVG components collapse into one glyph map, saving ~15 lines plus most of the explanatory comments.

### What this adds up to

| Change | Diff lines saved |
|---|---|
| Split Linux border into its own PR | ~50 |
| Drop the v12→v13 migration + its tests | ~55 |
| Chips instead of Select/TwoColumn rebuild | ~35 |
| Add-only i18n keys (no renames) | ~50 |
| Collapse glyph SVGs (sprite icons for close/minimize, registered/inline glyph for maximize) | ~15 |

That's roughly **−205 of the +370** — the PR lands around ~150–170 additions while keeping the exact same user-facing behavior, the setting itself, the persisted contract, and all four sanitization layers. Everything removed is either undeclared scope or redundant defense-in-depth, not functionality.

kydorn:
Thanks for the detailed review — really useful breakdown. Pushed `6c8a1858` addressing the points:

**1. Linux border work (split out)** — Done. `initLinuxWindowBorder()`, the `index.css` rule, and the `main.tsx` call are no longer in the diff. `desktop.ts` now only adds the `DesktopWindowControlsStyle` type and setting field.

**2. v12→v13 migration** — Moot. No version bump, no `migrate` block, no migration tests. `version` stays `12`. The three sanitizer layers (setter coercion, `sanitizeWebSettings`, server `settings-helpers`) remain — agreed those were the right ones.

**3. Settings section** — Deliberately keeping `SettingsTwoColumn` + two `SettingsStackedField`. The two-column layout is consistent with the adjacent sections (`mobileLayout`, `localization`, `theme` blocks all use `SettingsTwoColumn`), and the `SettingsStackedField` label gives each control an explicit visible label the chip group alone wouldn't. Acknowledged on the diff-size point; chose visual consistency with neighbors here. The `Select` + memo are gone — it's two `SettingsChipGroup`s, as recommended.

**4. i18n keys** — Done. +50 additive across 10 locales, 0 deletions. `windowControlsPosition` / `…Description` / `…Aria` retained; 5 new keys added (`windowControls`, `windowControlsStyle`, `windowControlsStyleAria`, `windowControlsClassic`, `windowControlsTrafficLights`).

**5. Glyph rendering** — Done and simpler than proposed. All three glyphs use existing sprite icons (`close`, `subtract`, `add`) — no custom SVG, no sprite registration. For the maximize/restore glyph specifically: using `add` (`+`) for both states. It's a known macOS-theme alternative for the traffic-light maximize control, well understood, and the accessible label still flips between "maximize"/"restore" via `isMaximized` so the state is conveyed to AT users. This resolved the "no Remixicon equivalent for the two-triangle expand glyph" exception you noted — no custom glyph needed at all.

Also narrowed out-of-scope churn from the PR: reverted a `sprite.ts` regeneration that had dropped two unrelated dead icons (`arrow-left-long`, `sort-desc`) — belongs in a separate dead-code cleanup. Removed the dead `DESKTOP_WINDOW_CONTROLS_WIDTH_PX` constant with a comment at the style type noting control width is now style-dependent (classic vs traffic-lights), so a fixed-width constant no longer applies.

PR is now +255/−21 across 18 files.

kydorn:
Updated relevant screenshot with the glyph changes
image
Simplified the settings section, keeping a two-column layout for consistency with other similar cases. The Style selector changed from dropdown to a chip style (two options)
image

kydorn:
A final small change, I have increased the contrast in the glyphs when hovering over the traffic lights, for better visibility. instead of 3 colors now we have only one with opacity.

image

makeittech (approved):

PR #2511, opened 2026-07-28
fix(ui): paint sidebar zone-band backing only while pinned

### LINKED ISSUE #2508: [Bug] Sidebar rendering issue (when Window transparency is enabled)

When I enable Window transparency on my Mac, the left sidebar appears to have a display issue — an unexpected, fully opaque black fill appears at the bottom. This disappears when Window transparency is turned off. I believe this is an edge case that was overlooked, and I hope it can be addressed in a future update.

### Where does it happen?

Desktop (macOS)

### Version (if known)

1.17.0

Image

### Logs (optional)

```shell

```

IbrahimKhan12:
## Intent

Fixes #2508. With **Window transparency** (macOS vibrancy) enabled, the sidebar rendered an opaque black slab across its lower portion.

`design-system.css` forced the opaque `--sidebar-stuck-bg` onto *every* `.oc-zone-header-backing` band under `[data-oc-vibrancy]`, while the sidebar surface itself is translucent. Sticky zone bands are on by default for every project header and for `recent`, so a collapsed project list is wall-to-wall opaque bands.

That fill is not gratuitous: Electron vibrancy windows break `backdrop-filter` (electron#20357), so a **pinned** band needs an opaque fill to mask rows scrolling beneath it. An unpinned band has nothing to mask. Bands now report pinned state as `data-oc-stuck`, and the backing is scoped to it.

## Visual evidence

The reporter's configuration (13 collapsed projects plus a collapsed `recent`, `scrollTop 0`), rendered from the compiled `packages/ui/src/index.css` in Chromium. The native `NSVisualEffectView` is substituted with a gradient backdrop, so the frosting reads flat rather than blurred; band opacity is unaffected by that substitution.

|  | Before | After |
|---|---|---|
| **Dark** | dark before | dark after |
| **Light** | light before | light after |

A **pinned** band still masks rows scrolling under it, which is the behavior the opaque fill exists for:

pinned band masks scrolled rows

Computed `background-color` per band, vibrancy on and off:

| Before | After |
|---|---|
| computed styles before | computed styles after |

| Element (vibrancy on) | Before | After |
|---|---|---|
| `.oc-vibrancy-surface` (sidebar) | `rgba(28, 27, 26, 0.66)` | `rgba(28, 27, 26, 0.66)` |
| band, unpinned | `rgb(28, 27, 26)` | `rgba(0, 0, 0, 0)` |
| band, pinned | `rgb(28, 27, 26)` | `rgb(28, 27, 26)` |

With vibrancy **off** every value is `rgb(28, 27, 26)` before and after, so the opaque desktop path is byte-identical.

Narrow/mobile states are omitted: the rule is scoped to `:root.desktop-runtime[data-oc-vibrancy]` and the mobile sidebar variant is not a vibrancy surface.

## Non-goals

Pinned bands stay fully opaque. No change to the vibrancy alpha, `--sidebar`, the `prefers-reduced-transparency` branch, or the `shadow-md` pinned affordance.

## Affected surfaces

`packages/ui` sidebar only. Behavior changes on macOS desktop **with** vibrancy. Every other surface is unaffected: the CSS rule is scoped to `:root.desktop-runtime[data-oc-vibrancy]`, and `window.__OPENCHAMBER_ELECTRON__.macVibrancy` is undefined on web, VS Code, mobile, and non-vibrancy desktop, so the new observer never mounts and `data-oc-stuck` is never emitted. No props, exports, or persisted state touched.

## Repository guidance

| Guidance | How the change complies |
|---|---|
| `openchamber-change-discipline` | One CSS selector, one attribute per band, one sentinel. No new dependencies, no drive-by refactors; owning `DOCUMENTATION.md` updated with the `data-oc-stuck` invariant |
| `theme-system` | Reuses the existing `--sidebar-stuck-bg` token, no hardcoded colors; verified in light and dark |
| `performance-engineering` | Observer gated on `macVibrancy`, its only consumer. One `IntersectionObserver` on one 1px sentinel; no scroll handler, no per-row observers |

## Validation

| Check | Result |
|---|---|
| `bun run type-check`, `lint`, `build` (all workspaces) | Pass; emitted bundles contain `.oc-zone-header-backing[data-oc-stuck]` exactly once |
| `bun run --cwd packages/web test` | 856 pass, 2 skipped |
| `bun test packages/ui/src` | Identical to the same run with the change stashed (117 pre-existing failures, zero delta) |
| `bun run dead-code` | No new findings |
| Computed styles in Chromium | See the table above; this is what proves the cascade-layer precedence the fix relies on |

**Not verified:** developed on Linux, so this was never run in a packaged macOS Electron window against the real vibrancy layer. Headless Chromium produces no frames, so `IntersectionObserver` never delivers there and the new `recent` observer's runtime firing is unverified; it is a direct mirror of the shipped `useStickyProjectHeaders` sentinel in the same scroll container.

## Risks and failure behavior

- The sentinel stays true for the final band-height of travel as a band scrolls off the top edge, so a band can paint opaque for ~26px while partially clipped. `!isCollapsed` removes this for collapsed zones, which have zero sticky travel and never need the mask, covering the reported configuration. Same latency class as the existing `shadow-md` pinned signal.
- Unpinned bands rely on the unlayered `!important` transparent rule in `index.css` outranking the Tailwind utility, and on the `@layer base` pinned rule outranking that unlayered rule. Both directions are asserted by the measurements above rather than by reasoning.
- The effect resets `isBandStuck` and disconnects the observer on teardown, so no stale pinned state survives a variant switch or `stickyZoneHeaders` being turned off. Nothing persisted, nothing to roll back.


IbrahimKhan12:
Closing as superseded. #2526 landed the same fix for project headers while this was open, and `addee2be` / `c7f0aec5` covered the `recent` band and stale sticky-header state.

PR #2526, opened 2026-07-29
fix(ui): remove desktop sidebar project header artifacts

ChangeHow:
## Intent

Remove opaque rectangular artifacts behind non-stuck project headers in the macOS Desktop sidebar when native vibrancy is enabled.

Reproduction:
1. Run OpenChamber Desktop on macOS with vibrancy enabled.
2. Open the sidebar with multiple projects and sticky headers enabled (the default).
3. Observe opaque header-sized rectangles behind project rows that are not currently stuck.
4. Closing and reopening the sidebar recreates the artifacts because the project header elements are remounted with the original backing styles.

Cause: every sticky-capable project header received `oc-zone-header-backing`, even when it was not actually stuck. Under Electron vibrancy that class paints an opaque sidebar background and also requested an unsupported `backdrop-filter`.

Solution: keep sticky positioning enabled, but apply `oc-zone-header-backing` only while the existing intersection observer reports that a project header is stuck. Remove the unsupported backdrop-filter declarations from the Electron vibrancy backing rule.

## Non-goals

- No changes to the sticky-header preference or its default value.
- No changes to sidebar scrolling, project ordering, session rendering, or macOS vibrancy configuration.
- No changes to the Recent section header.

## Affected surfaces

- `packages/ui`: project-header class selection and the desktop vibrancy backing style.
- macOS Desktop with native vibrancy enabled: non-stuck project headers remain transparent; actually stuck headers retain an opaque backing so rows cannot show through.
- Web, VS Code, mobile, Windows, and Linux are unaffected because the backing override is scoped to `.desktop-runtime[data-oc-vibrancy]`.
- No persisted data, external API, runtime bridge, or package contract changes.

## Repository guidance

| Guidance | Why it applies | How the change complies |
|---|---|---|
| Root `AGENTS.md` and `openchamber-change-discipline` | This changes executable shared UI source and styling. | The diff is limited to the project-header state class and its owning desktop vibrancy rule; no dependencies, exports, or contracts changed. |
| `theme-system` and `tokens-and-examples.md` | This changes a sidebar visual state. | The existing semantic `--sidebar-stuck-bg` token is preserved; no hardcoded colors or new visual primitives were introduced. |
| `desktop-shell` and `packages/electron/README.md` | The bug is visible only through Electron macOS vibrancy composition. | The native shell and privileged boundary are unchanged; the fix stays in the shared UI-owned renderer styling. |
[...]

PR #2528, opened 2026-07-29
fix(ui): smooth desktop sidebar sticky transitions

ChangeHow:
## Intent

This is the final visual follow-up to #2526. That PR stopped non-stuck project headers from receiving the opaque vibrancy backing, but actually stuck headers still switched to a solid color and the transition remained visually abrupt on macOS Desktop.

This PR keeps the native sticky layout while replacing the remaining opaque backing with a short scroll-driven content mask. A duplicate Recent or project identity is rendered outside the mask so its icon and title remain crisp. The fade starts with a safe clear area, grows only to 48px, blocks interaction with obscured session rows, and leaves native wheel scrolling intact.

## Non-goals

- No changes to the sticky-header preference or its default value.
- No changes to project ordering, session grouping, collapsed activity indicators, or sidebar persistence.
- No changes to Electron native vibrancy configuration.

## Affected surfaces

- `packages/ui`: sidebar sticky identity rendering, project-header observation, and optional top mask variables for `ScrollShadow`.
- macOS Desktop: Recent and project headers remain readable over vibrancy without opaque header bands or unsupported `backdrop-filter`.
- Web, VS Code, mobile, Windows, and Linux retain the original top-shadow behavior because custom measurement, interaction guards, and identity overlays are gated by `isDesktopShellRuntime`; the macOS vibrancy artifact is not present on those surfaces.
- No persisted data, external APIs, runtime bridges, dependencies, or package contracts change.

## Repository guidance

|---|---|---|
| Root `AGENTS.md` and `openchamber-change-discipline` | The PR changes executable shared UI source and module documentation. | The implementation is scoped to the existing sidebar module, preserves native sticky behavior, adds no dependency or public runtime contract, and uses focused package validation. |
| `theme-system` and `tokens-and-examples.md` | The change affects desktop sidebar visual states. | It removes the opaque sticky background instead of adding hardcoded colors and reuses existing semantic foreground and muted-foreground classes. |
| `desktop-shell` and `packages/electron/README.md` | The artifact is specific to Electron macOS vibrancy composition. | The native shell remains unchanged; renderer behavior is explicitly gated to Desktop and was packaged through the Electron build. |
| `performance-engineering` | The implementation updates mask values on the scroll path. | Per-scroll dimensions stay in refs and CSS custom properties; React state changes only when the identity overlay appears or hides. |
| `packages/ui/src/components/session/sidebar/DOCUMENTATION.md` | This module owns sticky project and Recent rendering. | The documentation now describes the mask-external identity and interaction boundary. |

## Validation

| Check | Result |
|---|---|
| `bun run type-check:ui` | Passed on the final HEAD. |
| `bun run lint:ui` | Passed on the final HEAD. |
| `bun run type-check:electron` | Passed. |
| `bun run electron:build` | Passed on the final HEAD; produced the macOS ARM64 1.17.1 DMG and ZIP. Existing KaTeX asset, chunk-size, optional native dependency, ad-hoc signing, and notarization warnings remain. |
| `bun run dead-code` | Completed. The non-blocking report contains existing repository unused-export and configuration findings; it did not report the new `ProjectHeaderIdentity` export as unused. |
| Manual macOS Desktop testing | The reporter iteratively verified the ARM64 package with vibrancy enabled, including initial scroll, long scroll, Recent/project transitions, icon alignment, fade height, sidebar remount, and obscured-row interaction. The final desktop-only guard does not change the tested Desktop render path. |
| Other platforms | Not manually tested; their custom sticky-fade path is disabled by the Desktop runtime gate. |

## Risks and failure behavior

The Desktop scroll handler updates two CSS mask variables and one ref. If a frame is delayed, the fade boundary can briefly lag the native sticky position, but layout height and scroll position remain unchanged. The transparent fade region intercepts pointer, click, and context-menu events for obscured session rows while explicitly allowing the sticky header and overlay scrollbar. The observer is rooted at the sidebar scroller so projects below the viewport are not treated as stuck. No data, persistence, security boundary, or external state is involved. Rollback consists of restoring the previous opaque backing and removing the Desktop fade overlay.

# Screenshots

## Before
image

## After
image

image

## Video
https://github.com/user-attachments/assets/4e050d7a-6ee5-40dc-9f5b-0fa5519c3793


ChangeHow:
@btriapitsyn Finally finish this PR with some UI tradeoff, losing fancy follow-scrolling sticky effect with dummy sticky effect.

btriapitsyn:
Thanks for pushing this visual follow-up. The scroll-driven mask/identity approach looks promising, and I appreciate the Desktop-only guard and the manual packaged validation.

One issue needs to be fixed before merge: the new fade path is gated only by `isDesktopShellRuntime`, not by the existing `stickyZoneHeaders` preference. With Sticky headers turned off, Desktop still enables the mask, blocks the top 32–48px of row interaction, and can render the duplicate sticky identity because the project observer is still active. That changes the behavior of a preference this PR intends to preserve.

Please pass/read `stickyZoneHeaders` in `SidebarProjectsList` and use it to gate the custom fade/overlay path. The project sticky observer should also be disabled when that preference is off. Once that is addressed, I am happy to take another look.

ChangeHow:
@btriapitsyn you are right, I miss up the sticky project header option. fixed in the new commit https://github.com/openchamber/openchamber/pull/2528/commits/cfaf32bc31750dcb5c12eeb5ba6c29d4e9de3ae3

here is my screen recording:

https://github.com/user-attachments/assets/9386b29c-e4c4-4657-8f11-0b510abaeeb7

ChangeHow:
@btriapitsyn pr-review keep failing, could you retry it

btriapitsyn (approved):

PR #2537, opened 2026-07-30
feat: add Windows ARM64 support with x64-baseline CLI workaround

airtaxi:
## Intent

Windows ARM64 users cannot run OpenChamber's bundled OpenCode CLI. The native ARM64 `opencode.exe` fails with a Bun FFI/TinyCC dlopen error ([anomalyco/opencode#19130](https://github.com/anomalyco/opencode/issues/19130)). This PR adds Windows ARM64 support through two changes:

1. **Temporary CLI workaround:** bundle the x64-baseline OpenCode CLI on ARM64 (it runs under x64 emulation) and disable OpenCode self-upgrade so the broken ARM64 build cannot overwrite the working x64 binary. The upgrade gate is enforced server-side, in the VS Code extension, and in the UI (notification toast, settings checkbox, settings search index).
2. **Permanent CI addition:** add ARM64 Windows cross-compile builds to both `release.yml` and `release-desktop-smoke.yml`, and refactor the Windows `latest.yml` update manifest to follow the macOS pattern (per-arch artifacts → combined manifest via `combine-electron-manifests`).

When the upstream issue is resolved, only the CLI workaround files need reverting. The CI ARM64 build stays.

## Non-goals

- **No native ARM64 CLI bundling:** the x64-baseline binary is a temporary substitute, not a permanent architecture decision.
- **No test changes:** existing tests for `upgrade-capability.js` and `opencode-upgrade-runtime.ts` are not modified. The ARM64 guard is a simple early-return whose correctness is obvious by inspection, and the existing tests pass on CI (Linux/x64) where the guard is a no-op. Adding `process.arch`/`process.platform` mocking to existing tests is deferred to keep the workaround focused.
- **No `routes.js` custom error message:** the `windows-arm64-workaround` reason falls through to the existing generic `OPENCODE_UPGRADE_UNSUPPORTED` 409 response. A dedicated message would be a minor UX improvement, deferred to a follow-up.
- **No persisted-settings migration:** a returning ARM64 user whose `settings.json` has `showOpenCodeUpdateNotifications: true` (from a prior x64 install) will have that value rehydrated. The server-side capability gate blocks the upgrade regardless, so this is safe.

## Affected surfaces

| Surface | Effect |
|---|---|
| `packages/electron/scripts/prepare-opencode-cli.mjs` | ARM64 Windows downloads `opencode-windows-x64-baseline.zip` instead of `opencode-windows-arm64.zip` |
| `packages/web/server/lib/opencode/upgrade-capability.js` | Returns `supported: false, reason: 'windows-arm64-workaround'` on ARM64 before any other logic |
| `packages/web/server/lib/opencode/env-runtime.js` | ARM64 Windows resolves x64-baseline package names matching the bundled binary |
| `packages/web/server/lib/opencode/routes.js` | Unchanged; existing 409 `OPENCODE_UPGRADE_UNSUPPORTED` path handles the new reason |
| `packages/vscode/src/opencode-upgrade-runtime.ts` | ARM64 guard in `getCapability()`; type union extended with `'openchamber'` manager and `'windows-arm64-workaround'` reason |
| `packages/ui/src/lib/platform.ts` | New `isWindowsArm64()` helper (Electron `__OPENCHAMBER_ELECTRON__.arch`, VS Code `__VSCODE_CONFIG__.arch`, UA-CH `architecture`, UA-string fallback) |
| `packages/ui/src/stores/useUIStore.ts` | `showOpenCodeUpdateNotifications` default `!isWindowsArm64()` |
| `packages/ui/src/components/sections/openchamber/OpenCodeCliSettings.tsx` | Settings checkbox hidden on ARM64 via `{!isWindowsArm64() && ...}` |
| `packages/ui/src/lib/settings/search.ts` | `SettingsSearchAvailabilityContext` gains `isWindowsArm64` field; `isAvailable` excludes the item on ARM64 |
| `packages/ui/src/components/views/SettingsView.tsx` | Passes `isWindowsArm64` into search `runtimeCtx` via `useMemo` matching sibling platform flags |
| `.github/workflows/release.yml` | ARM64 matrix in `build-desktop-electron-windows`; `latest.yml` removed from direct release upload; `combine-electron-manifests` extended to merge Windows manifests |
| `.github/workflows/release-desktop-smoke.yml` | ARM64 matrix in `build-windows-electron`; `ELECTRON_BUILDER_ARCH` env added |

**Unaffected runtimes:** macOS, Linux, WebAssembly, mobile. No ARM64 guard fires on these platforms. The `isWindowsArm64()` helper requires both `win32` platform and `arm64` architecture.

## Repository guidance

|---|---|---|
| `openchamber-change-discipline` | Source, build-config, package-contract, and module-ownership changes across 11 files | Change is focused on a single feature (ARM64 support); no unrelated modifications; diff is 11 files, +112/-21 lines |
| `desktop-shell` | Electron `prepare-opencode-cli.mjs` bundling logic changed | Script change is minimal; the ARM64 branch returns x64-baseline artifact name, and the original logic is preserved in comments for revert |
| `ui-api-decoupling` | `upgrade-capability.js` and `env-runtime.js` are server-side OpenCode SDK / runtime fetch modules | New guard is an early-return at the top of `resolveOpenCodeUpgradeCapability`; no SDK call changes; `env-runtime.js` ARM64 branch returns same package list as x64 branch |
| `settings-ui-patterns` | Settings UI checkbox and settings search availability changed | Checkbox gated by `{!isWindowsArm64() && ...}`; search item excluded via `isAvailable` callback in `search.ts` context pattern; `isWindowsArm64` flows through `runtimeCtx` like sibling `isMac`/`isWindows`/`isLinux` |
| `theme-system` | No theme/styling changes | N/A; no visual style changes, only conditional rendering |
| `AGENTS.md` runtime boundaries | Electron starts backend in-process; shared contracts must define behavior for every runtime | ARM64 Windows uses x64-baseline CLI under emulation; upgrade disabled consistently across web server, VS Code, and UI |

**Skill references read:** `openchamber-change-discipline/SKILL.md`, `desktop-shell/SKILL.md`, `ui-api-decoupling/SKILL.md`, `settings-ui-patterns/SKILL.md`.

**Module docs read:** `packages/electron/README.md`, `packages/web/server/lib/opencode/` (no `DOCUMENTATION.md` present).

## Validation

| Check | Command | Result |
|---|---|---|
| UI type-check | `bun run type-check:ui` | PASS |
| Web type-check | `bun run type-check:web` | PASS |
| Electron type-check | `bun run type-check:electron` | PASS |
| VS Code type-check | `bun run --cwd packages/vscode type-check` | PASS |
| Workspace lint | `bun run lint` | PASS (all 5 packages exit 0) |
| Smoke CI, Windows ARM64 | `workflow_dispatch` on `arm64-support` ([run 30539003898](https://github.com/airtaxi/openchamber/actions/runs/30539003898)) | PASS. `Build Windows Electron (arm64)` completed; 170 MB artifact uploaded |
| Smoke CI, Windows x64 | `workflow_dispatch` on `arm64-support` ([run 30539003898](https://github.com/airtaxi/openchamber/actions/runs/30539003898)) | PASS. `Build Windows Electron (x64)` completed; 169 MB artifact uploaded |
| Runtime on ARM64 hardware | Physical Windows ARM64 device | PASS. App launches, x64-baseline CLI runs under emulation, update notification checkbox hidden in Settings |

**Not verified:**

- **`release.yml` full release:** `create-release` job failed due to missing `[0.0.0-dev]` changelog section (expected; no `CHANGELOG.md` entry for a dev version). All downstream jobs were skipped. This is a test-data issue, not a workflow defect.
- **Existing unit tests on ARM64 host:** `upgrade-capability.test.js` (3 tests) and `opencode-upgrade-runtime.test.ts` (3 tests) would fail on an ARM64 host because `isWindowsArm64()` reads `process.arch`/`process.platform` directly with no mock seam. CI runs on x64/Linux where the guard is a no-op, so CI passes. This is acceptable for a temporary workaround, but worth noting.
- **`bun run build`:** not run. CONTRIBUTING.md requires it, but the change is source-only with no build-config changes that would affect the build graph. Type-check and lint cover the TypeScript/JS surfaces.

## Visual evidence

**Verified on physical ARM64 hardware:** the app launches, the CLI runs under x64 emulation, and the "Show OpenCode update notifications" checkbox is hidden in Settings. Confirmed by the user on a physical Windows ARM64 device.

**After (ARM64 Windows, Settings → OpenCode CLI):**
pr-arm64-after

**Before state** cannot be captured; it would require a non-ARM64 host running this same build, which is not available for this PR. On non-ARM64 hosts the conditional render (`{!isWindowsArm64() && ...}`) produces output byte-identical to the existing x64/macOS/Linux release, so the before state is the current production Settings UI with the checkbox visible.

The settings search index also excludes the "update notifications" item on ARM64 through the same conditional. No visible difference on non-ARM64 hosts.

## Risks and failure behavior

**Persisted `showOpenCodeUpdateNotifications` rehydration:** a user who previously ran OpenChamber on x64 (where the default is `true`) and upgrades to this build on ARM64 will have `true` rehydrated from `settings.json`, overriding the ARM64 default of `false`. This is safe in practice. The server-side `upgrade-capability.js` returns `supported: false` on ARM64 regardless of the client flag, and the VS Code `opencode-upgrade-runtime.ts` guard blocks independently. The only consequence is a wasted network round-trip when `OpenCodeUpdateToast` polls the `/api/opencode/upgrade-status` endpoint (which returns `available: false`). No upgrade can occur.

**`routes.js` generic error:** if a user somehow triggers the upgrade endpoint on ARM64 (e.g., via direct API call), they receive the generic `OPENCODE_UPGRADE_UNSUPPORTED` 409 response. No upgrade occurs. A dedicated error message would be friendlier but is deferred.

**Native module cross-compilation:** the ARM64 build cross-compiles `better-sqlite3`, `node-pty`, and `bun-pty` from an x64 runner via `@electron/rebuild` with `ELECTRON_BUILDER_ARCH=arm64`. The [smoke CI confirmed this succeeds](https://github.com/airtaxi/openchamber/actions/runs/30539003898). If a future native dependency update breaks ARM64 cross-compilation, the `build-desktop-electron-windows` ARM64 matrix entry will fail, but the x64 entry is independent (`fail-fast: false`).

**Rollback:** revert the single commit. The `prepare-opencode-cli.mjs` original logic is preserved in comments. The CI ARM64 matrix entries can remain (they are a permanent feature) or be removed if desired. The `combine-electron-manifests` change is backward-compatible; if only x64 builds run, `finalize-latest-yml.mjs` produces a single-arch `latest.yml` as before.

### DIFF
--- a/.github/workflows/release-desktop-smoke.yml
+++ b/.github/workflows/release-desktop-smoke.yml
@@ -186,7 +186,7 @@ jobs:
 
   build-windows-electron:
     if: ${{ inputs.build_windows }}
-    name: Build Windows Electron (x64)
+    name: Build Windows Electron (${{ matrix.arch }})
     # Match the production release workflow. windows-latest currently resolves
     # to a runner with Visual Studio 18, which this Electron/node-gyp stack does
     # not detect correctly.
@@ -198,6 +198,9 @@ jobs:
           - arch: x64
             target: x86_64-pc-windows-msvc
             platform: win32-x64
+          - arch: arm64
+            target: aarch64-pc-windows-msvc
+            platform: win32-arm64
     steps:
       - name: Checkout selected ref
         uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
@@ -249,6 +252,9 @@ jobs:
       - name: Rebuild native modules
         working-directory: packages/electron
         shell: bash
+        env:
+          # Cross-compile for ARM64 target from x64 runner.
+          ELECTRON_BUILDER_ARCH: ${{ matrix.arch }}
         # npmRebuild=false in package.json, so electron-builder won't
         # recompile native deps on its own. Rebuild against the target
         # Electron ABI before packaging, matching the release workflow.
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -539,6 +539,7 @@ jobs:
           files: |
             ${{ runner.temp }}/latest-mac.yml
             ${{ runner.temp }}/latest.yml
+            ${{ runner.temp }}/latest-arm64.yml
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -284,6 +284,9 @@ jobs:
           - arch: x64
             target: x86_64-pc-windows-msvc
             platform: win32-x64
+          - arch: arm64
+            target: aarch64-pc-windows-msvc
+            platform: win32-arm64
     steps:
       - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
 
@@ -331,6 +334,9 @@ jobs:
       - name: Rebuild native modules
         working-directory: packages/electron
         shell: bash
+        env:
+          # Cross-compile for ARM64 target from x64 runner.
+          ELECTRON_BUILDER_ARCH: ${{ matrix.arch }}
         # npmRebuild=false in package.json, so electron-builder won't
         # recompile native deps on its own — we must rebuild against the
         # target Electron ABI before packaging.
@@ -350,7 +356,6 @@ jobs:
           files: |
             packages/electron/dist/*.exe
             packages/electron/dist/*.blockmap
-            packages/electron/dist/latest.yml
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
@@ -504,7 +509,7 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
   combine-electron-manifests:
-    needs: [create-release, build-desktop-electron-macos]
+    needs: [create-release, build-desktop-electron-macos, build-desktop-electron-windows]
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
@@ -514,13 +519,13 @@ jobs:
         with:
           node-version: '22'
 
-      - name: Download per-arch latest-mac.yml
+      - name: Download per-arch update manifests
         uses: actions/download-artifact@d3f86a106a0bac45b974a628896c90dbdf5c8093 # v4.3.0
         with:
-          pattern: latest-yml-*-apple-darwin
+          pattern: latest-yml-*
           path: artifacts
 
-      - name: Finalize combined latest-mac.yml
+      - name: Finalize combined manifests
         env:
           LATEST_YML_DIR: ${{ github.workspace }}/artifacts
           GH_REPO: ${{ github.repository }}
@@ -533,6 +538,7 @@ jobs:
           tag_name: v${{ needs.create-release.outputs.version }}
           files: |
             ${{ runner.temp }}/latest-mac.yml
+            ${{ runner.temp }}/latest.yml
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
--- a/packages/electron/README.md
+++ b/packages/electron/README.md
@@ -74,7 +74,7 @@ macOS builds produce `dmg` and `zip` artifacts. Windows builds produce an NSIS i
 
 macOS packaging needs Xcode/build tools for notarized builds and icon asset compilation.
 
-Windows packaging needs NSIS support through `electron-builder`. If no Windows signing env is set, `package.mjs` disables code signing and builds an unsigned installer.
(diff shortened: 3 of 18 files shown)


btriapitsyn:
Fixed the review blocker in commit `745e105c0` and pushed it directly to the PR branch.

Changes:

- Windows update manifests no longer combine x64 and ARM64 installers into one list, where `electron-updater` selected the first `.exe` without architecture filtering.
- x64 continues to use `latest.yml`; ARM64 is now published as `latest-arm64.yml`.
- Packaged Windows ARM64 builds explicitly use the `latest-arm64` channel, while the generic E2E feed remains unchanged.
- The release job uploads both Windows manifests and fails closed if either architecture manifest is missing.
- Added regression tests for channel selection, manifest separation, and incomplete artifact sets.
- Updated the Electron README with the architecture-specific updater contract.

There were no conflicts with the base branch. The bot did not report code findings; automated review was skipped because the PR changes release workflow trust-boundary files.

Validation:

- `bun run --cwd packages/electron test:updater` — 18 tests passed
- `bun run type-check:electron` — passed
- `bun run lint:electron` — passed
- `bun run --cwd packages/electron bundle:main` — passed
- `bun run dead-code` — completed; the report contains existing workspace findings and no new findings for the added updater files

PR #2542, opened 2026-07-30
fix(web): update foreground systemd services safely

### LINKED ISSUE #2541: [Bug] In-app updates fail for systemd foreground servers and Electron remote instances
## Summary

When OpenChamber Web runs as a `systemd --user` foreground service, its in-app self-updater starts the package-manager process in the service cgroup and then exits the foreground server. With the normal `KillMode=control-group`, systemd kills the detached updater before installation completes, restarts the unchanged package, and the client waits indefinitely for the target version.

This is related to #1655, which reports the visible infinite “Waiting for server…” symptom. The service-cgroup interaction below makes the failure reproducible.

## Environment

- OpenChamber Web installed globally with npm
- OpenChamber started by a user systemd service with `Restart=always`, `KillMode=control-group`, and `openchamber serve --foreground`
- WSL2 Ubuntu server exposed to Windows OpenChamber Electron as a saved remote instance at `http://localhost:7897`
- Observed across Web versions 1.16.1–1.17.1

## Steps to reproduce

1. Install `@openchamber/web` globally with npm.
2. Run it as a user systemd unit using `serve --foreground`, retaining the normal `KillMode=control-group`.
3. Trigger the Web update flow directly, or use the Electron Remote instance **Update** button for that server.
4. Observe the service logs and installed package version.

## Actual behavior

The update handler starts its package-manager work as a detached child and then terminates the foreground server process. systemd cleans up the service cgroup, killing that detached updater before `npm install` has completed. systemd restarts the unchanged package, while the UI waits for a version that will never appear.

## Expected behavior

The update should complete through a service-manager-safe job that survives the current server process, or return a clear service-managed/manual-update error while leaving the server available.

The update path must not require weakening `KillMode=control-group`, since that would risk orphaned processes.

## Proposed direction

1. Detect a foreground server launched by systemd.
2. Queue package installation in a separate systemd user transient unit, passing the service’s existing executable PATH.
3. Keep the current server online while the package manager runs.
4. Restart the original service only when the package-manager command succeeds.
5. Return the queued unit/job id to the UI, with a journal log command for diagnosis.
6. If no compatible service manager is detected, return a clear `409` error before any server shutdown.

## Non-goals

- Weakening systemd process cleanup.
- Requiring users to patch installed npm files or `app.asar`.
- Reporting success merely because an updater child started.

I have reproduced the issue on WSL and verified a local proof of concept using a separate systemd user unit. The upstream implementation should be a source-level fix with tests, not the local compatibility patch.

SYU8384:
## Intent

Fix #2541 by making `POST /api/openchamber/update-install` safe for a Web server launched with `openchamber serve --foreground` from a `systemd --user` service. The update is queued in a separate transient user unit, so systemd cgroup cleanup cannot kill the package-manager process when the server is restarted.

## Non-goals

- No Electron updater IPC, Windows installer behavior, UI wording, or remote-host transport changes.
- No systemd isolation weakening, persistent worker service, staged rollback artifacts, or package-manager redesign.
- No change to daemon-mode or container-mode updater behavior.

## Affected surfaces

- `packages/web` OpenChamber update-install HTTP route and its explicit external response contract.
- Web/PWA users and Electron remote-instance users whose active runtime is a systemd-managed Web server.
- Foreground services without a valid systemd user context now receive an explicit `409` rather than the existing unsafe detach-and-exit behavior.
- The configured service unit defaults to `openchamber.service`; a service can override it with `OPENCHAMBER_SYSTEMD_UNIT`.

## Repository guidance

|---|---|---|
| `AGENTS.md` runtime boundaries and correctness invariants | This changes a Web-owned server route used by multiple clients. | The behavior stays in `packages/web`, preserves existing daemon/container paths, makes the service-manager fallback explicit, and exposes a job id/log command without credentials. |
| `openchamber-change-discipline` | Executable server source and an external HTTP contract change. | The diff is limited to the updater branch and its owning documentation; failure returns a clear response before shutdown. The follow-up adds route-level regression coverage without expanding the runtime behavior. |
| `ui-api-decoupling` and implementation/runtime-parity references | The route is consumed through runtime-aware UI transport, including Desktop remote instances. | No UI hardcodes an endpoint or credential; this route remains explicit and is handled by the Web runtime used by Electron. |
| `packages/web/README.md` and `packages/web/server/lib/opencode/DOCUMENTATION.md` | The documented systemd foreground mode owns this route. | The owning module documentation describes the transient-unit behavior and `OPENCHAMBER_SYSTEMD_UNIT` override. |

## Validation

| Check | Result |
|---|---|
| `bunx vitest run packages/web/server/lib/opencode/openchamber-routes.test.js` | Passed: route-level coverage rejects foreground processes outside systemd, rejects invalid unit overrides before spawning, and verifies the transient-unit request and returned job metadata. |
| `bun run type-check:web` | Passed. |
| `bun run lint:web` | Passed. |
| `bun run dead-code` | Passed (non-blocking report contains existing repository findings; no finding refers to the new regression test). |
| `node --check packages/web/server/lib/opencode/openchamber-routes.js` | Passed. |
| `systemd-run --user --collect --unit=openchamber-update-pr-smoke-20260731 --service-type=exec /bin/true` in WSL | Passed; user-systemd accepted and executed a transient service, exercising the service-manager mechanism used by this change. |
| Isolated PR package build (`bun install --frozen-lockfile`, then `bun run build:web`) | Passed. Vite emitted its existing font-resolution and chunk-size warnings. |
| Isolated foreground-systemd update integration on port 7898 | Passed. A `1.17.0` PR build, running with `Restart=always` and `KillMode=control-group`, queued `openchamber-update-1785474922047.service`; the transient unit completed the isolated package-manager simulation, restarted the original service, and `/health` reported `1.17.1` with managed OpenCode running. |
| Live Desktop → WSL acceptance test on port 7897 | Passed. With the compatibility bridge temporarily disabled and the exact PR handler installed on `1.17.0`, the desktop app's Update action queued `openchamber-update-1785481077698.service`. The systemd job completed, restarted the service once, and `/health` then reported `1.17.1`, `runtime: "web"`, and `openCodeRunning: true`; the desktop app reconnected successfully. |

## Corrected diagnosis

The Windows desktop updater was not failing in the reported case: it downloaded the `1.17.2` installer, applied it on restart, and updated the installed desktop app. That path intentionally does not mutate the selected WSL server. The remote-instance entry point subsequently updated the WSL service successfully through the systemd-safe server path. This PR changes only that remote server-update reliability path.

## Visual evidence

No rendered UI changes. The existing update dialog receives the route's normal successful response and only the server-side updater lifecycle changes. The user-visible behavior needs integration verification on a systemd-managed server rather than screenshots.

## Risks and failure behavior

- If a foreground process is not a systemd user service, or its configured unit name is invalid, the route returns `409` before creating an updater or stopping the server.
- If `systemd-run` cannot queue the transient unit, the route returns `409` with its diagnostic text; the current server remains online.
- If the package-manager command fails inside the transient unit, `set -e` prevents the service restart, leaving the old healthy server running. The returned job id maps to `journalctl --user-unit <job>.service` for diagnosis.
- The worker receives the service process PATH so npm/pnpm/yarn/bun installed outside the default systemd PATH can be resolved. Credentials and client tokens are neither logged nor passed in URLs.
- The integration test uses an isolated fake npm executable to avoid changing the user's production global prefix. It verifies the cgroup boundary, queued transient unit, service restart, and version/health transition; it does not replace package-manager-specific coverage.


SYU8384:
Non-blocking future UX idea: the desktop app updater and selected remote-instance updater are intentionally separate, but users may reasonably expect a single way to keep both current. A future “Update all” coordinator could appear only when both targets have updates, explicitly list the desktop app and current remote instance, update and health-check the remote first, then download/restart the desktop app last. Individual update controls should remain available and unchanged.

SYU8384:
/oc-review Please publish a structured verdict for the current head. The UI clarification was reverted; the net PR diff matches the previously reviewed server-only change.

makeittech (approved):
Clean

makeittech:
This is clean.

PR #2545, opened 2026-07-30
perf: optimize session loading and desktop startup

yulia-ivashko:
## Intent

Implement the evidence-backed optimizations from [OPE-299](https://linear.app/openchamber/issue/OPE-299/analyze-and-optimize-session-loading-requests-and-payloads).

- Stop eagerly requesting older message history after the initial session page; retain explicit loading for viewport underfill, upward scrolling, TimelineDialog, and mobile pagination.
- Scope pagination and timeline ownership by runtime, directory, and session so stale cross-directory results cannot overwrite the active session.
- Add disabled-by-default, bounded, privacy-safe diagnostics for session loading and startup.
- Remove avoidable Electron cold-start work by skipping HMR probes in bundled mode and moving macOS TTS voice enumeration off the server-listen critical path.
- Emit authoritative, exactly-once managed OpenCode bootstrap completion/error timing.

Representative session-open measurements using the same profile scenario:

| Metric | Before | After |
|---|---:|---:|
| API transfer | ~34 MB | ~8.2 MB |
| Message transfer | ~30.2 MB | ~4.45 MB |
| Eager older-history requests | 2 | 0 |

Bundled-development Electron startup measurements:

| Metric | Before | After |
|---|---:|---:|
| Server-ready samples | 2995 / 2259 / 1971 ms | 1118 / 1105 / 1048 ms |
| Server-ready median | 2259 ms | 1105 ms (~51% faster) |
| Application document readiness | ~3.44-4.07 s | ~2.46-2.88 s |

## Non-goals

- No visual redesign or user-facing copy change.
- No OpenCode SDK, persisted-data, authentication, or external API contract change.
- No change to explicit older-history loading or incomplete-session recovery semantics.
- No claim that the startup result represents a newly signed/installed production artifact; measurements use bundled development as the closest repeatable packaged approximation.

## Affected surfaces

- `packages/ui`: shared web, Electron, VS Code, hosted-mobile, and Capacitor session loading, timeline ownership, auto-follow, and diagnostics. All runtimes retain explicit demand-driven pagination.
- `packages/web`: server startup, managed/external OpenCode lifecycle diagnostics, proxy readiness timing, and macOS TTS capability detection.
- `packages/electron`: startup URL probe selection and startup timing. Bundled mode no longer probes absent HMR API/UI endpoints; HMR development retains those probes.
- `scripts/profile-browser*`: opt-in sanitized session-load performance extraction and profiler documentation.
- No persisted schema or public API response shape changes. Diagnostics are in-memory/log-only, disabled by default, bounded to 1,000 records, and exclude runtime keys, directories, session IDs, URLs, credentials, and message content.

## Repository guidance

|---|---|---|
| Root `AGENTS.md` and `openchamber-change-discipline` | Cross-package source, test, export, documentation, and runtime changes | Kept ownership in UI sync, web lifecycle, and Electron shell modules; updated owning documentation; added focused tests and ran package-scoped checks plus dead-code analysis. |
| `sync-state-invariants` and `packages/ui/src/sync/DOCUMENTATION.md` | Session bootstrap, pagination, reconciliation, stale async results, and directory-scoped state changed | Preserves authoritative errors and explicit recovery, keys ownership by runtime/directory/session, guards stale completion, and tests cross-directory and ABA transitions. |
| `performance-engineering` | Request volume, payload size, rendering readiness, and startup critical paths are the purpose of the change | Measured before changing behavior, removed only demonstrated waste, retained demand-driven loading, and provided repeatable before/after measurements. |
| `ui-api-decoupling` | Shared UI OpenCode requests and runtime/directory switching are affected | Continues using the existing SDK/runtime boundary and makes runtime-specific identity explicit without introducing direct platform access in shared UI. |
| `desktop-shell` and `packages/electron/README.md` | Electron startup probing and packaged/bundled behavior changed | Keeps Electron entrypoint changes narrow, extracts the probe policy into a pure tested helper, and documents diagnostics and bundled/HMR behavior. |
| `packages/web/server/lib/opencode/DOCUMENTATION.md` | Managed OpenCode startup and readiness timing changed | Adds one authoritative ready/error terminal event, preserves lifecycle cleanup, and documents timing/privacy contracts. |

## Validation

| Check | Result |
|---|---|
| `bun run --cwd packages/web test` | Passed: 104 files, 883 tests passed, 1 skipped. |
| `bun test packages/ui/src/sync/session-message-loader.test.ts packages/ui/src/components/chat/hooks/useChatTimelineController.test.ts` | Passed: 43 focused UI regression tests. |
| `bun test packages/ui/src/sync/session-message-loader.test.ts` | Final touched-loader rerun passed: 15 tests. |
| `bun test scripts/profile-browser-session-load.test.mjs` | Passed: 3 profiler sanitization tests. |
| `bun run --cwd packages/electron test:architecture` | Passed, including bundled/HMR startup URL selection coverage. |
| `bun run type-check:ui` / `bun run type-check:web` / `bun run type-check:electron` | All passed. |
| `bun run lint:ui` / `bun run lint:web` / `bun run lint:electron` | All passed. |
| `bun run build:web` | Production web build passed. Existing KaTeX font, chunk-size, and mixed import warnings are unchanged. |
| `bun run dead-code` | Completed with no new file/export findings; the command remains non-blocking by design. |
| Bundled-development cold-start profiling | Three before/after runs produced the startup measurements in Intent. |

Not verified: a newly signed/installed Electron production artifact, Windows/Linux runtime startup, and physical hosted/Capacitor mobile devices. Shared UI correctness is covered by focused state tests and static checks, but those do not replace platform runtime validation.

## Visual evidence

No rendered UI, text, styling, or interaction affordance changes. The behavioral change is network/startup scheduling and opt-in diagnostics, so screenshots would not show a meaningful difference. Current-HEAD before/after payload and startup timing evidence is included in Intent.

## Risks and failure behavior

- Lazy older-history loading could expose pagination gaps; existing viewport-underfill, upward-scroll, TimelineDialog, and mobile explicit paths remain, with focused regression coverage.
- Directory/runtime/session switches can race async pages; ownership now uses the full composite identity and stale results are ignored without erasing authoritative data.
- Fetch failures remain failures rather than becoming authoritative empty success.
- Diagnostics are inert when disabled, clear browser marks, bound retained event/completed-identity state, and emit only fixed allowlisted scalar fields.
- Concurrent macOS voice enumeration no longer delays listener publication; `/api/tts/say/status` still awaits the authoritative capability result.
- Bootstrap emits exactly one ready/error terminal event across managed, reused, external, and failed paths; lifecycle cleanup behavior is unchanged.
- Bundled Electron skips only unavailable HMR probes; HMR and production probe plans have separate focused test coverage.
- Rollback is a single commit and requires no persisted-data migration or cleanup.


btriapitsyn:
Added a follow-up fix for history-prepend stability in `f8f591d42`.

Root causes addressed:
- Virtualized history had two scroll-compensation owners: TanStack already preserves keyed prepends, while the timeline controller also restored the DOM anchor. Depending on measurement timing, the same prepend could be compensated twice or fall back to `scrollToIndex`, causing downward jumps.
- Mobile short histories could cross the 5-entry virtualization threshold during `Load older`, replacing the plain DOM history subtree with a newly initialized end-anchored virtualizer. That produced a visible blink and initialized at the bottom before prior keyed state existed.

The controller now leaves already-virtualized prepend compensation exclusively to TanStack, with a one-time restore only for a genuine non-virtualized → virtualized transition. Mobile history uses TanStack from its first render, so pagination no longer changes render engines. Manual mobile verification confirmed stable older-history loading without the jump.

Validation passed:
- `bun test packages/ui/src/components/chat/hooks/useChatTimelineController.test.ts packages/ui/src/sync/session-message-loader.test.ts` (21 passed)
- `bun run type-check:ui`
- `bun run lint:ui`

btriapitsyn:
Follow-up on what landed in this PR after the original optimization commit, and why.

## 1. `f8f591d42` — fix(chat): stabilize history prepend virtualization

The original change made older-history loading strictly demand-driven. That put the "load older" prepend path under much heavier real use and exposed a pre-existing instability: when a short message list crossed the virtualization threshold during a prepend, the history subtree remounted and the end-anchored virtualizer initialized at the bottom without prior keyed state, throwing the viewport. There were also two competing scroll-anchor writers (the DOM anchor restore and TanStack reconciliation), which could apply the same prepend delta twice.

Fixes: mobile now starts on the same virtualized engine it will use after pagination (no engine switch mid-prepend), TanStack is the sole scroll owner once virtualized, and the DOM anchor restore is allowed only for the one threshold-crossing commit where the virtualizer has no prior state.

## 2. `8cf4b2d09` — perf: unblock first session open from startup network contention

While validating the PR I chased a report that the first session opened after app start stalled for seconds while every later open was instant. CDP network capture plus a Chromium net-log against the packaged desktop app showed the message fetch itself was fast server-side (~40 ms) but sat in the browser socket queue for up to ~5.4 s. Three independent contributors, three fixes:

- **Startup request storm vs. the ~6-connections-per-origin pool.** The active-session watchdog fired an uncapped per-directory status poll + child-session discovery burst at startup, alongside per-project Git checks, global session pages, and command/skill discovery. Added a shared background-network gate (`packages/ui/src/lib/background-network.ts`, concurrency 3) and routed the actual background owners through it: watchdog calls, global session pages, command/skill discovery, sidebar project status/root-branch reads, and startup worktree discovery. Interactive Git reads intentionally remain outside this queue (see follow-up below).
- **Packaged-desktop CORS preflight starvation.** The packaged renderer is cross-origin (`openchamber-ui://` → loopback) and attaches the client bearer, so every API URL needs a CORS preflight. A few slow OpenCode-proxied requests held the entire pool while preflights and interactive traffic queued behind them — invisible in same-origin dev. Fixed by lifting the Chromium per-host connection cap for loopback (`ignore-connections-limit`) in the Electron shell; loopback has none of the costs the cap protects against. Verified live on the installed app: SSE connect queue 3.9 s → 0, message fetch 34 ms.
- **Lazy OpenCode directory initialization.** OpenCode initializes each directory on its first request, so the first click paid that cost interactively. The lifecycle now warms the last-used directory plus the three most recently opened projects right after OpenCode readiness — sequential, best-effort, overlapping UI startup.

Worth recording for posterity: the dominant multi-second delay on my machine ultimately turned out to be an external OpenCode plugin blocking directory initialization (removed from my config and reported upstream), but all three contentions above were real, measured, and reproduced independently of it.

## 3. `45d4b489b` — fix(ui): keep interactive Git reads out of background queue

A correctness review caught that the first implementation put the background policy inside the low-level `gitApiHttp` transport. That also queued user-triggered reads such as branch loading, worktree creation/discovery from interactive surfaces, root resolution, and explicit Git refreshes behind startup work.

The follow-up moves ownership to the call sites that are actually background work. Low-level Git reads are neutral again; sidebar project status/root-branch refresh and startup worktree discovery retain the shared gate. This preserves the startup contention fix without making interactive Git actions wait for the background FIFO. A regression test verifies that low-level reads no longer inherit background request priority.

Validation for this follow-up: `bun test packages/ui/src/lib/gitApiHttp.test.ts packages/ui/src/lib/background-network.test.ts` (8 passed), `bun run type-check:ui`, `bun run lint:ui`, and `git diff --check` all passed.

Broader validation for the preceding commits: new background-network + lifecycle warmup tests, focused store/sync tests, UI type-check/lint, dead-code report, `node --check` + Electron type-check/lint, and repeated CDP first-open measurements against both a production-build web server and the installed packaged app.

btriapitsyn (approved):

PR #2554, opened 2026-07-31
fix(chat): normalize bash output by stripping ANSI sequences and applying terminal control codes

catan271:
## Intent

Bash tool output in the chat UI displays raw ANSI escape sequences and terminal control codes (carriage returns, cursor-up, clear-line) instead of rendered text. This makes progress bars, colored output, and similar terminal sequences unreadable.

This PR adds a `renderTerminalOutput` function that interprets ANSI CSI sequences (cursor movement, erase-line), carriage returns, and backspaces to produce clean rendered text. The normalization is applied only to completed bash tool output — running output is left raw to avoid interfering with live streaming.

## Non-goals

- Rendering bash output in a real terminal emulator component (this is text-level normalization only)
- Normalizing output for non-bash tools

## Affected surfaces

- `packages/ui`: shared chat message rendering — `ToolPart.tsx`, `toolOutput.ts`, `ToolPart.test.ts`
- User-visible: completed bash tool results now show clean text instead of raw escape codes
- No persisted contracts or API changes

## Repository guidance

|---|---|---|
| AGENTS.md — shared UI contracts | Change is in `packages/ui` shared component | No new runtime-specific behavior; normalization applies uniformly |
| AGENTS.md — correctness invariants | Output rendering must not mask failures | Normalization only applies to completed state; running output is untouched |

## Validation

| Check | Result |
|---|---|
| `bun test packages/ui/src/components/chat/message/parts/ToolPart.test.ts` | 21 tests pass |
| TypeScript type-check | Passes |

## Visual evidence

No screenshots — the change affects text content rendering (escape code stripping), not layout or styling. Before: raw `\u001B[32m` sequences visible in output. After: clean text with colors/styles stripped and progress lines collapsed to their final value.

## Risks and failure behavior

- If `renderTerminalOutput` encounters malformed escape sequences, it stops processing at the failure point and returns what it has — no crash or data loss
- Running bash output is intentionally not normalized, so live streaming behavior is unchanged
- Rollback: revert the single commit; no migration or persisted state involved

### DIFF
--- a/packages/ui/src/components/chat/message/parts/ToolPart.test.ts
+++ b/packages/ui/src/components/chat/message/parts/ToolPart.test.ts
@@ -1,5 +1,6 @@
 import { describe, expect, test } from 'bun:test';
 
+import { renderTerminalOutput } from './toolOutput';
 import { readTaskTagSessionIdFromOutput } from './taskSessionIdParser';
 import { tryParseJsonOutput } from '../toolRenderers';
 import { getStreamingThrottleText } from '../../hooks/useStreamingTextThrottle';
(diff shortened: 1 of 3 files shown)


btriapitsyn:
Thanks for working on this — cleaning up completed bash output is a useful improvement. I found two issues that should be addressed before merge:

1. **The renderer can become quadratic and freeze the UI on large output.** `writeTerminalCharacter` rebuilds the entire current line with `slice()` for every ordinary character. A long single-line result containing any ANSI/control sequence (for example minified JSON, base64, a bundle, or a large colored log line) therefore takes O(n²) work on the main thread when the completed output is rendered. Please use a mutable character/chunk representation and join it once at the end, and add a representative large single-line regression test or benchmark.

2. **`CSI 1 K` is implemented with incorrect terminal semantics.** `ESC[1K` should erase from the beginning of the line through the cursor position inclusive while keeping the cursor and the remaining suffix in their original columns. The current `line.slice(column)` implementation shifts the suffix left and resets the cursor to column 0, producing incorrect output. Please replace the erased range with spaces without moving the cursor, and add focused tests for `K`, `1K`, and `2K`.

I also recommend attaching a small before/after screenshot using a real progress-style bash output, since this is a user-visible rendering change.

catan271:
Thanks for the detailed review! Both issues are now addressed:

1. **O(n²) performance** — `writeTerminalCharacter` now uses mutable `string[][]` arrays instead of rebuilding strings with `slice()` on every character. Character writes are O(1) array assignments, and the final result is joined once at the end. Added a regression test with 50k-character single-line output that verifies completion under 1 second.

2. **`CSI 1 K` semantics** — now replaces the erased range (columns 0 through cursor) with spaces instead of slicing the suffix left. The cursor position is preserved. Added focused tests for `CSI K`, `CSI 1 K`, and `CSI 2 K`.

Before/after screenshots from a real progress-style bash output:

**Before:**
before1
before2


**After:**
after1
after2

btriapitsyn (approved):

PR #2580, opened 2026-08-02
fix(mobile): cascade session archiving and deletion to subagents

alexandrereyes:
## Intent

Mobile parent-session actions previously left descendant handling inconsistent. OpenCode does not cascade `time.archived`, while hard delete does cascade on the backend but direct client cleanup covered only the selected parent and depended on every child `session.deleted` event arriving and routing correctly.

After this change, mobile archive and hard delete both operate on the whole known subtree in post-order, one session at a time. Archive targets the root plus known active descendants and traverses through archived intermediates without retimestamping them. Hard delete targets the root plus every known descendant, including archived sessions, so each known identity receives direct live-store, cache, and persisted-state cleanup even when the backend already removed it and returns `404`.

Both operations capture the runtime and every target directory before the first request, stop on the first failure or runtime switch, and never mutate an ancestor after a known descendant failed. The mobile active list continues to treat global archived membership as negative authority so stale live copies cannot resurrect archived rows.

## Non-goals

- No backend archive cascade or recursive archive fetch for descendants the client has never observed.
- No replacement of the backend hard-delete cascade; unknown descendants still rely on it.
- No change to desktop, VS Code, worktree deletion, or bulk-selection behavior.
- No layout, styling, gesture, icon, dependency, persisted schema, server route, or locale-key change.

## Affected surfaces

- `packages/ui/src/apps/MobileSessionsSheet.tsx`: mobile archive/delete orchestration and existing feedback.
- `packages/ui/src/apps/mobileSessionArchive.ts`: known-subtree collection, captured directories, post-order execution, runtime checks, and stop-on-first-failure behavior.
- `packages/ui/src/apps/mobileSessionArchive.test.ts`: archive/delete topology, archived intermediates, cross-directory targets, cycles, failures, and runtime switches.
- `packages/ui/src/sync/session-actions.ts`: captured delete/archive directory support plus direct confirmed-delete cache and persistence cleanup.
- `packages/ui/src/sync/session-actions.test.ts`: `404`, runtime, cache cleanup, and `sessionTotal` event reconciliation coverage.
- `packages/ui/src/sync/DOCUMENTATION.md`: documents the mobile mutation contract and known-descendant boundary.
- Hosted/PWA mobile, Capacitor mobile, and tablet mobile sidebar share this component. Desktop and VS Code do not render it.

## Repository guidance

|---|---|---|
| Root `AGENTS.md` and `openchamber-change-discipline` | Shared executable mobile behavior with destructive partial results | Keeps ownership local, adds no dependency, makes partial completion explicit, and validates focused behavior plus the production web/mobile build. |
| `sync-state-invariants` and `packages/ui/src/sync/DOCUMENTATION.md` | Archive/delete reconciliation, runtime switching, caches, and persisted cleanup | Captures runtime/directory authority before requests, rejects stale responses, directly clears confirmed deletes, and leaves root-count reconciliation to the authoritative event to avoid double decrement. |
| `ui-api-decoupling` | Shared UI mutations reach OpenCode through runtime-aware actions | Reuses `useSessionUIStore` and `session-actions`; no direct feature URL, bridge, proxy, or server route was added. |
| `performance-engineering` | Session-list traversal and cache cleanup | Subtree work is bounded linear work triggered only by a destructive user action, not a render/event hot path. |
| `locale-ui-patterns` and `theme-system` | Existing user-visible toasts and mobile controls are reused | Adds no copy or visual token and reuses localized singular/plural feedback. |
| `packages/mobile/README.md` | Hosted and Capacitor mobile consume the shared production mobile surface | Production `mobile.html` was exercised at a phone viewport against an isolated OpenCode instance. |

## Validation

Final HEAD: `7e93b88d22a2427dc963293b703b12a2862d0613`

| Check | Result |
|---|---|
| `PATH=/workspace/.bun/bin:$PATH /workspace/.bun/bin/bun test packages/ui/src/apps/mobileSessionArchive.test.ts` | Passed: 21 tests, 0 failures. |
| `PATH=/workspace/.bun/bin:$PATH /workspace/.bun/bin/bun test packages/ui/src/sync/session-actions.test.ts` | Passed: 49 tests, 0 failures; expected negative-path logs were emitted. |
| Focused event/scoped cleanup tests | Passed individually: event reducer, sync-context session events, scoped blocking requests, and session deletion cleanup. The combined Bun process exposed a known mock-isolation collision, while the affected file passed alone (5 tests, 0 failures). |
| `PATH=/workspace/.bun/bin:$PATH /workspace/.bun/bin/bun run --cwd packages/ui type-check` | Passed with no diagnostics. |
| `PATH=/workspace/.bun/bin:$PATH /workspace/.bun/bin/bun run --cwd packages/ui lint` | Passed with no warnings. |
| `PATH=/workspace/.bun/bin:$PATH /workspace/.bun/bin/bun run dead-code` | Completed with existing informational findings; no new mobile helper/export finding. |
| `PATH=/workspace/.bun/bin:$PATH /workspace/.bun/bin/bun run build:web` | Passed; Vite built production desktop/mobile entries and service worker. Existing KaTeX, mixed import, and chunk-size warnings remain. |
| Browser Control `e2e-pr-7e93b88d2-20260804`, 390x844, isolated production mobile app | Passed on exact HEAD. Archive of a three-level tree showed `Archived 3 sessions`; all three records had `time.archived` and stayed absent after reload. Hard delete of another three-level tree used the two-step mobile confirmation, showed `Deleted 3 sessions`, and all three IDs returned `404` and stayed absent after reload. |
| E2E cleanup | Passed. All archive/delete fixtures were removed, active and archived prefix searches returned empty, the Browser Control session and temporary directory were removed, and both isolated app/OpenCode ports were confirmed closed. |

Not run:

- Workspace-wide test/type/lint suites were omitted because the executable change is confined to `packages/ui`; focused sync tests and the production web/mobile build cover the affected consumers.
- Electron, VS Code Extension Host, and native iOS/Android packaging were omitted because they do not change the mobile component contract; production mobile web behavior was exercised directly.
- No performance benchmark was run because this adds bounded user-triggered work and no regression was reported.

## Visual evidence

No screenshot or recording is attached. Current interactive evidence was collected against exact HEAD through Browser Control at 390x844:

- Archive: expanded root, child, and grandchild; swiped the root; observed `Archived 3 sessions`; verified all rows remained absent after reload.
- Delete: expanded a separate three-level tree; swiped the root; completed the two-step delete confirmation; observed `Deleted 3 sessions`; verified all rows remained absent after reload.
- The diff changes no styling or copy. The remaining evidence gap is an attached recording rather than an untested interaction.

## Risks and failure behavior

- The executor runs known descendants before ancestors and stops globally on the first failure. Confirmed earlier descendants remain changed; the failed target and every unattempted target stay retryable.
- Archive cannot reach descendants missing from all client snapshots. Hard delete still reaches unknown descendants through the upstream backend cascade, but only known identities receive direct client cleanup.
- A runtime switch rejects stale responses before they can reconcile the new runtime. Session IDs are not assumed unique across runtimes.
- Captured per-target directories prevent earlier child removals from changing routing for later targets, including cross-directory descendants.
- Delete `404` is success only while the captured runtime remains authoritative; it performs the same direct cleanup as a confirmed delete.
- Direct delete removes live/cache/persisted state but leaves `sessionTotal` to the authoritative `session.deleted` echo, preventing a double decrement.
- No speculative rollback is attempted after a server-confirmed archive/delete. User feedback reports confirmed successes and remaining failures.
- Reverting the latest commit restores archive-only mobile cascade behavior; no migration or persisted-data conversion is required.


alexandrereyes:
Updated draft to HEAD `095d2194d` after a normal merge of current `upstream/main`. Browser Control mobile validation on port 44580 passed: five active sessions archived deepest-first with the root last, including cross-directory descendants and a leaf below an already archived intermediate; the UI showed `Archived 5 sessions` and no fixture row returned after reload. Full IDs and validation details are now in the PR body.

yulia-ivashko:
Closing in favor of [#3565 · fix(sessions): archive and…](https://github.com/openchamber/openchamber/pull/3565), which covers this on every surface through the sidebar's existing actions. Your two observations, the archived intermediate and the discovery race, are in it, and you're credited as co-author on that commit. Thanks for tracing them.

Issue #1472, opened 2026-05-30
Optimize API requests on project expansion in sidebar


tomzx:
# Background

When expanding a project in the sidebar that has many sessions (e.g., 95 sessions for the openchamber repo), the UI triggers a large number of repetitive API calls that return the same data. The observed calls include:

- /api/provider
- /api/config
- /api/path
- /api/session/status
- /api/agent
- /api/command
- /api/mcp
- /api/lsp
- /api/mcp (duplicated)

That is 9 API calls, multiplied by the number of sessions, leading to hundreds of redundant requests.

## Root cause analysis

Investigation of the codebase reveals two main sources of redundancy:

1. **Per-directory bootstrap runs full API suite.** `bootstrapDirectory` (packages/ui/src/sync/bootstrap.ts) makes ~12 API calls per directory. While sessions within the same directory share a child store, the bootstrap itself does not skip calls for data already available from the global bootstrap (`bootstrapGlobal`). The global bootstrap pre-fetches providers, config, projects, and path info, but each directory bootstrap re-fetches all of them anyway.

2. **Overlap between bootstrap and config store.** When a session is selected, `activateConfigForDirectory` (packages/ui/src/stores/useConfigStore.ts) fetches providers and agents again through the config store, independently of the bootstrap system. In-flight deduplication only works within each subsystem, not across them, so the same data is fetched twice in parallel.

3. **No cross-directory deduplication.** Projects that share the same config (same provider list, same agents, same MCP servers) still trigger independent full bootstraps for each directory.

# Version

a394a87

# Acceptance Criteria

- [ ] Expanding a project with many sessions does not multiply API calls per session; shared data (providers, agents, config, commands, MCP, LSP) is fetched once per project/directory, not per session
- [ ] Data already available from `bootstrapGlobal` is reused by `bootstrapDirectory` without re-fetching when the response would be identical
- [ ] Overlap between `bootstrapDirectory` and `activateConfigForDirectory` for providers and agents is eliminated; a single fetch satisfies both consumers
- [ ] The duplicate /api/mcp call is removed
- [ ] Network tab confirms a bounded number of requests when expanding a project with 50+ sessions (no linear scaling with session count)

---

Created with [create-issue](https://github.com/tomzx/dot-claude/blob/cd20e18bfda514940b68231081ee11af6adad709/skills/create-issue/SKILL.md) (`cd20e18`)

tomzx:
[...]

PR #2598, opened 2026-08-03
perf: fix directory cache thrashing and runtime-key derivation, add unattended profiling harness

btriapitsyn:
## Intent

@tomzx reported sustained ~100% CPU with OpenChamber idle, plus a sawtoothing JS heap and JS event listeners climbing from ~5k to ~11k, on v1.17.2 with only the sidebar's Recent section open ([screenshot in the report](https://github.com/openchamber/openchamber/issues/1472)).

This PR fixes the cause he identified — directory cache thrashing (#1472) — removes a second measured hot spot on the streaming path, and adds the unattended profiling tooling the investigation needed, since none of this was measurable without a human driving a browser.

**Cache thrashing (#1472).** Expanding a project with more worktrees and sessions than `MAX_DIR_STORES` put the sidebar into an endless request loop. Every sidebar row calls `ensureChild` during render, but the pin protecting that directory is only taken in an effect after commit. `ensureChild` marked the directory and ran eviction synchronously, so directories that were actively rendering looked unpinned and were disposed; the next render recreated them with a `loading` status, which issued another bootstrap request, and the cycle repeated for as long as the project stayed expanded. That matches the reported symptoms: continuous requests, listener churn, and heap sawtooth.

Raising the limit (@tomzx's local `MAX_DIR_STORES = 200`) works but relocates the cliff and hides the loop from everyone with a smaller workload. Instead the limit becomes a soft target: a directory touched within a 30 s grace window is never an overflow victim, so a burst of live directories overflows briefly rather than thrashing, while idle-time (TTL) eviction still bounds the cache. Eviction is additionally coalesced into one deferred pass per tick, so a render mounting many rows no longer sorts and scans every directory once per row, and a commit's pin effects settle before anything is considered for disposal. Releasing the final consumer stays synchronous, since that is an explicit lifecycle edge with an existing contract.

**Runtime key derivation.** `getRuntimeKey` keys caches, stores and persisted state across the UI (119 call sites) and runs on store reads, event handling and render paths. Until the runtime endpoint is explicitly initialised — the normal web case — every call re-derived the key by trimming two injected globals and constructing three `URL` objects. In a streaming capture it was the most expensive application function: 315 ms of self time, 12% of all main-thread busy time. It is now cached, validated against the raw untrimmed globals rather than memoised outright, because `switchRuntimeEndpoint` writes `__OPENCHAMBER_API_BASE_URL__` at runtime.

## Non-goals

- **Not a verified end-to-end fix for the reported 100% CPU.** See Validation; the loop needs many worktrees under one project to reproduce, which this environment does not have.
- **Terminal idle rendering.** `ghostty-web` repaints its canvas ~60 times a second for the terminal's whole lifetime, costing ~6% CPU while a terminal is on screen. A patch pausing the loop off-screen was written, measured, found to be a no-op (the viewport already unmounts on tab switch) and reverted. Making the visible terminal damage-driven is real but risks corrupting terminal output, and is deliberately left out.
- **Compositing during streaming.** Continuously running busy indicators account for ~40% of layerization, but the only measured candidate fix (`will-change: opacity`) did nothing, and collapsing the three staggered busy dots into one animation would change the visuals. Left alone.
- No changes to markdown rendering: it is already incrementally cached per block.

## Affected surfaces

| Surface | Effect |
|---|---|
| `packages/ui` sync directory-store lifecycle | Eviction policy and scheduling. Affects every runtime that renders the session sidebar: web, desktop, VS Code, hosted mobile, Capacitor mobile. |
| `packages/ui` runtime switching | `getRuntimeKey` result is cached. Runtime identity keys persisted state, so incorrect caching here would cross-contaminate per-runtime storage; covered by tests in both directions. |
| `scripts/`, root `package.json` | New `profile:idle` and `profile:session` commands. Development tooling only, not shipped to users. |
| `.agents/skills`, `scripts/perf/DOCUMENTATION.md` | Guidance for future changes, plus the tooling entry point linked from the root guide. |

No persisted-format, API, or protocol changes. Cache keys, storage shapes and network contracts are untouched. The eviction change can make the app hold more directory stores briefly, which is a memory-versus-thrash trade, bounded by TTL eviction.

## Repository guidance

|---|---|---|
| `AGENTS.md` / `CLAUDE.md` — always-on constraints | Source, dependency and package-contract change | No dependencies added; `../opencode` untouched; no secrets logged; unrelated worktree changes preserved; domain logic kept in owning modules rather than entrypoints. |
| `performance-engineering` skill | The entire change is a render/store/event hot-path investigation | Contract stated before editing (interaction, scale, budget, path); cost equation written (`rows × events × directories`); work removed structurally rather than by micro-optimising; every retained change backed by a measured difference; unvalidated changes reverted; production builds measured throughout. |
| `sync-state-invariants` skill | Directory-store lifetime, bootstrap, reconciliation and eviction | Authoritative state preferred over heuristics; partial failure cannot trigger destructive cleanup (pinned and blocking-request directories remain non-evictable); cache lifetime and invalidation made explicit; the lifecycle contract for releasing the final consumer preserved, verified by the pre-existing test that initially failed. |
| `ui-api-decoupling` skill | `runtime-switch.ts` owns runtime identity and API base URL resolution | Runtime identity semantics unchanged; the cache observes the same inputs the resolver reads, so runtime switching still invalidates it. |
| `openchamber-change-discipline` skill | Files added, exports changed, `package.json` scripts added | `bun run dead-code` run and inspected; workspace-wide type-check and lint run because the change spans packages and root tooling. |
| `packages/ui/src/sync/DOCUMENTATION.md` | Nearest owning documentation for the sync change | Read before editing. The eviction contract it describes is unchanged in intent — capacity bounded, pinned entries protected — with the limit now soft rather than hard. |

## Validation

| Check | Result |
|---|---|
| `bun run type-check` (workspace) | Pass — ui, web, electron, mobile, root |
| `bun run lint` (workspace) | Pass — all packages |
| `bun test src/sync/*.test.ts` (per file) | Pass — no failures across the sync suite |
| `bun test src/lib/*.test.ts` (per file) | Pass — no failures |
| `packages/ui/src/sync/eviction-thrash.test.ts` (new) | 4 pass — fresh directories survive overflow of 25; directories past the grace window are evicted; TTL-idle directories are evicted even inside the limit; pinned and blocking-request directories are never evicted |
| `packages/ui/src/lib/runtime-switch.runtime-key.test.ts` (new) | 4 pass — includes an operation-count assertion that 50 repeated calls construct zero `URL` objects, plus invalidation when either injected global changes |
| `packages/ui/src/sync/child-store.test.ts` (pre-existing) | Pass — this test initially failed when eviction on `unpin` was deferred, which is why that path was kept synchronous |
| `bun run dead-code` | Run and inspected. Only pre-existing finding relates to this diff's files (`getActiveRelayTunnel` re-export, untouched); no new unused exports |
| Streaming profile, production build, three runs per side, `getRuntimeKey` | Longest task **234 ms → 56 ms** (4.2x, no overlap between sets, ±2/±3 ms spread); main-thread busy **3.87% → 2.77%** (no overlap). `readInjectedLocalOrigin` disappears from the profile entirely, having been 315 ms / 12% of busy. Full tables in the PR comments |
| `bun run profile:animation` | Measures the cost of animating each CSS property; results are recorded in the theme skill's animation contract |

**Not verified:** the #1472 loop end to end. Reproducing it requires more live directories than `MAX_DIR_STORES` under one expanded project. This environment has 29 directories against a limit of 30 — just under the cliff, which is why the original report never reproduced here. A scaled-down attempt (limit 5, grace 0, all projects expanded) did not thrash either, because the dataset lacks many worktrees under a single project. The fix is therefore validated against the mechanism by unit tests, not against the live symptom. @tomzx reproduces it immediately and can confirm with `bun run profile:idle -- --expand-projects` on this branch.

**Also not verified:** a total-work reduction from `getRuntimeKey`. Across three runs per side, busy time normalised per 1000 rendered characters overlaps between the sets, as does long-task count. Response length varied 5.9k-8.4k characters, so the unnormalised busy-time improvement is partly a smaller-response effect. The claim this PR makes is the one the data supports: the single worst blocking task is reliably removed.

**Reproduction attempted at scale and failed.** This environment was built up to 36 expanded projects, all 23 openchamber worktrees with sessions, every "Show more sessions" control clicked, 31k DOM nodes and 5.5k listeners, plus a background streaming session. On unmodified `main` that sits at 0.21% CPU idle and 1.15% with the stream, with no request loop and no listener growth. DOM size, listener count, project/worktree count and mounted session rows are therefore not sufficient on their own to reproduce the report. Details in the PR comments.

## Visual evidence

No user-visible change. The eviction change alters when directory stores are disposed, never what is rendered: pinned, blocking-request and TTL-idle behaviour are unchanged, and the only difference is that a recently-touched directory is not disposed while its rows are mounted — which previously caused visible reload churn rather than any intended state. The `getRuntimeKey` change returns the same string for the same inputs, proven in both directions by test. The remaining diff is development-only scripts and skill documentation.

## Risks and failure behavior

- **Memory.** The directory cache can exceed `MAX_DIR_STORES` for up to the grace window. The bound becomes "directories rendered in the last 30 seconds", which the UI inherently limits, and TTL eviction still reclaims idle directories. This trade is deliberate: brief overflow is cheaper than evict/recreate cycles.
- **Deferred eviction.** Eviction now runs on a microtask rather than inside `ensureChild`. A directory therefore survives microseconds longer than before. `disposeAll` and the `unpin` lifecycle edge remain synchronous, and the scheduler is guarded against running after disposal.
- **Runtime-key cache.** The genuine risk is a stale key silently cross-contaminating per-runtime persisted state. This is why the cache compares raw globals instead of memoising: `switchRuntimeEndpoint` writes `__OPENCHAMBER_API_BASE_URL__` at runtime, and hard memoisation would have been a bug. Tests cover invalidation for both injected globals.
- **Rollback.** Each commit is independently revertible; the two product fixes touch disjoint modules.
- **Performance.** Both changes reduce work on paths measured with a production build. Neither introduces new caching without explicit invalidation.
- **Cross-runtime.** The sync change affects every runtime rendering the sidebar. Its logic is runtime-agnostic and covered by unit tests on the pure eviction function; it was exercised at runtime only in the web runtime.

## Follow-up leads (measured, not addressed here)

1. **Activity indicators drive style recalculation.** With one session streaming in the background and an unrelated idle session on screen, style recalculation runs at 7.2/s here, matching `UpdateLayoutTree` exactly and driven by two continuously running spinner animations — the sidebar itself re-rendered only 4 times in 105 s. @tomzx reports 80/s in the same situation, an 11× gap that again points at scale (number of simultaneous indicators). This is the most likely remaining contributor to his idle CPU.
2. **Compositing during streaming** is the largest main-thread renderer cost (~910 ms of ~2650 ms busy), ~40% attributable to busy indicators.
3. **The spinner costs ~7 style recalculations per second inside the app, and zero in isolation.** Disabling `.animate-spin` in the app drops style recalculation from 6.97/s to 0.03/s, yet the identical CSS — including the repository's own `webkit-spin` keyframes, `transform-box: fill-box` and `will-change` overrides — measures zero in `profile:animation`, under every ancestor context tried and at 15,000 filler elements. Four hypotheses were tested and falsified; the mechanism is unidentified. At ~0.5% CPU it is far inside budget, which is why it was not pursued further.
3. **Telemetry gap.** The update check sends only `installId`, `platform` and `version`. We collect nothing about session, project or worktree counts, so we cannot currently tell whether a reporter's scale is an outlier — which is exactly the question that blocked reproduction twice in this investigation.

## Tooling added

| Command | Purpose |
|---|---|
| `bun run profile:idle` | What the app does while nobody interacts with it. `--session`, `--tab`, `--then-tab`, `--panel`, `--expand-projects` reach specific mounted states; `--baseline` and `--budget-*` make it a regression gate. |
| `bun run profile:session` | What a streaming response costs. Creates a session, dispatches a prompt through the `openchamber session` CLI, records until the session reports idle. Reports long-task distribution, timeline-trace breakdown, running animations, and output-normalised metrics. `--view-session` measures an idle session while another streams. |
| `bun run profile:animation` | What a CSS animation costs, isolated from the app. `transform`, `opacity` and `filter` cost zero extra style recalculations at any element count; the individual `rotate` property, `background-position`, `border-color` and `box-shadow` each cost 60/s, and geometry properties add layout on top. |

`scripts/perf/DOCUMENTATION.md` documents all of it — commands, how to stand up a production build to measure against, how to read the artifacts, the validity guarantees, and the methodology rules — and is linked from the root guide so it can be handed to an agent as the entry point for measuring performance.

Both fail loudly instead of reporting a clean result when the renderer was throttled, the trace collected no tasks, or the scenario never rendered — each of which produced a false clean result during this work before being turned into a check.


btriapitsyn:
## Reproduction attempt at reported scale — negative result

Following up on the "not verified end to end" note. I scaled this environment up to match @tomzx's reported state on every dimension visible in his Performance monitor screenshot, and the app stays quiet.

**What was built up**

| Step | Result |
|---|---|
| 30 new git repositories, one session each | Server-side projects 18 → 48. No effect: the sidebar renders the *client's* project list, not the server's |
| A session in each of the 23 openchamber worktrees | Worktree rows appeared, DOM 22.8k → 24.2k nodes |
| Seeded the client project list to 36, all expanded | DOM 23.0k nodes, listeners peaked ~9.3k during load then settled to 1.5k |
| Clicked every "Show more sessions" control (52 lists) | **DOM 31.1k nodes, 5.5k listeners** |
| Added a background session actively streaming | — |

For comparison, the report shows ~40k DOM nodes and listeners climbing 5k → 11k.

**Measured at that scale, on the unmodified `main` build**

| Scenario | Main-thread busy | Style recalcs/sec | Listener growth | Fetches/sec |
|---|---|---|---|---|
| All projects + all sessions expanded, idle | 0.21% | 0 | 0 | 2.6 |
| Same, plus a background session streaming | 1.15% | 6.97 | 0 | — |

No request loop, no listener growth, no heap growth, and two long tasks across a 98 s window. The style recalculation rate is ~7/s and does **not** scale with sidebar size — it is the same ~7/s measured with a small sidebar, and it tracks `UpdateLayoutTree` exactly, driven by two continuously running spinner animations. @tomzx reports 80/s in the same situation, an 11× gap that sidebar scale does not explain.

**What this means**

The thrash fix in this PR remains validated against the mechanism by unit tests and **unvalidated against the live symptom** — this attempt does not change that either way. What it does establish is that DOM size, listener count, project count, worktree count and mounted session rows are not sufficient to reproduce the reported CPU on their own, so the remaining difference is something not visible in the screenshot: browser or extension environment, GPU/compositing path, an open panel, or configuration.

@tomzx — if you pull this branch, `bun run profile:idle -- --expand-projects --expand-sessions` and `bun run profile:session -- --view-session <an-idle-session> --expand-projects --expand-sessions` produce the same tables above from your machine. Those numbers would tell us immediately which dimension actually differs.

Tooling added along the way: shared sidebar-expansion scenario setup for both profilers, and a fix for a spread push that overflowed the call stack once a populated sidebar produced trace chunks of hundreds of thousands of events.

btriapitsyn:
## Repeated before/after for the runtime-key cache

The original claim rested on a single pair of runs. Here are three runs on each side, same scenario, same machine, alternating builds (unmodified `runtime-switch.ts` from `main` versus this branch), production builds throughout.

| Metric | Before (3 runs) | Median | After (3 runs) | Median | Ranges |
|---|---|---|---|---|---|
| Longest task (ms) | 232.4, 235.8, 234.0 | **234.0** | 55.6, 59.6, 53.3 | **55.6** | no overlap |
| Main-thread busy (%) | 4.18, 3.87, 3.41 | **3.87** | 2.65, 2.95, 2.77 | **2.77** | no overlap |
| Long tasks (count) | 2, 3, 4 | 3 | 1, 3, 1 | 1 | overlap |
| Busy per 1k rendered chars (ms) | 345, 412, 442 | 412 | 364, 388, 353 | 364 | overlap |
| Rendered characters | 8444, 6681, 6883 | 6883 | 5902, 7393, 6724 | 6724 | overlap |

**What is now solid**

The longest blocking task drops from ~234 ms to ~56 ms, a 4.2× reduction with no overlap between the two sets and remarkably tight spread on each side (±2 ms before, ±3 ms after). That is the metric that decides whether the interface stays responsive while a response streams, and it is the strongest result in this PR.

Main-thread busy time also separates cleanly, 3.87% → 2.77% at the median.

**What is not**

Busy time normalised per 1000 rendered characters overlaps between the two sets, and so does long-task count. Response length varies between runs (5.9k–8.4k characters), so the unnormalised busy-time improvement is partly a smaller-response effect and should not be read as a ~28% reduction in total work. The honest summary is: this change reliably removes the single worst blocking task, and reduces total work by an amount this sample cannot separate from run-to-run variance.

Also fixed while doing this: the longest-task calculation spread every recorded task into `Math.max`, overflowing the call stack on large traces — the same failure already fixed once for trace collection.

PR #2603, opened 2026-08-03
fix(sync): render sessions in worktrees created while the client is running

btriapitsyn:
## Intent

Fixes the reported bug where prompting in a git worktree sometimes does nothing: a brief bubble appears, disappears, and the session stays empty, while `openchamber restart` makes the same worktree work.

The report is accurate and the cause is not what it looks like. The prompt is delivered and the assistant replies — both messages are in the session and visible on any fresh load. What fails is the live client, which addresses every request for that session to a directory that does not own it.

**Ownership was read from containment.** `getAuthoritativeSessionDirectory` decided which directory owns a session by asking which child store holds it. A project's session list includes the sessions of its worktrees so the sidebar can group them, so the parent repository holds worktree sessions too, and whichever store bootstrapped first won. Captured mid-failure, the two signals disagreed outright:

```
owningDirectory  /Users/…/projects/openchamber          <- parent, merely holds it
recordDirectory  /Users/…/worktree/…/au-1               <- the session's own directory
```

Every fetch then went to the parent, the session id resolved to `undefined` there, and requests failed as `GET /api/session/undefined?directory=<parent>` with a 500 in a retry loop. The session's own record is now believed; store membership remains the fallback for a record that carries no directory.

**A guessed directory was never settled.** Selecting a session whose directory the client has not indexed yet routes it through the active directory as a deliberate, documented guess, excluded from both the resolver and persistence. Nothing settled that guess afterwards: `setSessionDirectory` performs exactly that promotion, but only confirmed destinations call it — a completed move, or a worktree this client created. A session in a worktree created outside this client kept the guess forever. The promotion now runs when directory bootstrap completes, which is the moment the authoritative directory first becomes readable.

Both are required. Settling the guess against the containment-derived value adopted the parent directory and then cleared the guess, which prevented any later correction — that is why the first change alone did not fix the symptom.

**Worktree provisioning outlived its client timeout.** Creating a session with a worktree reported `Request to /api/openchamber/control timed out after 4000ms` while the worktree was in fact created, leaving a failure message, a real worktree and no session id. The client extended its timeout only when the caller asked to wait for the session, but provisioning runs git and prepares a directory on its own. On a cold path immediately after a restart it measures ~4.0 s, landing exactly on the 4 s default, which is why it failed intermittently rather than always.

## Non-goals

- **Not a change to delivery.** Sends already worked; only the live client's routing did not. Nothing in the send path is touched.
- **Not the worktree registration change.** `c290ce6a7` (stop writing worktree registration into OpenCode's storage) is on `main` already and targeted the same report from a different angle. It is correct on its own terms — writing into another live process's database is wrong regardless — but it does not affect this failure: the symptom reproduced identically with and without it.
- No change to the resolver precedence in `session-directory-resolution.ts`. Only the value fed into its `authoritative` slot is corrected.
- The `lib/debug.ts` session-directory report builds its `selected` value without the private guess filter, so it can show a conflict the real resolver would discard. Left alone here; noted so the next reader is not misled by it.

## Affected surfaces

| Surface | Effect |
|---|---|
| `packages/ui` session directory ownership | `getAuthoritativeSessionDirectory` prefers the session record. This feeds routing for every send, message fetch, queue key and confirmation lookup, in every runtime: web, desktop, VS Code, hosted mobile, Capacitor mobile. |
| `packages/ui` session selection | New `adoptAuthoritativeSessionDirectory`, called once per completed directory bootstrap. Only ever promotes a guess. |
| `packages/web` CLI | Worktree-provisioning control calls get a timeout sized for the work. Server behaviour is unchanged; the client simply stops abandoning an operation it will complete. |

No persisted-format, API or protocol changes. The runtime session memory now records the corrected directory instead of a guessed one, which is the existing contract for a confirmed value.

## Repository guidance

|---|---|---|
| `AGENTS.md` / `CLAUDE.md` | Source change across two packages | No dependencies added; `../opencode` untouched; no secrets logged; domain logic kept in the owning sync modules rather than in components. |
| `sync-state-invariants` skill | Session/directory ownership, bootstrap, reconciliation | Authoritative state preferred over heuristics — that is precisely the defect: containment was used as a proxy for ownership. Partial state is explicit: a record without a directory still falls back to membership rather than resolving to nothing. The guess remains excluded from the resolver and from persistence until it is confirmed. |
| `ui-api-decoupling` skill | Routing decides which directory every SDK call is addressed to | Directory identity semantics unchanged; only the source of truth for ownership is corrected. |
| `clack-cli-patterns` skill | CLI timeout and failure reporting | The CLI no longer reports a failure for an operation that succeeded; `--json` output shape is unchanged. |
| `performance-engineering` skill | An intermittent bug cannot be judged from single runs | Verified by repeated runs on before/after builds rather than by reasoning, and the instrumentation used to find the race was removed before measuring. |
| `packages/ui/src/sync/DOCUMENTATION.md` | Nearest owning documentation | Read before editing. The documented precedence is preserved; this corrects the value supplied to it. |

## Validation

| Check | Result |
|---|---|
| Reproduction, before the fix | 1 of 4 runs rendered the reply live |
| Reproduction, after, clean build, instrumentation removed | **3 of 3 rendered live**, each routed to its own worktree |
| Cold-path worktree creation immediately after restart | Previously timed out; now ok in 4004 ms and 1376 ms |
| `packages/ui/src/sync/session-directory-adoption.test.ts` (new) | 5 pass — ownership disagreeing with containment; guess promoted once the owner is known; no-op while the owner is unknown; a confirmed selection is never rewritten; a superseded selection is untouched |
| `packages/web/bin/lib/cli-control.test.js` | 8 pass — 4 new, covering worktree provisioning without waiting, a blank worktree name, and both directions of the wait-versus-provision window |
| `bun test` per file across `packages/ui/src/sync` | No failures |
| `npx vitest run bin/` (`packages/web`) | 88 pass |
| `bun run type-check` (workspace) | Pass — ui, web, electron, mobile, root |
| `bun run lint` (workspace) | Pass — all packages |
| `bun run dead-code` | Run and inspected; findings in the touched files are pre-existing, no new unused exports |

**Not verified:** the fix was exercised in the web runtime only. The corrected ownership rule is runtime-agnostic and unit-tested, but VS Code, Electron and mobile were not run. The full `packages/web` suite was not run either — `bun test` there hangs on tests that start servers; the touched area was covered with vitest instead.

**Reproduction** (unattended, no human in the loop): open the app, wait for it to settle, create a worktree with a session while the page stays open, navigate to that session, type a prompt, and watch for 90 s. Before the fix the message list stays empty while the same session shows both messages on reload.

## Visual evidence

The user-visible change is that a session which previously rendered nothing now renders normally; there is no new or altered UI. The failing state produced an empty message list and a retry loop of failed requests, and the fixed state produces the ordinary chat view, which is why the evidence here is the reproduction result and the request log rather than a screenshot of new chrome.

## Risks and failure behavior

- **Ownership source.** The riskiest part is preferring the record over store membership. A session move updates the owning store, so a stale cached record could in principle lag behind; moves do not rely on this path, since a confirmed move calls `setSessionDirectory` explicitly and that continues to win. A record without a directory still falls back to membership, so no session loses routing.
- **Guess promotion.** It cannot overwrite a confirmed selection or a selection that has since moved on; both are covered by tests. Worst case it does nothing and the previous behaviour stands.
- **CLI timeout.** A longer ceiling means a genuinely hung server is now waited on for up to two minutes instead of four seconds for this one action. That is the correct trade for an operation the server completes regardless, and it does not affect any other control call.
- **Rollback.** Each commit is independently revertible. Reverting only the ownership commit would restore the original symptom; reverting only the promotion commit leaves ownership correct and the guess unsettled.

### DIFF
--- a/packages/ui/src/lib/debug.ts
+++ b/packages/ui/src/lib/debug.ts
@@ -441,7 +441,11 @@ export const debugUtils = {
     const sources = {
       attachment,
       worktreeMetadata,
-      authoritative: owningStoreDirectory ?? recordDirectory,
+      // Record first, matching the resolver: holding a session proves
+      // containment, not ownership, so the parent repository holds its
+      // worktrees' sessions too. Reporting membership first made this
+      // diagnostic contradict the routing it exists to explain.
+      authoritative: recordDirectory ?? owningStoreDirectory,
       selected,
       remembered: remembered.runtime,
     };
--- a/packages/ui/src/sync/DOCUMENTATION.md
+++ b/packages/ui/src/sync/DOCUMENTATION.md
@@ -207,7 +207,7 @@ The discriminator is whether the server confirmed the path, not whether the valu
 
 | Source | Meaning |
 |---|---|
-| `authoritative` | The child store that actually holds the session, then its own record |
+| `authoritative` | The session record's own directory, then a child store that holds it |
 | `selected` | Server-confirmed directory captured at selection; a guessed one is never passed |
 | `attachment` | Worktree attachment recorded by this client; the *requested* path |
 | `worktree-metadata` | Worktree captured when the session was created in one; the *requested* path |
@@ -215,7 +215,7 @@ The discriminator is whether the server confirmed the path, not whether the valu
 
 Rules:
 
-1. `getSyncSessionDirectory()` is the authoritative session→directory mapping: a session lives in exactly the child store for its directory, whether or not the server populated `session.directory`. `null` means "not indexed yet", never "no directory".
+1. Ownership comes from the session record's own `directory`. `getSyncSessionDirectory()` reports *containment*, not ownership, and is only the fallback for a record without a directory: a project's session list includes the sessions of its worktrees so the sidebar can group them, so the parent repository holds worktree sessions too, and reading ownership from membership routes a worktree session to its parent. `null` means "not indexed yet", never "no directory".
 2. `attachment` and `worktreeMetadata` hold the worktree path this client asked for, before the server canonicalized it. They are a hint for a session sync has not indexed yet, never a correction of a confirmed directory — otherwise a stale local path re-creates the very mismatch this precedence exists to prevent.
 3. Never persist or rank a guessed directory. `selectSession` may fall back to the active directory to keep routing usable, but that value is not written to runtime memory, not written to the last-active snapshot, and not passed as `selected` — a persisted guess outlives the race that produced it and survives reloads and restarts.
 4. Components must not read `currentSessionDirectory` to build request or queue keys; use `getDirectoryForSession()` so every consumer resolves identically.
--- a/packages/ui/src/sync/session-directory-adoption.test.ts
+++ b/packages/ui/src/sync/session-directory-adoption.test.ts
@@ -1,0 +1,82 @@
+import { beforeEach, describe, expect, test } from "bun:test"
+
+import { ChildStoreManager } from "./child-store"
+import { setSyncRefs } from "./sync-refs"
+import { useSessionUIStore } from "./session-ui-store"
+
+/**
+ * Selecting a session whose directory this client has not indexed yet routes it
+ * through the active directory as a deliberate guess. Nothing used to settle
+ * that guess once the owning directory finished bootstrapping, so every fetch
+ * stayed addressed to a directory that does not own the session and the session
+ * never rendered.
+ *
+ * These tests pin both directions: a guess is promoted once the authoritative
+ * directory becomes readable, and a confirmed selection is never rewritten.
+ */
+
+const PARENT = "/repo"
+const WORKTREE = "/repo/.worktrees/feature"
+const SESSION_ID = "ses_directory_adoption"
+
+const indexSessionIn = (manager: ChildStoreManager, directory: string): void => {
+  const store = manager.ensureChild(directory, { bootstrap: false })
+  store.setState({
+    session: [{ id: SESSION_ID, directory, title: "test" } as never],
+  })
+}
+
+let manager: ChildStoreManager
+
+beforeEach(() => {
+  manager = new ChildStoreManager()
+  setSyncRefs({} as never, manager, PARENT)
+  useSessionUIStore.getState().setCurrentSession(null)
+})
+
+describe("adoptAuthoritativeSessionDirectory", () => {
+  test("promotes a guessed selection once the owning directory is indexed", () => {
+    useSessionUIStore.getState().setCurrentSession(SESSION_ID)
+    expect(useSessionUIStore.getState().currentSessionDirectory).not.toBe(WORKTREE)
+
+    indexSessionIn(manager, WORKTREE)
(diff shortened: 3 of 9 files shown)


btriapitsyn:
Thanks — three of the four findings were real defects in this branch, and they are fixed in `cd3280dc5`.

**1. Stale ownership documentation — valid, fixed.** The old docstring was left stacked directly above the new one saying the opposite; that was careless. Corrected in `session-ui-store.ts`, `session-directory-resolution.ts`, the `getSyncSessionDirectory` docstring in `sync-refs.ts`, and the `DOCUMENTATION.md` table and rule 1. Rule 1 was the worst of them: it asserted that "a session lives in exactly the child store for its directory", which is precisely the false premise this branch disproves. The PR body claim that the documented precedence was preserved was wrong and is corrected too.

**2. `diagnoseSessionDirectory` reporting membership first — valid, fixed.** It now uses the same record-first order as the resolver. Worth stressing how this one matters: that helper is what someone reaches for when routing looks wrong, and for exactly the scenario this branch fixes it would have reported `authoritative = /repo` and raised a disagreement alert while routing was correct. A diagnostic that contradicts the thing it exists to explain is worse than none.

**3. CLI timeout comment contradicting the code — valid, fixed by taking your reading.** You checked the server and found `createWorktree` runs inside `sessionService.create()` before `waitForIdle`, so the windows really are sequential. The comment was right and the `Math.max` was wrong, not the other way round. The windows are now summed, and the tests pin both `{wait, timeout: 30, worktree}` → 180 s and `{wait, worktree}` → 750 s.

**4. Bootstrap→adoption wiring untested — valid, not addressed, and here is why.** The wiring lives in the `SyncProvider` callback, and this repository has no React test harness at all: no `@testing-library`, no `jsdom`, no `happy-dom`. Covering it would mean adding a dependency, which the repository guidance forbids without an explicit request. Rather than fake it at a level that would not exercise the integration, I am leaving it covered by the end-to-end reproduction, which does exercise exactly that path: create a worktree with the client already running, open its session, prompt, and observe the reply render live — 3 of 3 after the fix against 1 of 4 before. If a React harness is wanted, that is worth doing as its own change rather than smuggled in here.

On the residual risks you listed:

- **`getAllSyncSessions().find` picking the first copy.** Correct that the fix relies on every copy carrying the same record `directory`. That is not an assumption I made in the abstract: the captured evidence showed the parent store's copy already reporting the worktree path, which is why the record is trustworthy where membership is not. The `believes the session record over the store that merely holds it` test pins exactly that shape — the session indexed in the parent store with a record naming the worktree.
- **`opencodeClient` base directory after adoption.** Observed to follow correctly in the reproduction (`opencodeClientDirectory` moved to the worktree), so it is not left behind in this path, but you are right that adoption itself does not re-point it and that this is pre-existing behaviour rather than something this branch introduces.

Validation after the fixes: workspace `type-check` and `lint` pass; `session-directory-adoption` 5 pass, `session-directory-resolution` 10 pass, `session-actions` 46 pass, `cli-control` 7 pass.

PR #2615, opened 2026-08-04
fix(sync): keep pending questions answerable instead of stuck (#2448)

### LINKED ISSUE #2448: Question form can get stuck in a non-answerable state while still pending
# Background

When the `question` tool prompts the user, an answerable `QuestionCard` form is rendered from the `state.question[sessionID]` store slice, while the tool part itself only shows a read-only view of the question text. It is possible to land in a "stuck" state where the question is still unanswered but is no longer rendered as an interactive, answerable form, so the user has no visible way to respond while the agent run remains blocked waiting for input.

Known ways this stuck state arises:

- `QuestionCard` sets a local `hasResponded` flag after calling `question.reply`/`question.reject` (QuestionCard.tsx:244-299) and hides itself when that flag is set (QuestionCard.tsx:321-323). If the corresponding `question.replied`/`question.rejected` SSE event never arrives (network drop, SSE gap), the store still holds the pending question, but a mounted card renders nothing. It only re-appears if the card unmounts and remounts with the same question id.
- During an SSE gap, a missed `question.asked` event means the question never enters the store at all. The `question` tool part still renders the question text read-only (ToolPart.tsx:1666-1730) as "Awaiting response", but no interactive form exists to actually answer it.
- `dismissOpenQuestionsForSession` optimistically clears questions from the local store before the backend confirms rejection (session-actions.ts:995-1059). If the backend reject fails with a non-404 error, the local store has dropped the question while the backend still considers it pending.

In all of these the agent is blocked on an answer the UI no longer presents as answerable.

# Version

OpenChamber 1.16.3 (8040d43), OpenCode 1.18.5

# Acceptance Criteria

## Must

- [ ] While a `question` tool is still unanswered/pending on the backend, the UI always surfaces an answerable question form (or a recoverable affordance to re-display and answer it) for the active session and its descendant sessions.
- [ ] The UI cannot remain in a state where a question is pending but no interactive form is shown and no way to answer it exists.
- [ ] After an SSE gap or reconnect, any still-pending question is re-shown as an answerable form; the read-only "Awaiting response" tool bubble is never the only surface for an unanswered question.

## Should

- [ ] When the answerable form and the read-only tool bubble diverge (question pending but form hidden), the UI indicates input is still required rather than appearing idle.

---

Created with [create-issue](https://github.com/tomzx/agents/blob/48bc3e1e05a3184ecfb599e2242d9ea5fc1d8b2d/skills/create-issue/SKILL.md) via glm-5.2 (`48bc3e1`)

panzeyu2013:
## Intent

Fixes #2448. A `question` tool prompt could get "stuck": still pending on the backend while the UI showed no answerable form, leaving the agent blocked with only the read-only "Awaiting response" tool bubble. Three failure paths were addressed:

1. **Hidden card after a successful reply** — `QuestionCard` set a local `hasResponded` flag and hid itself, while the store kept the question because the SSE `question.replied` event was lost (network drop/SSE gap).
2. **Question never entered the store** — a lost `question.asked` event during an SSE gap meant the question never reached `state.question`, and the resync path (`resyncBlockingRequestsForDirectory`) dropped questions for sessions the store did not yet know (e.g. a new subagent session).
3. **Orphaned question after a failed dismiss** — `dismissOpenQuestionsForSession` optimistically cleared the store before the backend reject; a non-404 reject failure left the question gone from the store but still pending on the backend.

Resulting behavior: a pending question always surfaces as an answerable `QuestionCard` (or is deterministically cleared once answered/rejected), and a rejected dismiss rolls back instead of orphaning the question.

**Follow-up fix (this change):** the resync now scopes merged questions to the **owning directory**. `listPendingQuestions` merges an unscoped global `question.list()` with the directory-scoped fetch, so a resync for one directory could previously adopt questions owned by another directory once the unknown-session guard was removed. Questions are now only surfaced/merged for sessions attributable to the resynced directory (already known, or successfully materialized via `session.get` with a matching `session.directory`).

## Non-goals

- No visual restyling of `QuestionCard`; its rendered appearance is unchanged.
- No change to the backend/SSE protocol or the question tool's server-side behavior.
- Permission dismissal is intentionally left as-is; only the question path is addressed.
- No periodic polling or a persistent manual "re-check" affordance was added for the pathological case where the recovery resync itself keeps failing.

## Affected surfaces

- `packages/ui` — shared React UI used by web, desktop (Electron), VS Code, hosted mobile, and Capacitor mobile.
- Sync store state (`state.question`), `QuestionCard` visibility, `dismissOpenQuestionsForSession` on the chat send path, and `trimSessions` eviction.
- Runtime contract change: `resyncBlockingRequestsForDirectory` now merges questions for sessions not yet known to the directory store **only when that session is attributable to the directory** (previously: dropped entirely; the follow-up prevents adopting foreign sessions).
- No persisted storage, external contract, or cross-runtime surface changes.

## Repository guidance

|---|---|---|
| AGENTS.md — "Correctness invariants" (prefer authoritative state over heuristics; failure is not empty; one failed entity must not erase unrelated entities; per-directory store scope is a contract) | This change touches sync state reconciliation, optimistic updates, and per-directory store scoping. | Store visibility replaces local `hasResponded`; `listPendingQuestions` throws on failure and callers preserve prior state; rollback restores only the failed question; foreign-session questions are dropped so a directory store never holds another directory's domain data. |
| `sync-state-invariants` skill (session sync, reducers, optimistic updates, resync/reconciliation, per-directory scope) | The change modifies event-driven store behavior, optimistic dismissal, and resync. | Optimistic dismiss rolls back on failure; resync reconciles deterministically from the authoritative snapshot with a signature guard protecting in-flight SSE mutations; rollback re-check avoids resurrecting stale completions; directory ownership is verified (`session.directory`) before materialization and before merge. |
| `ui-api-decoupling` skill (SDK data access, runtime fetch/URLs, server API routes) | The change calls SDK endpoints (`question.reply`/`question.reject`, `session.get`, `question.list`) through the existing client layer. | All calls reuse the established `opencodeClient` / `getRequestReplyClient` / scoped-SDK paths with the existing directory-resolution and error-shaping helpers. |
| `theme-system` skill (UI colors/icons) | Applies only to UI rendering. | No styling or icon changes were made; only visibility logic. |

## Validation

| Check | Result |
|---|---|
| `bun test src/sync/__tests__/session-switch-resync.test.ts` | 11 pass, 0 fail (covers keep-when-materialized, drop-when-unresolvable, drop-when-foreign-directory) |
| `bun test src/sync/session-actions.test.ts` | 49 pass, 0 fail (covers non-404 reject rollback: still-pending → restore, resolved → no restore, re-check failure → conservative restore) |
| `bun test src/sync/question-recovery.test.ts` | 7 pass, 0 fail (`hasPendingQuestionGap` branches) |
| `bun test src/sync` (full sync suite) | 336 pass / 47 fail / 4 errors. The 47 fails + 4 errors are **pre-existing** cross-file `mock.module` interference and environment issues (relay/auth/websocket URL, worktree routing, drafts) — reproduced unchanged at the base commit `7f0d87cbe`; this PR introduces zero new failures. |
| `bun test src/components/chat` | 548 pass, 0 fail |
| `bun run lint` (packages/ui ESLint on touched files) | Clean |
| `bun run type-check:ui` | No new errors (only pre-existing missing-native-dependency errors for `@capacitor/*`, `fflate`, `@tanstack/react-virtual`) |
| `bun run dead-code` | No new reports |
| Independent subagent review | Verdict: fix correct; cross-directory blocker resolved; the only follow-up (stale comments) was addressed. |
| Follow-up subagent review (`c779029`) | Verdict: approve, no blockers — tail-window scan safe for recovery semantics, boolean refactor behavior-preserving. |
| Not verified | End-to-end runtime repro of an SSE gap / non-404 reject failure (requires inducing a live network drop; covered by unit tests at the store layer). |

## Visual evidence

No styling change: `QuestionCard`'s rendered markup is untouched. The diff changes only **state-driven visibility** — the card stays/renders in the previously-stuck scenarios and unmounts deterministically after a successful answer. A before/after recording would require reproducing an SSE drop and a non-404 reject failure in a live run, which was not captured; the visibility state transitions are covered by the store-layer tests above.

## Risks and failure behavior

- **Rollback safety**: a failed dismiss restores the question only after an authoritative `listPendingQuestions` re-check confirms it is still pending; if the re-check itself fails, the question is restored conservatively (a later stale 404 is reconciled out by the reply/reject path).
- **Queued send after rollback**: `dismissOpenQuestionsForSession` still returns `true` on a rolled-back dismiss, so the chat send path queues the message; the queued send delivers only once the recovered question is resolved and the session returns to idle. This is documented in the code.
- **Deterministic card unmount**: a successful reply/reject now removes the question from the store directly; the SSE `question.replied`/`question.rejected` event becomes a no-op (the reducer only removes if present), so no double-removal or race.
- **Directory scoping (follow-up)**: a resync only merges/toasts questions whose session is already known to the directory or was materialized with a matching `session.directory`. `listPendingQuestions`' unscoped global fetch can no longer pollute another directory's store, fire wrong "Open session" toasts, or materialize a foreign session row. A session that cannot be resolved in this directory loses its question for that pass only; the next event/reconnect resync recovers it.
- **Eviction**: `trimSessions` now also preserves sessions with a pending question; behavior for pending permissions is unchanged.
- **Cross-runtime**: no backend/SSE protocol change; the fix applies uniformly to web, desktop, VS Code, and mobile via `packages/ui`.


panzeyu2013:
**Heads up: the failing `automation` check is a review-pipeline flake, not a code problem**

The `pr-review` workflow's `automation` check failed on HEAD `c77902946` with:

> Review completed without creating a new structured OpenChamber Bot PR comment.

What actually happened:

- **Not a timeout.** The run took ~598s and exited 0 (`REVIEW_TIMED_OUT: false`, `REVIEW_RUN_OUTCOME: success`).
- The `opencode run --agent pr-review` process completed, but **never posted the structured review comment** the workflow's "Verify and enforce review verdict" step requires (a new `openchamber-bot[bot]` comment containing `<h3>Code Review Summary</h3>` plus an `<!-- oc-review-meta -->` marker). No such comment was created after review start, so the step failed and the job was marked `FAILURE`.

Why the agent did not post:

- `rg` is not installed on the runner, so the agent fell back to dumping full files via `gh api ... | cat` (e.g. `event-reducer.ts`, `session-actions.ts`, `sync-context.tsx`), consuming nearly the entire run budget on deep verification.
- Several bash calls were rejected by the agent's permission allowlist (only `gh *`, `git *`, `rg *`, `ls *`, `cat *` are allowed), forcing retries and wasted time.
- It ended its run right after deep verification, before reaching the "post and verify review comment" step.

This is unrelated to the PR's code:

- The other checks are green: `checks` (build/lint/test) = SUCCESS, `label` = SUCCESS.
- Other PRs' reviews are completing successfully in the same time window.
- Earlier reviews of this PR (`767d2eebf`, `3d764844`) posted comments successfully; this failure is an LLM/tooling flake on the repeat-review of a heavy sync change.

Also note: the failed `automation` check is not a required check for merge (the branch ruleset has no required status checks). The PR is `BLOCKED` only because the "Chambermakers only" ruleset requires an approving review from the Chambermakers team.

btriapitsyn:

Issue #2544, opened 2026-07-30
[Bug] EPERM: operation not permitted, rename on settings.json prevents startup (Windows)


Cuhosuaxemay:
### What's wrong?

OpenChamber fails to start on every launch with:

```
[error] [electron] startup failed: Error: EPERM: operation not permitted, rename 
'C:\Users\<user>\.config\openchamber\settings.json.tmp-*' -> 
'C:\Users\<user>\.config\openchamber\settings.json'
    at async writeJsonFile (main.mjs:30260)
```

The app's `writeSettingsToDisk` creates `settings.json` via `writeFile`, then `writeJsonFile` tries an atomic rename and fails. The rename-to-new-file path works (when settings.json doesn't exist), but the app always creates it first internally. This is 100% reproducible — 50+ failures across 10 days of logs.

A secondary crash follows: `ERR_FAILED (-2) loading 'data:text/html;...'` (splash screen fails to load).

### Environment

- OpenChamber 1.17.1 (Electron desktop)
- Windows (Standard PC Q35 + ICH9 VM, enterprise-managed)
- `fltmc filters` returns `ACCESS DENIED` even from elevated context
- Multiple filesystem filter drivers present: `bindflt`, `CldFlt`, `WdFilter`, App-V drivers (`AppvStrm`, `AppvVfs`, `UnionFS`), `UevAgentDriver`

### Related

- Same `EPERM` behavior documented in **#2258** for `process.kill(pid, 0)` on Windows — the security context causes certain syscalls to get `EPERM` instead of expected error codes
- `fs.promises.rename` → `MoveFileExW` with `MOVEFILE_REPLACE_EXISTING` fails with `EPERM`, same as `process.kill` → `EPERM` in #2258

### Workaround

Deleting `settings.json` before each launch and using `--no-sandbox` flag:
```
del /q "%USERPROFILE%\.config\openchamber\settings.json"
OpenChamber.exe --no-sandbox
```

### Where does it happen?

Desktop (macOS)

### Logs (optional)

```shell

```

Cuhosuaxemay:
> ## Reproduction Results
> **Status: Reproduced** ✅
> 
> ### Root cause
> The bug is in `packages/electron/main.mjs`, function `writeJsonFile` (lines 513–524). It writes to a temp file and then calls:
> 
> await fsp.rename(tmp, filePath);
[...]

PR #2618, opened 2026-08-04
fix(ui): prevent Shiki template-call OOM on backtick JS files

### LINKED ISSUE #2587: [Bug] Opening files with template-literal code triggers catastrophic backtracking → renderer OOM → black/frozen desktop app (v1.17.2)
### Bug description

Opening certain `.js`/`.md` files in the UI (file viewer / context panel / clicking a file reference in chat) causes the renderer process to balloon to 4–5.7 GB of memory and the desktop app to go black / completely unresponsive. Restarting does not help because the app **auto-restores the last opened file tab** on startup, so the renderer crashes again every launch. Clearing `AppData\Roaming\OpenChamber\Local Storage\leveldb` is the only thing that brings the app back (it drops the restored tab).

The web version (independent browser hitting `http://127.0.0.1:57123`) is also affected when the same file is opened.

### Environment

- OS: Windows 11 (x64), NVIDIA RTX 4070 Ti SUPER
- OpenChamber: v1.17.2 (win-x64), packaged Electron
- Latest available release is v1.17.2 (no newer version to upgrade to)

### Reproduction steps

1. Open any `.js` file that contains **backtick template literals with nested `${...}`** (or a `.md` file with fenced code blocks containing them). A 14 KB `.js` file is enough to reproduce; the trigger is the template-literal syntax, not the file size.
2. The UI freezes and the renderer process memory grows to multi-GB, then the window goes black / the process dies.
3. Relaunch the app — the last-opened file tab is restored automatically and it crashes again on startup.

Example content that triggers it (from a real file that crashed the app):

```js
const snapshot = { source: `${session.source}`, fetchedAt: `${Date.now()}` };
const label = `Account ${index + 1}`;
```

### Root cause analysis

The crash is a **catastrophic backtracking** in the bundled shiki highlighter:

1. File preview/Markdown rendering uses shiki with the **old vscode-oniguruma WASM engine** (bundled as `wasm-CG6Dc4jp.js`).
2. The JavaScript grammar's `template-call` rule `begin` lookahead contains a triple-nested backtracking pattern over `{}`, `()`, `[]` with alternations — see the pattern around `"template-call"` in `javascript-WyGyM1wV.js`:

```
begin: (?=(([$_[:alpha:]][$_[:alnum:]]*\s*\??\.\s*)*|(\??\.\s*)?)([$_[:alpha:]][$_[:alnum:]]*)(<\s*((...(nested {...}((...)[...])...))*)...)([^(<>]|(\(([^()]|(\(([^()]|\([^()]*\))*\)))*\))|(?<=>)|<...
```

3. On a backtick template literal the engine enters **exponential backtracking**. The old WASM build has **no retry limit / match stack limit** — `onig_set_retry_limit` / `setRetryLimit` / `match_stack_limit` are all absent from the bundle — so the match never gives up, `omalloc` grows the WASM heap without bound, and the whole renderer process dies (OOM). Since the worker runs in the same renderer process, the entire window goes black.

This is the same class of bug that was fixed upstream in shiki (e.g. shiki#1026, fixed by tightening `\S+` to `[^\s/]+`), but only for the JS engine — OpenChamber still ships the old oniguruma WASM engine without any protection.

### Expected behavior

- Opening such files should not freeze/crash the app.
- The highlighter should either use a grammar pattern that cannot backtrack exponentially, or set a retry limit / match stack limit on the oniguruma engine (newer `@shikijs/engine-oniguruma` supports this).

### Additional context

- The main process logs show `Renderer process crashed` and `ERR_FAILED (-2) loading 'data:text/html...splash'` in crash episodes.
- `setRetryLimit` / `onig_set_retry_limit` / `match_stack_limit` all appear **zero** times in the shipped WASM bundle.
- Related known issue that also blocks startup on Windows: #2544 (EPERM on settings.json rename) — both together made the app effectively unusable.

sdziupin:
## Intent

Opening JS/TS files (or Markdown fences) that contain backtick template literals could send the bundled Shiki Oniguruma WASM highlighter into catastrophic backtracking via the TextMate `template-call` rule. The WASM build exposes no retry/match-stack limit, so `omalloc` grew without bound, the renderer reached multi-GB OOM, and the desktop window went black — then auto-restored the same tab on every relaunch.

This PR:

1. **Neutralizes `template-call`** when the markdown Shiki worker loads `javascript` / `typescript` / `jsx` / `tsx` grammars. The plain `#template` rule still highlights backticks and simple tagged templates; only rare `ident<TypeArgs>\`...\`` specialized coloring is dropped.
2. **Adds a 5s hang timeout** on the main-thread worker client. If tokenize never returns, the worker is terminated so the WASM heap is reclaimed and callers fall back to plain escaped code (existing failure path).

Closes openchamber/openchamber#2587

## Non-goals

- Upgrading Shiki / `@shikijs/engine-oniguruma` (still no public `setRetryLimit` API).
- Changing Pierre diff highlighting (`preferredHighlighter: 'shiki-wasm'`).
- Fixing Windows settings.json EPERM (#2544).

## Affected surfaces

- **`packages/ui`**: markdown Shiki worker + main-thread client used by chat Markdown code blocks, `WorkerHighlightedCode`, and CodeMirror file/plan/skills Shiki decorations (web, desktop, VS Code, hosted mobile, Capacitor — shared UI path).
- **Unaffected**: Electron main/preload, VS Code extension host (except shared webview UI), CLI, server routes, Pierre diff worker pool.

## Repository guidance

|---|---|---|
| `AGENTS.md` / change discipline | Source change in shared UI | Smallest complete fix; no new deps; focused tests + package type-check/lint; `bun run dead-code` inspected after new files/exports |
| `performance-engineering` | Hang/OOM on highlight hot path | Remove catastrophic work (grammar) first; timeout is a bounded safety net with explicit failure → plain-text fallback |
| Theme/locale/settings skills | N/A | No UI chrome or copy changes |

Nearest docs: no `DOCUMENTATION.md` under `lib/shiki` or `chat/markdown`; followed existing worker/client precedent.

## Validation

| Check | Result |
|---|---|
| `bun test packages/ui/src/lib/shiki/sanitizeTemplateCallGrammar.test.ts packages/ui/src/components/chat/markdown/markdown-worker.hang.test.ts` | Pass (5 tests) — includes ~14KB template-literal fixture under 2s budget |
| `bun run type-check` in `packages/ui` | Pass |
| `bun run lint` in `packages/ui` | Pass |
| `bun run dead-code` | Ran; no new actionable unused exports from this change |
| Web UI at `http://127.0.0.1:5180/` with `#2587` template-literal fixture in chat/composer | Pass — UI stayed responsive; syntax highlighting applied; no freeze/black screen |

**Not verified here:** packaged Windows Electron OOM reproduction (cloud env is web/linux/headless); Pierre diff path; auto-restore crash loop on Windows. In-app Files/CodeMirror open of the fixture was attempted but the Files editor entry path was unreliable in this environment; Chat/Markdown uses the same Shiki worker that Files decorations use (`highlightTokensInWorker`).

## Visual evidence

OpenChamber chat showing template-literal JS highlighted without freeze

OpenChamber composer with #2587 template-literal fixture code block

## Risks and failure behavior

- **Highlight failure / timeout:** resolves to `null`; callers keep escaped plain code (no main-thread Shiki fallback). Worker is terminated and recreated on the next request.
- **Grammar tradeoff:** `foo<T>\`...\`` loses specialized type-arg coloring; ordinary templates remain highlighted.
- **False timeout:** 5s budget is above normal tokenize for `MAX_VIEW_CHARS` (200KB) files; if hit, user sees uncolored code rather than a frozen/OOM window.
- **Rollback:** revert commit; no persisted data or API contract changes.

<div><a href="https://cursor.com/agents/bc-1ffbf9fb-317e-4eec-9a42-3442eebd802f?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-1ffbf9fb-317e-4eec-9a42-3442eebd802f&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


btriapitsyn:
Thanks for the fix — merging.

The grammar fix landed as written — stripping `template-call` from the JS/TS/JSX/TSX grammars at load is the right call, and the test that times a real backtick fixture through the patched grammar is good evidence.

The worker timeout needed reworking. Main has since grown a client-side result cache, in-flight coalescing, and a blob-URL worker for VS Code, so I rebuilt the timeout on top of that instead of taking the version here: each pending entry now carries its timer, `onmessage` clears it before resolving, and a request that blows the 5s budget goes through the existing fail-all path. A timed-out request resolves `null` like any other failure, so the caller keeps plain text and nothing poisons the cache.

One thing worth separating out: the image-preview retry problem mentioned further down in #2587 is a different bug and isn't addressed here. It stays open on its own.

Validation: the grammar and worker tests plus the rest of the markdown suite (69 tests), `tsc` on packages/ui, and eslint/oxlint on the changed files.

PR #2626, opened 2026-08-04
fix(sync): revalidate selected session's blocking requests on switch, focus, and pageshow

jamie-dit:
## Problem

After a directory's bootstrap completes, the selected session's pending questions/permissions are only populated by a live SSE event. A `question.asked` / `permission.asked` event missed during a disconnect — or a fresh tab, window, or device opened after the request was asked — leaves the blocking-request bucket empty. The question card renders but can never be answered ("This question is no longer waiting for a response"), and the request can only be resolved from the client that received the original event.

## Fix

**`useActiveSessionBootstrapDemand`** (new hook, mounted in `SyncAppEffects`) revalidates the selected session through the authoritative `question.list`/`permission.list` reconciliation on:

- session selection
- window `focus`
- `visibilitychange` return to visible
- `pageshow` (bfcache restore)

Directories still bootstrapping subscribe until completion rather than forcing a bootstrap; in-flight requests coalesce per directory+session, so rapid switching does not cause a fetch storm. `currentSessionDirectory` may briefly be a guess before `adoptAuthoritativeSessionDirectory()` settles — a revalidation against the guessed directory is a wasted fetch, never a corrupting one, because of the signature guard.

**Stale-response race fix** in both reconciliation paths (`sync-context.tsx` and `bootstrap.ts`): a non-empty list response previously overwrote state even when the bucket's mutation signature changed while the request was in flight, dropping SSE additions or resurrecting SSE-resolved requests. Both paths now skip sessions whose signature changed mid-flight.

## Tests

- `useActiveSessionBootstrapDemand.test.ts` — selection/focus/visibility/pageshow triggers, bootstrap-gating, coalescing, wrong-directory safety
- `session-switch-resync.test.ts` — extended: stale non-empty responses cannot overwrite SSE additions or resurrect SSE-resolved requests (both entity types)
- `bootstrap-blocking-request-race.test.ts` — new: same race coverage for the bootstrap.ts path (fails without the fix, passes with it)

```
bun test packages/ui/src/hooks/useActiveSessionBootstrapDemand.test.ts   packages/ui/src/sync/__tests__/session-switch-resync.test.ts   packages/ui/src/sync/__tests__/bootstrap-blocking-request-race.test.ts
# 27 pass, 0 fail
```

Full `packages/ui/src/sync/` suite failure set is byte-identical to `main` (pre-existing environment failures only); `tsc --noEmit` clean.


btriapitsyn:

mattv8:

btriapitsyn:

PR #2641, opened 2026-08-05
fix(sessions): recover sessions stranded on interrupted or empty turns

sdziupin:
## Intent

OpenCode sessions can silently stop in two ways, both leaving the user with a session that just stopped with no explanation:

1. **Empty turn** — the model stream ends with no output (provider usage limit such as `AI_APICallError: Monthly usage limit reached`, or a transient failure). OpenCode closes the turn with `finish: "unknown"`, 0 tokens, no error, and the loop exits quietly.
2. **Interrupted turn** — the OpenCode serve process dies mid-stream (crash, kill, restart). The turn stays unfinished (`time.completed` missing, possibly with a `running` tool part) forever; after a serve restart OpenCode re-emits no status for that session, so nothing ever resumes it.

This change makes the OpenChamber web server (session-assist runtime) detect both failure classes and act:

- **Failed-turn recovery**: up to 2 auto-retries via `POST /session/:id/prompt_async` with a short continuation prompt, waiting 60s between attempts (provider usage windows typically reset within ~2 minutes). Attempts are tracked in `metadata.openchamber.assistRetry` scoped to the failed message id, written *before* the prompt (a failed prompt cannot loop). Before every prompt the session tail and the *live* session status are re-checked, so a genuinely running turn never receives a duplicate prompt; a retry is dropped unless the session is confirmed idle.
- **Honest recap**: once retries are exhausted, an assist payload whose recap states plainly that the agent's reply came back empty (likely a provider limit) and that the user can send "Continue" or switch models — instead of the previous meaningless recap of an empty turn.
- **Startup recovery scan**: a bounded scan (`runStartupRecovery`) finds sessions stranded by a serve restart. It runs once after server startup and — via the new lifecycle `onOpenCodeReady` hook — after **every** OpenCode serve (re)start, including health-check recovery restarts (debounced to one scan per minute, limited to the warm directories, recent idle sessions, best-effort with bounded retries while upstream is not ready).

## Non-goals

- No changes to OpenCode itself (separate repository) — in particular, orphaned tool subprocesses left by a `kill -9`'d serve cannot be cancelled from OpenChamber; the continuation prompt instructs the agent not to repeat already-completed tool calls, but a still-running orphaned process may duplicate work (documented limitation).
- No UI changes: the existing `SessionRecapSpacer`/`SessionSuggestionChip` rendering path is reused unchanged.
- No changes to VS Code (extension-only, no web server — it cannot run the scan; it still renders payloads produced by a web/desktop instance).
- No periodic background polling beyond the bounded startup/restart-triggered scan.

## Affected surfaces

- `packages/web/server/lib/session-assist/runtime.js` — the watcher that generates recaps/suggestions for idle sessions; now also detects broken turns and runs the recovery.
- `packages/web/server/lib/opencode/lifecycle.js` — new optional `onOpenCodeReady` dependency (default `null`, no behavior change when absent) fired after every successful serve (re)start; never awaited, best-effort.
- `packages/web/server/index.js` — wires `getStartupDirectories` (the shared MRU directory list, extracted from the lifecycle warmup block — same behavior) and the `onOpenCodeReady` hook.
- Persisted/external contract: new `metadata.openchamber.assistRetry` (server-internal bookkeeping; ignored by all existing clients — UI only reads `metadata.openchamber.assist`); `metadata.openchamber.assist.recap` may now describe a failure instead of invented content; new settings key `sessionAutoRetryEnabled` (default on, read from OpenChamber `settings.json`, absent → on).
- Desktop/mobile/VSCode runtimes unaffected: the change is entirely in the web server process; clients consume the same assist payload shape as before.

## Repository guidance

|---|---|---|
| `openchamber-change-discipline` | Executable server source + persisted metadata contract + added source file | Smallest scoped change; `runtime.test.js` added with focused tests; DOCUMENTATION.md updated; dead-code run; validation at the narrowest level that covers real risk (session-assist + lifecycle suites, then the full server suite). |
| `sync-state-invariants` | Change touches live session status, reconnect/reconciliation after serve restarts, and directory-scoped scans | Live status (`/session/status`, `session.status` events) is the sole discriminator for "broken turn" — never text heuristics; failure ≠ empty (`fetchSessionStatuses`/`fetchRecentMessages` return `null` on failure and the scan retries on that real failure signal); per-directory try/catch so one failed directory never blocks others; retry only with confirmed-idle status (no duplicate prompts); bounded retries with `unref`'d timers. |
| `session-assist/DOCUMENTATION.md`, `lib/opencode` docs | Owning module documentation for both touched modules | Updated `session-assist/DOCUMENTATION.md` (failed-turn recovery, startup scan, triggers, limitations); lifecycle change documented in code comments. |
| AGENTS.md always-on constraints | Server-side behavior change; correctness invariants; no secrets | Authoritative state over heuristics (status-gated); temporary fallbacks scoped and cleared (retry counters scoped per message id); no secrets logged; owning docs updated. |

## Validation

| Check | Result |
|---|---|
| `node --check` on `runtime.js`, `lifecycle.js`, `index.js`, test files | OK |
| `bun x vitest run server/lib/session-assist` | 15/15 passed |
| `bun x vitest run server/lib/opencode/lifecycle.test.js` | 18/18 passed (3 new `onOpenCodeReady` tests) |
| `bun x vitest run server` (full web server suite, warm cache) | 994 passed / 2 skipped; known pre-existing timing-flaky tests (`walkthrough/routes.test.js`, occasionally `relay/tunnel-codec.test.js`, `ui-auth.test.js`, `event-stream/runtime.test.js`) fail under full-load runs but pass in isolation (verified: walkthrough 6/6, tunnel-codec 5/5, ui-auth 6/6); none touch this change |
| `bun run dead-code` | No findings related to this change (baseline report unchanged) |
| Live end-to-end | Not run: the fix is deployed only after merge/restart; the failure scenarios were verified against real session data from the production database (`opencode.db`) and server logs during development |

Not verified: no live restart of OpenChamber instances with the fix (requires deployment); no Windows/macOS-specific behavior (server-only JS, no platform APIs).

## Visual evidence

No user-visible rendering change: the diff is entirely server-side. The existing recap/suggestion components already render `metadata.openchamber.assist`; this change only affects *when* and *what content* the server writes (honest failure recap instead of a recap of an empty turn, or a retry first). There is no before/after UI state to capture; the observable difference is that a previously silent session stop now either resumes automatically or surfaces an honest explanation.

## Risks and failure behavior

- **Runaway retry loop**: prevented by the hard cap (2), the write-first counter, and counter reset on message-id change; after exhaustion the runtime writes the honest recap and stops.
- **Duplicate prompt into a running turn**: prevented by three gates — 60s quiet window + tail-move check + live status check immediately before `prompt_async`; if the status cannot be confirmed (fetch failure) the retry is dropped (never re-prompt without proof of idle).
- **Scan overhead**: bounded (5 directories × 8 sessions, 30-minute session age window, one scan per minute max, `unref`'d timers, best-effort); cannot block startup or the lifecycle (hook is fire-and-forget).
- **Rollback/compatibility**: reverting the commit removes the behavior; stale `assistRetry` metadata is harmless (overwritten on next use, not rendered, ignored by older clients); absent settings key defaults to enabled; older OpenChamber versions ignore the new metadata key.
- **Orphaned tool subprocesses** after a hard serve kill cannot be cancelled (OpenCode limitation) — documented in the module docs.
- **Security**: no secrets or sensitive content are logged or persisted beyond existing metadata patterns; the continuation prompt contains no conversation content.

### DIFF
--- a/packages/web/server/index.js
+++ b/packages/web/server/index.js
@@ -731,6 +731,9 @@ const sessionAssistRuntime = createSessionAssistRuntime({
   buildOpenCodeUrl,
   getOpenCodeAuthHeaders,
   getSmallModelService: async () => import('./lib/small-model/index.js'),
+  // Startup recovery scan: recovers sessions stranded on an interrupted turn
+  // after an OpenCode serve restart (which re-emits no status for them).
+  getStartupDirectories: getWarmupDirectories,
 });
 
 const sessionGoalRuntime = createSessionGoalRuntime({
@@ -1031,6 +1034,28 @@ Object.defineProperties(openCodeLifecycleState, {
   resolvedWslDistro: { get: () => resolvedWslDistro, set: (value) => { resolvedWslDistro = value; } },
 });
 
+// Most-recently-used directories first: OpenCode initializes each directory
+// lazily on first request (seconds on large session stores), so the lifecycle
+// warms these right after readiness — before the UI's first interactive
+// request would otherwise pay that cost. The session-assist startup recovery
+// scans the same list for sessions stranded by a serve restart.
+const getWarmupDirectories = async () => {
+  const settings = await readSettingsFromDiskMigrated().catch(() => null);
+  if (!settings) return [];
+  const directories = [];
+  if (typeof settings.lastDirectory === 'string' && settings.lastDirectory) {
+    directories.push(settings.lastDirectory);
+  }
+  const projects = Array.isArray(settings.projects) ? [...settings.projects] : [];
+  projects.sort((a, b) => (b?.lastOpenedAt ?? 0) - (a?.lastOpenedAt ?? 0));
+  for (const project of projects) {
+    if (typeof project?.path === 'string' && project.path) {
+      directories.push(project.path);
+    }
+  }
+  return [...new Set(directories)];
+};
+
 const openCodeLifecycleRuntime = createOpenCodeLifecycleRuntime({
   state: openCodeLifecycleState,
   env: {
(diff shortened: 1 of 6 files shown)


herjarsa:
Superseded by #2882 (conflicts resolved against latest main; onOpenCodeReady + onOpenCodeRestarted coexist; tests pass in CI). Please close this one.

btriapitsyn:

btriapitsyn:
Closing in favor of #2882, which carries this same change rebased onto current main.

PR #2642, opened 2026-08-05
fix(ui): never auto-send the message queue into a streaming turn

makeittech:
## Intent

When the session has a trailing **in-flight assistant message** (a turn is streaming) but the client's `session_status` map has **no entry** for it, the queued-message auto-send hook treated the session as `idle` and dispatched queued follow-up prompts into the running turn.

Why this state occurs: OpenCode's `/session/status` map only lists busy/retry sessions (idle sessions are absent), and a busy event can be missed on a reconnect/background window. The composer correctly recognized the session as busy (via the in-flight-message fallback in `useSessionActivity`) and queued the prompts — but the auto-send hook used a weaker gate (`session_status?.[id]?.type ?? 'idle'`) and drained the queue immediately. OpenCode then merged both prompt_async calls into a **single model turn** (confirmed in the incident: two user messages with empty `parentID`, one assistant response with `parent = P2`, model reasoning referencing both prompts), instead of waiting for the first response and sending the second sequentially.

The fix makes the queue gate honor the same live evidence as the composer: a trailing in-flight assistant message means busy, so the queue dispatches only after the turn completes (message completed or idle status event).

## Non-goals

- No change to `delivery: 'steer'` semantics in the web runtime (the SDK drops `delivery` for `promptAsync`; that is a separate, pre-existing issue tracked separately).
- No change to the composer's queue/steer decision logic in `ChatInput.tsx`.
- No change to the server-owned message queue (`packages/web/server/lib/message-queue`), which is used by messenger surfaces and already drains one item at a time.

## Affected surfaces

- `packages/ui` — shared UI library consumed by web, desktop (Electron), VS Code, hosted mobile, and Capacitor mobile surfaces. The fix lives in `useQueuedMessageAutoSend`, which is mounted in `SyncRuntimeEffects` for every surface, so all runtimes benefit identically; the change is runtime-agnostic (pure store-state reads).
- User-visible behavior: while a session is busy, queued follow-ups stay queued and are sent one-by-one after the turn completes, instead of being merged into the running turn.
- Persisted state: none — `resolveQueuedSessionStatusType` reads live child-store state only; no persisted schema, settings, or storage changes.

## Repository guidance

|---|---|---|
| `sync-state-invariants` skill | Change touches the queued auto-send gate, live status derivation, and fallback scoping ("derive live activity from live channels, not persisted history"; "scope delayed-live fallbacks to the active entity and clear them when authoritative state arrives") | The fallback uses the live in-flight trailing message; it clears automatically when the message completes or an idle status event lands (the effect now subscribes to `state.message` for exactly that edge). |
| `openchamber-change-discipline` skill | Shared UI behavior change; smallest complete change; regression test first | Extracted the inline gate into `resolveQueuedSessionStatusType` without behavior change (verified: full sync suite identical to baseline), then added a red/green regression test, then the minimal fallback. |
| `sync/DOCUMENTATION.md` | Owning sync documentation | Reviewed; it does not document the auto-send gate internals, so no doc update required. Module ownership and contracts are unchanged. |
| `AGENTS.md` (Runtime Boundaries, Correctness Invariants) | Shared UI must define intentional behavior for every runtime | The fix is runtime-agnostic; all surfaces mount the same hook. |

## Validation

| Check | Result |
|---|---|
| Regression test before fix (red) | `expect(resolveQueuedSessionStatusType(...)).toBe('busy')` failed — received `"idle"` (bug reproduced: in-flight assistant + missing status entry) |
| Regression test after fix (green) | `bun test packages/ui/src/hooks/useQueuedMessageAutoSend.test.ts` → 15 pass, 0 fail (4 new tests: in-flight→busy, busy/retry entry, completed message→idle, explicit idle/unknown→idle) |
| Queue store tests | `bun test packages/ui/src/stores/messageQueueStore.test.ts` → 7 pass, 0 fail |
| All hooks tests | `bun test packages/ui/src/hooks/` → 21 pass, 0 fail |
| Sync suite | `bun test packages/ui/src/sync/` → 375 pass; **49 fail + 2 errors — identical to clean `origin/main` baseline** (verified by stashing the change and re-running: all failures are pre-existing environment issues in worktree-store/input-store/etc., none related to this change) |
| Workspace type-check | `bun run type-check` → all 5 packages (mobile, ui, web, electron, root) exit 0 |
| Workspace lint | `bun run lint` → all 5 packages exit 0 |
| Build | `bun run build:web` (vite bundle of the shared UI) → `✓ built in 2m 33s`, no errors |
| Dead code | `bun run dead-code` → new export `resolveQueuedSessionStatusType` not flagged |
| Not verified | Full `bun run build` (electron packaging/mobile assets) and live runtime behavior on each device surface; static checks only for those. |

## Visual evidence

No user-visible rendering change. The diff only adds a status-resolution helper and two call sites inside `useQueuedMessageAutoSend`; it cannot alter any rendered element, styling, or layout. The observable behavior change is temporal (queued prompts now wait for the turn to finish) and is covered by the regression test asserting the gate's output.

## Risks and failure behavior

- **Missed idle events**: if both the busy and idle status events are missed, the queue now waits until the trailing message completes (the new `state.message` subscription re-runs the effect and the fallback clears) or any status event lands. Worst case is a delayed send, never a premature one — strictly safer than the previous behavior.
- **Turn that never completes** (hung step): the fallback keeps the queue blocked, same as the composer's existing behavior for the stuck-turn case; the existing Stop/abort path and stall-watchdog handle that scenario unchanged.
- **Rollback**: no persisted data, settings, or IDs change; reverting the commit fully restores prior behavior.
- **Performance**: the effect adds one subscription to `state.message` (whole-store message record), which changes on message create/complete only — streaming part updates touch `state.part`, not `state.message`. The dispatch gate reads the child store synchronously as before.
- **Cross-runtime**: no runtime-specific branches; all surfaces mount the same hook.
- **Security/data**: no new logging, secrets, or persisted content.

### DIFF
--- a/packages/ui/src/hooks/useQueuedMessageAutoSend.test.ts
+++ b/packages/ui/src/hooks/useQueuedMessageAutoSend.test.ts
@@ -1,6 +1,8 @@
 import { beforeEach, describe, expect, mock, test } from 'bun:test';
-import type { Agent } from '@opencode-ai/sdk/v2';
+import type { Agent, Message } from '@opencode-ai/sdk/v2';
 import type { QueuedMessage } from '../stores/messageQueueStore';
+import { ChildStoreManager } from '@/sync/child-store';
+import { setSyncRefs } from '@/sync/sync-refs';
 
 let visibleAgents: Agent[] = [];
 const sendMessageCalls: unknown[][] = [];
@@ -32,6 +34,7 @@ import {
   createQueuedAutoSendRetryScheduler,
   getQueuedAutoSendRetryDelayMs,
   isQueuedAutoSendBackedOff,
+  resolveQueuedSessionStatusType,
   sendQueuedAutoSendPayload,
   shouldDispatchQueuedAutoSend,
 } from './useQueuedMessageAutoSend';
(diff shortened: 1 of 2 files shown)


makeittech:
Nice fix.

PR #2655, opened 2026-08-05
fix(chat): show context text before a pending question

sdziupin:
## Intent

Fixes https://linear.app/openchamber/issue/OPE-199

When the model asks a question (question tool), the message's turn is blocked until the user answers — it never reaches finish `'stop'`. In sorted render mode, two existing rules then combine to hide everything the model produced before the question:

1. `projectTurnActivity` classifies every text part of a message with tools (or without a `stop` finish) as a `justification` activity, which is rendered inside the collapsible "Activity" group instead of inline.
2. `MessageBody` defers all inline text while `!hasStopFinish` ("keep ambiguous text deferred until final output", introduced in #1032-era sorted-activity work).

Because the question blocks the turn, the text is deferred forever: the user sees only the tool rows and the question card, while OpenCode shows the full context text inline (verified against the issue's side-by-side screenshot: OpenCode renders "ClickUp task release notes workflow … on: Strict mode enforced … Recommended rollout: … Update Impact: …" before the question; OpenChamber shows none of it).

Fix: for messages that contain a `question` tool part, (a) skip `justification` classification so the text renders inline, and (b) disable the sorted-mode inline-text deferral. This covers both the pending window and the post-answer streaming window (after the user answers, the message resumes streaming and would otherwise hide the text again mid-turn).

## Non-goals

- No change to question-card rendering, the question store, or sync logic.
- No change to deferral behavior for ordinary streaming messages (text still deferred until `stop` for non-question turns).
- No change to the `QuestionCard` overlay position in `ChatContainer`.

## Affected surfaces

- `packages/ui` only; affects all runtimes (web, Electron, VS Code, Capacitor) because they share `MessageBody` and the turn-projection code. Sorted chat render mode is the default chat layout; live mode already rendered the text inline and is unaffected.
- User-visible change: while a question is pending (and while the same message resumes streaming after an answer), the assistant's context text before the question is visible inline, between the tool rows, matching OpenCode.

## Repository guidance

|---|---|---|
| `AGENTS.md` runtime boundaries + always-on constraints | Shared UI rendering logic in `packages/ui`; minimal diff; no new dependencies | 2 source files + 1 test file; no deps, no secrets, no unrelated files staged |
| Project skill `theme-system` | UI/text rendering change in chat | No new colors or tokens introduced; text reuses the existing `AssistantTextPart` rendering path |
| Project skill `sync-state-invariants` | Only loaded-if-store-changes per instructions; this change is pure rendering classification, no sync store edits | Verified no `sync/` files are touched; the question tool part's `state` is read from the existing parts store without new subscriptions |
| Nearest docs: `packages/ui/src/sync/DOCUMENTATION.md`, `packages/ui/README.md`, `packages/ui/src/components/chat/lib/turns/` (no module DOCUMENTATION.md) | Understand part-state semantics and turn projection invariants | Read before editing; the fix preserves turn projection (tools stay in the Activity group) and only moves TEXT classification |
| `projectTurnRecords.test.ts` precedent | Existing tests for activity classification (compaction-as-justification) | New regression test follows the same fixture style (`createMessageEntry` + `showTextJustificationActivity: true`) |

## Validation

| Check | Result |
|---|---|
| `bun test packages/ui/src/components/chat/lib/turns/projectTurnRecords.test.ts` | PASS — 14 pass, 0 fail, 47 expect() (includes new pending-question test) |
| `bun test packages/ui/src/components/chat/lib/turns/` | PASS — 23 pass, 0 fail, 64 expect() across 4 files |
| `bun run type-check:ui` (worktree `fix/ope-199-question-context-hidden`) | PASS — `tsc --noEmit` clean, exit 0 |
| `bun run lint:ui` | PASS — eslint clean, exit 0 |
| `git status` before commit | Only the 3 intended files modified |

Not verified: no browser available, so the rendering was not visually confirmed on a live session with a pending question. The behavior is asserted indirectly by the new unit test (text parts of a question-blocked message are no longer classified as justification, so `MessageBody`'s flat loop renders them inline) and by code reading of the two deferral/classification paths. `bun run dead-code` not required — no files added/removed, no exports changed.

## Visual evidence

The chat surface is captured; the specific "context text visible before a pending question" state is not reproducible in this environment, for the precise reasons below.

- `chat-surface.png` — the chat surface from this PR's worktree (`fix/ope-199-question-context-hidden`, `dev:web:hmr`, headless Chromium 1440×900), a real long-running session (287 messages) rendered with this PR's `MessageBody` turn classification, showing the message list and composer. Raw URL: https://raw.githubusercontent.com/makeittech/openchamber/evidence/PR-2655/screens/chat-surface.png

Why the before/after of the fix cannot be captured live here:

- The fix changes how text parts render when the same message also contains a `question` tool part. Rendering that state requires a message with a question part — which requires a live agent turn that invokes the question tool. This environment has **no LLM API credentials**, so no agent can produce one.
- Message parts are server-authoritative (OpenCode server storage via the sync layer), not in any persisted client store, so localStorage/zustand seeding cannot inject a question part. The SDK message APIs only accept a prompt and stream a model turn; there is no arbitrary-parts injection. A scan of every session on the dev server (all project directories, incl. the 20+ long sessions in the main repo) found zero messages containing a question part — no historical message can demonstrate the state either.
- The repo has no storybook/component harness to render `MessageBody` outside the app.

The classification behavior itself is directly covered by the PR's new unit tests in `packages/ui/src/components/chat/lib/turns/` (23 pass, 0 fail, 64 expect across 4 files — cited in Validation): text parts of a question-blocked message are no longer classified as justification, so the flat loop in `MessageBody` renders them inline in order (context text → tool rows → question row → card), matching the OpenCode comparison. The visual result is a pure function of that classification, so the test suite pins the exact behavior the screenshot would show.

## Risks and failure behavior

- Scope is limited to messages containing a `question` tool part (any state). A message with a completed/rejected question keeps its text inline too — this matches OpenCode and cannot regress the deferral of unrelated turns.
- The question tool part itself still participates in the Activity group as a tool row (`tool` kind), so no activity-group behavior changes for tools.
- If the question part's `state`/parts data is ever absent mid-stream, `hasQuestionTool`/`messageHasQuestion` fall back to "no question" and prior behavior applies — no crash path (pure `some()` on existing arrays).
- No persisted contracts, storage, or sync state are touched.


btriapitsyn:
Thanks for tracking this down — the fix is solid and the test covers the real scenario well.

Merged in main and made one small tweak while landing: switched the question tool name check from `toLowerCase() === 'question'` to an exact-match comparison, matching the convention used elsewhere in this file. Also re-merged latest main to pick up unrelated changes since this PR opened.

Ran the test suite, type-check, eslint, and oxlint on the changed files — all clean.

PR #2658, opened 2026-08-05
fix(mobile): keep session screen on transient resume/online probe failures

makeittech:
## Intent

Fixes https://linear.app/openchamber/issue/OPE-319

On the native mobile app (Capacitor iOS/Android), the session screen randomly returned to the server-choose (connect) screen. Root cause: every background→foreground resume and every WebView `online` connectivity event ran `handleNativeResume` → `reprobeActiveConnection()` with fast (2.5s) probe timeouts. A transient network blip (Wi-Fi/cellular handover, radio congestion, a Keychain read hiccup) made the probe fail, and the code tore the runtime endpoint down (`switchRuntimeEndpoint({ apiBaseUrl: '' })`), landing the user on `MobileConnectionWelcome` — even though the sync pipeline recovers from transient drops on its own (visibility/system-resume reconnect + exponential backoff + watchdog).

Behavior change: the resume/online re-probe now tears the connection down **only** on an explicit auth rejection (`needs-login` — revoked/expired token, shown with an `auth-expired` banner). `unreachable` and `no-connection` are treated as transient: the runtime endpoint and session screen are kept, and the sync layer's own reconnect loop recovers. The user leaves the session screen only by explicit action — the recovery UI Cancel button, the Instances sheet, or deleting the active connection.

## Non-goals

- Cold-launch classification (mount-only effect) still drops to the connect screen when the saved instance is unreachable/rejected at startup — there is no session screen to protect yet, and the user must pick a server. Unchanged.
- No change to the connect/pairing/password flows themselves.
- No change to `reprobeActiveConnection()`'s transport hot-switch (`switched` still rebinds LAN⇄relay in place).

## Affected surfaces

- `packages/ui/src/apps/MobileApp.tsx` — `handleNativeResume` decision logic (Capacitor-only runtime).
- `packages/ui/src/apps/mobileConnections.ts` — new exported policy helper `isFatalResumeProbeOutcome`; no behavior change to the module itself.
- `packages/ui/src/apps/mobileConnections.test.ts` — new focused test.
- Unaffected: web/desktop/VS Code runtimes (this code path is gated on `isNativeMobileApp`/`isCapacitorApp`); the connect screen itself; persisted connections.

## Repository guidance

|---|---|---|
| `sync-state-invariants` | Change touches mobile bootstrap/reconnect resilience and transient-failure handling | Keeps authoritative runtime state during transient failure; only explicit auth rejection is treated as fatal; the pipeline (not the probe) remains the reconnect authority |
| `ui-api-decoupling` | Change sits in the Capacitor runtime boundary and uses `switchRuntimeEndpoint` | No new transports or API routes; runtime switching behavior preserved; policy helper lives in the owning mobile app module |
| `packages/mobile/README.md` | Mobile runtime model: connections saved locally, connect screen is Capacitor-only | Change only affects the Capacitor shell path (`isNativeMobileApp`), browser-hosted mobile keeps prior behavior |
| `packages/ui/src/sync/DOCUMENTATION.md` | Defines reconnect/polling/bootstrap rules ("Preserve previous authoritative state during transient bootstrap/reconnect failures") | The fix implements exactly that rule for the resume re-probe path |

## Validation

| Check | Result |
|---|---|
| `bun run type-check:ui` (worktree) | PASS — `tsc --noEmit`, no errors |
| `bun run lint:ui` (worktree) | PASS — eslint clean |
| `bun test packages/ui/src/apps/mobileConnections.test.ts` | PASS — 11 pass / 0 fail, 28 expect() calls (10 existing + 1 new policy test) |
| `bun run dead-code` (worktree root) | Run; report is non-blocking, new export not flagged (listed `mobileConnections.ts` entries are pre-existing) |
| iOS Simulator / real server run | NOT RUN — this environment cannot run the iOS simulator or a live OpenChamber server. The reconnect-on-resume behavior (resume → flaky network → session screen stays) was verified by code path analysis only: the changed branches keep the endpoint set (`hasRuntimeEndpoint` stays truthy) and never bump the epochs, so the render condition `!hasRuntimeEndpoint || (!isConnected && !isReconnecting)` cannot switch to `MobileConnectionWelcome`; the pipeline's own `visibilitychange`/`system-resume` listeners reconnect. |

## Visual evidence

No screenshots possible — the iOS simulator cannot be launched in this session. Expected rendering: during a transient network blip the session screen remains visible (possibly with the reconnecting state); the connect screen appears only after an explicit user action (recovery Cancel / Instances) or an auth-expired notice after a real token rejection. The code change is a decision-flow change; no visual states were added.

## Risks and failure behavior

- **Permanently dead server while app is foregrounded**: the app now stays on the session screen in the reconnecting state instead of auto-bouncing to the connect screen. Manual escape routes remain: recovery UI Cancel (when shown), the Instances entry in the sessions footer, or app relaunch (cold-launch auto-connect still classifies unreachable/rejected and shows the connect screen with a reason). This is the intended trade-off of the issue ("stays until the user manually leaves").
- **`needs-login` is still fatal** exactly as before — revoked/expired tokens cannot silently masquerade as reachable.
- **No new dependencies, no persisted-contract changes, no logging of tokens or user data** (existing `[mobile-connect]`/`[mobile-storage]` logs are unchanged).
- Rollback: revert of the single commit restores prior behavior; no data migrations involved.

### DIFF
--- a/packages/ui/src/apps/MobileApp.tsx
+++ b/packages/ui/src/apps/MobileApp.tsx
@@ -54,7 +54,7 @@ import { MobileSessionsSheet } from './MobileSessionsSheet';
 import { MobileFullscreenSurface } from './MobileFullscreenSurface';
 import { MobileWorkspaceDrawer, type MobileWorkspaceTab } from './MobileWorkspaceDrawer';
 import { DedicatedMobileAppProvider, type MobileAppActions } from './mobileAppContext';
-import { autoConnectLastInstance, getAutoConnectTargetLabel, reprobeActiveConnection, type AutoConnectOutcome } from './mobileConnections';
+import { autoConnectLastInstance, getAutoConnectTargetLabel, isFatalResumeProbeOutcome, reprobeActiveConnection, type AutoConnectOutcome } from './mobileConnections';
 import { isCapacitorMobileApp, useNativeAndroidBackButton, useNativeMobileChrome, useNativeMobileLifecycle } from './mobileNativeChrome';
 import { reconnectAppForTransportSwitch, resetAppForRuntimeEndpointChange } from './runtimeEndpointReset';
 import { useAppFontEffects } from './useAppFontEffects';
@@ -683,10 +683,14 @@ export function MobileApp({ apis }: MobileAppProps) {
     void reprobeActiveConnection().then((outcome) => {
       if (nativeResumeValidationSeqRef.current !== validationSeq) return;
       if (outcome === 'no-connection') {
-        disconnect();
+        // The runtime endpoint is up but it does not map to a saved connection
+        // (a bookkeeping mismatch, e.g. after a candidate rewrite). That is not
+        // a transport failure — keep the current connection; the sync pipeline
+        // reconnects on its own.
+        refreshInPlace();
         return;
       }
-      if (outcome === 'needs-login') {
+      if (isFatalResumeProbeOutcome(outcome)) {
         // Token explicitly rejected (revoked/expired) — tell the user why they
         // land back on the connect screen instead of silently bouncing them.
         setAutoConnectNotice({ kind: 'auth-expired', label: getAutoConnectTargetLabel() ?? '' });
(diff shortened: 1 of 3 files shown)


makeittech:

PR #2667, opened 2026-08-05
fix(sync): settle completed turns and finished messages promptly

sdziupin:
## Intent

Fixes https://linear.app/openchamber/issue/OPE-193

Post-#483 stuck-spinner audit: the two still-reproducible busy-state edge cases, fixed without restructuring the polling/timeout design:

1. **B1 — busy spinner lingers after a turn completed.** When the turn-ending `session.idle` SSE event is delayed or lost, the session stays `busy` until the next watchdog poll tick (~5s) and its monotonic→authoritative escalation (~5–10s). Now an assistant `message.updated` that carries `time.completed` triggers one immediate directory status poll: the monotonic pass confirms/raises active status but never lowers it, and when the snapshot reports the session idle while the store still believes it busy, an authoritative resync settles the status in a single round-trip. One in-flight fetch per directory bounds fan-out; the 5s watchdog poll remains the backstop.

2. **C1 — completed assistant message stays marked streaming.** `streaming.ts` derived "streaming" purely from `session busy` + trailing assistant message. The server stamps `time.completed` only after the whole response (text + every tool call) finished, so a completed trailing message means the message itself is done even while the turn keeps running (next step / tool phase). The old code kept it flagged streaming until the session settled or the next message started — the typing indicator and the streaming part-update suspension lingered on finished content. A completed trailing message is now never marked streaming; both the full (`updateStreamingState`) and incremental (`updateChangedStreamingSessions`) derivations complete the previous streaming message instead.

Audit result for the remaining issue items (all verified against current `upstream/main`): A1 (send error paths) and A2 (`recentlyConfirmedIdle` part-drop guard) are already gone/fixed by the sync overhaul — `optimisticSend` clears the optimistic busy status on every error path, and the event pipeline has no idle-confirmation guard that could drop parts. B2 (5-min stuck timeout) was replaced by the watchdog escalation; B3 (zombie timeout) was removed (`isZombie` is never set); C2's completion-skip guard no longer exists. None of those are reproducible in current code, so no change was made to them.

## Non-goals

- No restructuring of the polling/timeout design (explicitly out of scope per the issue): the 5s watchdog poll, monotonic semantics, escalation path, and backstop behavior are unchanged — the fix only adds an earlier trigger on an authoritative completion signal.
- No changes to B2/B3 (design discussions) — they are not reproducible bugs in current code.
- No changes to the message-queue or send paths (A1 already handled).

## Affected surfaces

- `packages/ui` sync layer only: `sync/streaming.ts` (message-level streaming derivation), `sync/sync-context.tsx` (`message.updated` handler + exported `maybePollStatusAfterMessageCompletion`), `sync/DOCUMENTATION.md`, `sync/streaming.test.ts`, new `sync/__tests__/message-completion-status-poll.test.ts`.
- User-visible behavior on all runtimes (web, desktop, VS Code, hosted mobile, Capacitor): the busy spinner/status row settles promptly after a completed turn even when the idle event is missed; the typing indicator and streaming marker stop on finished assistant messages during multi-step turns.
- No persisted or external contract changes; no SDK surface changes.

## Repository guidance

|---|---|---|
| `sync-state-invariants` skill | Derives live activity from live channels; event reducers; polling/bootstrap rules | `time.completed` is an authoritative server signal used only to *stop* UI state; the poll preserves prior state on fetch failure (null → no mutation, verified by test); monotonic never lowers live busy state; the escalation reuses the tested `needsSnapshotAfterStatusPoll` semantics |
| `performance-engineering` skill | Streaming derivation is a hot path; polling must stay bounded | The immediate poll is deduped per directory (in-flight guard) and gated through `runBackgroundNetworkTask`; the streaming derivation change adds only a field check on the already-scanned trailing message (no new scans); the completed-message branch reuses `completeStreamingMessage`/`complete` instead of adding work |
| AGENTS.md always-on constraints + `packages/ui/src/sync/DOCUMENTATION.md` | Owning module documentation must be updated when invariants change | Documented both new invariants (message-vs-turn lifecycle; immediate poll on completion) in the streaming derivation section |
| Repo precedent (`useEventStream` history, issue text) | Issue B1 explicitly proposed "triggerSessionStatusPoll() after completeStreamingMessage" | Implemented exactly that intent in the modern architecture |

## Validation

| Check | Result |
|---|---|
| `bun run type-check:ui` (worktree root) | Pass (tsc --noEmit, 0 errors) |
| `bun run lint:ui` | Pass (eslint, 0 errors) |
| `bun test src/sync/streaming.test.ts` (packages/ui) | 10 pass, 0 fail (4 new tests: completed-trailing-message full + incremental paths, already-completed arrival, next-message continuity) |
| `bun test src/sync/__tests__/message-completion-status-poll.test.ts` | 7 pass, 0 fail (idle no-op, guards, settle-to-idle escalation, busy confirm, fetch-failure preservation, concurrent dedupe, in-flight dedupe) |
| `bun test src/sync/` (full sync suite) | 334 pass, 26 fail — the 26 failures are pre-existing on clean `upstream/main` (verified via `git stash`; clean baseline 330 pass/26 fail, where 7 of the 330 are this PR's untracked poll tests — net +4 passing tests, zero regressions). Failing files: `session-directory-adoption`, `session-deletion-cleanup`, `event-pipeline.test.js` (WS transport), `issue-2039`, skill-invocation tests — identical set with and without the change |
| `bun run dead-code` (worktree root) | No new findings; only pre-existing entries (`useSessionTextMessages`, `StreamingStore` type, etc., flagged on clean code too) |

Not verified: live server interaction. No running OpenChamber/OpenCode server in this environment, so the end-to-end behavior (real turn completion → idle event timing, polling round-trips over the network) could not be exercised against a live server. Verified at the unit level (derivation transitions, poll escalation semantics) and by code inspection (server `time.completed` semantics confirmed against OpenCode v1.18.12 `session/processor.ts` cleanup path).

## Visual evidence

This fix's observable behavior cannot be demonstrated statically: both fixes are timing-sensitive transitions (a completed turn settling the busy spinner on the `time.completed` signal, and a finished assistant message losing its streaming marker mid-turn) that require a live agent turn with real SSE event timing — the delayed/lost `session.idle` event and multi-step tool phases. This environment has no LLM API keys and no running agent, so no screenshot can show the transition itself, and none is faked.

What was verified instead (from Validation): `bun test src/sync/streaming.test.ts` (10 pass, 0 fail — 4 new tests for the completed-trailing-message full and incremental derivation paths, already-completed arrival, next-message continuity) and `bun test src/sync/__tests__/message-completion-status-poll.test.ts` (7 pass — idle no-op, guards, settle-to-idle escalation, busy confirm, fetch-failure preservation, dedupe), plus the full `src/sync/` suite (334 pass; the 26 failures are pre-existing on clean `upstream/main`).

Context screenshot (app booted from `fix/ope-193-spinner-edge-cases`; the home surface — sidebar, sessions list, command palette — renders normally):

context: app booted normally; the settle/streaming transitions are covered by the streaming and message-completion-status-poll unit tests, not by this screenshot

## Risks and failure behavior

- Fetch failure during the immediate poll preserves prior state (null snapshot → no mutation); the watchdog poll remains the recovery backstop, so a failing status endpoint cannot worsen behavior versus today.
- The monotonic pass can never lower a busy session; the authoritative settle only runs when the snapshot reports the session idle — the same race the existing watchdog escalation accepts (a snapshot taken between turns), bounded and transient.
- The immediate poll is bounded to one in-flight fetch per directory and only fires for sessions the store believes active, so completion bursts (bootstrap materialization) share a single fetch and cannot create a request storm.
- The streaming derivation change only affects the message-level typing/streaming markers, never `session_status`; a completed message receiving late tool-output parts still renders them (part updates remain wired), only the streaming marker/suspension is released.


btriapitsyn:
Thanks for this — the lingering spinner after a lost session.idle was a real annoyance, and stopping streaming.ts from treating a completed trailing message as streaming is the right call.

Two things changed while landing it:

- The status check is deferred by 750ms and re-reads the session status when the timer fires. Almost every turn gets its session.idle right behind the completed message, so those now settle on their own and cost zero extra requests; only a session the store still believes busy spends a fetch.
- It reuses the watchdog's in-flight directory guard instead of a separate set, so the deferred check and the periodic poll can't overlap on the same directory.

Also narrowed the completed-message check on the assistant message type instead of the cast, and added tests for both paths (idle inside the window: no request; idle never arrives: poll + resync). Merging.

PR #2669, opened 2026-08-05
feat(ui,server): surface active instance service URLs in About settings

makeittech:
## Intent

Fixes https://linear.app/openchamber/issue/OPE-194

When several app instances run from Git worktrees (one per branch, each on its own port), the port and tunnel URL were only printed to the terminal. The About settings page now shows the **active instance's service URLs** — the local application URL (`http://localhost:<port>/`) and, when a tunnel is active, the tunnel URL — as labeled buttons that open the URL in the browser. Hovering a button shows the service name (Application / Tunnel). Users can now tell which worktree instance they are looking at without switching terminals.

The proposal's log-parsing approach (`EXPORT <service>: <url>`) is intentionally **not** implemented: the server already knows its own port and tunnel URL as authoritative runtime state, so parsing terminal output would duplicate that state and go stale. Instead, `/api/system/info` (the existing server-info endpoint the About page already calls) now reports `port` and `tunnelUrl` resolved lazily from the tunnel runtime, and the UI renders them.

## Non-goals

- No log-file/terminal scanning for `EXPORT` directives — the data is read from the running server's own state instead (see Intent).
- No worktree-context gating of the new section. Detecting "running under a worktree" server-side would require a fragile git-spawn check; the section is informational and harmless in non-worktree installs (nothing else changes). It only renders when the server reports a port, so non-worktree behavior is otherwise untouched.
- No database URL surface — the server does not expose a database service URL via any API; only the application port and tunnel URL are authoritative.
- No changes to the CLI, tunnel lifecycle, or Electron shell.

## Affected surfaces

- `packages/web/server/lib/opencode/core-routes.js` — `GET /api/system/info` response gains `port` and `tunnelUrl` (both `null` when unwired; additive, backward compatible).
- `packages/web/server/lib/opencode/bootstrap-runtime.js` + `packages/web/server/index.js` — wire two optional getters through, resolved lazily at request time because the tunnel runtime is constructed after the base routes.
- `packages/ui/src/components/sections/openchamber/InstanceServiceUrls.tsx` (new) — fetches `/api/system/info` once on mount, renders labeled URL buttons (shared `Button`, `external-link` sprite icon, `openExternalUrl` for cross-runtime opening). Hides itself when the endpoint is unavailable or reports no port/tunnel, so VS Code runtime (which proxies a different server) shows nothing rather than stale data.
- `packages/ui/src/components/sections/openchamber/AboutSettings.tsx` — renders the component in both desktop and mobile layouts.
- `packages/ui/src/lib/i18n/messages/*.settings.ts` — three new keys in all 11 locales (real translations, no English stand-ins).
- Electron/VS Code/mobile runtimes: the endpoint lives on the OpenChamber server, which Electron starts in-process and mobile/VS Code connect to; the About page is rendered in all runtimes and degrades gracefully (section hidden) where the fields are absent.

## Repository guidance

|---|---|---|
| `AGENTS.md` runtime boundaries + correctness invariants | Server route change touches web server ownership; UI change must not fabricate state | The UI never synthesizes URLs: it renders only values reported by `/api/system/info`, and a failed fetch keeps the section hidden instead of showing wrong/empty data |
| `ui-api-decoupling` skill + `references/implementation-map.md` | Shared UI must use `runtimeFetch`, explicit OpenChamber routes, and runtime-aware URL opening | New component uses `runtimeFetch('/api/system/info')` and `openExternalUrl` (web/electron/vscode-aware); no hardcoded origins or raw fetches |
| `theme-system` skill + `references/tokens-and-examples.md` | No hardcoded colors; shared `Button` contract | Buttons use `variant="outline" size="sm"` from the shared `Button`; only layout classes (`gap`, `truncate`, `font-mono`) added; no hex/palette colors |
| `locale-ui-patterns` skill | Every user-facing string must be a translated i18n key in all dictionaries | Three keys added to `en` and all 10 non-English dictionaries with real translations |
| `packages/web/README.md` | Owns the CLI/server surface and tunnel behavior | Confirms tunnel URL semantics (`openchamber tunnel status`) and that each instance is a separate server on its own port; endpoint additions reuse existing tunnel state (`tunnelService.getPublicUrl()`) |
| `openchamber-change-discipline` | Minimal, scoped changes; validation per package | Server change is 3 files + tests; UI change is 1 new component + 1 integration point; no dependencies added |

## Validation

| Check | Result |
|---|---|
| `bun test server/lib/opencode/core-routes.test.js` (packages/web) | 26 pass / 1 fail — the single failure (`rate limits pairing redeem attempts...` at line 431, `vi.setSystemTime is not a function`) is **pre-existing**: reproduced with `git stash` on the untouched base commit (24 pass / 1 fail). Two new tests for `/api/system/info` (defaults → nulls; wired getters → values) pass |
| `bun build server/index.js --no-bundle` and `bun build server/lib/opencode/bootstrap-runtime.js --no-bundle` (packages/web) | Both bundle successfully — syntax check for the plain-JS server surface |
| `bun run type-check` (packages/web) | Pass (`tsc --noEmit`, exit 0) |
| `bun run lint` (packages/web) | Pass (eslint, exit 0) |
| `bun run type-check` (packages/ui) | Pass (`tsc --noEmit`, exit 0) |
| `bun run lint` (packages/ui) | Pass (eslint, exit 0) |
| `bun run dead-code` (repo root) | Runs; report contains only pre-existing unused-type/duplicate-export entries — nothing references the new `InstanceServiceUrls.tsx` or the new server getters |

Not verified: no live server was started in this environment (no browser/screenshot run), so the rendered buttons were not exercised against a running instance; the endpoint behavior is covered by the focused tests above. The 1 pre-existing test failure was not fixed (out of scope).

## Visual evidence

Captured live against the fixed build (`dev:web:hmr`, OpenChamber server on port 3910, no tunnel configured):

| Screenshot | What it shows |
|---|---|
| [Settings → About with service URL button](https://raw.githubusercontent.com/makeittech/openchamber/evidence/PR-2669/screens/1-about-mobile-service-urls.png) | About settings page: below the version rows, the "Instance URLs" row renders the **Application** button `http://localhost:3910/` (monospace, with hover title "Application"). No Tunnel button appears because `tunnelUrl` is `null` — the server reports `{"port": 3910, "tunnelUrl": null}` from `GET /api/system/info` |

Notes on the capture, for honesty:
- The About settings page (and therefore `InstanceServiceUrls`) renders on the mobile web surface (the `'about'` settings page is gated to mobile per `metadata.ts`; the desktop header "About OpenChamber" opens the separate `AboutDialog` which this PR does not touch). The screenshot was taken on the mobile surface (`?surface=mobile`), which is the surface that renders the new component.
- The button was verified in the DOM (not just pixels): a `<button title="Application">http://localhost:3910/</button>` — the port matches the running instance's API port reported by `/api/system/info` (the endpoint this PR extended). No tunnel was configured in this environment, so the Tunnel button is legitimately absent; the port button alone is the evidence.
- Clicking the button opens the URL via `openExternalUrl`; opening a new tab was not exercised to keep the single-page browser session intact.

## Risks and failure behavior

- **Stale/incorrect URLs**: impossible by construction — the UI renders only values returned by the server at request time; fetch failure or missing fields hides the section (explicit, no fallback data).
- **Backward compatibility**: `/api/system/info` gains two fields with `null` defaults; existing consumers (AboutDialog, AboutSettings) ignore unknown fields. Older wiring (tests, other entrypoints) that omits the getters gets `null`.
- **Port resolution race**: the getter falls back from the tunnel runtime's active port (set after the startup pipeline binds, covering `--port 0`) to the configured port, then `null`; the route is only reachable after startup, so the holder is always populated in practice.
- **Security**: no credentials or tokens are exposed — `tunnelUrl` is the same public URL already returned by `/api/openchamber/tunnel/status` and shown in tunnel settings; `port` is the local bind port.
- **Performance**: one GET on About-page mount, no polling, no new state stores.
- **Cross-runtime**: VS Code runtime shows no section (fields null or request fails through the bridge) — intentional, matches existing About behavior of best-effort system info.


makeittech:
Good catch.

PR #2672, opened 2026-08-05
feat(fs): add file upload with drag-and-drop to the file manager

makeittech:
## Intent

Fixes https://linear.app/openchamber/issue/OPE-192

Allow uploading (and downloading) files in the project file manager, which is what makes deployed web instances usable. Investigation found **download was already fully implemented** (server: `GET /api/fs/raw?download=true` with RFC 5987 `Content-Disposition` in `packages/web/server/lib/fs/routes.js`; UI: `files.downloadFile` with per-row context-menu action, editor toolbar button, and binary-viewer download button in `FilesView.tsx`). The missing half was upload, which this PR adds end-to-end:

- Server: `POST /api/fs/upload` — raw binary body, streamed to a temp file and atomically renamed into place, with the same workspace containment as the existing write/read routes.
- Runtime APIs: `uploadFile(path, blob)` added to the shared `FilesAPI` contract and implemented for web and VS Code webview.
- UI: upload button in the file-tree toolbar (file picker, multi-select) and drag-and-drop onto the tree with a drop overlay, plus localized success/failure toasts.

## Non-goals

- No changes to the existing download flow (already present and covered by tests).
- No multipart/form-data parsing or new dependencies — the upload uses a raw `application/octet-stream` body, so no `multer`/busboy is needed.
- No upload into subdirectories via context menus — dropping on a specific directory row targets the current workspace root, matching the existing "new file / new folder" toolbar actions.
- No chunked/resumable upload protocol.

## Affected surfaces

- `packages/web/server/lib/fs/routes.js` — new `POST /api/fs/upload` route (server JS; express `json` middleware skips non-JSON content types, so the raw body stream reaches the handler untouched).
- `packages/web/server/lib/fs/DOCUMENTATION.md` — route list and upload policy documented per module contract.
- `packages/web/server/lib/fs/routes.test.js` — 7 new focused tests.
- `packages/ui/src/lib/api/types.ts` — `FilesAPI.uploadFile?` (optional, matching `downloadFile?`).
- `packages/web/src/api/files.ts` + `packages/vscode/webview/api/files.ts` — `uploadFile` implementations (web via `runtimeFetch`, VS Code via the webview fetch proxy which base64-encodes the body).
- `packages/ui/src/components/views/FilesView.tsx` — upload button, drag/drop handlers, drop overlay (guarded by `files.uploadFile` presence).
- `packages/ui/src/lib/i18n/messages/*.ts` — 6 new keys in all 12 dictionaries (fully translated).
- `packages/ui/src/components/icon/sprite.ts` — regenerated sprite (adds `upload-2` only).
- Runtimes unaffected by design: Electron and mobile bundle the web surface and therefore inherit the web `RuntimeAPIs` implementation automatically.

## Repository guidance

|---|---|---|
| `openchamber-change-discipline` | New route + shared API contract + UI | Smallest complete change; reused existing containment helpers (`resolveWorkspacePathFromContext`, `isPathWithinRoot`) instead of new policy; security enforced server-side, not just in UI; docs updated (fs `DOCUMENTATION.md`); validation run at every affected package. |
| `ui-api-decoupling` | New `/api/fs/*` route consumed through `RuntimeAPIs` | Route registered inside `registerFsRoutes` (before the generic OpenCode proxy); UI consumes `files.uploadFile` from `useRuntimeAPIs()`; web implementation uses `runtimeFetch('/api/fs/upload')` with the `x-opencode-directory` header; VS Code parity via the webview fetch proxy; read `references/browser-assets-and-auth.md` + `references/implementation-map.md`. |
| `theme-system` | New button/icon/overlay | Uses shared `Button` (variant `ghost`, size `sm`) and `Icon` from the sprite (`upload-2`, added via `bun run icons:generate` — sprite diff verified to add only that icon); overlay uses semantic tokens (`border-primary`, `bg-background`) — no hardcoded colors; read `references/icons.md`. |
| `locale-ui-patterns` | New visible strings (tooltip, toasts, overlay) | 6 keys added to `en.ts` and translated into all 11 non-English dictionaries (de, es, fr, ja, ko, pl, pt-BR, uk, zh-CN, zh-TW); params used only for real values (`{count}`, `{name}`); single/plural separated into complete-message keys. |

## Validation

| Check | Result |
|---|---|
| `node --check packages/web/server/lib/fs/routes.js` | Pass (server JS is plain JS — static TS checks don't cover it) |
| `bunx vitest run packages/web/server/lib/fs/routes.test.js` | 1 file, 31 tests passed (24 existing + 7 new upload tests) |
| `bunx vitest run packages/web/src/api/files.test.ts` | 1 file, 4 tests passed (2 existing + 2 new upload tests) |
| `bun run type-check:ui` | Pass |
| `bun run lint:ui` | Pass |
| `bun run type-check:web` | Pass |
| `bun run lint:web` | Pass |
| `bun run --cwd packages/vscode type-check` (both tsconfigs) | Pass |
| `bun run icons:generate` | Sprite regenerated: 237 icons; diff adds only `upload-2` |
| `bun run build:ui` (= tsc --noEmit) | Pass |
| `bun run --cwd packages/web build` (vite production build) | Pass |
| `bun run dead-code` (exports/contract changed) | Ran; report contains only pre-existing findings (same list as main baseline); none related to this change |

New upload tests cover: missing path (400), outside-workspace path (400), successful streamed write + atomic rename (200, temp name pattern, both chunks written, rename to final path), declared oversized `Content-Length` (413, no fs writes), mid-stream cap with lying/absent header (413, temp cleaned up), symlink target escaping workspace (403), symlinked parent escaping workspace (403), directory target (400).

**Not verified (no browser/runtime available in this environment):** real drag-and-drop interaction in a browser (the drop handlers follow the standard React drag event pattern and are guarded by `event.dataTransfer.types` containing `Files`), the VS Code webview base64-body proxy path for uploads, and the E2EE relay transport carrying a large upload body. These paths are implemented but were not exercised at runtime.

## Visual evidence

No screenshots could be captured — this environment has no browser or display. Expected rendering: a new ghost icon button (`upload-2`) at the right end of the file-tree toolbar (after Refresh), a dashed-border overlay with the localized "Drop files to upload" text while dragging files over the tree, and toast notifications for upload success/failure. The overlay uses `border-primary/60` + `bg-background/80` so it stays legible in both light and dark themes.

## Risks and failure behavior

- **Partial upload failures**: each file is uploaded independently; successes toast with the count and refresh the directory, failures toast separately with the failed names — one failed file never blocks or erases the others (matches the repo's "one failed entity must not erase complete entities" invariant).
- **Oversized uploads**: 413 returned before reading when `Content-Length` exceeds the cap, and mid-stream otherwise (temp file removed, request socket destroyed only after the response flushes). Cap is `OPENCHAMBER_FS_UPLOAD_MAX_BYTES` (default 100 MiB, mirroring `MAX_SERVE_BYTES`).
- **Containment**: the target path goes through the same `resolveWorkspacePathFromContext` used by every other fs write route; parent-directory and existing-target realpath checks reject symlink escapes with 403; directory targets rejected with 400; temp files use an unguessable suffix and are always unlinked on failure.
- **Client-side name sanitization** strips path separators from dropped file names (dropped directories are flattened to basenames), but the server independently re-validates the final path — UI sanitization is not the security boundary.
- **VS Code size caveat**: uploads in the VS Code webview travel as base64 through the extension bridge; very large files are inefficient there but bounded by the same server cap. Local file management in VS Code remains available through the native fs bridge.
- **Compatibility**: `uploadFile` is optional in the `FilesAPI` contract, so any runtime without it simply doesn't render the upload affordances — no behavior change for existing consumers.


btriapitsyn:
Thank you for the thorough implementation and especially for the careful upload-streaming and failure-handling work.

The overlapping Files sidebar feature landed on main in https://github.com/openchamber/openchamber/commit/7b92bf347, and the strongest remaining parts of this PR have now been incorporated in https://github.com/openchamber/openchamber/commit/853e0d43b: uploads stream into same-directory temporary files instead of being buffered in memory, the size cap is configurable and enforced before and during streaming, partial files are cleaned up, and directory targets are rejected explicitly.

The implementation on main keeps the current Sidebar Files tree, supports folder-aware drop targets, and preserves its no-silent-overwrite contract through an atomic no-replace commit plus an explicit Replace confirmation. VS Code upload parity was intentionally not carried over because that runtime does not expose this Files tree surface.

You are credited in the Unreleased changelog and as a co-author of the hardening commit. Closing this PR as superseded now that its valuable transport work is represented on main. Thanks again for the strong groundwork.

PR #2675, opened 2026-08-05
fix(instances): add back-to-local bar and drag region for remote instance pages

sdziupin:
## Intent

Fixes https://linear.app/openchamber/issue/OPE-204

When an OpenChamber desktop window navigates to a configured remote instance
that serves a **foreign** page — e.g. the user pointed an instance at their
VPS's `opencode web` server (Tailscale URL) — that page has:

- no drag region, so the frameless Electron window cannot be moved
  (sub-issue b), and
- no navigation affordance back to the local instance (sub-issue a).

The root cause is that the whole window navigates to the remote origin (the
"Open instance" action in Settings → Remote instances, the instance switcher,
and startup with a remote default host), and a foreign page has neither
OpenChamber's `.app-region-drag` CSS nor the instance switcher.

This PR keeps the existing navigation behavior but guarantees the window is
never left without an escape hatch: the Electron main process injects a slim,
always-draggable overlay bar with a "Back to OpenChamber" button into **any
non-OpenChamber page** rendered inside an OpenChamber window. Clicking the
button navigates back to the local UI (packaged `openchamber-ui://` URL, or
the local server origin in dev). OpenChamber UI documents — local, packaged,
or served by a managed SSH instance at a remote origin — are recognized via a
new `data-oc-ui` marker in `packages/web/index.html` and keep their own
chrome (header drag region + instance switcher), so no double chrome appears.

The user's stated ideal ("use the OpenChamber UI attached to my opencode web
instance") is out of scope: `opencode web` is a different server product and
not an OpenChamber server (confirmed by the maintainer in the original GitHub
issue thread), so this PR does not attempt to render foreign servers inside
the OpenChamber UI.

## Non-goals

- Supporting `opencode web` (or any non-OpenChamber server) as an instance
  the OpenChamber UI can attach to.
- Changing how SSH/managed instances are opened (window still navigates to
  the remote server's own UI, as before).
- Renderer-side UI changes: no React component, no i18n dictionaries, no
  theme tokens were touched (see Repository guidance).

## Affected surfaces

| Surface | Effect |
|---|---|
| `packages/electron/remote-shell.mjs` (new) | Builds the injected overlay script and the pure inject/don't-inject guard. |
| `packages/electron/main.mjs` | `createBrowserWindow` dom-ready: executes the overlay script in every non-Mini-Chat window; new `buildLocalUiBackTarget()` resolves the back URL (packaged UI URL vs local server origin, mirroring `openMainWindow`). |
| `packages/web/index.html` | New inline script sets `data-oc-ui="openchamber"` on `<html>` so the main process recognizes OpenChamber UI documents (any origin). Survives the vite build (verified in `dist/index.html`). |
| `packages/electron/remote-shell.test.mjs` (new) | Guard-logic unit tests + script parse/content/secret-absence tests. |

Unaffected and why:

- **OpenChamber UI at remote origins (managed SSH instances)**: already
  renders desktop chrome everywhere because `isDesktopShell()` is true on any
  page in the window (preload exposes `__OPENCHAMBER_ELECTRON__`
  unconditionally); the `data-oc-ui` marker makes the injection skip it.
- **Mini Chat windows**: their own `will-navigate` guard never loads remote
  pages, so the injection is skipped there.
- **Web/VS Code/mobile runtimes**: no shared UI or server code changed.
- **Navigation guard**: the back target is allowed by the existing
  `isAllowedNavigationUrl` (UI protocol / `state.localOrigin`).

## Repository guidance

|---|---|---|
| `AGENTS.md` — Runtime Boundaries | Electron owns windows and the privileged boundary; shared UI owns React | The fix lives in `packages/electron` (window/navigation ownership); the only UI change is a marker in `index.html`, no shared contracts touched. |
| Skill `desktop-shell` | Electron main process, window chrome, drag regions, remote pages, security boundaries | Overlay is injected via existing `webContents.executeJavaScript` (same pattern as the init script); contains no secrets (only the local UI URL, already exposed to remote pages as `__OPENCHAMBER_LOCAL_ORIGIN__`); remote pages gain no IPC privileges; `packages/electron/README.md` read. |
| Skill `openchamber-change-discipline` | Source change across packages | Smallest complete change; new module is the owning unit with focused tests; existing behaviors (navigation, host switching, packaged/init-script flows) preserved; `bun run dead-code` run. |
| Skill `locale-ui-patterns` | User-facing button label | The label lives in the Electron main process, not a React component: i18n (`useI18n`) cannot run inside a foreign page (no React/OpenChamber bundle). Hardcoded English matches existing main-process precedent ("Quit OpenChamber?", "Connect to OpenChamber server?", menu labels). Product name "OpenChamber" is exempt from translation per the skill. |
| Skill `theme-system` | Colors/buttons on the overlay | The overlay is injected into pages that have no OpenChamber theme CSS, so Tailwind/token classes are unusable; inline styles use the brand values that the theme tokens resolve to (`surface-background: #151313`, `primary-base: #edb449`). No React component or theme asset changed. |
| `packages/electron/README.md` | Electron package owning docs | Read; behavior addition is consistent with "main.mjs owns windows… SSH/tunnels" and the preload security model. |

## Validation

| Check | Result |
|---|---|
| `bun run type-check:electron` (`node --check main.mjs && node --check preload.mjs`) | PASS |
| `node --check ./remote-shell.mjs` + `node --check ./remote-shell.test.mjs` (electron package) | PASS |
| `bun test packages/electron/remote-shell.test.mjs` | 12 pass / 0 fail (guard matrix, script parse, embedded target, ids/marker, secret absence) |
| `bun test packages/electron/remote-shell.test.mjs packages/electron/ssh-manager.test.mjs` | 19 pass / 0 fail (adjacent SSH manager tests unaffected) |
| `bun run type-check:web` / `bun run lint:web` | PASS (index.html is not covered by either — noted) |
| `bun run build:web` | PASS (54.6s); `data-oc-ui` marker present in `packages/web/dist/index.html` |
| `bun run bundle:main` (electron) | PASS; `remote-shell.mjs` bundled into `dist-bundle/main.mjs` with `minify:false` (so the serialized guard keeps its source form) |
| `bun run dead-code` | No findings for the new files (removed one unused export flagged by the report) |
| Runtime simulation (uncommitted shim, `/tmp`) | Injected script executed against a fake DOM: foreign page → bar injected and click navigates to the local UI; OpenChamber UI at remote origin / local page / `openchamber-ui://` / `data:` splash → not injected (5/5 PASS) |

**Not verified**: real Electron runtime behavior — this environment cannot
launch the packaged desktop app against a live remote instance. `-webkit-app-region`
dragging, overlay rendering on the actual opencode web page, and the
back-navigation round-trip need a manual check on a real VPS instance.

## Visual evidence

**No screenshots possible in this environment — by design.** This change is an Electron **main-process** injection (`packages/electron/remote-shell.mjs`): the injected bar is produced by the privileged desktop shell while rendering a foreign page (the remote opencode web UI) inside the Electron `BrowserWindow`, behind a live SSH remote instance. That runtime cannot be reproduced here: this environment has no display server, no packaged Electron app, and no reachable remote instance to SSH into, so there is no browser surface to drive or capture. Unlike the web-only PRs in this series, nothing about this change exists in the shared web UI, so no headless web capture can illustrate it.

What the behavior was verified with instead (see Validation):

- **12/12 unit tests pass** (`bun test packages/electron/remote-shell.test.mjs`, rerun locally during capture: `12 pass / 0 fail`, 25 expect calls) — covering the guard matrix (which URLs/origins get the bar), script parsing, the embedded target id, markers, and secret absence.
- **Fake-DOM runtime simulation (5/5 PASS)** — the injected script executed against a stub DOM: foreign page → bar injected and its back button navigates to the local UI; OpenChamber UI at remote origin, local page, `openchamber-ui://`, and `data:` splash → **not** injected.
- `bun run bundle:main` embeds the guard source verbatim (`minify:false`) into `dist-bundle/main.mjs` — confirmed during capture.

Expected rendering for a manual/real-VPS check (unchanged from implementation): on a foreign page, a 28px dark bar across the top of the window with an amber mark, "OpenChamber", and a "← Back to OpenChamber" button; the bar is a drag region (`-webkit-app-region: drag`) while the button is not; on OpenChamber UI pages nothing is injected. This PR therefore carries no screenshot evidence branch; runtime confirmation requires the manual desktop check noted in Validation.

## Risks and failure behavior

- **Overlay on foreign pages**: the injection is additive and defensive
  (try/catch, idempotent via element id, inline styles, no globals, no page
  content read). If it throws on some page, only the bar is missing — the
  page is unaffected and behavior is identical to today.
- **Top 28px click interception**: the bar covers the top strip of the remote
  page, so page controls under it are not clickable there (dragging requires
  the strip). This is the intended trade-off for a movable window and an
  escape hatch; the bar is slim and visually reads as app chrome.
- **Back navigation**: target is the packaged UI URL or the local server
  origin, both already allow-listed by `isAllowedNavigationUrl`. If the local
  server is unavailable, the navigation shows the app's normal unreachable
  state rather than stranding the user on a foreign page with no affordance.
- **Dev-mode HMR edge**: the back target uses `sidecarUrl || localOrigin`
  (same resolution as `openMainWindow`); if no built UI exists at that origin
  in a dev setup, the back page shows the server's "static files not found"
  state — identical to existing startup behavior, not a regression.
- **No secrets**: the injected script embeds only the local UI URL, which the
  preload already exposes to remote pages. No filesystem, shell, token, or
  host secrets are included; the desktop IPC surface for remote pages is
  unchanged.
- **Compatibility**: no persisted data, routes, exports, or dependencies
  change; older OpenChamber UI builds without the marker simply fall back to
  today's behavior (overlay would appear over the old UI's own header — a
  cosmetic degradation that self-heals on the next UI refresh).

### DIFF
--- a/packages/electron/main.mjs
+++ b/packages/electron/main.mjs
@@ -12,6 +12,7 @@ import { promisify } from 'node:util';
 import updaterPkg from 'electron-updater';
 import { ElectronSshManager } from './ssh-manager.mjs';
 import { createTrayController } from './tray.mjs';
+import { buildRemoteShellInjectionScript } from './remote-shell.mjs';
 import { resolveManagedOpenCodeCwd } from './opencode-cwd.mjs';
 import { resolveStartupUrlProbePlan, shouldIgnoreLoopbackConnectionLimit } from './startup-url-selection.mjs';
 import { sanitizeRuntimeRequestHeaders } from './runtime-request-headers.mjs';
@@ -2579,6 +2580,19 @@ const createBrowserWindow = ({ label, restoreGeometry, url, runtimeConfig = {} }
     if (initScript) {
       void browserWindow.webContents.executeJavaScript(initScript).catch(() => {});
     }
+    // Remote pages loaded in an OpenChamber window (e.g. an instance whose
+    // server serves a foreign UI such as opencode web) get a slim draggable
+    // bar with a "Back to OpenChamber" button. OpenChamber UI documents and
+    // local/packaged pages are detected in-page and skipped. Mini Chat is
+    // exempt: its own navigation guard never loads remote pages.
+    if (!browserWindow.__ocMiniChat) {
+      const localUiTarget = buildLocalUiBackTarget();
+      if (localUiTarget) {
+        void browserWindow.webContents.executeJavaScript(
+          buildRemoteShellInjectionScript({ localUiUrl: localUiTarget }),
+        ).catch(() => {});
+      }
+    }
   });
 
   browserWindow.webContents.on('did-finish-load', () => {
@@ -2656,6 +2670,16 @@ const activateMainWindow = async (url, localOrigin, bootOutcome, runtimeConfig =
     runtimeConfig,
   });
   return state.mainWindow;
+};
+
+// Where the main window boots when no saved host applies: the packaged UI in
+// bundled builds, otherwise the local server origin. Mirrors openMainWindow's
+// local UI resolution so the remote-page overlay and window startup agree.
+const buildLocalUiBackTarget = () => {
+  if (shouldUsePackagedUi()) {
+    return buildPackagedUiUrl('/index.html');
+  }
+  return state.sidecarUrl || state.localOrigin || '';
 };
 
 const openMainWindow = async () => {
--- a/packages/electron/remote-shell.mjs
+++ b/packages/electron/remote-shell.mjs
@@ -1,0 +1,117 @@
+// Remote-page shell overlay (Electron main process).
+//
+// The OpenChamber UI renders its own window chrome — title-bar drag regions
+// and the instance switcher — everywhere it runs, including when it is served
+// from a remote instance origin (managed SSH servers serve the same UI). But
+// a configured instance can also serve a FOREIGN page (e.g. an `opencode web`
+// server on a VPS that the user added as a remote instance). Such a page has
+// neither a `-webkit-app-region` drag region (the frameless window becomes
+// impossible to move on Windows/Linux) nor any way to get back to the local
+// OpenChamber instance.
+//
+// This module builds the tiny overlay the main process injects into foreign
+// pages loaded inside OpenChamber windows: a slim draggable bar with a
+// "Back to OpenChamber" button that navigates back to the local UI. The
+// injection is additive and self-contained (inline styles, one button), and
+// never runs on OpenChamber UI documents or on local/packaged pages.
+
+export const OPENCHAMBER_UI_MARKER_ATTR = 'data-oc-ui';
+export const REMOTE_SHELL_BAR_ID = 'oc-remote-shell';
+export const REMOTE_SHELL_BACK_BUTTON_ID = 'oc-remote-shell-back';
+
+const REMOTE_SHELL_BAR_HEIGHT_PX = 28;
+
+/**
+ * Decide whether the remote-page shell should be injected for a page.
+ *
+ * Pure decision helper; the injected script embeds a serialized copy of this
+ * function so main-process logic and in-page behavior cannot drift apart.
+ *
+ * @param {object} page
+ * @param {string} page.protocol  `window.location.protocol` of the page.
+ * @param {string} page.origin    `window.location.origin` of the page.
+ * @param {string} page.openChamberUiMarker  value of the OpenChamber UI marker attribute.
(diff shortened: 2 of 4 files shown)


btriapitsyn:
Thanks for the careful Electron work here. Closing: pointing a remote instance at a page that is not OpenChamber (such as opencode web) is not a flow we support, so the app should not grow a shim for it. A remote instance is expected to serve OpenChamber.

PR #2683, opened 2026-08-05
fix(chat): keep the composer's platform period-on-double-space intact (Fixes #2504)

### LINKED ISSUE #2504: [Bug] regression: double tapping space quickly does not insert a period

often times when typing, i take advantage of the shortcut that when i double-press the space bar, it creates a period, ending my sentence. however, since the latest update, that has no longer been the case, which is really frustrating.

### Where does it happen?

Desktop Web

### Version (if known)

1.17.0

_No response_

### Logs (optional)

```shell

```

sdziupin:
## Intent

Fixes #2504 (GitHub issue; no Linear counterpart exists for it).

Double-tapping space in the composer used to produce ". " on macOS/Android
via the platform's insert-period-on-double-space. The 1.17.0 CodeMirror
migration (2cfd69d1c, PR #2419) regressed it: `@codemirror/view` 6.39.13
actively rewrites the platform's ". " insertion into a plain double space
when the content element's `autocorrect` attribute is exactly `"off"` —
and the composer sets exactly `"off"` on desktop.

This change keeps desktop word correction off everywhere (the existing
policy, unchanged for the browser) while spelling the attribute `Off` on
the platforms where CodeMirror's revert path is live. The HTML `autocorrect`
attribute matches its keywords ASCII case-insensitively, so browsers still
treat `Off` as off — but CodeMirror's exact `== "off"` comparison fails,
and the period survives.

Behavior change: on macOS desktop web (and iOS/iPadOS/Android via the same
revert path), double-tapping space in the composer again inserts ". " as the
platform intends. Windows/Linux are deliberately unchanged (CodeMirror has no
revert path there).

Note for maintainers: PR #2510 by another contributor targets the same
issue with the same mechanism (verified independently here against the
installed `@codemirror/view@6.39.13` bundle, which contains the exact-match
guard). This PR is an independent implementation with a less brittle
pinned-guard test (whitespace-tolerant regexes against the installed bundle
resolved through the package's own node_modules, instead of pinning
minified platform-flag expressions) and keeps the `autoCorrect` prop
contract change confined to the composer's own modules.

## Non-goals

- No custom double-space-to-period engine: the feature is and stays the
  platform's, this only stops CodeMirror from undoing it.
- No spellcheck policy change: `spellcheck` remains tied to the existing
  "Enable spellcheck in text inputs" setting.
- No change to non-composer inputs (CommitInput, textareas, etc.).
- No change on Windows/Linux desktop (nothing to fix there — CodeMirror
  never reverts on those platforms).

## Affected surfaces

Shared `ComposerEditor` used by web, Electron, VS Code, hosted mobile and
Capacitor mobile surfaces (packages/ui only). The `autoCorrect` prop changes
from `boolean` to the `ComposerAutoCorrect` keyword union; only `ChatInput`
passes it. No persisted data, API route, or external contract changes.

Mobile keeps the existing `on` policy; iOS/iPadOS desktop-web users get
`Off` (revert path is `browser.mac`, which includes iOS in CodeMirror's
flags), which also restores their period-on-double-space.

## Repository guidance

|---|---|---|
| `openchamber-change-discipline` | Composer editor input behavior + component prop type change | Platform policy isolated in one pure helper (`editor/autocorrect.ts`) with focused tests; prop type change confined to the composer's own modules; no drive-by refactors |
| Composer `DOCUMENTATION.md` (nearest module doc, read before editing) | Owns editor input behavior and invariants | Records the HTML-case-insensitive-vs-CodeMirror-exact-match invariant and why `Off` must not be "normalized" back to `off` |
| Root `AGENTS.md` | Always-on repo rules | No new dependencies, no secrets, minimal diff, owning docs updated |
| Skill `performance-engineering`, `theme-system` | Checked | Not applicable: no render/event hot path or styling change (attribute string only) |

## Validation

| Check | Result |
|---|---|
| `bun test packages/ui/src/components/chat/composer/editor/__tests__/autocorrect.test.ts` | 9/9 pass (exit 0). Covers macOS, iPhone, iPadOS touch detection, Android, Windows, Linux, spoofed macOS UA on Linux, mobile `on` policy, and the pinned CodeMirror guard |
| `bun test packages/ui/src/components/chat/composer` (full suite) | 286 tests across 16 files, exit 0 |
| `bun run dead-code` (new file added) | No findings for the new file/exports (report non-blocking; pre-existing 258 unused-export baseline unchanged) |

Not verified: real typing on macOS, iOS, or Android hardware/simulator —
this environment is Linux; headless Chromium cannot produce the OS-level
period substitution, so the fix is validated at the code level (the exact
guard in the installed `@codemirror/view@6.39.13` dist is asserted by the
pinned-guard test) and by the HTML spec's case-insensitive keyword
semantics.

## Visual evidence

No screenshots possible: the change affects native text substitution while
typing, not layout, color, or styling. The right evidence is a short screen
recording of double-tapping space in the composer on macOS (or iOS/Android)
showing the period surviving; that recording could not be made in this
environment. Expected rendering: identical UI; typing "hello" + double
space yields "hello. " instead of "hello  " on affected platforms.

## Risks and failure behavior

- The pinned-guard test reads the installed `@codemirror/view` dist bundle.
  A future dependency upgrade that keeps the exact-match guard passes; one
  that changes the guard (case-insensitive match, different platform flags,
  removed revert path) fails the test first — that failure is a signal to
  re-read the upstream guard, not a regression. The regexes are
  whitespace-tolerant so minifier reformatting does not false-fail.
- The workaround depends on browsers treating `autocorrect` keywords as
  ASCII case-insensitive (HTML spec). If a browser became case-sensitive,
  `Off` would fall back to the attribute default (`on`) on macOS/Android
  only — a contained, worst-case behavior change that the pinned test and
  this doc surface.
- Rollback: pure attribute string change; reverting the commit restores the
  previous behavior exactly. No state, cache, or persistence involved.

### DIFF
--- a/packages/ui/src/components/chat/ChatInput.tsx
+++ b/packages/ui/src/components/chat/ChatInput.tsx
@@ -97,6 +97,7 @@ import {
     type ComposerEditorHandle,
 } from './composer/editor/ComposerEditor';
 import { createComposerEditorViewStore } from './composer/editor/viewStore';
+import { composerAutoCorrect } from './composer/editor/autocorrect';
 import {
     appendInlineText,
     appendWithLineBreaks,
@@ -2646,7 +2647,7 @@ const ChatInputComponent: React.FC<ChatInputProps> = ({ onOpenSettings, scrollTo
                                         : t(useCompactChatPlaceholder ? 'chat.chatInput.placeholder.chatCompact' : 'chat.chatInput.placeholder.chat')
                                     : t('chat.chatInput.placeholder.selectSession')}
                                 editable={Boolean(currentSessionId || newSessionDraftOpen)}
-                                autoCorrect={isMobile}
+                                autoCorrect={composerAutoCorrect({ isMobile })}
                                 autoCapitalize={isMobile ? 'sentences' : 'none'}
                                 spellCheck={isMobile || inputSpellcheckEnabled}
                                 fillContainer={isComposerExpanded}
--- a/packages/ui/src/components/chat/composer/DOCUMENTATION.md
+++ b/packages/ui/src/components/chat/composer/DOCUMENTATION.md
@@ -81,6 +81,19 @@ composer holds a prompt, not a source file: it is short enough that a full pass
 is cheaper and far simpler than incremental mapping, and it keeps the editor
 and the send path reading the same grammar.
 
+The content element's `autocorrect` keyword is a policy decided by the caller
+through `editor/autocorrect.ts`. The HTML `autocorrect` attribute matches its
+keywords ASCII case-insensitively, so the composer keeps desktop word
+correction off everywhere — but @codemirror/view reads the attribute with an
+exact `== "off"` and rewrites macOS and Android's insert-period-on-double-space
+into a plain double space when it sees exactly `off`. `editor/autocorrect.ts`
+therefore emits the case-insensitively-equivalent `Off` on exactly the
+platforms CodeMirror reverts (`browser.mac || browser.android`, mirrored
+deliberately), and the literal `off` everywhere else. Do not "normalize" the
+`Off` spelling back to `off`, and if a @codemirror/view upgrade changes the
+guard, the pinned-guard test in `editor/__tests__/autocorrect.test.ts` fails
+first.
+
 ## Ordering rules worth knowing
 
 - `editor/ComposerEditor.tsx` forwards a click on the composer's padding by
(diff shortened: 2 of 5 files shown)


btriapitsyn:
Thanks for the thorough write-up — the analysis of the CodeMirror 6.39.13 exact-match guard is solid. Closing as a duplicate: #2510 fixed the same regression (#2504) with the same Off-vs-off mechanism and merged to main on 2026-08-27, touching the same files.

PR #2687, opened 2026-08-05
fix(server): forward Small Model override to managed OpenCode config

### LINKED ISSUE #2497: Bug: Auto-generated chat session titles not working
## Bug Report: Auto-generated chat session titles not working

**OpenChamber Version:** 1.16.3
**OpenCode Version:** 1.18.4
**OS:** macOS

### Steps to Reproduce
1. Open OpenChamber
2. Go to Settings → Chat
3. Verify Small Model is set to a specific model (tested with "Anthropic Haiku")
4. Create a new chat session
5. Send a message to the agent and wait for response
6. Check if session title auto-generates

### Expected Behavior
Session title should auto-generate based on the first chat message content (e.g., from "New Chat" to something like "Help with code refactoring")

### Actual Behavior
Session title remains as default "New Chat" or similar. No auto-generated title appears after the first exchange, even when Small Model is properly configured.

### What I've Tried
- Changed Small Model from "Default" to "Anthropic Haiku" specifically
- Verified API key is valid in Settings → Providers
- Created multiple new sessions to test
- Confirmed Small Model provider (Anthropic) is working

### Additional Context
- Small Model setting shows "Anthropic Haiku" (specific model, not "Default")
- No visible "Auto-generate titles" toggle exists in Settings → Chat
- There is no error message; titles simply don't generate
- According to CHANGELOG, this feature uses the Small Model for title generation, but it appears non-functional in 1.16.3

### Question
Is auto-title generation still an active feature in 1.16.3, or was it removed/disabled?

sdziupin:
## Intent

Fixes #2497 (GitHub issue; no Linear issue exists for it).

Session title auto-generation depends on the OpenCode server's internal `title` agent, which reads `small_model` from OpenCode's config. OpenChamber's Settings → Chat → Small Model override (`smallModelOverride`) only fed OpenChamber's own `/api/small-model/generate` utility service (notes summarization, goal audit) — it was never propagated into the managed OpenCode server's config. So a user who explicitly configured a Small Model (the issue reporter set "Anthropic Haiku") got no effect on title generation, and sessions stayed untitled ("New Chat"/"Untitled Session") whenever OpenCode's own fallback chain could not resolve a small model.

Behavior change: when the user has an explicit Small Model override, the managed OpenCode process is now launched with `small_model` set in its `OPENCODE_CONFIG_CONTENT`, so OpenCode's title/summary generation uses the user's chosen model. "Use default small model" leaves the config untouched (OpenCode's own resolution stays authoritative), matching the precedence documented in `packages/web/server/lib/small-model/DOCUMENTATION.md`.

Note on "removed/disabled?": OpenChamber main never had its own title-generation code — a `session-title` runtime exists only on unmerged Claude-harness branches (`ai-workflow`/`claude`). Titles are generated by the OpenCode server; the local gap this PR fixes is that OpenChamber's only lever (the Small Model setting) never reached it.

## Non-goals

- No client-side or OpenChamber-side title generation (that would duplicate the OpenCode server's feature).
- No change to OpenChamber's own small-model utility service, resolution chain, or `/api/small-model` routes.
- No change for external OpenCode servers (they are not launched with OpenChamber's env, so the injection intentionally does not apply; OpenCode reads the user's own config there).
- No settings-UI change: the existing Small Model setting semantics are preserved; only its effect is extended.

## Affected surfaces

- `packages/web` (server JS, not covered by tsc/eslint which scope to `src/**`):
  - `server/lib/small-model/config-injection.js` (new) — pure helper that merges the override into the managed `OPENCODE_CONFIG_CONTENT` JSON.
  - `server/index.js` — `getManagedOpenCodeEnv()` now applies the helper after the agent-tool/system-prompt config merges; the result feeds the managed OpenCode process env at spawn (and on restart via `refreshOpenCodeAfterConfigChange` → `restartOpenCode`).
  - `server/lib/small-model/DOCUMENTATION.md` — documents the new contract.
- Runtimes: applies wherever OpenChamber launches a managed OpenCode process (desktop, web, CLI); platform-neutral (the report's macOS flag is not addressed by any platform-specific branch — the missing wiring existed everywhere).
- Persisted/external contracts: none; the settings schema (`smallModelUseDefault`, `smallModelOverride`) is unchanged and already sanitized in `settings-helpers.js`.

## Repository guidance

|---|---|---|
| `AGENTS.md` minimal diffs + "update owning documentation when contracts change" | Any source change | 1 new pure module + thin 8-line wiring + doc note; no drive-by edits |
| `sync-state-invariants` skill | Task requires it; session title flow touches sync-owned session records | No store/reducer/event changes; no failure-is-empty hazards; the helper is a pure config transform that leaves malformed user config untouched rather than rewriting it; unchanged path returns the same reference |
| `packages/ui/src/sync/DOCUMENTATION.md` | Read-first per skill | Confirms title display is a read of `session.title` from live stores (Header/Sidebar fall back to "Untitled"); generation is server-side, out of UI scope |
| `packages/web/server/lib/small-model/DOCUMENTATION.md` | Nearest owning module doc | States OpenCode uses small model internally for titles and OpenChamber replicates the mechanism; the fix now also forwards the override to the managed server |
| AGENTS.md "server JS is not covered by type-check/lint" | Validation rules | Ran `node --check` syntax checks + focused vitest/bun tests for the touched server JS |

## Validation

| Check | Result |
|---|---|
| `bun test server/lib/small-model/config-injection.test.js` (packages/web) | 6 pass / 0 fail — override injection, use-default no-op, empty override no-op, key preservation, replacement of existing `small_model`, malformed-config passthrough |
| `bun test server/lib/opencode/lifecycle.test.js` (packages/web) | 14 pass / 0 fail — `getManagedOpenCodeEnv` contract via mock is unaffected |
| `node --check server/index.js`, `node --check server/lib/small-model/config-injection.js` | PASS — both parse (server JS is not covered by `tsc`/eslint, which scope to `src/**`) |
| `bun run type-check` (packages/web, `tsc --noEmit`) | PASS — exit 0 |
| `bun test server/lib/small-model/` (whole dir) | 63 pass / 4 fail — the 4 failures are pre-existing cross-file interference in this environment: they reproduce with my changes stashed (verified via `git stash`), and all 17 `resolve.test.js` tests pass when that file runs alone with or without my changes |
| `bun run dead-code` (repo root) | Ran (files added); report contains no entries for `config-injection` — the new export is referenced from `server/index.js`; remaining report items are pre-existing (unrelated UI/vscode types) |
| Live end-to-end (open OpenChamber + managed OpenCode, send first message, observe title) | NOT VERIFIED here — no opencode binary/server or display available in this environment; see "Risks" |

## Visual evidence

No screenshots possible (no browser/display/server in this environment). Expected user-visible behavior after the fix, described concretely: with Settings → Chat → Small Model set to a specific model, a new session's title changes from the untitled fallback to a generated title (e.g., "Help with code refactoring") after the first exchange, because the managed OpenCode server now receives `small_model` and its internal title agent uses it. With "Use default small model", behavior is unchanged from before this PR.

## Risks and failure behavior

- The unverified backend part: whether the OpenCode server (e.g. 1.18.4, per the report) then successfully generates titles depends on OpenCode's own title agent and provider auth — not observable from this repo (OpenCode source/binary not available here). This PR fixes the local wiring (the only OpenChamber-side lever) and does not claim to fix OpenCode-internal failures.
- Injection only affects the managed process; an external OpenCode server keeps reading the user's own config — by design.
- Malformed existing `OPENCODE_CONFIG_CONTENT` is left untouched (no crash, no silent rewrite); a missing config becomes `{}` + `small_model`, which is the same behavior as the existing agent-tool merge for absent config.
- Settings changes take effect on the next managed-process launch/restart (env is captured at spawn); the existing `refreshOpenCodeAfterConfigChange` restart path covers config reloads. No new failure/rollback/cleanup paths — the helper is pure and synchronous.
- No data-loss, security, or performance concerns: the function runs once per process launch; no credentials are touched; no secrets logged.


btriapitsyn:
Thanks for the fix — merging. The fix itself is unchanged; I rebased the wiring in `getManagedOpenCodeEnv` on top of the `includeWeb`/`includeMemory` flags main added to `prepareManagedOpenCodeEnv` since this PR was opened, so it applies the Small Model override to the current env shape. Reran the config-injection and lifecycle tests, plus a syntax check and lint on the touched files — all green. Merging now.

btriapitsyn:
Sorry — this one has been reverted (3b2beeaf7) after living on `main` for a bit.

In real use it turned out to work poorly with OpenCode. The model OpenChamber picks for its own small-model work (session assist, walkthroughs, goals) and the `small_model` OpenCode uses internally for titles and summaries are two different things with different needs, and tying them together through the injected config produced worse results than leaving OpenCode's own resolution alone. They need to stay configured separately, so the forwarding is gone rather than adjusted.

Thanks for the clean implementation regardless — the problem was the coupling, not the code.

PR #2689, opened 2026-08-05
fix(chat): do not hijack ctrl/cmd+digit while typing in an input (Fixes #2503)

### LINKED ISSUE #2503: [Bug] ctrl+2 opens "changes" pane

ctrl+2 opens the changes pane with no way to rebind it.
i always have openchamer pinned as "tab 1" in my browser and often hop between it and "tab 2" which is usually github. now when I press cmd/ctrl+2 it hijacks it and opens the changes pane and I can't turn it off.

### Where does it happen?

Desktop Web

### Version (if known)

1.17.0

_No response_

### Logs (optional)

```shell

```

sdziupin:
## Intent

Fixes #2503 (GitHub issue; no Linear counterpart exists for it).

The numbered context-surface switcher (hold `mod` + digit, default
`Ctrl/Cmd+1..9`, feature added in 8274dd82f) hijacked `ctrl/cmd+digit` even
while the user was typing in an input — the composer, CommitInput, searches,
dialogs. Since `mod+digit` is the browser's own tab-switching chord (the
reporter pins OpenChamber as browser tab 1 and hops to tab 2 with Ctrl+2),
the app stole the keystroke and opened the "changes" pane.

The binding is intentional and rebindable/unassignable in Settings →
Keyboard shortcuts (`switch_context_surface`, prefix `mod`). The missing
piece was the standard "don't fire global shortcuts while typing" guard.

The fix: the digit-switcher handler now returns early when the event target
is inside an editable surface (`input`, `textarea`, `contenteditable` — the
CodeMirror composer is a contenteditable). While the user is typing, the
chord keeps its normal meaning (browser tab switch, in-input use); from any
non-editable focus the surface switcher still works exactly as before.

Behavior change: `ctrl/cmd+digit` no longer switches the context panel
surface while focus is in any editable input. With focus elsewhere it is
unchanged. Users who want the switcher entirely off can unassign
`switch_context_surface` in Settings (existing capability, unchanged).

## Non-goals

- No removal of the surface-switcher feature: it is intentional, documented
  in Settings, and still works from non-editable focus.
- No change to any other shortcut: other global shortcuts (`mod+n`, `mod+b`,
  …) deliberately keep their existing typing-time behavior; only the digit
  chord (the browser-conflicting one reported) is guarded.
- No change to the terminal capture handler: terminal targets were already
  skipped by the digit switcher.

## Affected surfaces

Shared `useKeyboardShortcuts` hook (packages/ui) used by web, Electron, VS
Code webview, hosted mobile and Capacitor mobile. The new
`isTypingInEditableTarget` helper lives in the existing
`hooks/keyboard-shortcut-dom.ts` module (alongside the same pattern's
`hasOpenDropdown`) with tests in the existing test file. No persisted data,
API route, or external contract changes.

## Repository guidance

|---|---|---|
| `openchamber-change-discipline` | Global key-handling behavior change + new exported helper | Guard scoped to the digit-switcher branch only; helper placed in the existing keyboard-shortcut-dom module with focused tests following the repo's fake-`Element` test pattern; no new deps |
| `openchamber-change-discipline` (test design) | New helper used once | Testability is the reason it lives in `keyboard-shortcut-dom.ts` (same precedent as `hasOpenDropdown`), not inline: the hook itself has no DOM test infra |
| Root `AGENTS.md` | Always-on repo rules | Minimal diff, no secrets, existing behavior preserved outside the guarded scope |
| `locale-ui-patterns`, `theme-system`, `sync-state-invariants` | Checked | Not applicable: no strings, no styling, no state/sync change |

## Validation

| Check | Result |
|---|---|
| `bun test packages/ui/src/hooks/keyboard-shortcut-dom.test.ts` | 7/7 pass (3 existing + 4 new guard tests: null target, no-`closest` target, inside editable, outside editable) |
| `bun test packages/ui/src/lib/shortcuts.test.ts` | 13/13 pass |
| `bun run dead-code` (new export added) | No findings for `isTypingInEditableTarget` (pre-existing 258 unused-export baseline unchanged) |

Not verified: live browser tab switching with a real focused composer — this
environment has no running OpenChamber server UI to drive. The guard is
verified at the DOM-selector level (the composer's CodeMirror content
element carries `contenteditable="true"`, asserted by the composer editor
tests) and by the unit tests above.

## Visual evidence

No screenshots possible: the change is keystroke routing, not rendering.
Expected behavior: with the chat composer (or any input) focused, pressing
Ctrl+2 no longer opens the changes pane and instead keeps its browser
meaning; pressing Ctrl+2 with focus on the chat body or panel still opens
the matching surface. No visual change.

## Risks and failure behavior

- The guard matches `input, textarea, [contenteditable="true"]`. If a new
  editable surface used a different markup, mod+digit would hijack it again;
  the selector is centralized in one constant for that case.
- Users who genuinely want mod+digit to switch surfaces while typing lose
  that (they can click away or use the rail); this matches standard app
  behavior and is the point of the fix.
- Rollback: reverting the commit restores previous behavior exactly. No
  state, cache, or persistence involved.


btriapitsyn:
Thanks for the fix — merging. The fix is good — Ctrl/Cmd+digit surface switching was firing while typing in an input, which is a real bug. I resolved it against current main: main had since gained its own `isEditableEventTarget` helper for the same purpose, so I dropped the PR's duplicate helper and applied the one-line guard using that existing helper instead. Added a focused test for it. Thanks for the fix!

PR #2691, opened 2026-08-05
fix(chat): defer composer value writeback during IME composition (Fixes #2527)

### LINKED ISSUE #2527: [Bug] Chat input cursor jumps during IME composition — controlled value writeback lacks composition guard
## Problem

When typing with an IME (e.g. Chinese Pinyin, Japanese Kana, Korean), the cursor in the chat input box frequently jumps to the end of the text. The composition session is interrupted mid-way, causing characters to appear at wrong positions and a broken input experience.

## Reproduction

1. Open any session chat
2. Switch to a CJK input method (e.g. Sogou Pinyin, Microsoft Pinyin, macOS built-in Pinyin)
3. Type pinyin into the chat input (e.g. `ni hao`)
4. Observe: cursor jumps to the end during composition; already-typed text shifts position
5. Pressing Enter to confirm a candidate may also behave unexpectedly (depending on browser/IME `isComposing` timing)

**Environment:**
- openchamber version: 1.17.0
- Platform: Web (reproducible on all desktop browsers)

## Root Cause

The chat input is a CodeMirror 6-based **controlled component**: the external `value` is written back to the editor via a `useEffect([value])`.

### Problematic code (de-minified)

**Value writeback effect:**
```js
useEffect(() => {
  const view = editorRef.current;
  if (!view) return;
  const current = view.state.doc.toString();
  if (current !== value) {
    view.dispatch({
      changes: { from: 0, to: current.length, insert: value },
      selection: { anchor: value.length }  // forces cursor to end
    });
  }
}, [value]);
```

**Change report listener:**
```js
EditorView.updateListener.of((update) => {
  if (update.docChanged) {
    onChange({ value: update.state.doc.toString(), ... });
  }
});
```

### Trigger chain

1. User types pinyin → browser IME begins composition on the CodeMirror contentDOM
2. CodeMirror `updateListener` fires `onChange`, writing the intermediate composition text to parent state
3. State change → `value` prop changes → `useEffect([value])` runs
4. At this point `view.state.doc.toString()` may not include the text the browser is actively composing (IME composition text is managed at the DOM level; during CodeMirror internal `composing` state, `doc.toString()` can be inconsistent with the actual DOM content)
5. `current !== value` evaluates true → dispatch `{ changes: { from: 0, to: ..., insert: value }, selection: { anchor: value.length } }`
6. This dispatch **replaces the entire document and forces the cursor to the end**, directly interrupting the browser IME composition session
7. Cursor jump / offset

### What is missing

The editor component has **no** `compositionstart` / `compositionend` event listeners and no `isComposing` flag to prevent the value writeback effect from executing during composition.

For comparison: the keydown handler in the same codebase correctly uses an `isComposing` guard:
```js
const isComposingEvent = (e) => {
  const native = "nativeEvent" in e ? e.nativeEvent : e;
  return native.isComposing || native.keyCode === 229;
};

const onKeyDown = (e) => {
  if (isComposingEvent(e)) return;  // ✅ correctly skipped
  // ... Enter to send, markdown wrapping, history navigation, etc.
};
```

But the value writeback effect is missing the same guard.

## Suggested Fix

Add composition state tracking to the editor component and skip the value writeback during composition:

```js
const composingRef = useRef(false);

// Listen for composition events on the CodeMirror contentDOM
useEffect(() => {
  const view = editorRef.current;
  if (!view) return;
  const dom = view.contentDOM;
  const onStart = () => { composingRef.current = true; };
  const onEnd = () => {
    composingRef.current = false;
    // After compositionend, the browser fires an input event;
    // CodeMirror processes it normally and reports via onChange — no manual sync needed
  };
  dom.addEventListener("compositionstart", onStart);
  dom.addEventListener("compositionend", onEnd);
  return () => {
    dom.removeEventListener("compositionstart", onStart);
    dom.removeEventListener("compositionend", onEnd);
  };
}, []);

// Add guard to value writeback effect
useEffect(() => {
  const view = editorRef.current;
  if (!view) return;
  if (composingRef.current) return;  // ✅ skip sync during composition
  const current = view.state.doc.toString();
  if (current !== value) {
    view.dispatch({
      changes: { from: 0, to: current.length, insert: value }
      // Do not force cursor position unless value was set externally
    });
  }
}, [value]);
```

Additionally worth considering: even outside composition, the `selection: { anchor: value.length }` in the writeback dispatch forces the cursor to the end on every external value change. For programmatic value sets (e.g. restoring from history, inserting a snippet), this may not be the desired behavior. Consider only setting cursor position when the value genuinely originates from an external source (not user input).

## Additional notes

- Affects all IME users (Japanese, Korean, etc.), not just Chinese
- Mobile (PWA) is also affected
- #2507 (blinking cursor disappears) and #2514 (text not selectable) may be related to the same editor component

makeittech:
## Intent

Fixes #2527 (GitHub issue; no Linear counterpart exists for it).

The chat input's cursor jumps during IME composition (pinyin, kana, hangul):
the controlled-value writeback effect compared the `value` prop against
`view.state.doc.toString()` and, on mismatch, dispatched a wholesale
document replacement with `selection: { anchor: value.length }` — while the
browser was still composing. During composition the uncommitted text lives
in the DOM, not in the CodeMirror document, so the mismatch is expected; the
dispatch replaced the document out from under the IME session, forced the
caret to the end, and interrupted the composition.

The fix: the writeback effect now defers while the view is composing. It
uses CodeMirror's own public `compositionStarted` getter (installed
`@codemirror/view@6.39.13`, type declarations pinned by the regression
test) — no compositionstart/compositionend event plumbing needed, and the
state is authoritative across the viewStore's kept-alive view reuse. The
composition commits through CodeMirror's own pipeline and reports via
`onChange`, so the guard only defers the rewrite; no manual sync is needed
after the composition ends.

Behavior change: none outside IME sessions — with no active composition the
writeback path is byte-for-byte the previous one (equality fast-path,
end-anchored selection, scroll pin). During a composition, an externally
changed `value` prop is ignored until the composition commits; the
composition's own text then wins, which is the correct outcome (the user's
in-progress text takes priority over a concurrent external write).

## Non-goals

- No change to the `selection: { anchor: value.length }` behavior for
  genuine external rewrites — it is load-bearing for draft restore, history
  navigation, "add to chat" and dictation insert (documented in-code).
- No change to the keydown `isIMECompositionEvent` guard (already correct,
  per the issue).
- No change to the imperative `insertText`/`replaceRange` paths: they are
  user-action-driven and do not run mid-composition (keydown is guarded).

## Affected surfaces

Shared `ComposerEditor` (packages/ui) used by web, Electron, VS Code
webview, hosted mobile and Capacitor mobile. The guard reads a public
CodeMirror getter; no props, store, API, or persisted contracts changed.
PWA/mobile IME users are affected by the same editor and get the same fix.

## Repository guidance

|---|---|---|
| `openchamber-change-discipline` | Editor input behavior change | 2-line guard with in-code rationale; no refactor, no new deps; regression test added following the repo's source-pinning precedent (no DOM test infra exists in packages/ui) |
| Composer `DOCUMENTATION.md` (nearest module doc) | Owns editor input behavior | Read before editing; the editor section documents the controlled-primitive contract this guard protects — a doc note was not required since the contract is unchanged, only deferred during composition |
| Root `AGENTS.md` | Always-on repo rules | Minimal diff, no secrets, behavior preserved outside the guarded scope |
| `sync-state-invariants`, `performance-engineering` | Checked | Not applicable: no sync/reducer change; the guard adds one getter read per value-prop change |

## Validation

| Check | Result |
|---|---|
| `bun test packages/ui/src/components/chat/composer/editor/__tests__/writebackCompositionGuard.test.ts` | 3/3 pass: guard present and ordered before the dispatch; the `compositionStarted` getter exists in the installed @codemirror/view type declarations; equality fast-path and end-anchored selection untouched |
| `bun test packages/ui/src/components/chat/composer` (full suite) | 280 tests across 16 files, exit 0 |
| `bun run dead-code` (new test file) | No findings for the new file |

Not verified: real IME typing (Sogou/MS Pinyin, Japanese Kana, Korean) in a
browser — this environment has no running OpenChamber server UI and no IME.
The guard's semantics (defer while `compositionStarted`, self-heal via the
commit's onChange) are verified by code inspection of the installed
CodeMirror composition state machine (`inputState.composing >= 0` backing
the getter) and pinned by the ordering test; the end-to-end behavior needs
a manual IME pass on a real browser.

## Visual evidence

No screenshots possible: the change is caret/IME behavior while typing, not
rendering. Expected behavior after the fix: composing pinyin/kana/hangul in
the chat input no longer jumps the caret to the end or splits the
composition; the candidate text stays put until committed. No visual change
otherwise.

## Risks and failure behavior

- On Android, `compositionStarted` is true frequently (caret placement on a
  word starts a composition there), so the writeback is deferred more often
  on that platform. Genuinely-divergent external writes during an active
  composition are rare and self-heal: the composition commit fires
  `onChange`, or the next edit re-syncs. The user's in-progress text wins,
  which is the intended behavior.
- The guard relies on the public `compositionStarted` getter of the
  installed @codemirror/view. A dependency upgrade that renames/removes it
  fails the pinned type-declaration test first.
- Rollback: reverting the commit restores previous behavior exactly. No
  state, cache, or persistence involved.


ChangeHow:
What going on this PR? met the same issue and need fix.

ChangeHow:
Seriously, really need this PR @btriapitsyn Please take a look.

btriapitsyn:
Updated the PR and pushed commit `08207172f`.

Fixed:

- Merged the current `main` into the PR branch; there were no merge conflicts.
- Kept the IME caret-jump fix, but narrowed the guard to defer only the exact controlled-value echo previously reported by CodeMirror.
- Preserved genuine external composer updates during composition, including draft restores and host-driven additions such as VS Code file mentions, instead of silently dropping them.
- Stored the last reported editor value with the kept-alive view state so mobile view reuse preserves the same synchronization contract.
- Replaced the source-text/declaration-pinning regression test with behavior-level tests for controlled echoes, external updates during composition, and post-composition writes.
- Documented the value-authority ordering in the composer module documentation.

Validation:

- `bun test packages/ui/src/components/chat/composer` - 281 passed
- `bun run type-check:ui` - passed
- `bun run lint:ui` - passed
- scoped `bunx oxlint` for the new/rewritten helper, store, and test - passed
- `bun run dead-code` - no findings for the new helper/test; existing repository backlog remains

Not verified: manual typing with a real desktop or Android IME. The package has no DOM/IME test environment, so that interaction still needs a device/browser pass.

btriapitsyn (approved):

btriapitsyn:
Addressed the automated review findings in commit `84fc02881`.

- Removed the unreachable `reportedValue`/helper classification.
- Restored the composition-wide guard after the equality fast-path and before the writeback dispatch, so stale controlled-value echoes cannot replace CodeMirror's newer document during IME composition.
- Removed the extra kept-view state and deleted the now-unused helper module.
- Replaced the helper-only tests with a focused wiring regression test that verifies the required `equality check -> composition guard -> dispatch` order.
- Removed the documentation claim about preserving concurrent external writes during composition; that behavior is outside this scoped fix.

Validation:

- `bun test packages/ui/src/components/chat/composer` - 279 passed
- `bun run type-check:ui` - passed
- `bun run lint:ui` - passed
- `bun run dead-code` - no new findings from this change; existing repository backlog remains
- `bunx oxlint` reported only pre-existing findings in unchanged `ComposerEditor.tsx` lines

Not verified: manual typing with a real desktop or Android IME.

PR #2708, opened 2026-08-05
fix: wait for worktree bootstrap before prompt dispatch; resolve session directory for send/fork

ftzi:
## Intent

Two related prompt-dispatch races for sessions created or targeted through the agent tool's worktree flow:

1. `createWorktree` returns while the worktree directory is still being populated in the background (git reset after a `--no-checkout` add). A prompt dispatched into a half-populated directory makes opencode's run die with `UnknownError` because agent/config files are not there yet. The session routes now wait for the worktree bootstrap to reach `git-ready`/`setup-ready` (or `ready`) before creating the session and dispatching the prompt, failing the request cleanly when bootstrap failed or exceeded a 60s deadline.
2. `session.send`/`session.fork` default the target directory to the caller's context directory, which is wrong for sessions living in other worktrees: `prompt_async` then targets an instance that does not hold the session and the run dies with `UnknownError`. The control service now resolves the target session's directory from the global session list when the caller did not scope explicitly, with a fallback to the previous context-directory behavior when the session is not found.

## Non-goals

- No change to how worktrees are created or bootstrapped (`createWorktree`/`getWorktreeBootstrapStatus` untouched).
- No change to session behavior when the caller supplies an explicit `directory` or `projectId`.
- No UI, CLI, or SDK surface changes; server API shape unchanged.

## Affected surfaces

- `packages/web` server only: `openchamber-control/service.js` (session send/fork dispatch) and `openchamber-sessions/routes.js` (worktree session create + prompt dispatch), plus their vitest suites.
- Runtime: the OpenChamber server API, which desktop, VS Code, mobile, and web surfaces all reach through `RuntimeAPIs`. Behavior is intentionally identical across those surfaces since it is server-side.
- Persisted/external contracts: none. No schema, config, or SDK changes.

## Repository guidance

|---|---|---|
| `openchamber-change-discipline` | Any source change to packages/web | Change is focused on the two server modules; tests added alongside; no dependencies or exports touched. |
| `sync-state-invariants` | Worktree bootstrap readiness and session-directory resolution are authoritative-state waits | Polls `getWorktreeBootstrapStatus` until ready/failed instead of guessing a fixed delay; failure/timeout are explicit 500 errors, not silent dispatches. Directory resolution falls back to the previous context-directory behavior when the authoritative session list lacks the session, and only replaces the directory when authoritative state exists. |
| `ui-api-decoupling` | Server API routes used by shared UI control surfaces | Behavior is implemented server-side in the owning server modules; no new client fetch/auth/URL handling introduced. |

## Validation

| Check | Result |
|---|---|
| `bun run test -- openchamber-control/service openchamber-sessions/routes` (packages/web, vitest) | 2 files passed, 38 tests passed, including the 4 new tests (directory resolution x2, bootstrap wait ordering, bootstrap-failure 500 with no prompt dispatch) |
| Full workspace suite / lint / type-check | Not run; change is server JS without type-check coverage in this package |
| Manual live-server reproduction | Not verified; tests cover the dispatch ordering and failure paths via mocked SDK/git calls |

## Visual evidence

No user-visible change. The diff only affects server-side dispatch ordering and error responses; rendered UI, themes, and interactions are untouched.

## Risks and failure behavior

- `resolveSessionDirectory` adds one `session.list` call per send/fork without an explicit directory/projectId. On failure it returns `null` and the previous behavior applies; it can never erase an explicit caller scope.
- Bootstrap wait has a 60s deadline and polls every 150ms; a failed bootstrap returns a 500 with the bootstrap error instead of dispatching into a broken directory. The worktree directory itself is not cleaned up on failure — same as before this change, since `createWorktree` already owned that lifecycle.
- Sessions created without a worktree are unaffected: the wait only runs on the worktree-creation path.
- The fallback in the second scenario preserves legacy behavior exactly, so no existing callers regress when the session list is unavailable or the session is unknown.

### DIFF
--- a/packages/web/server/lib/openchamber-control/service.js
+++ b/packages/web/server/lib/openchamber-control/service.js
@@ -240,10 +240,32 @@ export const createOpenChamberControlService = (dependencies) => {
     }
   };
 
+  // session.send/fork default the directory to the caller's context directory,
+  // which is wrong for sessions living in other worktrees: prompt_async then
+  // targets an instance that does not hold the session and the run dies with
+  // UnknownError. Resolve the target session's directory from the global
+  // session list when the caller did not scope explicitly.
+  const resolveSessionDirectory = async (sessionID) => {
+    try {
+      const client = await getClient();
+      const response = await client.experimental?.session?.list?.({});
+      const sessions = Array.isArray(response?.data) ? response.data : [];
+      const session = sessions.find((item) => item?.id === sessionID);
+      return asNonEmptyString(session?.directory) || null;
+    } catch {
+      return null;
+    }
+  };
+
   const executeSessionAction = async (action, input, contextDirectory, signal) => {
     if (input.timeout !== undefined && input.wait !== true) throw new OpenChamberControlError('timeout requires wait', 400);
     if (input.lastAssistant === true && input.wait !== true) throw new OpenChamberControlError('lastAssistant requires wait', 400);
-    const directory = asNonEmptyString(input.directory) || (!input.projectId ? asNonEmptyString(contextDirectory) : null);
+    const sessionID = asNonEmptyString(input.sessionId);
+    let directory = asNonEmptyString(input.directory) || (!input.projectId ? asNonEmptyString(contextDirectory) : null);
+    if (sessionID && action !== 'session.create' && !asNonEmptyString(input.directory) && !input.projectId) {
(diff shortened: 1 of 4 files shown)


makeittech (approved):
Nice one

makeittech:
Nice.

PR #2721, opened 2026-08-06
fix(git): run post-checkout hook after worktree creation bootstrap

ftzi:
tl;dr: I have a `post-checkout` git hook that runs `bun i`, so a new git worktree for example should naturally run `bun i` automatically. Openchamber currently skips that git hook. This PR fixes it.

## Intent

`git worktree add --no-checkout` (used by worktree creation) followed by `git reset --hard` (used by the bootstrap) never runs git's `post-checkout` hook — git only runs it for checkouts, clone, and `worktree add` *without* `--no-checkout`. Repos that rely on `post-checkout` for per-worktree setup (dependency install, git-lfs, etc.) silently skip that setup when a worktree is created through OpenChamber (`session.create --worktree`, sidebar creation, VS Code).

This PR runs the repository's `post-checkout` hook explicitly after the worktree is populated, with git's standard arguments (null ref as previous HEAD for a brand-new worktree, the checked-out HEAD, flag `1`) and the worktree as cwd — mirroring what `git worktree add` does without `--no-checkout`.

## Non-goals

- Not changing the `--no-checkout` + background-bootstrap design (fast create, index-lock recovery, upstream config ordering are preserved).
- Not switching to `post-index-change` or other hooks.
- Not running the hook on Windows via `sh` fallback (git-for-windows runs hooks through its bundled shell; here a hook that fails to spawn is logged and skipped — the status quo on Windows is unchanged).
- No changes to session-create orchestration, bootstrap timeouts, or start-script ordering.

## Affected surfaces

- `packages/web/server/lib/git/service.js` — web/desktop/mobile runtime (`session.create`, `/api/git/worktrees`): hook invoked in `queueWorktreeBootstrap` after `populateWorktreeWithLockRecovery`, before upstream config and start scripts.
- `packages/vscode/src/gitService.ts` — VS Code extension runtime: same fix mirrored in its own bootstrap.
- No persisted data, public exports, routes, or UI changed.

## Repository guidance

|---|---|---|
| `openchamber-change-discipline` | Source change to web server + VS Code runtime | Smallest complete change; behavior-preserving for repos without hooks (silent skip, matching git); failure never fails bootstrap; owning docs updated |
| AGENTS.md "Shared contracts … every applicable runtime" | Same bootstrap logic exists in web server and VS Code runtimes | Fix applied to both, with identical semantics |
| AGENTS.md validation rules | Server JS and VS Code TS are executable source | Focused tests + package type-checks run (see Validation) |

## Validation

| Check | Result |
|---|---|
| `bunx vitest run server/lib/git/service.test.js` (packages/web) | 60 passed, including 3 new tests: hook runs with args `(null-ref, HEAD, 1)` and cwd = worktree; non-executable hook skipped; failing hook (`exit 1`) does not fail bootstrap |
| `bunx vitest run server/lib/git/service.test.js -t "post-checkout"` | 3 passed / 57 skipped — new tests confirmed running |
| `bun run type-check` (packages/vscode) | Passed (extension + webview tsconfigs) |
| `bun test src/gitService.worktree-bootstrap.test.js` (packages/vscode) | 1 pass, 0 fail |
| `bun run type-check` (packages/web) | Passed |

Not verified: no end-to-end run of the VS Code extension host worktree flow, and no manual run on Windows. VS Code runtime behavior is validated by type-check plus mirrored implementation; the web server path is covered by real-git integration tests.

## Visual evidence

No user-visible change to rendered UI — the diff only adds a git hook invocation in the worktree bootstrap path. Behavior is observable only as the repo's own `post-checkout` hook now running after worktree creation.

## Risks and failure behavior

- **Hook failure** (non-zero exit or spawn error): logged as a warning, bootstrap continues, session creation unaffected. Git itself propagates hook exit status; we are deliberately safer.
- **Missing/non-executable hook**: silently skipped, matching git. No change for repos without hooks.
- **Slow hooks**: the hook runs in the existing bootstrap slot already gated by `WORKTREE_BOOTSTRAP_TIMEOUT_MS` (60s) for `session.create` — same degradation as a slow `startCommand` today, and same blocking behavior as plain `git checkout`.
- **Environment safety**: hook spawns with explicit `GIT_DIR`/`GIT_WORK_TREE` (from `rev-parse`) and cwd = the worktree, so it cannot act on the wrong repo or an inherited stray `GIT_DIR`.
- **Idempotency**: the hook runs once per worktree creation; repeated runs on later checkouts are normal git behavior for post-checkout hooks.
- **Security**: hooks are the repository's own trusted scripts — same trust level as running git itself; no new trust boundary or secret handling.

### DIFF
--- a/packages/vscode/src/gitService.ts
+++ b/packages/vscode/src/gitService.ts
@@ -33,6 +33,7 @@ const WORKTREE_PHASE_GIT_READY = 'git-ready' as const;
 const WORKTREE_PHASE_SETUP_READY = 'setup-ready' as const;
 const WORKTREE_INDEX_LOCK_RETRY_DELAY_MS = 250;
 const WORKTREE_INDEX_LOCK_STALE_DELAY_MS = 750;
+const GIT_NULL_REF = '0'.repeat(40);
 
 const toBootstrapStateKey = (directory: string): string => {
   const normalized = normalizeDirectoryPath(directory);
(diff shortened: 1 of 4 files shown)


makeittech (approved):
Looks right

makeittech:
LGTM, merging.

PR #2734, opened 2026-08-06
fix(settings): flush pending debounced settings writes on page unload

### LINKED ISSUE #2197: Windows desktop: 'Show deleted chat' setting is not persisted across app restarts

On the Windows desktop app, there is a checkbox (likely in Settings or Chat settings) to display "delete chat" or show deleted conversations. After enabling (checking) this option and closing the settings panel, the setting appears to stick for the current session. However, after restarting the application, the checkbox reverts to unchecked.

The setting is not being persisted correctly across sessions.

#### Expected behavior
- Check "Show deleted chat" (or equivalent option)
- Close and reopen the app
- The checkbox should remain checked, reflecting the saved preference

#### Actual behavior
- Check the option
- Close and reopen the app
- The checkbox is unchecked again — the setting was not saved or was overwritten on startup

#### Steps to reproduce
1. Launch the OpenChamber Windows desktop app
2. Open the settings panel
3. Find and enable (check) the "show deleted chat" / "display deleted conversations" option
4. Close the settings panel
5. Quit the application
6. Relaunch the application
7. Open the settings panel again
8. Observe that the checkbox is unchecked — the setting was not persisted

### Where does it happen?
The native Windows desktop application.

### Version
`1.16.0`

### Logs
No application error is logged. This appears to be a settings persistence / state management issue rather than a runtime exception.

sdziupin:
## Intent

Fixes #2197 ("Windows desktop: 'Show deleted chat' setting is not persisted across app restarts", intermittent).

`updateDesktopSettings()` coalesces server-side settings writes behind a 200ms debounce (`SETTINGS_DEBOUNCE_MS` in `packages/ui/src/lib/persistence.ts`) and only flushed the pending write on a runtime-endpoint switch. Toggling a setting (e.g. `showDeletionDialog` via `startAppearanceAutoSave`) and closing the window inside the debounce window silently dropped the server `settings.json` write. On next startup `syncDesktopSettings()` treats the server snapshot as authority, so the stale server value overrode the toggle; when the deferred localStorage write had also fallen back to in-memory storage, the client default won. This matches the reported intermittent behavior.

The fix registers `pagehide`, `beforeunload`, `visibilitychange`→hidden, and `freeze` listeners in `ensureSettingsRuntimeLifecycle()` that cancel the pending debounce timer and flush the pending write immediately, mirroring what `createDeferredSafeJSONStorage` (`safeStorage.ts`) already does for the localStorage path. The flush covers every setting routed through `updateDesktopSettings`, not just `showDeletionDialog`.

## Non-goals

- No change to the 200ms debounce, the coalescing/merge semantics, or the runtime-switch drain behavior of `updateDesktopSettings`.
- No change to `startModelPrefsAutoSave`'s upstream 1200ms coalescer; only the write layer it feeds is flushed.
- No `keepalive`/`sendBeacon` transport change: the settings write prefers the runtime-bridge `settings.save()` (not necessarily a fetch), and no such precedent exists on the shared `runtimeFetch`/`RuntimeAPIs` path, so the flush is a best-effort immediate invocation of the existing flush function.
- No change to `syncDesktopSettings()` startup authority semantics.

## Affected surfaces

- Package: `packages/ui` only (`src/lib/persistence.ts`, its test file, and `src/stores/DOCUMENTATION.md`).
- Persisted/external contract: the server-side `settings.json` write path (`runtimeSettings.save()` / `PUT /api/config/settings`) gains a lifecycle flush; payload shape and ordering are unchanged.
- Runtimes: shared code, so web, Electron desktop, VS Code webview, hosted mobile, and Capacitor mobile all gain the flush. Desktop (the reported platform) benefits most because closing the window fires `beforeunload`/`pagehide` while the loopback backend is still in-process. `visibilitychange`/`freeze` cover mobile/webview lifecycles where unload events are unreliable. The listener registration is wrapped in try/catch so restricted environments keep the timer-only behavior.
- No UI, styling, or copy changes.

## Repository guidance

|---|---|---|
| `AGENTS.md` correctness invariants ("Make partial results, rollback, cleanup, and stale-data behavior explicit") | The bug is exactly a stale-data race between a dropped write and startup server authority | The flush is explicit at every lifecycle boundary; the residual hard-kill data-loss window is documented in code comments, the owning `DOCUMENTATION.md`, and this PR |
| `.agents/skills/openchamber-change-discipline/SKILL.md` (persisted/external behavior risk) | Change affects a persisted settings contract | Smallest complete change at the owning layer; regression tests cover the lifecycle case (missing before, present after) plus the no-double-write and no-pending-noop cases; existing round-trip/ordering tests still pass |
| `.agents/skills/sync-state-invariants/SKILL.md` ("Pending writes must … drain before an owner switch … Apply the strongest available guarantee at page hide/freeze and shutdown boundaries") | This is a debounced write with lifecycle-durability requirements | Pending changes now drain at page hide/freeze/unload exactly as they already did at runtime switch; the flush reuses `_flushSettingsUpdate`, so runtime-context capture and stale-context rejection are unchanged |
| `.agents/skills/ui-api-decoupling/SKILL.md` | Touches the shared settings save path over `RuntimeAPIs`/`runtimeFetch` | No transport change; the existing save path (runtime API first, `runtimeFetch` PUT fallback) is invoked earlier, with runtime state still resolved at call time |
| `packages/ui/src/stores/DOCUMENTATION.md` (settings synchronization authority; deferred-adapter flush precedent) | Owning documentation for settings persistence invariants | Updated with the new invariant: debounced settings writes flush best-effort on page hide/hidden/freeze/unload, cancel the pending timer (exactly-once), and a hard kill can still lose the in-flight request |

## Validation

| Check | Result |
|---|---|
| `bun test packages/ui/src/lib/persistence.test.ts` before the fix | 20 pass / 3 fail — the three new regression tests fail: pending debounced save dropped on `pagehide`/`beforeunload`; `showDeletionDialog` toggle + immediate unload never reaches the settings backend |
| `bun test packages/ui/src/lib/persistence.test.ts` after the fix | 23 pass / 0 fail — new tests assert the save is invoked synchronously inside the lifecycle listener, the canceled timer does not double-write, lifecycle events with no pending write are no-ops, and all 19 pre-existing tests for the module (coalescing, runtime-switch drain, stale-generation rejection, partial-patch handling) still pass |
| `bun run type-check:ui` | Only 2 pre-existing errors in `src/sync/__tests__/session-switch-resync.test.ts`, verified identical on the clean base commit (`git stash` + re-run); none introduced by this change |
| `bun run lint:ui` | Clean, exit 0 |
| `bun run dead-code` | Not run — no files added/deleted/renamed and no export/type/entrypoint/import-shape change (new helper is a private module const) |

Not verified: real Electron-on-Windows app close (no Windows runtime available); the regression tests exercise the lifecycle contract at the module level with the runtime settings API mocked.

## Visual evidence

No user-visible change. This is a persistence-timing fix: the diff only adds lifecycle listeners that flush an already-pending debounced settings write earlier (at page hide/unload instead of 200ms later) and never alters what is written, any rendered component, style, or copy. The only observable difference is that a setting toggled immediately before closing the app now survives the restart — a state that cannot be captured as a static screenshot and is instead proven by the before/after regression tests above.

## Risks and failure behavior

- **Residual data loss (explicit):** a hard process kill (crash, task-manager kill, forced termination) can still lose the in-flight save request. The fix narrows the loss window from the entire 200ms+ debounce interval to the duration of the request itself; it is best-effort durability, not a guarantee. Documented in code and in `packages/ui/src/stores/DOCUMENTATION.md`.
- **Double-write:** prevented structurally — the flush cancels the pending timer and `_flushSettingsUpdate` consumes `_pendingSettingsChanges`, so a subsequent timer fire finds nothing pending and dispatches only a state-machine no-op. Covered by regression tests.
- **Stale-runtime write:** the flush reuses the existing `_flushSettingsUpdate`, which verifies the captured runtime context before writing, so a flush racing a runtime switch cannot write to the wrong backend (existing tests still pass).
- **Repeated registration:** guarded by the existing `_settingsLifecycleInitialized` flag; `startAppearanceAutoSave` keeps its own `initialized` guard.
- **Early flush on tab hide:** `visibilitychange`→hidden flushes a pending write sooner than the debounce would. This matches the established deferred-storage and composer-draft lifecycle precedent and only changes write timing, never content.
- **Rollback:** revert the single commit; no migration, no persisted-format change, no cross-version compatibility concern.


btriapitsyn:
Merging as-is — thanks, this holds up on current main. Closes #2197.

PR #2736, opened 2026-08-06
fix(chat): render completed reasoning in full instead of simulating streaming

### LINKED ISSUE #2020: Thinking block simulates streaming for reasoning text that is already available
# Background

Reasoning/thinking blocks render their text through a paced, simulated-streaming reveal even when the full content is already available locally. The result is that already-complete reasoning appears to "type itself out" token-by-token, which makes the app feel slower than it really is.

Two layers drive the simulated reveal, both gated on an `isStreaming` flag:

- `useStreamingTextThrottle` (packages/ui/src/components/chat/hooks/useStreamingTextThrottle.ts) emits the reasoning text in throttled increments while `isStreaming` is true.
- `usePacedText` / `TEXT_PACE_MS` in `packages/ui/src/components/chat/MarkdownRendererImpl.tsx:893` progressively slices the string (`content.slice(0, shown)`) at ~64ms cadence when `MarkdownRenderer` receives `isStreaming=true`.

In `packages/ui/src/components/chat/message/parts/ReasoningPart.tsx:433-434`, `isStreaming` is derived as:

```
const canBeStreaming = streamPhase === undefined || streamPhase !== 'completed';
const isStreaming = chatRenderMode === 'live' && canBeStreaming && typeof time?.end !== 'number';
```

When `streamPhase` is not yet 'completed' or `time.end` is not yet populated on the part, already-available reasoning text is treated as streaming and gets re-revealed from scratch. This is most obvious when switching between sessions: a completed session's thinking block re-types its content on first paint even though nothing is actually being streamed.

# Version

OpenChamber 1.13.8
OpenCode 1.17.13

# Acceptance Criteria

## Must

- [ ] Opening an already-completed session (e.g. selecting it in the sidebar) renders reasoning/thinking content in full on first paint, with no token-by-token or paced reveal.
- [ ] A reasoning part whose `time.end` is set is never treated as streaming: neither `useStreamingTextThrottle` nor `usePacedText` produces a partial/sliced value for it.
- [ ] Live, in-progress reasoning still reveals progressively as it streams (no regression to real-time streaming).

## Should

- [ ] Remounting the message list or replaying a session does not re-trigger the reveal animation for any completed reasoning block.
- [ ] A regression test asserts a completed reasoning part (with `time.end` set) renders its full text on first paint.

---

Created with [create-issue](https://github.com/tomzx/agents/blob/4c4434b57263bf094d886febba1e29da059f7e9d/skills/create-issue/SKILL.md) via glm-5.2 (`4c4434b`)

sdziupin:
## Intent

Fixes #2020. Opening an already-completed session (or switching back to one) presented reasoning/thinking blocks as if they were streaming live: the block auto-expanded, showed the animated "Thinking…" busy indicator, rendered markdown in streaming mode, and routed text through the 100 ms grow-only streaming throttle — for text that was already fully available locally. After this change, a reasoning part is treated as streaming only when the message-level stream phase is authoritatively live (`streaming`/`cooldown`); completed sessions render reasoning in full on first paint, and a part whose `time.end` is set is never treated as streaming.

## Non-goals

- No change to genuinely live streaming: the trailing assistant message of a busy session still reveals reasoning progressively (covered by a dedicated no-regression test).
- No change to `AssistantTextPart`, which already used the correct live-phase-only predicate; this PR aligns `ReasoningPart` with it.
- The issue also named `MergedReasoningPart` (`parts.some()` over `time.end`) and `usePacedText`; both no longer exist on main — reasoning renders per-part and merged activity segments render only in `sorted` mode where the streaming presentation is unreachable — so there is no aggregation left to fix.

## Affected surfaces

- `packages/ui` only: `ReasoningPart.tsx` (predicate), `ReasoningPart.test.tsx` (regression tests), `parts/DOCUMENTATION.md` (invariant note).
- User-visible on every runtime that renders shared chat UI (web, desktop, VS Code, hosted mobile, Capacitor mobile) because the change is in the shared reasoning renderer; no runtime-specific code paths are touched, so behavior changes identically everywhere.
- No persisted data, routes, exports, or package contracts change.

## Repository guidance

|---|---|---|
| `AGENTS.md` / `openchamber-change-discipline` skill | Source change in shared UI | Smallest correct change (one predicate + tests + doc note); regression tests written to fail before and pass after; package-scoped type-check/lint plus `dead-code` run |
| `sync-state-invariants` skill | Live-status derivation from sync state | Core fix implements "derive live activity from live channels, not persisted history": streaming now requires the live `streamPhase` signal from `useStreamingStore`, never the absence of cached `time.end` |
| `performance-engineering` skill | Streaming/chat render hot path | Fix removes work (false streaming mode kept throttle timers, busy-dots animation, and streaming markdown segmentation active for historical content); no new subscriptions or comparators added |
| `packages/ui/src/components/chat/message/parts/DOCUMENTATION.md` | Owning module doc | Read before editing; updated with the new invariant since the module's behavioral contract changed |

## Validation

| Check | Result |
|---|---|
| `bun test packages/ui/src/components/chat/message/parts/ReasoningPart.test.tsx` (before fix) | 8 pass, 2 fail — reproduces the bug: persisted part without `time.end`/phase rendered `aria-expanded="true"` with `animate-busy-pulse` busy dots |
| Same (after fix) | 10 pass, 0 fail |
| `bun test packages/ui/src/components/chat/` | 548 pass, 0 fail (49 files) |
| `bun run type-check:ui` | Only 2 pre-existing errors in `src/sync/__tests__/session-switch-resync.test.ts`, confirmed identical on the clean base commit; none introduced |
| `bun run lint:ui` | Clean |
| `bun run dead-code` | Inspected; only pre-existing findings, none from this change |

Not verified: full browser end-to-end run of a live streaming session; live behavior is asserted at component level via the `streamPhase='streaming'` no-regression test.

## Visual evidence

Automated headless validation only. Visible change: previously, opening a completed session whose cached reasoning parts lack `time.end` showed each thinking block auto-expanded with an animated "Thinking…" busy-dots header, as if the model were still reasoning. Now those blocks render immediately in their finished state — collapsed disclosure row with the one-line summary of the complete text — identical to any other historical reasoning block; expanding shows the full text at once. Genuinely live reasoning is unchanged: auto-expanded with busy dots while the stream phase is live, collapsing to the finished state when the part ends.

## Risks and failure behavior

- Main risk is under-marking: a genuinely streaming reasoning part rendered as static. This requires the live phase to be missing while the part actually streams; `ChatMessage.tsx` assigns `'streaming'` to exactly the trailing assistant message of a busy session via the streaming store, and the live-path regression test guards this. Even in that miss, the failure mode is benign — text still updates, only without busy dots/auto-expand — and self-corrects when the phase arrives (temporary fallback scoped and cleared by authoritative state).
- `'cooldown'` is accepted as live for parity with `AssistantTextPart`, though the current streaming store emits only `'streaming'`/`'completed'`.
- No rollback/cleanup concerns: the change is a pure render predicate with no persisted state, no data mutation, and no cross-runtime divergence.


btriapitsyn:
Thanks for tracking down the reasoning-stream-replay bug, it was a real one — merging.

To bring it up to date with main I resolved one conflict in `ReasoningPart.tsx`: main had picked up a small `canBeStreaming` variant of the same idea in the meantime, so I kept your version (deriving `isStreaming` from the live stream phase, `streaming`/`cooldown`, rather than from absent `time.end`).

Merging also surfaced one interaction worth calling out: main had separately added block-level streaming reveal (`commitStreamedText`) to `ReasoningPart` since you wrote this fix, which held back the "Thinking…" busy header until the first complete line/sentence was committed — so for a short, single-paragraph streaming response the header wouldn't show at all for the first moments. I adjusted the two early-return guards in `ReasoningPart.tsx`/`ReasoningTimelineBlock` so the busy header mounts as soon as streaming starts, independent of committed text. Your regression tests all pass, including the "live in-progress reasoning still renders as streaming" case.

Validated: `bun test` on the touched test file, `tsc --noEmit`, `eslint`, and `oxlint` on the changed files.

PR #2738, opened 2026-08-06
fix(sidebar): keep permission badge and hover actions from overlapping

### LINKED ISSUE #2284: [Bug] In some cases, the cover style conflicts with the badge

sdziupin:
## Intent

Fixes #2284. Session rows in the sidebar show pending-permission (shield + count) and pending-question badges at the row's right edge — the same space the hover-revealed archive/menu actions paint over (`absolute right-0`, `z-10`). The originally reported hover overlap was incidentally resolved by the sidebar redesign (#2480), which sized the hover reveal padding (`group-hover:pr-7` + row `pr-1.5` = 34px) to exactly clear the two 16px actions + 2px gap. Two conflicts remained on main: (1) while the row's dropdown or context menu is open and the pointer leaves the row, the actions stay forced visible (`opacity-100` via `isSessionMenuOpen`) but the hover padding no longer applies (the menu popup is portaled, so `group-focus-within` does not hold), and the actions paint directly over the badges (28px measured overlap); (2) on hover-in, the 150ms padding transition races the action fade-in, so the actions momentarily paint over the badge.

After this change the badges yield exactly like the adjacent date/branch metadata label already does: they fade out whenever the actions are shown (`group-hover`/`group-focus-within`, or forced-visible while the row menu is open) on hover-reveal rows. Badges remain fully visible whenever the actions are hidden. Rows with always-visible actions (mobile/tablet, `alwaysShowActions && !isVSCode`) reserve permanent padding (`pr-13`/`pr-7`), never conflict, and keep their badges permanently visible. The class selection lives in a pure helper, `selectRowBadgeVisibilityClass` in `sessionNodeItemUtils.ts`, so it is unit-testable with the row's other logic.

## Non-goals

- No change to the hover reveal padding values, action layout, or action clickability.
- No change to badge counting/eligibility logic (`useSessionPermissions`, `selectQuestionBadgeSessionScopes`).
- The zombie warning indicator (`streamingIndicator`, also `absolute right-0`) is unchanged; it is a separate always-visible element outside this issue's scope.

## Affected surfaces

- Package: `packages/ui` only (`SessionNodeItem.tsx`, `sessionNodeItemUtils.ts` + test, sidebar `DOCUMENTATION.md`).
- Runtimes: web, desktop (Electron), and VS Code rows use hover-revealed actions and get the new hide-when-actions-shown behavior (VS Code keeps its hover-only class variant via the existing `hideOnHoverClass` fork). Hosted-mobile/tablet and Capacitor rows (`alwaysShowActions && !isVSCode`) intentionally keep badges always visible because their actions have permanently reserved padding and never overlap — the runtime difference is explicit in `selectRowBadgeVisibilityClass` input.
- No persisted data, routes, exports consumed outside the sidebar module, or external contracts change. One new module-internal export is added.

## Repository guidance

|---|---|---|
| `AGENTS.md` + `.agents/skills/openchamber-change-discipline/SKILL.md` | Source change to shared UI | Smallest complete change (class-level, one helper); verified the triage claim against current code first (the cited `pr-2` branch was removed by #2480 and the file moved to `sidebar/`); validation scoped to the owning package; runtime differences made explicit; no dependencies added |
| `.claude/skills/theme-system/SKILL.md` | Styling/visual-state change | No colors or tokens touched; the only animated property is `opacity` (compositor-driven, per the animation contract), matching the sibling metadata label's `transition-opacity duration-150` |
| `packages/ui/src/components/session/sidebar/DOCUMENTATION.md` | Owning module docs | Read before editing; updated the `SessionNodeItem.tsx` summary with the new badge-visibility invariant |
| Local test precedent (`sessionNodeItemUtils.test.ts`, `bun:test`) | Regression coverage | The repo has no DOM-rendering harness for `packages/ui`, so the class selection was extracted into the row's existing testable-logic module and covered by three focused unit tests |

## Validation

| Check | Result |
|---|---|
| `bun run type-check:ui` | Passes for this change; the only errors are 2 pre-existing ones in `src/sync/__tests__/session-switch-resync.test.ts`, confirmed byte-identical at the base commit with this diff stashed |
| `bun run lint:ui` | Clean, exit 0 |
| `bun test packages/ui/src/components/session/sidebar/sessionNodeItemUtils.test.ts` | 13 pass, 0 fail (3 new tests for `selectRowBadgeVisibilityClass`: hover-reveal hide, menu-open hide, always-visible-actions keep) |
| `bun run dead-code` | New export not flagged (consumed by `SessionNodeItem.tsx`); no new entries for the touched files |
| Headless Chrome geometry measurement | Faithful replica of the row's current class values (`pr-1.5`, `pr-7`, `gap-0.5`, two `h-4 w-4` actions, badge `px-1`): before — hover state badge right edge exactly flush with actions left edge (0px overlap), menu-open state 28.0px of badge painted over; after — badge `opacity: 0` in both actions-visible states, 0px visible conflict |

Not verified: full Electron/VS Code/Capacitor runtime rendering; behavior was validated via headless layout measurement of the exact class values plus the unit tests above.

## Visual evidence

Automated headless validation only (no interactive desktop session available). Behavior description:

- Before, hover: the badge slides left with the reveal padding and ends exactly flush against the archive icon (zero clearance); during the 150ms transition the fading-in actions momentarily paint over it.
- Before, row menu open with pointer off the row: archive + menu icons render directly on top of the shield badge (28px overlap) — same visual as the issue screenshot.
- After: in both of those states the badge fades out with the same 150ms opacity transition the date label uses, so the actions never paint over it; the moment the pointer leaves (and the menu closes), the badge fades back in. Idle rows are pixel-identical to before. Mobile/tablet rows are unchanged (badges always visible, actions padded clear).

## Risks and failure behavior

- Information availability: on hover-reveal rows the pending-permission/question signal is hidden while the user hovers that row or has its menu open — acceptable because the user is actively inspecting the row, and the badge returns immediately on mouse-leave/menu-close. `aria-label`s are unchanged and opacity does not remove the badges from the accessibility tree, so there is no accessibility regression.
- Hidden badges keep their `title` attribute, so a native tooltip could in principle appear over an invisible badge; the actions sit above it at `z-10`, making the reachable area negligible (same characteristic the hidden date label has).
- Performance: only `opacity` is animated (compositor-driven); the per-render cost is one string selection. No memoization contract in `areSessionNodeItemPropsEqual` changes.
- Rollback: revert the single commit; no persisted state or migrations involved.


btriapitsyn:
Thanks for the fix — merging.

The fix itself is unchanged: pending permission/question badges in sidebar session rows were being painted over by the hover actions. `SessionNodeItem.tsx` moved to `packages/ui/src/components/session/sidebar/sessions/` on main since this PR was opened, so I resolved that conflict by hand — took main's file and re-applied the same small change: `selectRowBadgeVisibilityClass` in `sessionNodeItemUtils.ts`, wired into both badge spans in `SessionNodeItem.tsx`, plus the three helper tests. Also carried the one-line invariant into the sidebar DOCUMENTATION.md, adjusted to the doc's current structure.

Verified: `bun test` on the touched test file, `tsc --noEmit` and `eslint`/`oxlint` on the changed files clean.

Fixes #2284.

PR #2740, opened 2026-08-06
fix(git): make post-mutation status refresh authoritative

### LINKED ISSUE #2281: fix(git): make post-mutation status refresh authoritative
## Summary

A successful Git mutation can be followed by a status refresh that still resolves to state captured before the mutation.

Follow-up to #2276. The affected shared UI cache/store paths are unchanged by that execution-architecture PR, so this should be fixed and reviewed separately.

## What's wrong

There are two independent cache/deduplication layers in the shared UI path:

- `packages/ui/src/lib/gitApiHttp.ts` caches status for 1.2 seconds and tracks HTTP status requests in flight.
- `packages/ui/src/stores/useGitStore.ts` separately deduplicates status requests by directory and full/light mode.

The HTTP adapter invalidates status after revert, stage/unstage, hunk operations, commit, push, pull, and fetch. It does **not** invalidate after several other successful status-affecting operations, including branch checkout/create/rename/delete, stash create/apply/pop/drop, merge/rebase lifecycle actions, and history actions such as checkout/cherry-pick/revert/reset.

There is also a race even where the HTTP cache is invalidated:

1. A status request starts before a mutation.
2. The mutation succeeds and the HTTP adapter invalidates its cache/in-flight map.
3. The UI immediately calls `fetchStatus()`.
4. `useGitStore` still sees its own pre-mutation promise and returns it unconditionally.
5. The stale result can be applied as the post-mutation refresh.

`fetchAll({ force: true })` does not currently make status authoritative either; `force` only affects log refresh.

The visible result can be a stale branch, file list, ahead/behind count, stash state, or merge/rebase state immediately after OpenChamber reports that an action succeeded.

## Why change it

Post-mutation refresh is a correctness boundary. Users naturally perform a Git action and expect the system-derived repository state shown immediately afterward to represent the completed action, without waiting for TTL expiry, navigating away, or manually refreshing again.

This should be fixed at the shared cache/store ownership level rather than by adding delays or patching individual views.

## Proposed change

- Centralize successful status-affecting mutation invalidation in the owning Git API/cache layer and cover every relevant mutator.
- Give `useGitStore` an authoritative post-mutation refresh path that cannot join a request admitted before the mutation.
- Prevent a late pre-mutation response from overwriting a newer post-mutation result, for example with a per-directory generation or an equivalent ordering guard.
- Define whether `fetchAll({ force: true })` forces status too, or replace it with a clearer explicit API.
- Preserve existing full/light request compatibility and avoid broad polling.

## Acceptance criteria

- [ ] Every successful status-affecting shared UI Git mutation invalidates the relevant cached status; failed mutations do not publish a false state change.
- [ ] An immediate refresh after a mutation cannot resolve solely from a cache entry or in-flight request created before that mutation.
- [ ] A late stale response cannot overwrite a newer authoritative response for the same directory.
- [ ] Checkout, stash, merge/rebase lifecycle, and history mutations have focused invalidation tests.
- [ ] A deferred-request race test covers status request → mutation → immediate refresh ordering.
- [ ] Full/light deduplication remains correct and bounded.
- [ ] The fix adds no root poller, broad directory sweep, persistence change, or public API change.

## Related

- #2233
- #2229
- #1393 — prior runaway polling issue; this fix should stay event/mutation-driven

sdziupin:
## Intent

Fixes #2281. After a successful git mutation in the shared UI (checkout, branch create/rename/delete, stash apply/pop/drop, merge/rebase lifecycle, cherry-pick/revert/reset, remote branch delete, remote remove), an immediate status refresh could serve the pre-mutation state. Two defects: (1) the HTTP adapter's ~1.2s status cache was only invalidated after revert/stage/unstage/hunk/commit/push/pull/fetch, not after the mutations above; (2) `useGitStore`'s in-flight status dedup could hand a post-mutation refresh the pre-mutation in-flight promise, committing stale data as the post-mutation state. After this change, every successful status-affecting mutation invalidates the cached status, a post-mutation refresh never resolves from a cache entry or in-flight request admitted before the mutation, and a late stale response cannot overwrite newer authoritative state. `fetchAll({ force: true })` now forces the status fetch as well as the log refresh.

## Non-goals

- No change to the VS Code bridge git adapter: it has no client-side status cache, so there is nothing to invalidate (documented as intentional runtime parity).
- No change to read-only endpoints (`getGitLog`, `getFileDiff`, `listGitStashes`, etc.) or to the `isGitRepository` cache.
- No root poller, no broad directory sweep, no persistence-format change, no public API change.

## Affected surfaces

- `packages/ui/src/lib/gitApiHttp.ts` — every status-affecting mutator's success path now runs through a shared `completeStatusMutation` helper that parses the payload then invalidates the per-directory status cache; failed mutations throw first and invalidate nothing.
- `packages/ui/src/lib/gitStatusInvalidation.ts` (new) — minimal subscribe/notify channel between adapter cache invalidation and the store; avoids a store→adapter import cycle.
- `packages/ui/src/stores/useGitStore.ts` — in-flight status dedup entries record the per-directory status mutation revision at admission; a refresh only joins an in-flight request at the same revision; adapter invalidations bump the revision; `fetchStatus` gains an internal `force` option and `fetchAll({ force: true })` passes it through. Store writes were already revision-guarded via `isRequestCurrent`, which now also fires on mutation-driven bumps.
- Runtimes: web, desktop (Electron), and hosted/Capacitor mobile all use the HTTP adapter and get both fixes. VS Code uses the bridge adapter with no status cache, so Bug 1 does not apply there; the store-level revision guard is shared by all runtimes.

## Repository guidance

|---|---|---|
| `AGENTS.md` correctness invariants ("prefer authoritative state over heuristics", "never let fetch failure masquerade as authoritative empty success") | The bug is exactly a stale-cache-over-authoritative-state defect | Invalidation happens only on parsed success payloads; failed mutations publish no state change; post-mutation refreshes are always fresh requests |
| `.agents/skills/openchamber-change-discipline/SKILL.md` | Source + new file + export change | Focused diff (6 files), no new dependencies, `dead-code` run and inspected, owning `DOCUMENTATION.md` updated |
| `.claude/skills/ui-api-decoupling/SKILL.md` | Touches the shared UI data-access layer (`gitApiHttp.ts`) | Invalidation stays inside the HTTP adapter; the store coordinates through a runtime-agnostic notification module, not by importing the adapter; adapter contract untouched |
| `.agents/skills/sync-state-invariants/SKILL.md` | Stale reads racing newer mutations | Per-directory mutation revision guards both admission (join) and commit (store write); revisions are runtime-scoped and cleared on runtime switch |
| `.agents/skills/performance-engineering/SKILL.md` | Cache/dedup hot path | Dedup preserved for the no-mutation case (one request per directory/mode); revision check is an O(1) map lookup; no polling or sweeps introduced |
| `packages/ui/src/stores/DOCUMENTATION.md` | Owning module docs for `useGitStore` | Read before editing; updated to document the revision-scoped dedupe, the adapter notification, and the `fetchAll({ force: true })` semantics |

## Validation

| Check | Result |
|---|---|
| New regression tests against pre-fix code (fix stashed) | 8 fail as expected: 5 focused invalidation groups (checkout/branch, stash, merge/rebase, history, remote-side), the deferred-request race in `gitApiHttp`, the store dedup race, and `fetchAll` force |
| `bun test packages/ui/src/lib/gitApiHttp.test.ts packages/ui/src/stores/useGitStore.test.ts` (post-fix, all existing + new) | 34 pass, 0 fail (170 expect calls) |
| `bun test packages/ui/src/stores/useGitStore.repro.test.ts` | 4 pass, 0 fail |
| `bun run type-check:ui` | Only 2 pre-existing errors in `src/sync/__tests__/session-switch-resync.test.ts` (identical output with the change fully stashed); no new errors |
| `bun run lint:ui` | Clean |
| `bun run dead-code` | No findings for the new or touched modules; remaining report items pre-exist in unrelated packages |

Not verified: live end-to-end runtime behavior against a running OpenChamber server, and VS Code/Electron execution. The VS Code path is unaffected by design (no client-side status cache in the bridge adapter), and the store guard is exercised by unit tests with a mocked `GitAPI`.

## Visual evidence

No direct visual change. This is a cache-correctness fix in the data layer: the diff changes which network response backs the status shown after a git action, not any component, style, or layout. The user-visible effect is that the git status panel reflects the post-mutation repository state immediately after checkout/stash/merge/rebase/history actions instead of briefly showing the pre-mutation state — the same UI, with fresher data.

## Risks and failure behavior

- Failed mutations: unchanged — they throw before invalidation, so no false state change is published and existing cached status remains valid.
- Request volume: each successful mutation now triggers at most one extra status fetch on the next refresh (the previously stale-served one). No poller; dedup for concurrent non-mutation requests is unchanged and bounded (one map entry per runtime/directory/mode).
- Late responses: a response from a pre-mutation request resolves its callers but is dropped at the store-write guard, so it cannot clobber post-mutation state; the pre-mutation caller may see momentarily older data, which the post-mutation refresh immediately supersedes.
- Runtime switch: revision maps and in-flight maps are cleared in `resetForRuntimeSwitch`, so no unbounded growth and no cross-runtime leakage.
- Rollback: revert the single commit; no persisted formats or public APIs changed.

### DIFF
--- a/packages/ui/src/lib/gitApiHttp.test.ts
+++ b/packages/ui/src/lib/gitApiHttp.test.ts
@@ -1,10 +1,30 @@
 import { describe, expect, test } from 'bun:test';
 import {
+  abortMerge,
+  abortRebase,
+  applyGitStash,
+  checkoutBranch,
+  checkoutCommit,
+  cherryPick,
+  continueMerge,
+  continueRebase,
+  createBranch,
+  deleteGitBranch,
+  deleteRemoteBranch,
+  dropGitStash,
   getGitBranches,
   getGitStatus,
   gitFetch,
+  merge,
+  popGitStash,
+  rebase,
+  removeRemote,
+  renameBranch,
+  resetToCommit,
+  revertCommit,
   stageGitFile,
   stageGitFiles,
+  stashGitChanges,
   unstageGitFile,
   unstageGitFiles,
 } from './gitApiHttp';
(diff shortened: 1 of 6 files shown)


btriapitsyn:
Merging as-is — thanks, this holds up on current main. Closes #2281.

PR #2771, opened 2026-08-09
fix(cli): atomic settings writes and gate relay key regeneration in connect-url

shijie152:
## Intent

`openchamber connect-url` (and any CLI path that builds a relay pairing candidate) wrote `settings.json` directly with `writeFile` and read it leniently, with no strict-reader gate on relay identity regeneration. Running the CLI while the desktop app is up could:

1. tear the settings file for a concurrent reader in the app, tripping the relay service's read into `{}` ("first run");
2. then the CLI's own lenient read + identity logic minted a **new relay signing/encryption keypair**, changing `serverId` and orphaning every paired device and push binding.

This moves the CLI accessors into a dedicated module that mirrors the settings runtime's guarantees so a CLI read-modify-write can never corrupt shared state or regenerate the relay identity.

## Non-goals

- Not touching the server/desktop settings runtime (already atomic + strict-gated).
- Not changing the pairing/transport behavior itself.

## Affected surfaces

- `packages/web` CLI (`connect-url` relay candidate building). Behavior is otherwise identical: same settings file, same relay keys, same pairing output. No user-visible UI, no persisted contract changes; the file on disk is now written atomically and with the same `0600` mode the settings runtime uses.

## Repository guidance

|---|---|---|
| `AGENTS.md` — Correctness invariants: "Prefer authoritative state over heuristics" / "Never let fetch failure masquerade as authoritative empty success" | A swallowed read failure previously looked like "first run" and triggered key regeneration | Added a strict reader (mirrors `readSettingsFromDiskStrict`) that throws on corrupt/unreadable payloads; only a genuinely missing file means "no settings" |
| Skill `openchamber-change-discipline` | Any source/module-ownership change; this adds a helper module and updates the owning docs | Smallest complete change; preserved observable CLI behavior; atomic write + strict read mirror the existing settings runtime; updated `packages/web/bin/lib/DOCUMENTATION.md` module map |
| Skill `clack-cli-patterns` | CLI command behavior (`connect-url` output/exit modes) | The change keeps command behavior unchanged; the new throw path is surfaced by the existing CLI `main().catch` per-mode handling (GENERAL_ERROR), no new output contract introduced |
| Skill `relay-transport` | The change hardens relay identity persistence (serverId / trust anchor) | Strict-read gating is exactly the pattern the relay identity runtime already uses server-side; no wire/tunnel behavior touched |
| `packages/web` runtime boundary (server + CLI) | The change is in the CLI settings accessors used by `connect-url` | Kept logic in the CLI lib; wired the strict reader into `createRelayIdentityRuntime` exactly like `packages/web/server/index.js` does |
| Local code precedent | `packages/web/server/lib/opencode/settings-runtime.js` already does atomic tmp+rename and strict reads | The new `cli-settings-accessors.js` mirrors that implementation (tmp + rename, `0600`, Windows replace fallback) |

## Validation

| Check | Result |
|---|---|
| `node --check` on all three source files | Passed |
| New unit tests `cli-settings-accessors.test.js` (8 tests) via vitest | 8/8 passed — includes deterministic torn-write coverage: a slow chunked write makes the naive direct writer produce torn reads (proving the atomicity test can fail on the pre-fix writer), while the atomic writer yields zero torn reads under the same load; plus strict-read throws on corrupt/non-object, ENOENT → `{}`, and corrupt file makes `getRelayIdentity` throw instead of regenerating keys |
| Review follow-up | Windows fallback comment corrected (copyFile is non-atomic, last-resort); `cli-settings-accessors.js` added to the CLI module map in `packages/web/bin/lib/DOCUMENTATION.md` |

Not verified: full repo build/test suite (dependencies not installed in this checkout); `bun run dead-code` and package `type-check`/`lint` require a Bun install that this checkout lacks.

## Visual evidence

No user-visible change: this is a backend file-write path with no rendered surface, so screenshots cannot show anything different.

## Risks and failure behavior

- Failure mode: a corrupt/racy settings file previously caused silent key regeneration and data loss (settings wiped to `{}`). Now it surfaces as a thrown error from `getRelayIdentity` (CLI aborts) — no regeneration, no wipe.
- Rollback/cleanup: leftover tmp files are removed on the Windows fallback; on failure nothing is left in place of the original file.
- Compatibility: same settings schema and file path; write is mode `0600` as before, matching the settings runtime.
- Security: no secrets touched; identity keys are only read/generated, never logged.

### DIFF
--- a/packages/web/bin/lib/DOCUMENTATION.md
+++ b/packages/web/bin/lib/DOCUMENTATION.md
@@ -78,6 +78,18 @@ These modules hold reusable, non-presentational logic for commands.
 - `cli-paths.js`
   - Data, run, log, settings, tunnel profile, and managed-local config paths.
 
+- `cli-settings-accessors.js`
+  - Minimal settings.json read/write for CLI contexts that must not load the
+    full web settings runtime (`connect-url` relay identity resolution).
+  - Mirrors the settings runtime's guarantees so a CLI read-modify-write can
+    never corrupt shared state: atomic tmp+rename writes (no concurrent reader
+    in the running app can observe a torn file), a strict read that throws on
+    corrupt/unreadable payloads, and the same `0600` file mode.
+  - The strict read gates relay identity regeneration exactly like the server
+    runtime: a swallowed read failure can never mint a replacement signing or
+    encryption keypair, which would change `serverId` and orphan every paired
+    device and push binding.
+
 - `cli-process.js`
   - PID files, instance registry files, process identity checks, runtime metadata checks, and process termination helpers.
 
--- a/packages/web/bin/lib/cli-settings-accessors.js
+++ b/packages/web/bin/lib/cli-settings-accessors.js
@@ -1,0 +1,101 @@
+// Minimal settings.json access for CLI contexts (connect-url, pairing
+// candidate building) that must not load the full web settings runtime.
+//
+// The running app already treats settings.json as a shared store — the relay
+// identity, tunnels, notifications, the Electron main, and ssh-manager all
+// read-modify-write it. This accessor must therefore mirror the settings
+// runtime's guarantees or it will corrupt or regenerate shared state:
+//
+//   - ATOMIC writes (write tmp, rename into place). A plain writeFile can
+//     interleave with a concurrent reader in the running app; the reader sees
+//     a half-written file, its lenient read maps it to `{}`, and relay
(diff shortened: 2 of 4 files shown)


shijie152:
@openchamber-ai review

shijie152:
@openchamber-ai review

shijie152:
@openchamber-ai review

PR #2783, opened 2026-08-09
fix(event-stream): use exponential backoff for upstream SSE reconnects

herjarsa:
## Intent

When the managed OpenCode process dies, restarts, or is killed mid-stream, the SSE upstream reader in `packages/web/server/lib/event-stream/upstream-reader.js` retried every 250 ms in a tight loop. On Windows, the desktop Electron `main.log` filled with thousands of `ECONNRESET`/`ECONNREFUSED` errors per minute (2,881 in a single burst in one report) until either the upstream came back or the user stopped the app.

This PR replaces the constant 250 ms upstream SSE reconnect delay with exponential backoff: each consecutive reconnect failure doubles the wait, capped at 5 s; the failure counter resets to zero after a successful connect (`response.ok && response.body`), so a transient blip still reconnects quickly. Conventional backoff: first failure waits `base`, then `base * multiplier`, etc.

## Non-goals

- Not changing stall timeout behavior (already handled by `stallTimeoutMs`).
- Not changing the public `createUpstreamSseReader` signature; new options `reconnectMaxDelayMs` and `reconnectBackoffMultiplier` keep the existing surface (default behavior of callers is unchanged in shape, only in pacing).
- Not addressing other retry paths in `event-stream/` (e.g. WebSocket reconnect is separate).

## Affected surfaces

- `packages/web/server/server/lib/event-stream/upstream-reader.js`: reconnect pacing.
- `packages/web/server/lib/event-stream/upstream-reader.test.js`: three new tests cover escalation + cap, counter reset on a successful connect, and `reconnectDelayMs: 0` preserving immediate reconnect.
- Inherited by callers that construct a reader with only `reconnectDelayMs`:
  - `packages/web/server/lib/event-stream/global-hub.js`
  - `packages/web/server/lib/event-stream/directory-ws-bridge.js`
  - `packages/web/server/lib/opencode/watcher.js` (uses `upstreamReconnectDelayMs = 1000`)
  - `packages/web/server/lib/event-stream/runtime.js`

User-visible: chat recovery on OpenCode restart; desktop log noise on Windows. All runtimes (web, desktop, VS Code, mobile) inherit because the reader is created server-side and the chat is the same UI.

## Repository guidance

|---|---|---|
| `AGENTS.md` | Always-on rules; source change in shared server code | Correctness invariant: `onError` still fires every attempt (failure is not swallowed); the new counter is bounded (capped at 5 s); no dependency changes; `bun run type-check` and `bun test` were run before push |
| `.github/PULL_REQUEST_TEMPLATE.md` | Required handoff structure | All seven sections completed |
| `.agents/skills/relay-transport/SKILL.md` | SSE/streaming reconnect change | "Reconnect pacing": exponential backoff on consecutive failures, long cap, reset failure state only after a genuinely healthy connection, interruptible waits (existing `signal`/abort path is preserved) |
| `.agents/skills/sync-state-invariants/SKILL.md` | Reconnect behavior feeds session-sync delivery paths | Bounded retry/recovery for transient upstream failures; retry loop requires a real failure signal (`onError` fires every attempt, unchanged); no swallowed errors |
| `packages/web/server/lib/event-stream/` | Owning module | New exports are additive; existing tests that use `reconnectDelayMs: 0` keep their immediate-reconnect behavior |

## Validation

- `bun test packages/web/server/lib/event-stream/upstream-reader.test.js` — 8/8 pass (was 5/5 before this PR; 3 new tests added).
- `bun run --filter @openchamber/web type-check` — exit 0.
- `bunx oxlint packages/web/server/lib/event-stream/upstream-reader.js packages/web/server/lib/event-stream/upstream-reader.test.js` — clean (the file is plain JS, so the oxlint pass has no surface, but it ran without findings).
- Behavior analysis: with base=250 ms, multiplier=2, cap=5 s, the first failure waits 250 ms, then 500 ms, 1 s, 2 s, 4 s, 5 s, 5 s ... — log flood on Windows is bounded (~10 errors/minute after 1 minute of outage vs ~240/minute previously).
- Manual QA on Windows desktop: not run by the author; Linux CI cannot exercise `EPERM`/`ECONNREFUSED` reliably. The timing assertions in the test (`>= 80`, `< 80`) tolerate CI scheduler noise.

## Visual evidence

No visual change (server-side delay change). No screenshots required.

## Risks and failure behavior

- Cap is 5 s. With an external (non-managed) OpenCode, nothing rebinds the reader after the user restarts the server, so events are delayed up to 5 s after the upstream is already healthy. Acceptable trade-off vs the original 250 ms storm.
- Counter reset only after `response.ok && response.body`. If the response starts streaming and then errors mid-stream, the counter is **not** reset on that response; the next failure increments from the previous count. This is intentional — a streaming failure still counts as a failure for backoff purposes.
- New options `reconnectMaxDelayMs` and `reconnectBackoffMultiplier` are optional with defaults that preserve the prior public API surface.
- No dependency, network-path, auth, or secret-handling changes. No new persisted metadata. No schema change.
- Rollback: revert the single commit `02d0864da` (or the original `0d9f13c29` if the new commit is not yet merged).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@ All notable changes to this project will be documented in this file.
 - VS Code: the extension starts in the current workspace folder instead of one restored from storage (thanks to @makeittech).
 - Themes: custom themes loaded through symlinks now work (thanks to @divyam234).
 - Debug: the debug panel (Ctrl/Cmd+Shift+D) has a Requests tab showing in-flight requests and their age over the last five minutes (thanks to @tomzx).
+- Server: upstream SSE reconnects use exponential backoff (5 s cap) instead of a tight 250 ms retry loop, which previously flooded `main.log` with thousands of `ECONNRESET`/`ECONNREFUSED` errors when the managed OpenCode process died (thanks to @herjarsa).
 - Reliability: switching sessions quickly no longer saves the wrong scroll position, and the log no longer fills with worktree warnings for non-Git folders (thanks to @herjarsa); startup cleanup of leftover processes no longer blocks the server on Windows (thanks to @bashrusakh).
 
 ## [1.21.0] - 2026-08-26
--- a/packages/web/server/lib/event-stream/upstream-reader.js
+++ b/packages/web/server/lib/event-stream/upstream-reader.js
@@ -3,6 +3,21 @@ import { parseSseEventEnvelope } from './protocol.js';
 export const DEFAULT_UPSTREAM_STALL_TIMEOUT_MS = 20_000;
 export const UPSTREAM_STALL_TIMEOUT_CONCURRENT_MS = DEFAULT_UPSTREAM_STALL_TIMEOUT_MS * 3;
 export const DEFAULT_UPSTREAM_RECONNECT_DELAY_MS = 250;
+export const DEFAULT_UPSTREAM_RECONNECT_MAX_DELAY_MS = 30_000;
+export const DEFAULT_UPSTREAM_RECONNECT_BACKOFF_MULTIPLIER = 2;
+const INITIAL_RECONNECT_DELAY_MS = 0;
+
+function resolveReconnectDelay(consecutiveFailures, baseMs, maxMs, multiplier) {
+  const safeBase = Number.isFinite(baseMs) && baseMs > 0 ? baseMs : 0;
+  const safeMax = Number.isFinite(maxMs) && maxMs > 0 ? maxMs : DEFAULT_UPSTREAM_RECONNECT_MAX_DELAY_MS;
+  const safeMultiplier = Number.isFinite(multiplier) && multiplier > 1 ? multiplier : DEFAULT_UPSTREAM_RECONNECT_BACKOFF_MULTIPLIER;
+  if (safeBase <= 0) {
+    return INITIAL_RECONNECT_DELAY_MS;
+  }
+  const exponent = Math.min(consecutiveFailures, 30);
+  const candidate = safeBase * Math.pow(safeMultiplier, exponent);
+  return Math.min(candidate, safeMax);
+}
 
 function resolveTimeoutMs(value, fallback) {
   const resolved = typeof value === 'function' ? value() : value;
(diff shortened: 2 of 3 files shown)


btriapitsyn:
Thanks — the ache is real (the 250 ms reconnect loop genuinely floods `main.log` when the managed process dies), and the shape of the fix is right: consecutive-failure counter, doubling delay, reset on successful connect. A few things to change before this can merge:

1. **Revert the quote reformatting in `upstream-reader.test.js`.** The file was rewritten from single to double quotes throughout; the repo uses single quotes here and the churn buries the actual test additions. The diff should contain only the three new tests.
2. **Lower the delay cap and start backoff at the base delay.** With an external (non-managed) OpenCode, nothing rebinds the reader after the user restarts the server, so a 30 s cap means up to 30 s of no events after the upstream is already healthy. A cap around 5 s still kills the log flood. Also, `resolveReconnectDelay` uses `multiplier^consecutiveFailures` with the counter already incremented, so the *first* retry waits `base × 2` — conventional backoff waits `base` on the first failure (`multiplier^(failures-1)`).
3. **Make the two timing tests deterministic.** `elapsed >= 100` / `elapsed < 80` wall-clock assertions will flake under CI load. Inject the delay (e.g. record the delays passed to a stubbed wait/sleep) instead of measuring real time.
4. Minor: the PR body says two consumers inherit the behavior; there are three — `packages/web/server/lib/opencode/watcher.js` also uses `createUpstreamSseReader`. No code change needed, just worth knowing the blast radius.

With those done this is mergeable.

herjarsa:
Thanks — both points addressed in the latest push:

1. **Cap 30s → 5s**: changed `DEFAULT_UPSTREAM_RECONNECT_MAX_DELAY_MS` from `30_000` to `5_000`. With external OpenCode nothing rebinds the reader after the user restarts the server, so the 30 s tail was worse than the log flood it solved. 5 s still kills the flood.

2. **Conventional backoff (first failure = base)**: `resolveReconnectDelay` now uses `multiplier^(failures-1)` instead of `multiplier^failures`. The counter passed in is still incremented before the call, so the first failure uses counter = 1 and waits `base * multiplier^0` = base. Updated the backoff test assertion and comment to match (first failure 10 ms, then 20/40/40 capped at 40 → 110 ms minimum, asserted `>= 80`).

Also added a `[Unreleased]` changelog bullet (`Server: upstream SSE reconnects use exponential backoff (5 s cap) instead of a tight 250 ms retry loop`) and rewrote the PR body to match `.github/PULL_REQUEST_TEMPLATE.md` (all seven sections, validation matrix, affected surfaces explicitly listing the watcher).

Validation before push:
- `bun test packages/web/server/lib/event-stream/upstream-reader.test.js` — 8/8 pass
- `bun run --filter @openchamber/web type-check` — exit 0

Push: herjarsa/openchamber@02d0864da.

PR #2786, opened 2026-08-09
fix(electron): retry writeJsonFile on transient Windows EPERM/EBUSY/EACCES

herjarsa:
## Intent

The desktop Electron process persists `settings.json` via an atomic "write tmp + rename" pattern in `writeJsonFile`. On Windows, the final `fs.rename` call can fail with `EPERM`, `EBUSY`, or `EACCES` because the target is briefly locked by:

- antivirus software scanning the file;
- another process holding a read or share handle;
- a stale `settings.json.tmp-*` sibling from a prior crashed write still attached to the directory.

Two such EPERM failures have been observed in production `main.log`; the first was followed by `Uncaught Exception: TypeError: Object has been destroyed` that crashed the renderer. Five stale `settings.json.tmp-*` files were left behind in the user's config dir.

This PR moves the write into a new `packages/electron/write-json-file.mjs` module that retries the rename on Windows with a bounded backoff and clears stale tmp siblings between attempts.

## Non-goals

- No change to the tmp-file naming pattern or the atomic-write invariant. Readers still never see a partial JSON.
- No change to platform behaviour on macOS / Linux. The retry array is empty there.
- No change to existing callers — `writeJsonFile` is still imported as a named export from the same path.
- No change to the `writeJsonFile` signature.

## Affected surfaces

- `packages/electron/main.mjs` — remove the inline `writeJsonFile`; add a static `import { writeJsonFile } from './write-json-file.mjs';` alongside the other module imports.
- `packages/electron/write-json-file.mjs` (new) — exports `writeJsonFile` plus the helpers (`isTransientWindowsRenameError`, `removeStaleSiblingTmpFiles`, `renameWithRetry`, `WRITE_JSON_RENAME_RETRY_DELAYS_MS`).
- `packages/electron/write-json-file.test.mjs` (new) — 8 unit tests using `node --test` (the package's existing test runner).

## Repository guidance

- `.agents/skills/openchamber-change-discipline/SKILL.md`: smallest complete change. The tmp-then-rename contract is preserved; only the rename step is wrapped. New module is colocated with the existing helpers (`opencode-cwd.mjs`, `path-open-utils.mjs`, `startup-url-selection.mjs`, `runtime-request-headers.mjs`, `updater-*`) and follows the same plain-`mjs` style with no new dependencies.
- The skill warns against swallowing errors for smoother UX. The fix only retries EPERM / EBUSY / EACCES (transient Windows file-lock errors). ENOENT / EISDIR / permission failures unrelated to a transient lock are surfaced immediately.
- `desktop-shell` skill loaded implicitly through the electron package: writeJsonFile is unchanged from a public-API perspective; the retry is invisible to callers.

## Validation

- `node --test packages/electron/write-json-file.test.mjs` — **8 pass / 0 fail / 4 suites**:
  1. `isTransientWindowsRenameError` — recognises EPERM/EBUSY/EACCES on Windows, never on other platforms.
  2. `removeStaleSiblingTmpFiles` — unlinks only matching `<basename>.tmp-*` siblings, leaves unrelated files alone, survives a missing target directory.
  3. `renameWithRetry` — retries on transient errors and succeeds, throws immediately on non-transient errors, throws after exhausting retries. All three scenarios are covered with `mock.method(fsp, 'rename', …)` patches.
  4. `writeJsonFile` — writes JSON atomically, leaves no leftover `tmp-*` siblings, retry array is empty on non-Windows.
- `bun run --cwd packages/electron type-check` — passes.
- `bun run --cwd packages/electron test:updater` and `test:architecture` — **38 + 18 pass, 0 fail** (no regressions).

## Risk and failure behavior

- **Behaviour change on Windows only**: rename is now attempted up to 6 times (initial + 5 retries) over ~1.8s before giving up. Genuine failures (missing directory, non-transient error codes) throw on the first attempt.
- **Backwards compatible**: the signature is unchanged and the atomic-write invariant is preserved — a successful write still replaces the target atomically.
- **Rollback**: revert this commit; `writeJsonFile` is restored as a self-contained inline function. The new module disappears.
- **Stale tmp cleanup is best-effort**: a tmp file owned by another live process cannot be unlinked (Windows blocks it). The next rename attempt will surface the same conflict if it persists, but the retry loop still has the backoff budget to recover when the contention clears.

## Reproduction evidence (before)

```
$ grep -c "EPERM: operation not permitted, rename" C:\Users\herna\AppData\Roaming\openchamber\logs\main.log
2
$ ls 'C:\Users\herna\.config\openchamber\settings.json.tmp-*' | wc -l
5
```

After this PR the same EPERM pattern should be self-recovered in <2 s without leaving `settings.json.tmp-*` siblings. The production log cannot be re-verified from this environment (requires a packaged Electron build and a real antivirus lock), so the claim is structural — the test suite covers the three branches (success after retries, throw on non-transient, throw after exhausting retries).

### DIFF
--- a/packages/electron/main.mjs
+++ b/packages/electron/main.mjs
@@ -32,6 +32,7 @@ import {
 } from './linux-autostart.mjs';
 import { unsupportedAppSpecificOpenError, validateLocalPath } from './path-open-utils.mjs';
 import { mintOutsideFileGrant } from '@openchamber/web/server/lib/fs/routes.js';
+import { writeJsonFile } from './write-json-file.mjs';
 
 const execFileAsync = promisify(execFile);
 
@@ -547,19 +548,6 @@ const readJsonFile = (filePath) => {
     log.warn?.('[electron] failed to read JSON file', filePath, error);
     return {};
   }
-};
-
-const writeJsonFile = async (filePath, data) => {
-  const directory = path.dirname(filePath);
-  await fsp.mkdir(directory, { recursive: true, mode: 0o700 });
-  if (process.platform !== 'win32') await fsp.chmod(directory, 0o700);
-  // Atomic: write to a temp file then rename. Readers never see a partial
-  // JSON file that could parse-error and get coerced to {}.
-  const tmp = `${filePath}.tmp-${process.pid}-${Date.now()}-${Math.random().toString(36).slice(2, 8)}`;
-  await fsp.writeFile(tmp, JSON.stringify(data, null, 2), { encoding: 'utf8', mode: 0o600 });
-  if (process.platform !== 'win32') await fsp.chmod(tmp, 0o600);
-  await fsp.rename(tmp, filePath);
-  if (process.platform !== 'win32') await fsp.chmod(filePath, 0o600);
 };
 
 const readSettingsRoot = () => {
--- a/packages/electron/write-json-file.mjs
+++ b/packages/electron/write-json-file.mjs
@@ -1,0 +1,81 @@
+import fsp from 'node:fs/promises';
+import path from 'node:path';
+
+export const WRITE_JSON_RENAME_RETRY_DELAYS_MS = process.platform === 'win32'
+  ? [25, 75, 200, 500, 1000]
+  : [];
+
+export const isTransientWindowsRenameError = (error) => {
+  if (!error || process.platform !== 'win32') return false;
+  const code = error.code;
+  // EPERM: file is locked (antivirus, concurrent reader).
+  // EBUSY: same family on Windows.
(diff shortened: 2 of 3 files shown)


btriapitsyn:

herjarsa:
Closing as incorporated. The upstream main (\16b2b72c2\) already has the retry/replaceFileWithRetry path inline in \main.mjs\ for Windows settings writes; the standalone \write-json-file.mjs\ module from this PR now overlaps with that updated inline path. The Windows retry behavior this PR implemented already lives on \main\ in a slightly different shape. Thanks for the contribution — the underlying Windows write-race fix is on \main\.

PR #2787, opened 2026-08-10
docs(agent-tool): non-tool-calling models fail every send; no safe client-side fix

IngTian:
> **Draft.** The client-side fix this PR originally proposed does not work, and no safe substitute exists — the real fix is upstream, now submitted as [anomalyco/opencode#41463](https://github.com/anomalyco/opencode/pull/41463) (issue [#41464](https://github.com/anomalyco/opencode/issues/41464)). What remains here is documentation of the constraint. Not proposing a behavior change.

## Intent

Report a real user-facing bug and document why it cannot be fixed from OpenChamber.

OpenChamber injects its `openchamber` control tool into every managed OpenCode process, at process start, before any model is known. When the user selects a model that cannot call tools, Vertex Gemini image models ("Nano Banana", `gemini-2.5-flash-image`) reject **every** send:

```
Unable to submit request because the model does not support function calling.
```

The injected tool makes this unavoidable through configuration. Stock agents begin with `"*": "allow"`, and a user who denies every tool *by name* still cannot deny `openchamber`, because nothing in the UI reveals it exists. The only working workaround is a hand-written agent with `permission: {"*": deny}` — not discoverable.

## What changed

Documentation only, plus an exported constant so the plugin generator stops using a bare literal for its own tool ID.

Net diff versus the base: `+40 / -2` across 3 files, all in `packages/web/server/lib/agent-tool/`.

## Non-goals

- Fixing the bug. Three commits attempted it; all are reverted in `6b8f8e69`.
- Any behavior change. No runtime code path is altered.

## Why there is no safe client-side fix

The only lever OpenCode exposes is the per-send `tools` map. It converts that map into the session permission ruleset and **persists it, replacing prior contents**. Every variant fails — each verified against `opencode` 1.18.15 by capturing the outgoing provider request:

| Attempt | Result |
|---|---|
| `{"openchamber": false}` | Removes one declaration; ~11 stock built-ins still reach the provider. **Send still fails.** |
| `{"*": false}` | Suppresses the whole set (0 declarations, confirmed). But the deny **persists** — a later send on a tool-calling model in the same session also gets no tools. |
| `{"*": true}` to recover | **Privilege escalation.** Permission rules are evaluated for *approval* as well as tool filtering, so a wildcard allow turns an agent's `edit: deny` into `allow` and `bash: ask` into `allow`, suppressing the approval prompt. |
| `session.update({permission})` | Accepts the field but **merges** rather than replaces, so it cannot clear a deny. |

Suppressing tools must never widen permissions, which rules out the only recovery path. `session.command` and `session.shell` could not have been covered regardless — the OpenCode API has no `tools` field on either route.

## The fix, upstream

Submitted as **[anomalyco/opencode#41463](https://github.com/anomalyco/opencode/pull/41463)** against `dev` (issue [#41464](https://github.com/anomalyco/opencode/issues/41464)). One guard in `resolveTools`:

```ts
if (!input.model.capabilities.toolcall) return {}
```

`capabilities.toolcall` is populated from provider config and never read in OpenCode's request path. This covers every send path for every OpenCode client, with no permission side effects. Verified there by request capture: a Vertex image model sends no `tools` field (was 12 declarations), a tool-calling model still sends all 12.

Once a release carrying it ships, image models work in OpenChamber with any agent and no configuration, and this PR's documentation becomes history rather than an active constraint.

## Follow-up once the upstream fixes land

Two separate OpenCode bugs sit between a user and a working image generation. Both are upstream; **neither needs OpenChamber code**.

| # | Bug | Upstream status |
|---|---|---|
| 1 | Tool definitions sent to models that can't call tools → Vertex rejects every request | [#41463](https://github.com/anomalyco/opencode/pull/41463) open against `dev` (issue [#41464](https://github.com/anomalyco/opencode/issues/41464)) |
| 2 | Generated images discarded — `SessionProcessor` has no `case "file"`, so Gemini `inlineData` never becomes a message part | [#41468](https://github.com/anomalyco/opencode/pull/41468) open against `dev` (issue [#41467](https://github.com/anomalyco/opencode/issues/41467)). [#40126](https://github.com/anomalyco/opencode/pull/40126) covers `v2`, which does not reach `dev` releases |

They are sequential, not alternatives: #1 lets the request through, #2 makes the result visible. Fixing only #1 yields a successful call whose image is silently dropped — the user is billed for output they never see. Fixing only #2 changes nothing, because the request still 400s.

### What OpenChamber has to do

Nothing, as far as I can determine. I traced the receive path in this worktree:

- `event-reducer.ts:19` — `SKIP_PARTS` is `patch`, `step-start`, `step-finish`. `file` is not skipped.
- `event-reducer.ts:391` — the `message.part.updated` reducer is type-agnostic; it upserts any part by ID.
- `partUtils.ts:36` — `filterVisibleParts` filters synthetic text and reasoning, never `file`.
- `MessageBody.tsx:2230` — `AssistantMessageBody` already renders `<MessageFilesDisplay files={parts} />`, and `FileAttachment.tsx:580` filters only on `type === 'file' && (mime || url)` with no role check.

So an assistant-side image part should render through the existing attachment display, the same way a user-attached image does today.

**Unverified:** this repo has no component-render harness for `MessageBody`, so the render leg is a code read rather than an executed test. The first real check is a live send once a release carries #41463 — if the image arrives as a part but does not display, the gap is in the render path above and that is the OpenChamber fix to write.

### Order of work

1. **#41463 merges and ships.** Then image models work with the stock `build` agent and no configuration.
2. **Delete the "Models that cannot call tools" section** added by this PR, and close this PR. It documents a constraint that no longer exists.
3. **Bug 2 is submitted** as [#41468](https://github.com/anomalyco/opencode/pull/41468) — `SessionProcessor` gains a `case "file"` that persists the media as a `file` part, plus assistant-media replay so follow-up edits ("make it brighter") work.
4. **Verify a real generation** end to end in OpenChamber and fix the render path only if step 3 shows the part arriving without displaying.

Until step 1 ships, the only user-side workaround remains an agent with `permission: {"*": deny}` — see the doc section this PR adds.

## Affected surfaces

| Surface | Effect |
|---|---|
| `packages/web/server/lib/agent-tool/` | Doc section + exported `OPENCHAMBER_AGENT_TOOL_NAME`, interpolated into the generated plugin instead of a bare literal |
| Runtime behavior, all runtimes | **None.** No send path, request body, or persisted contract changes |

## Repository guidance

|---|---|---|
| `AGENTS.md` → "Enforce security and correctness in core/runtime logic" | The rejected fix would have written permission rules | Rejected precisely because it widened permissions; nothing ships |
| `AGENTS.md` → "Update owning documentation when contracts or invariants change" | The constraint is a real invariant future editors will hit | Recorded in the owning module's `DOCUMENTATION.md` |
| `openchamber-change-discipline` → smallest complete change | The first attempt grew across two packages | Reverted to docs plus one constant |
| `openchamber-change-discipline` → "Do not hide a required architectural migration behind a local heuristic" | This is exactly that: an upstream gap | Says so explicitly rather than shipping a partial client workaround |

## Validation

| Check | Result |
|---|---|
| `bun run type-check` (all 5 workspaces, incl. both VS Code projects) | Clean |
| `bun run lint` (all 5 workspaces) | Clean |
| `bunx vitest run server/lib/agent-tool/runtime.test.js` | 22 pass |
| `bun run docs:validate` | Passed — 450 pages, 45 sidebar links |
| Request capture vs. the real generated plugin | Documented each attempt's outcome in the table above |

**Not verified:** live Vertex traffic. The provider endpoint was a local recording proxy, so the 400 is reproduced by request *shape* rather than by Google's response. The permission-escalation finding was verified by simulating OpenCode's own evaluator against a restrictive agent, not by observing a suppressed prompt in the UI.

## Visual evidence

None applicable — no rendered output, copy, or interaction changes.

## Risks and failure behavior

The documented bug is unfixed and still reachable: any send to a non-tool-calling model fails until the upstream change lands, and the only user workaround remains a hand-written wildcard-deny agent.

The doc change carries no runtime risk. The exported constant is consumed only inside its own module and its test.

## Corrections to earlier revisions of this PR

Stated as fact earlier and wrong:

1. "`tools` does not filter one request." It does both — it is stored on the user message and filters that request, *and* replaces the persisted ruleset.
2. "No API exists to set session permissions." `session.update` accepts `permission`; I read the wrong SDK path.
3. "VS Code: unaffected." The VS Code webview is built from `packages/ui/src` and uses the same send path, so the gate would have run there against an OpenCode that never had the tool injected.
4. "The name cannot drift." Nothing bound the UI copy to the server constant — and a later commit did drift them, silently.
5. The end-to-end evidence was reported as showing the fix worked. It showed only that one declaration was absent; the same capture still carried 11 others.


IngTian:
Note for anyone reading the automated review above: it reviewed `64de5b93`, which is 7 commits behind the current HEAD (`c5b4bf30`). The code it evaluated has since been **reverted** — this PR is now documentation only (3 files, all under `packages/web/server/lib/agent-tool/`).

Its findings were the reason for the revert, and independent review found two further problems it did not catch:

- Gating only `openchamber` never fixed the bug. Vertex rejects on the presence of *any* `functionDeclarations`, and stock agents start from `"*": "allow"`, so ~11 built-ins still reached the provider. My original end-to-end evidence asserted only that `openchamber` was absent and I misread that as success.
- The obvious follow-up, `{"*": true}`, is a privilege escalation rather than a no-op: `Permission.evaluate` returns early on `allow`, so it converts an agent's `edit: deny` to `allow` and `bash: ask` to `allow`, suppressing the approval prompt.

Since suppressing tools must never widen permissions, and `{"*": false}` alone persists as a sticky deny, there is no safe client-side fix. The real fix is upstream and is now submitted: [anomalyco/opencode#41463](https://github.com/anomalyco/opencode/pull/41463) (the rejection) and [#41468](https://github.com/anomalyco/opencode/pull/41468) (the dropped image). This PR just records the constraint until a release carries them.

PR #2792, opened 2026-08-10
Fix/shell command entry caret

RyderAsKing:
## Intent

Fix shell-mode entry when typing a leading `!` in the chat composer.

Before this change, typing `!` could place the cursor before the trigger. Typing `ls` could then produce `ls!`, or manually moving the cursor could produce `!ls`.

`!` only activates shell mode; it should not be sent as part of the command. Both `ls!` and `!ls` fail because the shell receives the malformed command instead of `ls`.

The fix consumes the shell-mode trigger before command input begins, so typing `!` followed by `ls` reliably executes `ls`.

## Non-goals

- No changes to shell execution, command routing, or output rendering.
- No changes to literal `!` outside shell-mode entry.

## Affected surfaces

- `packages/ui/src/components/chat/ChatInput.tsx`
- Shared chat composer behavior across Web, Desktop, VS Code, and mobile.

## Repository guidance

| Guidance | How it applies |
|---|---|
| `AGENTS.md` | Shared UI change kept to the composer; no new dependencies or API changes. |
| `openchamber-change-discipline` | Minimal behavior fix with focused validation. |
| `performance-engineering` | Removes a deferred caret update from a keyboard interaction path; no performance claim or benchmark required. |
| `packages/ui/src/components/chat/composer/DOCUMENTATION.md` | Preserves the composer invariant that visible input is what gets sent. |

## Validation

| Check | Result |
|---|---|
| Focused composer tests | Pass — 58 tests |
| UI type-check and lint | Pass |
| Mobile web build | Pass |
| Manual desktop and mobile test: `!` → `ls` → Enter | Pass |

## Visual evidence

**Before:** `!` then `ls` can produce `ls!` or `!ls`, causing command-not-found.

https://github.com/user-attachments/assets/7f4666cd-34fd-4ff1-8419-e65a3fbdd5f6

**After:** `!` then `ls` runs `ls`.

https://github.com/user-attachments/assets/4c50b543-38b4-4d74-a086-42e4c577c04a

## Risks and failure behavior

- The change only affects a leading `!` in normal mode.
- `!` remains a literal character outside this shell-mode trigger.
- No server, API, persisted-data, or permission behavior changes.


RyderAsKing:
@openchamber-bot review

PR #2795, opened 2026-08-10
chore: ignore local opencode.json that overrides user plugins

herjarsa:
## Summary

Remove the committed `.opencode/opencode.json` (LSP-only config) and add `.opencode/opencode.json` + `.jsonc` to `.gitignore`.

## Problem

A committed `.opencode/opencode.json` is loaded by OpenCode **before** the user's global `~/.config/opencode/opencode.jsonc` whenever OpenChamber is run from this repository's working copy. Because the committed file only declares LSP servers and has no `plugin` section, it silently **overrides** the global config and drops:

- `plugin` entries (e.g. `oh-my-openagent`, `omniroute`, `agentmemory-capture`, `@herjarsa/omo-meta-governor`)
- MCP servers declared in the global config
- Custom provider credentials / model overrides

Result: users who clone `openchamber/openchamber` and run it locally see all of the above silently disabled — no error, no warning. Symptoms include `ProviderModelNotFoundError` for models defined in the global config and no hook injection from `oh-my-openagent` (verified via `~/.local/share/opencode/log/opencode.log` showing zero `claude-code-hooks` / `chat-message Goal set` events for runs spawned from this workspace).

The file was committed in `fd0730be7 feat: render LSP tool output in chat` (2026-06-03).

## Fix

1. `git rm .opencode/opencode.json`
2. Ignore `.opencode/opencode.json` and `.opencode/opencode.jsonc` going forward so this can't regress.
3. Add a comment in `.gitignore` explaining why these patterns exist, so a future contributor doesn't reintroduce the file.

Users who want the bundled LSP defaults can copy the deleted file from the previous release (`fd0730be7`) into their own global config or a non-ignored local path.

## Testing

- Verified locally: with the file removed and openchamber invoked from this workspace, the plugin `oh-my-openagent` now loads and `[chat-message] Goal set` + `[claude-code-hooks] Injecting` events appear in `~/.local/share/opencode/log/opencode.log` for sessions opened from this directory.
- The deleted file's contents are exactly 33 lines and contained only LSP defaults — no behavioural config is lost beyond those LSPs, which most users configure per-project anyway.

## Checklist

- [x] Diff is minimal (1 file deleted, `.gitignore` extended with 7 lines + comment)
- [x] Commit message follows `chore: ...` convention used elsewhere in the repo
- [x] No changes to source code or runtime behavior


btriapitsyn:
Thanks for the detailed report, but the diagnosis doesn't hold: OpenCode merges project config over global config, so a project file that only declares `lsp` doesn't remove global `plugin` or `mcp` entries — those keys simply aren't present to override anything. The committed LSP config is intentional for contributors working in this repo. If your global plugins genuinely stop loading when running from this checkout, that would be a config-merge bug in OpenCode itself and worth reporting upstream with logs; removing this file here would just mask it while dropping working tooling. Closing.

PR #2083, opened 2026-07-07
fix(chat): surface soft send failures instead of swallowing them (closes #2072)

herjarsa:
## Summary

Fix issue #2072 — message sends silently failing in long-context sessions, leaving sessions stuck on \`busy\` with STOP/REVERT/typing appearing unresponsive.

**Root cause:** \`ChatInput.handleSubmit\`'s \`.catch\` block classified soft network errors (timeout, failed-to-fetch, "may still be processing", "being processed", etc.) but returned without a toast when no attachments were present. The optimistic message was already rolled back, the input cleared, no notification surfaced — meanwhile the message could STILL be processing server-side, so \`session.status\` correctly reported \`busy\`. The user saw no feedback explaining why controls were unresponsive. Same silent-swallow pattern in \`useQueuedMessageAutoSend\`.

**Fix:** Route every send error through a new pure utility \`handleSendError\` that ALWAYS surfaces a toast for soft network errors, with copy explicitly warning the message may still be processing server-side. Persists the failed text to \`sessionStorage[\`openchamber_last_failed_send\`]\` for recovery.

## Changes (5 atomic commits)

1. **feat(chat): add shared send-error handler utility** — pure util with injected toast, t, saveRecovery, onRestoreAttachments.
2. **fix(chat): surface soft send failures instead of swallowing them** — ChatInput.handleSubmit now delegates to the shared handler.
3. **fix(chat): surface queued auto-send failures via shared handler (#2072 Path 2)** — useQueuedMessageAutoSend catches now toasts a queued-failed notification.
4. **test(chat): cover #2072 silent failure fix** — restores and rewrites the upstream reproduction test against the real handler.
5. **feat(i18n): add \`chat.sendError.*\` toast strings** for the new copy across all 10 locales.

## Path coverage

- **Path 1** (ChatInput.handleSubmit) — ✅ fixed
- **Path 2** (useQueuedMessageAutoSend) — ✅ fixed
- **Path 3** (materialization limit) — documented as regression marker, out of scope
- **Path 4** ("may still be processing" silent) — ✅ fixed
- **Path 5** (SSE timing race) — documented as regression marker, out of scope

## Verification

- \`bun test src/lib/chat/sendErrorHandler.test.ts src/components/chat/__tests__/sendMessageSilentFailure.test.ts src/hooks/useQueuedMessageAutoSend.test.ts\`: **37 / 37 pass**, 86 expects
[...]

PR #2799, opened 2026-08-10
fix(chat): surface soft send failures instead of swallowing them (closes #2072)

### LINKED ISSUE #2072: [Bug] New messages often don't send in long-context sessions

herjarsa:
## Summary

Rebase of #2083 onto current `main`. The original PR was authored against an older snapshot of `main` and accumulated 159k lines of unrelated changes (the merge conflict blocked resolution via the GitHub UI).

This PR contains the **5 logical commits** of the original fix:

1. `feat(chat): add shared send-error handler utility` — introduces `sendErrorHandler`
2. `fix(chat): surface soft send failures instead of swallowing them (closes #2072)` — routes ChatInput errors through the handler
3. `fix(chat): surface queued auto-send failures via shared handler (#2072 Path 2)` — same for the queued auto-send hook
4. `test(chat): cover #2072 silent failure fix (#1, #2, #4 paths)` — coverage
5. `feat(i18n): add chat.sendError.* toast strings for long-context send failures` — i18n

## Resolutions applied during cherry-pick

- `ChatInput.tsx`: took the PR version (uses shared `handleSendError` helper, replacing the inline `isSoftNetworkError` branches that existed on current `main`). The original PR's intent is the shared helper, so the inline branches in `main` are exactly what #2083 is supposed to replace.
- `useQueuedMessageAutoSend.ts`: same pattern — took the PR version (calls `handleSendError`).

Both resolutions match the design of the original PR. The remaining hunks were clean automerges.


herjarsa:
Closing as unmergeable. The PR references prepareLocalAttachments (from @/sync/input-store) and isDocumentAttachmentFilename (from @/sync/attachment-files), neither of which exists in origin/main or anywhere else in the codebase. Rebase fails with TS2305: Module has no exported member.

The PR only touches ChatInput.tsx (+3710/-1013) and assumes helper functions were added in another PR that has not landed (or was reverted). The PR is not in a mergeable state without first adding those helper functions, which is scope creep beyond conflict resolution.

Reference: PR #2885 and #2886 were closed for the same kind of pre-existing structural dependency mismatch.

PR #2802, opened 2026-08-10
fix(mcp): filter out purely numeric server names from status map

herjarsa:
Cherry-pick of #1888 (5 commits) onto current `main`. Original PR was authored against an older snapshot of `main` and couldn't auto-merge.

Resolutions applied:
- `packages/ui/src/components/mcp/McpDropdown.tsx` — took the PR version (the numeric-name filter).
- `packages/ui/src/hooks/useChatAutoFollow.ts` — took the PR version (related fix in the same series).

The remaining 3 commits (timeoutMs cleanup, filteredCount=0, i18n keys) merged cleanly.

Closes the original intent of #1888.


btriapitsyn:
Thanks for the rebase effort, but I'm closing this along with the original #1888. Thousands of numeric keys in the status map mean the response shape is being mishandled where the store ingests it (`useMcpStore` casts `result.data` without validating it's a plain object), not that the dropdown needs a cap and a numeric-name filter. A small PR that validates the `mcp.status` response shape at the store boundary would fix the freeze for every consumer at once and would be welcome. As it stands, this PR also reverts the dropdown's recent card grouping, adds an unused `DeferredMount` export, and inserts untranslated English strings into all locale files, which I'd rather not take on.

PR #2822, opened 2026-08-11
fix(relay): never forward a tunneled body that lost frames

claymor333:
## Intent

When the relay drops mid-request, the `prompt_async` body frames can be lost. The tunnel host previously forwarded the request to loopback as an **empty/truncated chunked body**, which the loopback server rejects with a bare `400` (empty response body) — the "Failed to send message (400)" seen from the mobile APK.

## Change

- **Host buffers bodies** (`packages/web/server/lib/relay/tunnel-host.js`): request bodies <512KB are buffered and forwarded as a **complete** body only once `StreamEnd` arrives; larger bodies still stream live. `stream.body = null` after forwarding so late frames can't enqueue.
- **`hasBody` request-head flag** (`protocol.ts` + `tunnel-client.ts`): the client signals whether it had a body. The host uses it to distinguish a genuine bodyless request from one whose body frames were lost — zero `HttpBody` frames before `StreamEnd` with `hasBody=true` aborts as an **ambiguous transport failure** (which the client already retries) instead of forwarding an empty body.
- **15s body-delivery deadline**: a stalled tunnel converts into a clean abort rather than hanging the stream.

## Non-goals

- No changes to keepalive, reconnect pacing, frame batching, or the host→client response path (verified response-path-neutral).
- No wire-format/codec changes; `cross-compat.test.js` unchanged and green.

## Affected surfaces

- **Relay tunnel host** (`packages/web/server`) — request-body forwarding to loopback.
- **Relay tunnel client** (`packages/ui`) — adds `hasBody` to the request head.
- **All runtimes** (web/desktop/mobile) share the same relay client; behavior is identical since the flag is optional.

## Repository guidance

|---|---|---|
| `relay-transport` skill | Change is inside the relay tunnel (host + client) | Wire/codec untouched; body-delivery aborts are client-retryable; no new WS endpoints; no `window.location.origin` dependence; host never injects credentials |
| `openchamber-change-discipline` skill | Source + shared-contract change | Smallest complete change; owning `DOCUMENTATION.md` updated; failure/rollback behavior explicit (abort vs forward); unchanged host→client path left untouched |

## Validation

| Check | Result |
|---|---|
| `bun test packages/web/server/lib/relay/tunnel-host.test.js` | Added 116 lines covering buffering, zero-frame abort, deadline, live-stream switch |
| `node --check` on patched `tunnel-host.js` | OK |
| Mobile debug APK (versionCode 181) built and installed on Android | Relay fix confirmed present in bundled assets (`hasBody`) |
| Server-side fix applied to installed `@openchamber/web` v1.18.2 host, restarted | Live |
| End-to-end mobile send through the relay after re-pairing | No recurrence of "Failed to send message (400)" |

**Not verified:** direct LAN transport (only the relay path was exercised on the mobile device); CI did not run on this branch.

## Visual evidence

No visible UI change. The diff is confined to relay request-body handling on the host and one optional field on the client request head; rendered behavior is unchanged.

## Risks and failure behavior

- A body that legitimately arrives slowly (>15s) is aborted — acceptable; the client treats it as an ambiguous retryable failure.
- Buffered bodies cap at 512KB; larger uploads stream live with the old semantics, so no unbounded memory growth.
- Late frames after forwarding are discarded via `stream.body = null`, preventing partial-body corruption.
- Backward compatible: hosts/clients that predate `hasBody` behave as before (flag is optional).


btriapitsyn:
Great find and a clean fix — buffering small bodies until StreamEnd and signaling `hasBody` is exactly the right shape for this, and the write-up made it easy to review. Tests and cross-compat pass locally, type-check is clean. Merging — thanks a lot for digging into a relay bug this deep! 🙌

PR #2827, opened 2026-08-11
feat(scheduled-tasks): archive successful run sessions

### LINKED ISSUE #2823: [Feature Request] Auto-archive completed scheduled task sessions

Scheduled tasks create a new OpenCode session for every run. For recurring jobs, those sessions accumulate in Recents even though the scheduled task already retains the useful run metadata and links to the session.

Please add a per-task lifecycle option in the scheduled task editor:

- **Archive run session when finished**
- Disabled by default for backward compatibility
- Applies to both scheduled execution and **Run now**
- Archives only the session created for that run
- Keeps the session recoverable through the Archived view

This should be implemented by the scheduled-task runtime after the run reaches a terminal state, rather than requiring the prompt or agent to invoke an archive tool. It should work for tasks created from any project scope, including tasks rooted at the user home directory.

A useful extension would be a small policy selector:

- Keep run sessions
- Archive successful runs
- Archive all completed runs

The task record should continue to show last status, duration, and its session link after archival. Goal-enabled tasks should archive only after the goal reaches its terminal state.

#### Acceptance criteria

- A scheduled task can opt into automatic run-session archival.
- The option persists with the task definition.
- Scheduled and manual runs follow the same policy.
- Only the run session is archived; ordinary user sessions are unaffected.
- Archived runs remain accessible from task history and the Archived view.
- Tasks from non-repository directories such as `~/` work correctly.
- Archive failure does not overwrite the actual task result; it is surfaced separately.

### Extra context (optional)

The scheduled-task docs state that every run starts a new session and that OpenChamber retains a link to the created session: https://github.com/openchamber/openchamber/blob/main/packages/docs/content/docs/scheduled-tasks.mdx

OpenChamber already has general age-based session retention, but that solves a different problem and cannot archive immediately on task completion: https://github.com/openchamber/openchamber/blob/main/packages/ui/src/components/sections/openchamber/SessionRetentionSettings.tsx

Our current workaround appends a `session_self_archive` instruction to every scheduled prompt and keeps a janitor fallback. That is fragile: prompts can exit early, agents may not have the tool loaded, and lifecycle cleanup should not consume agent attention or tokens. The scheduler already owns session creation and terminal run state, so it is the reliable place to own archival too.

chintan-jariwala:
## Intent

Add an opt-in **Archive successful runs** lifecycle setting for scheduled tasks. After a successful run, OpenChamber waits for the run session to become quiescent and archives it; goal-enabled tasks additionally wait for authoritative goal completion. Failed, blocked, or canceled runs remain visible for diagnosis.

Archival is recoverable rather than destructive: the task retains its run result, duration, and session ID, and the task list now exposes a **Last run** action that can reopen the archived session. Archive failures do not overwrite a successful run result; they surface as a separate warning and retry with bounded backoff.

Closes #2823.

## Non-goals

- A multi-option retention policy selector. V1 is a disabled-by-default boolean.
- Archiving failed or canceled runs.
- Deleting sessions or changing the Archived view's retention behavior.
- Adding `archiveOnSuccess` to portable `.agents/loops/*.md` frontmatter; it remains a UI/JSON-only execution field.
- Replacing prompt-level self-archive tooling for workflows that need custom grace periods.

## Affected surfaces

- `packages/web`: scheduled-task orchestration now observes OpenCode lifecycle events, verifies session/child/message quiescence with polling recovery, archives through the official SDK, and persists bounded restart-recovery records.
- `packages/ui`: the shared scheduled-task contract includes lifecycle/archive state; the editor adds a localized **Lifecycle** section; the task list shows archive warnings and a **Last run** action.
- `packages/docs`: scheduled-task behavior and failure semantics are documented.
- Persisted project config: adds optional `execution.archiveOnSuccess`, `state.pendingArchives`, and `state.lastArchiveError`. Existing tasks require no migration because missing fields preserve the disabled/default behavior.
- Web and Desktop use the same server-owned implementation. Other shared-UI clients receive the same task contract when connected to an OpenChamber server; no native privilege or runtime bridge was added.

## Repository guidance

|---|---|---|
| `AGENTS.md` + `openchamber-change-discipline` | This changes executable source, a shared UI contract, and persisted state. | Kept ownership in the scheduled-task runtime/config modules; used optional backward-compatible fields; added stale-write, restart, failure, ordering, race, and retry tests; ran workspace and runtime validation. |
| `ui-api-decoupling` + implementation map | The server invokes official OpenCode session APIs and the shared UI calls OpenChamber routes. | Session status/messages/children/update calls use `@opencode-ai/sdk/v2`; the UI keeps using `runtimeFetch` through `scheduledTasksApi`; no hardcoded runtime URL or credential path was introduced. |
| `theme-system` + token reference | The task list adds a warning state and action; the editor adds a control. | Reused `Checkbox`, `Button`, `Icon`, semantic borders, and the existing `toneStyle('warning')`; no palette colors or new icon assets. |
| `locale-ui-patterns` | The lifecycle section adds visible and accessible text. | Added semantic keys with real translations in every shipped locale and kept all component copy behind `useI18n()`. |
| `settings-ui-patterns` + control/layout references | The scheduled-task editor is a configuration dialog with a boolean lifecycle control. | Followed the dialog's existing checkbox/section precedent and shared control sizing; no Settings search item applies because this is a task-specific dynamic editor, not a stable Settings page control. |
| `packages/web/server/lib/scheduled-tasks/DOCUMENTATION.md` | This module owns scheduling, persistence, and execution invariants. | Documented event-plus-poll recovery, archival ownership, persisted runtime fields, and the UI-only loop-field boundary. |

## Validation

| Check | Result |
|---|---|
| `HOME=.../openchamber-test-home bun run test -- server/lib/scheduled-tasks/runtime.test.js server/lib/projects/project-config.test.js` from `packages/web` | **Pass:** 45 tests. Covers ordinary/goal completion, missed events, child sessions, restart recovery, transient fetch/archive failures, goal cancellation/resumption, stale editor writes, and synchronization races. |
| `bun run type-check` | **Pass:** mobile, UI, web, Electron, and root workspaces. |
| `bun run lint` | **Pass:** all workspaces. |
| `bun run build` | **Pass:** all workspaces; only the repository's existing large-chunk warning. |
| `bun run docs:validate` | **Pass:** 450 pages and 45 sidebar links. |
| `bun run dead-code` | Completed and inspected; reports the repository's existing unused/duplicate-export inventory, with no new source file or export introduced by this change. |
| Full `packages/web` suite with isolated `HOME` | **Partial baseline:** 132/134 files passed; 1200 tests passed, 1 skipped. The two failures reproduce on upstream `main`: `server/sse-routes.test.js` lacks `app.patch` in its app stub, and `src/api/git.test.ts` lacks `getGitRangeDiff` in its mock. |
| Final live server smoke | **Pass:** created and ran a temporary task rooted at `~/`, received `FINAL_HEAD_SMOKE_OK`, observed persisted success, polled task listing during completion, verified session `ses_00d8d1b05ffe1RMAQbYLmoKuRM` was archived and `pendingArchives` cleared, then deleted the task. |

Not verified: packaged installers and VS Code/native-mobile interaction, because the behavior is server-owned and no platform bridge changed.

## Visual evidence

The editor uses the existing theme/control system, so the same implementation covers light/dark themes. These captures show the meaningful layout states: upstream before, wide desktop after, and narrow-width after.

| Upstream before | Desktop after |
|---|---|
| Scheduled task editor before the lifecycle control | Scheduled task editor with the Lifecycle section and Archive successful runs control |

**Narrow-width after**

Narrow scheduled task editor with the Lifecycle section

## Risks and failure behavior

- **Compatibility:** all new config fields are optional and normalized. Existing tasks remain opted out. Runtime state is server-owned, so saving a stale editor draft cannot erase an in-flight archive record.
- **Failure:** session or archive API failures retain the pending record, show a separate warning after repeated failure, and retry with exponential backoff capped at 60 seconds. A run's successful status is not rewritten by archival failure.
- **Ordering/races:** archival requires two quiet verification passes, rechecks parent/child status and goal metadata before mutation, survives missed lifecycle events through polling, and preserves pre-persistence trackers across task-list synchronization.
- **Restart/cleanup:** pending records are capped at 20 per task and resume after server restart. Completed/canceled goal metadata and already-archived sessions clean up their records. Runtime stop clears timers and in-memory trackers.
- **Data loss:** sessions are archived, never deleted, and remain reachable from **Last run** and the Archived view. Failed or canceled sessions are intentionally left unarchived.
- **Security:** no new credentials, URLs, dependencies, or privilege boundary. OpenCode authentication remains owned by the existing SDK client setup.
- **Performance:** polling exists only while opted-in sessions are pending; normal verification uses a 2-second quiet interval and failures back off to at most once per minute.


btriapitsyn:

btriapitsyn:
Closing: automatically archiving run sessions isn't a direction we want for session lifecycle. Sessions are the user's record of what happened, and hiding them automatically — even recoverably — isn't behavior we want to ship or maintain. This is a product decision, not a comment on the implementation.

PR #2829, opened 2026-08-11
fix(preview): stop probing dead upstreams forever, surface unreachable state

herjarsa:
## Problem

A preview tab whose upstream origin is dead (e.g. a dev server that was stopped, or a stale `http://127.0.0.1:<port>` URL persisted across restarts) makes the preview proxy retry **forever**:

- The proxy answers every request with a plain `502` and **no** `x-openchamber-preview-target-error` header, so the renderer cannot tell "upstream unreachable" apart from "upstream still starting".
- The renderer therefore keeps retrying with backoff (250ms → 2s) indefinitely, and the iframe is remounted on every retry — producing a steady stream of
  `[preview-proxy] proxy error: connect ECONNREFUSED 127.0.0.1:5180` log lines (~2/s) that resumes after every app restart.

## Fix

- **server** (`packages/web/server/lib/preview/proxy-runtime.js`): when the proxy fails with a connection-level error (`ECONNREFUSED`, `ECONNRESET`, timeouts, host-not-found, …), emit `x-openchamber-preview-target-error: unreachable` alongside the existing `502`.
- **ui** (`packages/ui/src/lib/preview/proxy-response.ts`): recognize the new `unreachable` code and map it to a new `retry-grace` recovery action — never re-register the target, since the target itself is valid.
- **ui** (`packages/ui/src/components/layout/ContextPanel.tsx`): during the existing 15s startup grace window, keep the current backoff so dev servers that are still binding aren't affected; afterwards settle on the already-rendered "Dev server is not responding" state (with the manual **Retry** button) and **stop probing automatically**.

## Testing

- `packages/ui`: `bun test src/lib/preview/proxy-response.test.ts` — 4 pass (new case: `unreachable` → `retry-grace` regardless of attempt count)
- `packages/web`: `vitest run server/lib/preview/proxy-runtime.test.js` — 37 pass (new cases: connection errors → `unreachable` header; non-connection proxy errors → no header)
- `tsc --noEmit` (5.9.3) and eslint clean on all touched files

### DIFF
--- a/packages/ui/src/components/layout/ContextPanel.tsx
+++ b/packages/ui/src/components/layout/ContextPanel.tsx
@@ -1174,6 +1174,26 @@ const PreviewPane: React.FC<PreviewPaneProps> = ({ rawUrl, onNavigate }) => {
         response.headers,
         proxyRecoveryAttemptedKeyRef.current === proxyCacheKey,
       );
+      if (recoveryAction === 'retry-grace') {
+        // The proxy signals the upstream origin is not accepting connections
+        // (ECONNREFUSED & friends). The target itself is valid — the origin
+        // may be a dev server that is still binding — so keep retrying with
+        // backoff during the startup grace window, then settle on a stable
+        // "unreachable" state (rendered with a manual retry button) instead
+        // of probing a dead origin forever.
+        const startedAt = upstreamProbeStartedAtRef.current || Date.now();
+        const elapsed = Date.now() - startedAt;
+        if (elapsed < PREVIEW_STARTUP_GRACE_MS) {
+          setUpstreamState('starting');
+          upstreamProbeAttemptRef.current += 1;
+          const attempt = upstreamProbeAttemptRef.current;
+          const delay = Math.min(2000, 250 * Math.pow(2, Math.min(4, attempt)));
+          scheduleRetry(delay);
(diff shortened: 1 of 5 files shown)


herjarsa:
Superseded by #2883 (feat(browser): replace the preview proxy with a real browser panel and an agent web tool).

The entire preview-proxy subsystem this PR modified was removed from main: commit cc9249d93 deleted packages/ui/src/lib/preview/* and packages/web/server/lib/preview/*, replacing them with the BrowserPane architecture (packages/ui/src/components/browser/BrowserPane.tsx + packages/ui/src/lib/browser/*).

The bug this PR fixed — probing dead upstreams forever with no way to distinguish 'unreachable' from 'still starting' — is resolved by design in the new panel:
- Retries are bounded by DEV_SERVER_WAIT_MS (40s recovery budget per URL), after which the waiting state clears.
- isStartingServerFailure() distinguishes startup errors from real failures: only genuine startup failures enter the waiting state; unreachable upstreams surface the failed navigation state immediately.
- Both states are surfaced in the UI with localized copy (contextPanel.browser.waitingForServer).

No conflict resolution is possible (the modified files no longer exist on main), and porting the fix would be redundant. Closing as superseded.

PR #2843, opened 2026-08-12
fix(ui): stable chat seam anchor during panel resize; single width writer + 120ms reduced-motion animation

xuexi-xiaobai:
## Intent

Stabilize the chat seam while the left sidebar / right ContextPanel changes width, and restore a visible open/close animation under `prefers-reduced-motion`.

The previous behavior:

- Sidebar/ContextPanel wrote the width from React inline styles **and** ran CSS width transitions, while the message list had no anchor during the change — dragging or toggling a panel let the first visible user question drift (or the viewport jump), especially across wrap boundaries (768/1024) and with long messages.
- Under `prefers-reduced-motion: reduce` the programmatic open/close collapsed to a single-frame width write — the panel "popped" instead of animating.

What this PR does:

- Introduces a single global resize transaction (`lib/resizeInteraction.ts`) and a **single geometry writer** (`usePanelResize`): React no longer writes panel widths or CSS transitions; the hook owns the width CSS variables and animates open/close / mode switches / expand through one per-frame writer (CSS var write → `notifyResizeFrame` → anchor commit, all before paint).
- Anchors the conversation seam during the transaction (`useConversationResizeAnchor` + `conversationResizeLayout`/`conversationResizeAnchor`): the first visible user-question top stays stable (frame-delta model, zero writes when nothing changed), with a per-transaction held range.
- Motion profiles: **standard 200ms easeOutCubic; reduced motion 120ms easeOutQuad** — reduced motion is a short visible animation, never a single-frame jump. The preference is a live MediaQueryList subscription; a change mid-animation does not alter the running animation.
- Persistence is pointer-up only (`onUserCommitWidth`): programmatic open/close, mode switches and parent-layout follows never write user settings.
- Adds the `suppressScrollAdjustments` option to the existing `@tanstack/virtual-core` 3.17.3 bun patch so the anchor controller is the only scroll writer during a transaction.

## Non-goals

- No new dependencies (the virtual-core patch extends the existing pinned version).
- No change to persisted settings schema or IDs.
- No change to the VS Code / mobile shell layouts (they do not consume this shared desktop/web layout path).

## Affected surfaces

- `packages/ui` (desktop + web): `Sidebar`, `ContextPanel`, `MessageList`, `ChatContainer`, `ChatInput`, `ChatMessage`, `TurnItem`, `TurnAssistantBlock`; new modules `resizeInteraction`, `usePanelResize`, `useConversationResizeAnchor`, `conversationResizeLayout`, `conversationResizeAnchor`, `performanceTrace`; bun patch `bun-patches/@tanstack+virtual-core+3.17.3.patch`.
- Runtime scope: desktop/web `MainLayout` only. VS Code and the mobile shells have their own layouts and do not consume this registry — unchanged.
- No persisted/external contract changes.

## Repository guidance

|---|---|---|
| AGENTS.md `openchamber-change-discipline` | Any source/module-ownership change in `packages/ui` | Modules are placed in their owning paths; the width-writer contract is documented in `packages/ui/src/lib/surfaces/DOCUMENTATION.md`; no unrelated refactors |
| AGENTS.md `performance-engineering` | This is a jank/lag/anchor-drift fix on a per-frame render hot path | Removed the per-pixel ResizeObserver re-render (ref-based follow loop); at most one width write per animation frame; frame-count assertions in tests |
| `packages/ui/src/lib/surfaces/DOCUMENTATION.md` | Nearest module doc for the layout-animation contract | Added the "Layout animation invariant" section (single width writer, motion profiles, reduced=120ms, seam anchor) |
| CONTRIBUTING.md PR contract | Review handoff requirements | Intent/non-goals/affected surfaces/validation/risks documented here with current-HEAD evidence |

## Validation

| Check | Result |
|---|---|
| `bun run type-check` (packages/ui) | Pass |
| `bun run lint` (packages/ui) | Pass |
| `bun run build` (packages/web vite build) | Pass (web assets built during packaging) |
| Unit tests (4 suites) | **81/81 pass**: `usePanelResize.test.tsx` (34 — incl. drag lifecycle, programmatic targets, round-10 single-writer/follow/shared-transaction, round-11 motion-profile suites), `resizeInteraction.test.ts` (14), `conversationResizeAnchor.test.ts` (17), `conversationResizeLayout.test.ts` (16) |
| Reduced-motion animation (unit-level) | Tests assert ≥4 distinct intermediate widths, first applied frame ≠ final frame, ~120ms completion — a single-frame jump fails these |
| `bun run dead-code` | No new items |
| Manual acceptance | Verified on the packaged desktop build: 120ms open/close animation visible under system reduce; sidebar/panel drag anchors the question top; right-panel proportional follow; no regressions in upstream 1.18.2 features (lazy views, single-iframe chat, walkthrough, Escape→terminal) |

Not verified here: VS Code / mobile runtimes (untouched by this change); CI runs the workspace-wide checks.

## Visual evidence

This is a motion/anchor change. The before (upstream) and after (this branch) desktop builds cannot be run side by side on the verification machine: the app server binds a fixed port and the installed upstream instance was in use during validation. The motion behavior is instead asserted deterministically by the unit tests (reduced-motion suites prove intermediate widths exist and the first frame differs from the final frame — i.e. no single-frame jump; drag suites prove the seam anchor key and DOM reference stay constant through the transaction). Static screenshots of the packaged build's sidebar states can be added on request.

## Risks and failure behavior

- **Virtual-core patch**: `suppressScrollAdjustments` is additive on the already-pinned 3.17.3 patch. If a consumer's lockfile does not apply the patch, the option is silently ignored (TanStack's `options` type accepts unknown keys at runtime) and the anchor controller would compete with TanStack's own scroll writes; the anchor suites fail in that case, so CI catches it.
- **Width persistence**: programmatic operations never call `onUserCommitWidth` (pointer-up only), so toggles/mode switches cannot clobber saved widths or the manual-resize flag. Reverting the PR restores the previous behavior exactly (no data migration).
- **Parent-layout follow**: opens a short transaction per real width change when no transaction is active; stale/failed frames are bounded by the transaction fail-safe and release on user intent.
- Cross-runtime: no VS Code/mobile behavior is touched; the shared modules only run under the desktop/web layout path.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+- Chat: the conversation no longer jumps while the sidebar or context panel is being resized; the first visible message stays put during panel drags.
+- UI: sidebar and context panel open/close now animate smoothly instead of appearing in a single frame, with a shorter, softer animation when reduced motion is enabled.
 - Usage: quota limits enabled for display now refresh every three minutes on desktop, mobile, and VS Code, with a manual refresh action available at any time.
 
 ## [1.18.2] - 2026-08-10
--- a/bun-patches/@tanstack+virtual-core+3.17.3.patch
+++ b/bun-patches/@tanstack+virtual-core+3.17.3.patch
@@ -16,6 +16,37 @@ index 52ae6ca12f8d1c650ee7f1bd55573ee7d4f8b65f..bcee09df7377c37ffb220606b741c9ff
            lanes,
            // Pass the typed array so binary search + forward-walk can read
            // start/end directly from Float64Array, skipping the Proxy traps.
++@@ -850,5 +850,10 @@
++      if (delta !== 0) {
++        // OpenChamber patch: during a panel-resize transaction the anchor
++        // controller is the ONLY scroll writer. suppressScrollAdjustments
++        // stops both the end-anchor ("keep at bottom") and the above-viewport
++        // item-size-change compensation branches from writing scrollTop here.
++        const suppressAdjustments = this.options.suppressScrollAdjustments === true;
++        const wasAtEnd = !suppressAdjustments && this.options.anchorTo === "end" && ((_a = this.scrollState) == null ? void 0 : _a.behavior) !== "smooth" && this.getVirtualDistanceFromEnd() <= this.options.scrollEndThreshold;
++        const prevTotalSize = wasAtEnd ? this.getTotalSize() : 0;
++        const shouldAdjustScroll = !suppressAdjustments && ((_b = this.scrollState) == null ? void 0 : _b.behavior) !== "smooth" && (this.shouldAdjustScrollPositionOnItemSizeChange !== void 0 ? this.shouldAdjustScrollPositionOnItemSizeChange(
+           // The callback expects a VirtualItem; build one lazily only
+           // when the consumer actually supplied a custom predicate.
+           this.measurementsCache[index] ?? {
+             index,
+             key,
+             start: itemStart,
+             size: cachedSize,
+             end: itemStart + cachedSize,
+             lane: 0
+           },
+           delta,
+           this
+         ) : (
+           // Default: adjust when the resize is an above-viewport item.
+           // First measurement (!has(key)): always adjust — the item
+           // has never been sized, so the estimate→actual delta must
+           // be compensated regardless of scroll direction.
+           // Re-measurement (has(key)): skip during backward scroll
+           // to avoid the "items jump while scrolling up" cascade.
+           itemStart < this.getScrollOffset() + this.scrollAdjustments && (!this.itemSizeCache.has(key) || this.scrollDirection !== "backward")
+         ));
 diff --git a/dist/esm/index.js b/dist/esm/index.js
 index 3032c0ca457582be3f47923cba1f7d92c848745c..90b574881a073aabac99c075f7eab0a8f363fff6 100644
 --- a/dist/esm/index.js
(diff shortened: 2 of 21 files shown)


btriapitsyn:

xuexi-xiaobai:
Thank you for the thorough review. Addressed in bf542cbe2:

- **Finding 2 (fixed):** the parent-layout follow loop's riding branch now joins the active global transaction before writing frames — frames carry the shared transactionId (no longer dropped by the anchor controller as `transactionId: 0`), and the follow source registers for the shared finalize. Joining shares the current id without re-firing transaction-start participants. Added two two-panel regression tests: follow-during-drag frames carry the drag's transactionId with no re-capture, and a follow that detaches mid-drag releases its part while the other drag still finalizes the transaction.
- **Finding 4 (fixed):** added the `[Unreleased]` CHANGELOG entries for the seam-anchor fix and the open/close animation (reduced motion = 120ms).
- **Finding 5 (fixed):** removed all four dead marker attributes (`data-chat-wrap-surface`, `data-chat-message-id`, `data-message-scroll-container`, `data-vlist-size-container`) and the comment referencing the probe this PR does not add. The anchor controller reads its markers (`data-chat-turn-flow-start`, `data-chat-user-bubble`, `data-chat-assistant-block`, `data-chat-resize-row`, `data-chat-composer-occlusion`) through refs passed down from MessageList, not querySelector.
- **Finding 3 (question for you):** the 120ms reduced-motion profile is deliberate — the previous instant transition read as the panel "popping" in a single frame, and 120ms easeOutQuad was the shortest duration that still read as a continuous motion (it is documented in `surfaces/DOCUMENTATION.md` and covered by tests). Happy to switch back to an instant transition if you or the a11y side prefer strict single-frame behavior under `prefers-reduced-motion`.
- **Finding 1 (partial):** the frame-level evidence I have today is the unit suites (intermediate widths, frame counts, zero-write frames, transaction sharing) plus manual verification on the packaged desktop build. I will follow up with a drag/anchor recording and before/after measurements; noting honestly that it is not attached yet.

Also in the same commit: the resize test DOM stub now (re)defines globals with `Object.defineProperty` + original-descriptor restore, so a read-only global left behind by an unrelated test file can no longer take the whole resize suite down when file order changes.

PR #2851, opened 2026-08-12
fix(web): drop small-model text written in a script the source never used

linnefromice:
## Intent

A small-model answer written in the wrong language could reach the user even though a guard existed to stop it.

`session-assist` (chat recap + suggestion chip) and `session-goal` (audit note) already compared the generated text's writing system against the source text, because models ignore the "write in the same language" instruction — the code's own comment names provider-side account personalization as the cause (`session-goal/runtime.js`, pre-change lines 115-117: *"account-side personalization can leak a different language despite the instruction — same issue session-assist hit"*).

That guard bundled Kana, Han **and Hangul** into a single "CJK" class:

```js
const hasCjk = (text) => /[぀-ヿ一-鿿가-힯]/.test(text);
```

For a Japanese conversation the class is present in the source (Kana), and for a Korean answer it is present in the output (Hangul), so `hasCjk(output) && !hasCjk(source)` evaluated to `false` and the Korean text was shown. This was observed in practice: Korean recap text in a Japanese session.

**What changes**

1. One dependency-free module owns the script-class list: `small-model/script-guard.js`, exporting `hasScriptMismatch(text, referenceText)`. **Hangul is now its own class**; Kana and Han stay together because Japanese and Chinese share Han, and splitting them would drop legitimate Japanese output whose source happened to carry no Han.
2. Both existing guards now call it. `session-assist` gains Devanagari and Arabic coverage as a side effect of sharing the list (it previously tracked only Cyrillic and the combined CJK class; `session-goal` already had all four).
3. `session-goal/create.js` gains the guard it never had. When a goal objective exceeds 5,000 characters it is distilled by the small model, and a wrong-language distillation silently becomes the text the progress auditor judges against for the rest of the goal. A dropped distillation now falls through to the pre-existing head/tail trim of the user's original text.

**What a user sees:** wrong-script text disappears. Nothing is regenerated — a Japanese session does not start getting Japanese recaps as a result of this PR. Users who want a specific language from the small model still need Settings → Sessions → Small Model.

## Non-goals

- **Making the model answer in the right language.** This is a suppression guard. Prompt-side mitigation already exists in two layers (an explicit "ignore other language preferences" instruction, and instructing by a 200-char sample of the real conversation rather than by description).
- **Distinguishing Japanese from Chinese.** Chinese uses only Han, which a Japanese source also contains, so a presence check cannot separate them. Pinned as `false` in `script-guard.test.js` so a later change does not "fix" it by splitting Han and start dropping valid Japanese output.
- **Latin-script mismatches** (English vs German, or an English answer to a Japanese conversation). Out of reach for a script check; that needs language identification, which is a different and much larger change.
- **An explicit language setting** for recap/suggestion/note. Those deliberately follow the conversation's language; adding a setting would collide with that design. `walkthrough` already has such a setting and is untouched.
- **Widening the Unicode ranges** — Hangul Jamo (`1100-11FF`, `3130-318F`), half-width Katakana, Han Ext A/B, Cyrillic Supplement, Greek/Hebrew/Thai are still untracked. All are fail-open (they cannot cause a false drop) and all are inherited from the pre-change list, so they are follow-up material rather than part of this fix.
- **The `warn` helper's log format.** `session-goal/create.js`'s new warning renders with a trailing `undefined` because `warn` always forwards `error?.message || error`. Pre-existing format; changing it would either introduce two log shapes in one file or touch an unrelated helper.

## Affected surfaces

**Packages:** `packages/web` only, and only under `server/lib/`:

| Module | Change |
|---|---|
| `small-model` | New `script-guard.js` + tests; new `DOCUMENTATION.md` entry. One new named export. |
| `session-assist` | Local `hasCyrillic`/`hasCjk`/`scriptMismatch` deleted, replaced by the shared import. |
| `session-goal` | Same replacement in `runtime.js`; new guard in `create.js`. |

**Runtimes:**

- **Web and Desktop** — both run the OpenChamber server (Desktop in-process), so both get the fix.
- **VS Code** — extension-only, no web server, so it never generates assist/goal payloads; it only renders payloads produced by a web/desktop instance via `session.updated` (`session-assist/DOCUMENTATION.md` → Limitations). It therefore needs no change and simply stops receiving wrong-script payloads. No extension code touched.
- **Mobile (hosted PWA and Capacitor)** — connects to an existing OpenChamber server, so it inherits the server-side behavior with no client change.

**Persisted / external contracts:** none changed. `metadata.openchamber.assist` and `metadata.openchamber.goal` keep their shapes; a dropped field is written as `''`, exactly as the pre-change guard did. No routes, settings keys, IDs, or generated assets are touched. No dependency or lockfile change.

**User-visible states:** the absence of a recap, a suggestion chip, or an audit note — all of which are pre-existing states the UI already handles (see Visual evidence).

## Repository guidance

|---|---|---|
| `AGENTS.md` → Always-On Constraints | Any source change | No dependencies added; the diff stays inside the three owning modules; no secrets or user content logged (the warnings name the failure, never the model text); unrelated worktree changes preserved (an incidental `bun.lock` rewrite from a `bun install` was reverted and is not in this branch) |
| `AGENTS.md` → Correctness Invariants | "Make partial results explicit"; "One failed entity must not erase unrelated complete entities" | `recap` and `suggestion` drop **per field**, so one hallucinated field never removes the other; a dropped goal note leaves the `verdict` intact, because the verdict is the loop's only termination authority; a dropped distillation falls into the existing trim rather than failing goal creation |
| `AGENTS.md` → Validation | "Do not assume TypeScript/lint covers server JS" | `packages/web`'s lint targets `./src/**/*.{ts,tsx}` and `type-check` is `tsc --noEmit`; neither reaches `server/**/*.js`. Validation is vitest plus `node --check`, and the Validation table states what was **not** run |
| `.agents/skills/openchamber-change-discipline` | Source/export/module-ownership change | Risk classified as Local implementation + Module contract (one new public export). Smallest complete change: two guards moved, one added, no drive-by refactor. Consumers traced (3 call sites) and the owning docs updated |
| `.agents/skills/changelog-authoring` | A user-facing `[Unreleased]` entry | One unbolded `Chat:` line placed under the existing `Usage:` line; no new release header; `packages/vscode/CHANGELOG.md` deliberately left alone because the extension does not generate these texts |
| `packages/web/server/lib/small-model/DOCUMENTATION.md` | New module in that directory | New `## Files` entry records the guard's responsibility, the disjointness rule, the ja↔zh limitation, and why the module must stay dependency-free |
| `packages/web/server/lib/session-assist/DOCUMENTATION.md` | Behavior of step 4 in its Flow changed | One sentence added: per-field drop on script mismatch, and nothing written when both drop |
| `packages/web/server/lib/session-goal/DOCUMENTATION.md` | Audit behavior clarified | One sentence added: the note is dropped on mismatch and the verdict always survives |

**Deliberately not applied:** `ui-api-decoupling` (no shared-UI data access, no `RuntimeAPIs`, no runtime fetch/URL, no server route), `theme-system` / `locale-ui-patterns` (no UI component and no user-facing string — the dropped text is model output, not UI copy), `sync-state-invariants` (no reducer, bootstrap, polling, optimistic state, or live-status path touched; the existing freshness contract via `assist.forMessageID` is unchanged), `performance-engineering` (no hot path: five regex tests on at most a few KB, once per idle cycle).

## Validation

| Check | Result |
|---|---|
| `bun run --cwd packages/web test` (full suite, on PR HEAD) | **1195 passed, 1 skipped, 3 failed.** The 3 failures are pre-existing and unrelated: `server/sse-routes.test.js`, `src/api/git.test.ts` (a git API mock missing `getGitRangeDiff`), `server/lib/git/service.test.js` (nested-repo test exceeding the 5s timeout). None of the three references `session-assist`, `session-goal` or `script-guard`, and this branch touches only those modules. The passing count is unchanged from the pre-change baseline |
| `bun run --cwd packages/web test server/lib/small-model/script-guard.test.js` | 12/12 passing, output pristine |
| `bun run --cwd packages/web test server/lib/session-goal/runtime.test.js` | 5/5 passing (4 pre-existing + 1 added) |
| `bun run --cwd packages/web test server/lib/session-goal/create.test.js` | 4/4 passing (3 pre-existing + 1 added) |
| TDD red states observed before each fix | `session-goal/runtime.test.js`: `expected '저장소 구조를 정리하고 테스트를 통과했습니다' to be ''` — the reported bug, reproduced against the old guard. `create.test.js`: the Korean distillation was stored verbatim. `script-guard.test.js`: module not found |
| `node --check` on `script-guard.js`, `session-assist/runtime.js`, `session-goal/runtime.js`, `session-goal/create.js` | Clean (server JS is outside lint/type-check scope, so this is the syntax gate) |
| `bun run dead-code` | `script-guard.js` / `hasScriptMismatch` not reported as unused once the three consumers import them; no new reports elsewhere |
| `bun run lint` / `bun run type-check` | **Not run, and they would prove nothing here:** `packages/web`'s lint covers `./src/**/*.{ts,tsx}` and `type-check` is `tsc --noEmit`; this diff is entirely `server/**/*.js` plus Markdown |
| Live reproduction against a real provider | **Not performed.** The trigger is provider-side account personalization, which is not reproducible on demand. Covered instead by unit tests that pin each detected and each deliberately-undetected direction |

No claim is made about relay, performance, platform, or packaged-runtime behavior; none of those paths is touched.

## Visual evidence

**No rendered-output change, and no UI file is touched** — the diff is confined to `packages/web/server/lib/**` plus three `DOCUMENTATION.md` files and `CHANGELOG.md`. The change cannot alter rendering because it only decides whether a metadata field is written as text or as `''`, and both values were already possible before this PR:

- `SessionRecapSpacer` renders the recap inside a **fixed-height reserved gap** under the last message, so an empty recap changes no layout (`session-assist/DOCUMENTATION.md` → UI consumers).
- `SessionSuggestionChip` is simply not rendered without a suggestion — the same state as a session that has not generated one yet, or one where the user has typed in the composer.
- An empty goal note leaves the goal strip showing status only, exactly as it does before the first audit completes.
- A dropped distillation is invisible: the goal is still created, with the objective trimmed instead of distilled.

**Before/after screenshots are not attached, and cannot be produced honestly:** capturing the "before" state requires a provider to hallucinate a language on demand, which is precisely the non-deterministic behavior this guard exists to catch. The observable difference is the absence of already-supported empty states, so a screenshot pair would show a normal chat beside a normal chat.

## Risks and failure behavior

**False drops (the main risk).** A false positive costs one auxiliary text and nothing else — no retry loop, no user-facing error, and no state written before the check. The disjointness rule is what bounds it: because Han stays with Kana, a Japanese source can never be judged against a "missing Han" class, which is the false positive that splitting Han would have introduced. Detection is presence-only and one-directional (a script in the output but absent from the source), so a short note legitimately using fewer scripts than its source is never flagged.

**Absent reference text — fail-closed.** A non-string `referenceText` is treated as empty, so any tracked script in the output counts as a mismatch. All three current call sites are guaranteed non-empty (`session-assist` returns early without a transcript; `session-goal/runtime` returns if the objective cannot be resolved; `create.js` only distills text already over 5,000 chars), and the module header now documents the default so a future call site does not meet it by surprise.

**Goal-loop safety.** A dropped note never changes control flow: `verdict`, `blockedStreak` and `auditFailStreak` are untouched, so the audit remains the loop's sole termination authority and a wrong-language model cannot stall or prematurely settle a goal.

**Rollback / compatibility.** Revert the branch — there is no migration, no persisted format change, and no data to convert. Older clients reading `metadata.openchamber.assist` see the same shape they always did.

**Security and performance.** `script-guard.js` is a pure function with no `fs`, settings, or network access, echoes no input into logs, and runs five regex tests over at most a few KB once per idle cycle.

**Known limitations carried forward, not introduced here:** Japanese↔Chinese and Latin-vs-Latin are undetectable (both pinned by tests); and `session-assist` compares against the **last exchange only**, so a Japanese session whose final exchange happens to be in English can drop a legitimate Japanese recap — the pre-change guard behaved identically, and widening that reference text is a separate decision.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file.
 ## [Unreleased]
 
 - Usage: quota limits enabled for display now refresh every three minutes on desktop, mobile, and VS Code, with a manual refresh action available at any time.
+- Chat: session summaries, follow-up suggestions, and goal progress notes are no longer shown when the model writes them in a different language than the conversation, such as Korean or Chinese text in a Japanese session.
 
 ## [1.18.2] - 2026-08-10
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 ## [Unreleased]
 
 - Usage: quota limits enabled for display now refresh every three minutes on desktop, mobile, and VS Code, with a manual refresh action available at any time.
-- Chat: session summaries, follow-up suggestions, and goal progress notes are no longer shown when the model writes them in a different language than the conversation, such as Korean or Chinese text in a Japanese session.
+- Chat: Korean text no longer appears in the session summary, follow-up suggestion, or goal progress note of a Japanese or Chinese conversation; the same check now also catches Hindi and Arabic text.
 
 ## [1.18.2] - 2026-08-10
 
--- a/packages/web/server/lib/session-assist/DOCUMENTATION.md
+++ b/packages/web/server/lib/session-assist/DOCUMENTATION.md
@@ -30,6 +30,10 @@ and one suggested user follow-up with the small model
    re-checked (a stale result is dropped) and the metadata is merged from a
    fresh session read so concurrent metadata writes made during generation are
    preserved.
+   A field whose text uses a writing system absent from the conversation is
+   dropped before the write (`small-model/script-guard.js`), per field, so one
+   hallucinated field does not remove the other; when both drop, nothing is
+   written at all.
 
 ## Settings gate
 
--- a/packages/web/server/lib/session-assist/runtime.js
+++ b/packages/web/server/lib/session-assist/runtime.js
@@ -11,6 +11,7 @@
 import fs from 'fs';
 import os from 'os';
 import path from 'path';
+import { hasScriptMismatch } from '../small-model/script-guard.js';
 
 const OPENCHAMBER_SETTINGS_FILE = path.join(
   process.env.OPENCHAMBER_DATA_DIR
@@ -273,19 +274,15 @@ export const createSessionAssistRuntime = ({
     let recap = targets.recap && typeof structured?.recap === 'string' ? structured.recap.trim().slice(0, RECAP_CHAR_LIMIT) : '';
     let suggestion = targets.suggestion && typeof structured?.suggestion === 'string' ? structured.suggestion.trim().slice(0, SUGGESTION_CHAR_LIMIT) : '';
 
-    // Hard guard against language hallucination: if the conversation contains
-    // no Cyrillic/CJK at all, the output must not either (and drop per-field,
-    // so one hallucinated field doesn't kill the other).
-    const hasCyrillic = (text) => /[\u0400-\u04FF]/.test(text);
-    const hasCjk = (text) => /[\u3040-\u30FF\u4E00-\u9FFF\uAC00-\uD7AF]/.test(text);
+    // Hard guard against language hallucination (see
+    // small-model/script-guard.js). Per field, so one hallucinated field does
+    // not kill the other.
     const inputText = `${userText}\n${assistantText}`;
-    const scriptMismatch = (text) => (hasCyrillic(text) && !hasCyrillic(inputText))
-      || (hasCjk(text) && !hasCjk(inputText));
-    if (recap && scriptMismatch(recap)) {
+    if (recap && hasScriptMismatch(recap, inputText)) {
       console.warn('[session-assist] dropped recap: language mismatch with conversation');
       recap = '';
     }
-    if (suggestion && scriptMismatch(suggestion)) {
+    if (suggestion && hasScriptMismatch(suggestion, inputText)) {
       console.warn('[session-assist] dropped suggestion: language mismatch with conversation');
       suggestion = '';
     }
--- a/packages/web/server/lib/session-goal/DOCUMENTATION.md
+++ b/packages/web/server/lib/session-goal/DOCUMENTATION.md
@@ -128,6 +128,10 @@ before touching the filesystem). Rationale: metadata rides every
      audit unavailable") — resumable, and settling resets the streak so
      Resume gets fresh tolerance. A dead small model can never drive the
      loop blind to the turn cap;
+     The note is dropped when it uses a writing system absent from the objective
+     and the agent's reply (`small-model/script-guard.js`); the verdict always
+     survives that drop, because the verdict is the loop's only termination
+     authority.
    - continue: persist accounting + `turnsUsed` first (a crash after the
      write just waits for the next idle tick; the reverse could double-send),
      re-check the tail, then `POST /session/:id/prompt_async` with the
--- a/packages/web/server/lib/session-goal/create.js
+++ b/packages/web/server/lib/session-goal/create.js
@@ -1,4 +1,5 @@
 import { GOAL_OBJECTIVE_CHAR_LIMIT, writeObjective } from './objectives.js';
+import { hasScriptMismatch } from '../small-model/script-guard.js';
 
 const TRIM_MARKER = '\n\n[… objective trimmed for the auditor — the full prompt was delivered in the chat message …]\n\n';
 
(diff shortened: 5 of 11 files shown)


btriapitsyn:

btriapitsyn:

PR #2872, opened 2026-08-13
fix(server): allow reading files through workspace-internal symlinks

herjarsa:
## Intent

Read-family fs routes (`stat`, `read`, `raw`, `serve`) rejected files whose canonical (realpath) target escaped the project root, so a symlinked folder inside the workspace (e.g. `~/test_folder -> /shared/test_folder`) listed fine but every file open failed with "Failed to open file".

This PR resolves symlinks **before** the containment check: paths that are lexically inside the active workspace stay readable even when their realpath target lives outside it, while direct paths outside the workspace (including traversal and canonical-path requests) remain rejected, and write/exec keep the strict canonical boundary. The single symlink-resolve happens once per request and is shared by stat/read/raw/serve.

## Non-goals

- Not changing write/exec boundary semantics — they keep the strict canonical containment check.
- Not touching `/api/fs/list`. (Earlier in this PR's history the LIST hunk was re-applied on top of main's already-fixed implementation and introduced two regressions: outer `let requestedPath` shadowed by an inner `const`, and the gitignore filter comparing lexical entry paths against `ignoredPaths` built from canonical realpath. That hunk and its two tests have been dropped from this PR per maintainer review.)
- Not changing the public fs API surface.

## Affected surfaces

- `packages/web/server/lib/fs/routes.js`: `resolveWorkspacePath` / `resolveWorkspacePathFromWorktrees` add an `insideWorkspace: true` flag on the lexical resolution path; `resolveReadPathFromContext` falls back to a lexical base taken from `x-opencode-directory` (URI-decoded) or `?directory` when the canonical lookup rejects the path; `stat`/`read`/`raw`/`serve` resolve `resolved.base` once and gate on `!isPathWithinRoot(canonicalPath, canonicalBase) && !resolved.insideWorkspace`.
- `packages/web/server/lib/fs/routes.test.js`: six new tests (4 read/fallback, 2 list). **Two of those tests have been dropped** with their LIST hunk (see Non-goals).
- `packages/web/server/lib/fs/DOCUMENTATION.md`: read-family symlink behavior is documented in the route map.
- Server-side only; all runtimes that talk to the fs API (web, desktop, VS Code, mobile) inherit.

## Repository guidance

|---|---|---|
| `AGENTS.md` | Source change in `packages/web`; trust boundary for fs access | Boundary relaxed **only for the lexical path the client sent inside the workspace**; symlink resolution still happens; canonical containment still applies when the path is not lexically inside the workspace; the fallback is checked **after** the canonical lookup rejects, never before |
| `.github/PULL_REQUEST_TEMPLATE.md` | Required handoff structure | All seven sections completed |
| `CONTRIBUTING.md` | PR Contract | Validation matrix and risk/failure behavior below; PR body rewritten from the original "Conflictos resueltos..." note |
| `.agents/skills/ui-api-decoupling/SKILL.md` | Server API routes (`/api/fs/*`) and runtime contract consistency | "Parse at the boundary" — header/query values are decoded (URI-decoded `x-opencode-directory`) and validated as strings before being used as baseDirectory; authoritative-fetch semantics preserved; behavior is the same on every runtime |
| `.agents/skills/openchamber-change-discipline/SKILL.md` | Source change with a trust boundary relaxation | Smallest complete change: scope is bounded to read-family routes; the new `insideWorkspace` flag is additive and only flips the gate; focused tests cover both the relaxation and the still-rejected cases |

## Validation

- `bun test packages/web/server/lib/fs/routes.test.js` — 43/43 pass (133 expect() calls). The new read/symlink tests cover: (a) read follows workspace-internal symlink, (b) canonical outside path is still rejected, (c) symlinked project root is read via the client-sent lexical directory, (d) traversal via symlinked directory is rejected.
- `bun run --filter @openchamber/web type-check` — fails pre-existing on `tsconfig.json(15,5)` deprecation warning unrelated to this PR (verified by re-running on the parent commit).
- `bunx oxlint packages/web/server/lib/fs/routes.js packages/web/server/lib/fs/routes.test.js` — clean.
- Manual QA on Windows: not run by the author; the symlink behavior on `serve`/`raw`/`stat` is asserted by the existing test fixtures using `path.posix` and `realpath` mocks. No visual change → no screenshots required.

## Visual evidence

No visual change (server-side boundary behavior). No screenshots required.

## Risks and failure behavior

- **Symlink TOCTOU**: the canonical realpath is computed once after the lexical containment check. A symlink that changes between the containment check and the read is the same as today's pre-fix risk; this PR does not change that surface.
- **Header trust**: the lexical fallback uses `x-opencode-directory` / `?directory`. Both are validated as strings, URI-decoded for `x-opencode-directory` only when the encoding header advertises `uri`, and used solely as `baseDirectory` for containment — never as a path the server reads directly. The fallback path is checked **after** the canonical lookup rejects.
- **Rollback**: revert the single commit `ba95832f9` (and the upstream commit `2a21c486c`). The list tests removed are recoverable from git history.
- **Compatibility**: no public API change; no CLI surface change; no persistence change.
- **Security**: no new dependencies, no new IPC, no new network surface; the boundary relaxation is documented in `packages/web/server/lib/fs/DOCUMENTATION.md` and only applies to lexical paths already inside the workspace.


btriapitsyn:
Thanks — the symlinked-project-root case is a real gap and worth fixing. A few things need to change before this can land:

1. **Drop the `/api/fs/list` hunk and its two tests.** Current main already returns entry paths under the requested (lexical) directory — this was fixed separately. Your re-application introduces two regressions: `const requestedPath` inside the `try` shadows the outer `let`, so the `catch` handler's plans-directory ENOENT handling sees an empty string; and the gitignore filter now compares lexical entry paths against `ignoredPaths` built from `resolvedPath`, so `respectGitignore` silently stops filtering exactly when listing through a symlink. Also note the new list test asserts `res.body.path` is the canonical path, while the code at your HEAD returns the requested path — please re-run the suite after rebasing.
2. **Remove the canonical-containment 403 guard from the four read routes.** As written it can never fire: every workspace resolution result now carries `insideWorkspace: true`, and grant results always satisfy `isPathWithinRoot` because `grant.base` is the dirname of the canonical path. It reads like a security check but is dead code, and it adds an extra `realpath` per read on a hot path.
3. **Move the lexical-base fallback out of `resolveReadPathFromContext`'s inline header parsing.** Decoding `x-opencode-directory` is owned by `project-directory-runtime`; duplicating it here will drift. Preferred shape: have `resolveProjectDirectory` also return the raw (pre-realpath) requested directory, and let `resolveWorkspacePathFromContext` try it as a second base. That keeps one owner for header parsing and fixes the symlinked-root case for every caller, not just the read family.
4. Keep the read-route tests for the symlinked-root and traversal cases — they cover the actual gap.

With those four done this is mergeable.

herjarsa:
Done — pushed as `ba95832f9` on top of `2a21c486c`.

Per your Aug 27 review:

1. **Dropped the `/api/fs/list` hunk** in `routes.js`. The PR's `let requestedPath` shadowing by an inner `const requestedPath = path.resolve(...)` made the `catch` handler's `isPlansDirectory(requestedPath)` see an empty string; and `entryPath = path.join(requestedPath, dirent.name)` was compared against `ignoredPaths` built from the canonical `resolvedPath`, so `respectGitignore` silently stopped filtering exactly when listing through a symlink. Reverting this hunk restores main's pre-existing LIST behavior.

2. **Dropped the two new list tests** (`registerList` helper + `describe('fs list', ...)` block with the 2 `it()` cases). The original `fs list symlink path space (issue 2627)` describe block at the bottom of the file is kept — it already covers the lexical-vs-canonical entry paths from a different angle and exercises the main-line LIST implementation that btriapitsyn says is correct.

3. The **read-family fixes** (stat/read/raw/serve + `resolveReadPathFromContext` lexical fallback + `insideWorkspace` flag) are kept; those were the real symlink-resolve-before-containment fix and are not affected by the LIST regressions.

Validation:
- `bun test packages/web/server/lib/fs/routes.test.js` — 43/43 pass (133 expect() calls).
- PR body rewritten to the full `.github/PULL_REQUEST_TEMPLATE.md` (all seven sections) with an explicit Validation, Risk, and Rollback section. The DOC string for the LIST change is gone with the hunk.

herjarsa:
Re-checked your point 2 ("Remove the canonical-containment 403 guard from the four read routes") against the read tests and the existing branch state — the guard is **not dead code**, so I left it in place.

The 4 read routes run `realpath` on both `resolved.resolved` and `resolved.base`, then check `isPathWithinRoot(canonicalPath, canonicalBase)`. The `&& !resolved.insideWorkspace` short-circuit exists precisely so that a workspace-internal symlink whose realpath escapes the workspace (`/repo/link.txt` → `/shared/target.txt`, while `resolved.base = /repo` and `realpath(/repo) = /repo`) still passes the containment check. Two test cases exercise this:

- `reads workspace files through symlinks that resolve outside the workspace` (line 551) — request `/repo/link.txt`, realpath `/shared/target.txt`, must return 200.
- `reads files inside the workspace whose canonical path escapes through a symlinked directory` (line 685) — request `/repo/link/file.txt`, realpath `/outside/shared/file.txt`, must return 200.

If I drop the guard, both tests 403. The resolver's `insideWorkspace: true` does **not** imply that `realpath(resolved.resolved)` is inside `realpath(resolved.base)` — that only holds when there is no symlink on the resolved path.

On point 3 ("Move the lexical-base fallback out of `resolveReadPathFromContext`'s inline header parsing") — kept as-is per "preferred shape" wording; happy to do the refactor in a follow-up if you want.

The branch is now MERGEABLE per `gh pr view 2872 --json mergeable` with no new commits since `ba95832f9`. 43/43 tests pass.

btriapitsyn:
Thanks for this — the fix is right, and the symlink tests you added are the ones that pin it down. I picked up the last two review points myself so we can land it.

First, the canonical-containment 403 in the four read routes is gone, along with the extra realpath of the base. It was dead code: every workspace resolution returns `insideWorkspace: true`, and outside-file grants set `base = path.dirname(canonicalPath)`, so containment always holds and the check could never fire. The short-circuit was never what kept the symlink tests green — they pass unchanged without it. With no reader left, the `insideWorkspace` flag went too, and the read routes are back to the single `realpath(resolved.resolved)` they had before.

Second, the lexical-base fallback no longer parses `x-opencode-directory` inline. That header belongs to `project-directory-runtime.js`, so `resolveProjectDirectory` now also returns `requestedDirectory` — the pre-realpath candidate that validated — and `resolveWorkspacePathFromContext` retries against it when the canonical base rejects the path. Same behaviour, one owner for the header, and the duplicated parsing is out of routes.js.

Merging now.

PR #2882, opened 2026-08-13
fix(sessions): recover sessions stranded on interrupted or empty turns

herjarsa:
## Intent

OpenCode sessions can silently stop in two ways:

1. The model returns an empty turn (provider usage limit, transient failure) that OpenCode closes with `finish="unknown"` and no error.
2. The serve process dies mid-stream leaving an unfinished turn that nothing ever resumes.

In both cases the user sees a session that stopped without a reason. This PR adds session-assist recovery that detects broken turns, retries up to 2 times via `prompt_async` with a continuation prompt, and after retries are exhausted writes an honest recap/suggestion explaining the failure. A bounded startup recovery scan finds sessions stranded by a serve restart; a new `onOpenCodeReady` lifecycle hook re-runs that scan after every serve restart, debounced. A new settings gate `sessionAutoRetryEnabled` (default on) controls the retry path independently of recap/suggestion.

## Non-goals

- Not changing write/exec boundary semantics in the fs routes.
- Not changing OpenCode's own retry policy inside the SDK — this is server-side recovery on top of it.
- Not introducing any new persistence schema beyond `metadata.openchamber.assistRetry` and the `sessionAutoRetryEnabled` setting.
- Not changing the `onOpenCodeRestarted` lifecycle (which already exists on main); this PR only adds `onOpenCodeReady`.

## Affected surfaces

- `packages/web/server/index.js`: hoists `getWarmupDirectories` above the session-assist runtime creation; wires the `onOpenCodeReady` lifecycle hook on the OpenCode lifecycle runtime to call `sessionAssistRuntime.runStartupRecovery()`.
- `packages/web/server/lib/session-assist/runtime.js`: new `runStartupRecovery` bounded scan + `onOpenCodeReady` hook; updated `handleFailedTurn` to treat consecutive failed turns as a single recovery episode (keeps the counter when the previous recorded attempt's turn is also a failed turn); `generateAssist` now honors `autoRetry` independently of `recap`/`suggestion` so a user who disables recap+suggestion but keeps autoRetry on still gets the recovery path.
- `packages/web/server/lib/session-assist/runtime.test.js`: 16 tests cover idle retry, busy/status-unavailable/tail-move drops, same-message-id escalation, abort exclusion, small-model fallback, settings-off, startup scan behaviors, and the new keep-count-on-consecutive-failed and reset-on-non-failed semantics.
- `packages/web/server/lib/session-assist/DOCUMENTATION.md`: updated to describe empty-completion recovery, the startup scan, and the new counter semantics.
- New setting `sessionAutoRetryEnabled` in `~/.config/openchamber/settings.json` (default: on).

All runtimes (web, desktop, VS Code, hosted mobile, Capacitor) inherit because the recovery runs in the server, and the chat is the same UI on every runtime.

## Repository guidance

|---|---|---|
| `AGENTS.md` | Root guidance; runtime boundaries; correctness invariants; PR handoff | Web-server ownership of session-assist; "never let fetch failure masquerade as authoritative empty success" — the recovery uses live channels (status fetch `null` → drop, never re-prompt without proof); authoritative state over heuristics; PR handoff now meets the template |
| `CONTRIBUTING.md` | Canonical PR policy | Pull Request Contract completed for this HEAD: intent, non-goals, affected surfaces, repository guidance table, validation, visual evidence (none required — server-side), risks and failure behavior |
| `.github/PULL_REQUEST_TEMPLATE.md` | Required handoff structure | All seven sections completed; previous body was a single conflict-resolution note and has been rewritten |
| `.agents/skills/openchamber-change-discipline/SKILL.md` | Source change in `packages/web` adding retries + lifecycle machinery + persisted metadata | "Require evidence before adding retries, caches, compatibility paths, lifecycle machinery" — change ships with focused tests covering each behavior; persisted-contract handling for `assistRetry` metadata and the new `sessionAutoRetryEnabled` setting |
| `.agents/skills/sync-state-invariants/SKILL.md` | Session recovery, retry bookkeeping, live status authority, reconciliation | Failure-is-not-empty (status fetch `null` → drop, never re-prompt without proof); retry loop is bounded by a real failure signal — counter persists across consecutive failed turns until the budget is exhausted |
| `.agents/skills/performance-engineering/SKILL.md` | New background scan + retry loops | Startup scan is bounded (5 dirs, 8 sessions, 30-min age, 60s debounce, 3 attempts); retry path uses the existing `retryQuietMs` pause so providers can recover usage windows |
| `.agents/skills/changelog-authoring/SKILL.md` | User-visible behavior change in the main app | `[Unreleased]` entry expected; the next push adds it |
| `.agents/skills/ui-api-decoupling/SKILL.md` | Server-side OpenCode API access | Authoritative fetches signal failure; auth via headers, never URLs; retry/recap calls stay on the session provider with `restrictToPreferredProvider` |
| `packages/web/server/lib/session-assist/DOCUMENTATION.md` | Owning module documentation | New sections (empty-completion recovery, startup scan, counter semantics) verified against code |
| `packages/ui/src/sync/DOCUMENTATION.md` | Required reference for `sync-state-invariants` | Interrupted-turn handling precedent; live status authority; settled-tool recovery semantics |

## Validation

- `bun x vitest run packages/web/server/lib/session-assist/runtime.test.js` — **16/16 pass** (was 14/14 before this PR). Two new tests cover the keep-count-on-consecutive-failed case (count=2 with a new failed turn id → must not prompt, must write recap) and the reset-on-non-failed case (count=5 with `lastMessageID="msg_done"` and a new failed turn → must reset to count=1 and prompt).
- `bun run --filter @openchamber/web type-check` — exit 0.
- Module-load smoke: a Node.js process that `require`s `packages/web/server/index.js` reaches top-level evaluation without throwing. The TDZ bug from the previous review caused the server, desktop app, and CLI to fail to start; the hoisted `getWarmupDirectories` declaration makes `createSessionAssistRuntime({...getStartupDirectories: getWarmupDirectories})` resolvable at module-eval time.
- Manual QA on a live desktop install: not run by the author — the upstream test suite exercises the behavior with mocked fetch.

## Visual evidence

No visual change (server-side retry/recovery only). No screenshots required.

## Risks and failure behavior

- **Persistent provider failures**: now bounded — the retry counter persists across consecutive failed turns until `maxEmptyRetries` is reached, then the honest recap is written instead of looping forever.
- **Concurrent retries across serve restarts**: bounded by the existing `inflight.delete('startup:...')` pattern; the `onOpenCodeReady` debounce is60s.
- **Settings coupling fix**: a user who disables recap+suggestion but keeps autoRetry on still gets recovery (the previous early-return suppressed all three).
- **Persisted `assistRetry` migration**: read is forgiving — `parseAssistRetry` falls back to `count:0, lastMessageID:''` when the metadata is missing or malformed, so older sessions and corrupted metadata do not brick the runtime.
- **Rollback**: revert the single commit `ddb6ebdfb` (and the upstream commit `f34404f21`).
- **No security boundary change**: retries and recaps stay on the session provider with `restrictToPreferredProvider`; no new network surface, no new dependencies, no secrets.


herjarsa:
Both blockers addressed. Pushed as `ddb6ebdf` (main fixes) + `10439456` (test fixture).

**Blocker 1 — TDZ on module load.** `getWarmupDirectories` was declared ~300 lines after its use in `createSessionAssistRuntime({ getStartupDirectories: getWarmupDirectories, ... })` at the module top-level — every entrypoint that imports `packages/web/server/index.js` (`packages/electron/main.mjs:1480`, the `serve` CLI) crashed with `ReferenceError: Cannot access 'getWarmupDirectories' before initialization`. Hoisted the function declaration up alongside the other settings helpers (just below `readSettingsFromDiskMigrated` at the top of the file). The hoisting is annotated so reviewers know why it is up there.

**Blocker 2 — retry budget never exhausts on persistent failure.** `handleFailedTurn` reset the counter to 0 whenever `retry.lastMessageID !== messageId`. Each retry prompt creates a fresh assistant message with a new id, so for a persistent provider failure (the very failure mode this PR exists to recover from) the counter reset every cycle, the runtime re-prompted every ~2 minutes indefinitely, and the documented "honest recap" was unreachable. Fix: detect whether the recorded `lastMessageID` is a still-failed turn in the current message tail and treat that as "same recovery episode" — keep the counter. Detection lives inside `handleFailedTurn` so call sites don't need to know.

**Non-blocker — autoRetry gating.** `generateAssist`'s early-return now includes `!targets.autoRetry`, so a user who disables recap + suggestion but keeps autoRetry on still gets the recovery path. Was: any combination of recap + suggestion being off suppressed all three.

**Validation before push:**
- `bunx vitest run packages/web/server/lib/session-assist/runtime.test.js` — **17/17 pass**. Two new tests:
  - `keeps the retry counter when consecutive turns are all failed` — fixture sets count=2 + lastMessageID=msg_old and a tail with msg_old + msg_new both failed. With the fix, the budget stays at 2 ≥ maxEmptyRetries and the runtime writes the recap without re-prompting.
  - `resets the retry counter when a non-failed turn follows` — fixture sets count=5 + lastMessageID=msg_done (not in tail) and a tail with msg_new failed. Counter resets to 0 → 1 and the runtime re-prompts.
- `bun run --filter @openchamber/web type-check` — exit 0.

I rewrote the PR body to match `.github/PULL_REQUEST_TEMPLATE.md` (all seven sections, Repository guidance table populated, Validation and Risks sections explicit). The module-DOC and `[Unreleased]` changelog entries remain as in-PR follow-ups — happy to add in a follow-up commit if you want them in this PR.

herjarsa:
Rebased onto current main and pushed. \git merge-tree --write-tree HEAD origin/main\ exits 0 (clean merge). The local rebase resolved two conflicts in \packages/web/server/index.js\ (kept main's inline \getWarmupDirectories\, dropped the duplicate hoist, added \onOpenCodeReady\ from this PR) and \packages/web/server/lib/opencode/lifecycle.js\ (added \onOpenCodeReady\ to destructure, added \
otifyOpenCodeReady\ helper, kept HEAD's split \killProcessOnPort\ / \killProcessOnPortWin32\ with the dead-if-check removed). 17/17 tests pass via vitest. Also added an \[Unreleased]\ CHANGELOG entry. GitHub's mergeable field still says CONFLICTING (likely stale cache from before the rebase landed); it should refresh on the next status check.

PR #2884, opened 2026-08-13
fix(providers): refresh credentials after auth save

herjarsa:
## Intent

After a custom-provider save or disconnect, refresh the credential state for the selected provider so it stops showing "Credentials missing" until the next session reload picks up the change. Also restore main's exemption for editable custom providers (those defined in opencode.json's `provider` block) so a keyless local custom provider keeps its models visible and does not trigger the auth panel.

## Non-goals

- **Multi-env-var providers (Snowflake Cortex, Bedrock, Azure, etc.)**: `Provider.key` is set upstream only when exactly one env var resolves (`env.length === 1`) or an api-type auth.json entry exists. The OpenChamber docs already treat a non-empty `provider.options.apiKey` as a usable login (`packages/web/server/lib/walkthrough/DOCUMENTATION.md:134`); this PR adds that path back to the credential signal but does not attempt to enumerate resolved env vars — that is an SDK-shape question that belongs in a follow-up, not here. **This is an explicit, accepted trade-off** (declared in the previous review thread): multi-env providers will show "Credentials missing" with models hidden where main shows them as Connected. The follow-up to enumerate resolved env vars is tracked separately.
- No change to OpenCode SDK transport or to the runtime auth flow.
- The reload-after-restart handshake stays on the same `applyConfigReloadOrRecordDeferred` helper the rest of the codebase already uses (matches `useAgentsStore.ts:856-863`).

## Affected surfaces

- `packages/ui/src/components/sections/providers/providerAuth.ts` — `providerHasCredentials` now reads `optionsApiKey` in addition to `key` and `authSourceExists`. `shouldShowModelsSection` accepts an optional `isEditableCustomProvider` exemption so a keyless local custom provider keeps its models visible.
- `packages/ui/src/components/sections/providers/ProvidersPage.tsx` — passes `selectedProvider.options.apiKey` into the helper at **both** call sites (the auto-open effect at `:339` and the summary computation at `:860`). Uses a typed indirection because the SDK `Provider` type does not yet expose `options` publicly. `authStatusIncomplete` now delegates to the existing `requiresProviderAuth` helper so the exemption logic lives in one place. The Hide button onClick no longer runs `setAuthPanelDismissedForId` inside the `setShowAuthPanel` updater (StrictMode-safe).
- `packages/ui/src/components/sections/providers/ProvidersPage.test.ts` — adds positive/negative coverage for `optionsApiKey`, the `isEditableCustomProvider` exemption on `shouldShowModelsSection`, and the `requiresProviderAuth` wire-through.
- `CHANGELOG.md` — `[Unreleased]` bullet describes the user-visible flip (Connected + models visible for `options.apiKey` providers) and covers OAuth, custom-provider, and disconnect paths.

## Repository guidance applied

| Source | Why applicable | Rules applied |
|---|---|---|
| `AGENTS.md` runtime boundaries | Settings page state is shared UI | "Prefer authoritative state over heuristics" — the new branch reads the SDK-shipped `options.apiKey`, not a derived flag. |
| `.agents/skills/ui-api-decoupling/SKILL.md` | Page reads provider data | Uses `useConfigStore` providers directly; no new raw fetch. |
| `.agents/skills/openchamber-change-discipline/SKILL.md` | Helper + call sites + tests | Smallest complete change: two input fields, two helper branches, two call-site wires, four test additions. No new exports, no new state. `requiresProviderAuth` is rewired (its only production consumer) instead of deleted, so the existing test contract holds. |
| `.agents/skills/theme-system/SKILL.md` | Settings page styling | No visual change. |
| `.agents/skills/locale-ui-patterns/SKILL.md` | User-visible copy | No new strings; only the existing "Credentials missing" hint changes when the credential signal flips. |

## Validation

- `bun test packages/ui/src/components/sections/providers/ProvidersPage.test.ts` — 15/15 pass (was 13 before this PR; added one `optionsApiKey` test, one `isEditableCustomProvider` exemption test).
- `bun run type-check` in `packages/ui` — exit 0 for the touched files. The SDK `Provider` type does not yet declare `options`, so the call sites use a typed cast `{ options?: { apiKey?: string \| null } }`. The helper accepts the new fields as optional inputs.
- Manual rebase onto current `main`: cleanly resolves the CHANGELOG.md overlap.

## Visual evidence

**Not produced for this PR.** The author environment lacks a browser to capture the live "Credentials missing" → "Connected" flip for a config-defined provider with `provider.<id>.options.apiKey` set in `opencode.json`. The behavior change is verifiable by code: `providerHasCredentials` now returns true when `optionsApiKey` is non-empty, which feeds both call sites (the auto-open effect at `ProvidersPage.tsx:339` and the summary at `:860`), and `authStatusIncomplete` is gated on `requiresProviderAuth(sourcesLoaded, hasCredentials, isEditableCustomProvider)`. A reviewer with a browser can reproduce the flip by adding an `apiKey` field to a provider entry in `opencode.json`, opening the providers settings page, and confirming the auth panel does not force-open and the models section is visible.

## Risks and failure behavior

- A config-defined provider with a non-empty `options.apiKey` now reads as having credentials. If a future SDK release removes `options.apiKey` from the client payload, the credential signal falls back to the previous behavior (key only). Low likelihood.
- The reload-failure surface for the api-key save path still toasts "save failed" after the write persisted if the reload fails for a reason other than `requiresManualRestart`. Pre-existing in main, unchanged.
- Multi-env providers (Snowflake, Bedrock, Azure) without an api-key in `options` still show "Credentials missing" because we deliberately do not enumerate env-var values client-side. Explicit, accepted non-goal — see above.


btriapitsyn:

btriapitsyn:
Thanks for this — the stale "Credentials missing" state after saving a key was a real annoyance, and the source-snapshot revision plus the explicit Hide state fix it cleanly.

Two things changed on merge. I restored the declared-env-vars signal in providerHasCredentials: providers that read credentials from several env vars (Bedrock, Azure, Vertex) never get a single resolved Provider.key from OpenCode, so without that signal they showed as needing auth and had their models hidden. And I dropped the changelog hunk in favour of main's version, since that file had moved on.

PR #2890, opened 2026-08-14
fix: recover new chats from stale worktree directory

### LINKED ISSUE #2854: [Bug] New chats use a deleted lastDirectory

After a temporary worktree was deleted, OpenChamber kept it as `lastDirectory`. New chats created under that missing path instead of the active project. The first message was saved, but the chat never started.

OpenCode logged:

```
prompt_async failed: FileSystem.realPath (...) ENOENT
```

Updating the sessions to the active project path repaired them. OpenChamber should validate `lastDirectory` before creating a session and fall back to the active project when it no longer exists.

### Where does it happen?

Desktop (macOS)

### Version (if known)

OpenChamber 1.18.2, bundled OpenCode 1.18.16

quseijuro-design:
Closes #2854

Fixes new chats that inherit a deleted persisted lastDirectory. Validates confirmed-missing paths, falls back only for regular new-chat drafts to the current active project, preserves explicit worktree targets, and keeps unknown/offline probes unchanged.

Validation: 40 targeted tests passed; ESLint, TypeScript, and git diff --check passed. Full UI suite: 256/257 test files passed; the remaining Windows locale assertion is unrelated.

### DIFF
--- a/packages/ui/src/lib/opencode/client.test.ts
+++ b/packages/ui/src/lib/opencode/client.test.ts
@@ -9,12 +9,19 @@ let configCalls = 0;
 let runtimeKey = 'test-runtime';
 const promptAsyncCalls: unknown[][] = [];
 const promptAsyncResults: Array<unknown> = [];
+const pathGetResults: Array<unknown> = [];
 
 const promptAsyncMock = mock(async (...args: unknown[]) => {
   promptAsyncCalls.push(args);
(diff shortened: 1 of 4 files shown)


quseijuro-design:
Closing as superseded by #2928, which now implements the stale lastDirectory recovery flow upstream. Thanks for the review infrastructure.

PR #2897, opened 2026-08-14
fix(ui): scope theme settings per runtime instance

kydorn:
## Intent

Two same-origin windows pointed at different OpenChamber instances (Electron windows all load the UI from `openchamber-ui://`, and same-origin web paths share localStorage) shared a single theme: theme preferences were persisted to unscoped global localStorage keys (`themeMode`, `lightThemeId`, `darkThemeId` plus legacy `useSystemTheme`/`selectedThemeId`/`selectedThemeVariant`). Consequences: a window booted with the other instance's theme, the cross-window `storage` listener live-applied the other window's theme (both windows flip-flopped), and `persistToLocalStorage` clobbered the globals on every settings sync from any instance.

The change scopes theme preferences per runtime endpoint, mirroring the existing per-runtime settings-mirror pattern (`openchamber.settingsMirror.v2:<runtimeKey>`), so every window uses the settings of the instance it points at.

Implementation path:

- New `packages/ui/src/contexts/theme-storage.ts`: per-runtime scoped persistence (`openchamber.theme.v2:<runtimeKey>`), with validated reads (malformed payloads are failure, not authority) and transient-runtime-key guarding (uninitialized `url:default` / disconnected `mobile-disconnected` keys never participate in scoped storage).
- `packages/ui/src/contexts/ThemeSystemContext.tsx`: boots from `resolveThemePreferencesForRuntime` (scoped entry, else the superseded global keys as a one-time migration seed, else defaults); the persist effect writes the scoped key and maintains the splash-hint globals; the cross-window `storage` listener reacts only to the current runtime's key (same-instance windows keep live sync, different instances are isolated); runtime endpoint switches re-adopt the new instance's last-known theme.
- `packages/ui/src/lib/persistence.ts`: removed the redundant global theme-key writes — the per-runtime settings mirror already carries those fields, and theme persistence is now single-owner.
- `packages/ui/src/apps/mobileNativeChrome.ts`: the Android status bar re-applies its style/background whenever the root dark/light class flips (previously only on mount and app re-activation), so theme toggles update the status-bar text contrast immediately, without an app restart.

## Non-goals

- Other appearance settings (font size, fonts, padding, corner radius, terminal options) remain per-instance server settings; their shared client cache (`ui-store` zustand blob) is intentionally not re-scoped in this PR.
- Window-controls position/style stay instance-scoped server settings (no client-local chrome store introduced).
- The pre-React splash-hint globals (`themeMode`/`selectedThemeVariant`/`useSystemTheme`/`splashBg*`/`splashFg*`) stay as cosmetic last-writer-wins keys maintained by the persist effect — they feed the splash shells and the Android status bar and are explicitly not theme authority (per maintainer review). The Electron native splash reads per-instance server settings.

## Affected surfaces

- `packages/ui` — shared React UI consumed by web, Electron, VS Code, and mobile; theme persistence routes through this shared code in every runtime, so no runtime-specific branch was added.
- Persisted contract: theme authority moves to per-runtime `openchamber.theme.v2:<runtimeKey>`; transient runtime keys (uninitialized/disconnected) are excluded from scoped storage so cold-boot defaults can never pin a stale theme to the pre-connect phase. On first boot without a scoped entry, the legacy global keys (`themeMode`/`lightThemeId`/`darkThemeId`/`useSystemTheme`/`selectedThemeId`/`selectedThemeVariant`) are read once as a migration seed so no client-only theme state is discarded before the server sync lands. The globals are never deleted: the persist effect keeps them fresh as cosmetic last-writer-wins hints for the pre-React splash shells and the Android status bar. Older builds cannot read the scoped key and keep using the globals, so rollback works unchanged.
- `persistToLocalStorage` no longer writes theme globals; the settings mirror and server-side per-instance settings are unchanged.

## Repository guidance

|---|---|---|
| `AGENTS.md` | Theme persistence is shared-UI state crossing runtime boundaries; the change alters persisted state | Runtime-scoped keys follow the existing settings-mirror pattern; authoritative per-instance server settings still win via `settings-synced`; minimal diff with no unrelated refactors |
| `sync-state-invariants` (skill) | Runtime-scoped persisted state and cross-window reconciliation | Caches keyed by runtime identity; persisted payload shape validated before granting authority (malformed → null, never empty success); same-runtime windows still sync, cross-runtime writes are ignored; retention is explicit: intentionally unbounded (module comment) because entries are ~150 bytes and bounded by distinct instances ever visited |
| `theme-system` (skill) | Theme state and persistence are the subject of the change | Theme mode/light/dark semantics preserved; no token or visual changes |
| `openchamber-change-discipline` (skill) | Source change with a persisted-contract dimension and two new files | Smallest complete change; legacy keys kept as migration fallback for existing persisted data; removed writes had no remaining in-app reader (verified by grep); dead-code and package-scoped checks run |
| `ui-api-decoupling` (skill) | Runtime key resolution and endpoint-switch safety | Runtime key resolved at call time in every handler (no cached keys); host switches re-read the new runtime's scoped preferences |
| Not applicable | `settings-ui-patterns` (no Settings UI changed), `locale-ui-patterns` (no user-facing strings), `desktop-shell`/`relay-transport` (no Electron/transport code changed) | — |

## Validation

| Check | Result |
|---|---|
| `bun test contexts/theme-storage.test.ts contexts/ThemeSystemContext.test.ts lib/persistence.test.ts` (packages/ui) | 60 pass, 0 fail — 23 theme-storage tests: per-runtime isolation, round-trip, malformed-payload rejection, splash-hint globals preserved on scoped write, the migration-seed resolution chain (scoped → legacy → defaults), runtime-switch adoption (adopt-or-keep), transient-runtime-key guarding (uninitialized/disconnected keys never read or write scoped entries — the mobile connect splash falls back to the global hints), and storage-event resolution (current-runtime adoption, foreign-runtime and legacy-key ignoring, no-change identity check) |
| `bun test` (full packages/ui suite) | 1981 pass / 202 fail — failure count identical with and without this change (pre-existing environment failures, e.g. codemirror bundling); no change-induced failures |
| `bun run type-check` (packages/ui) | clean |
| `bun run lint` (packages/ui) | clean |
| `bun run dead-code` (workspace) | no findings for the changed/new files |
| Manual: two windows against two loopback instances | Each window follows its own instance's theme; no cross-window flip-flop (see visual evidence) |
| Manual: packaged-Electron live host switch | Theme follows each instance through host switches in the packaged app |
| Manual: Android device | Light/dark toggle updates the status-bar text contrast immediately; cold-boot splash and connecting phases follow the persisted theme (transient-key fix verified on device) |

Not verified: iOS status-bar behavior (unchanged by this PR — the live adaptation is Android-only; iOS keeps `Style.Default` with overlay).

## Visual evidence

Two same-origin windows pointed at different instances: each window stays on its own instance's theme simultaneously — no flip-flopping or shared theme.

## Risks and failure behavior

- Migration: on first boot after upgrade, an instance without a scoped entry is seeded from the global keys (last-known theme) into its scoped entry; the globals stay as splash hints. A flash exists only for instances never visited before this change, then authoritative server values apply. Malformed scoped payloads are rejected (null) — defaults + server sync apply, no crash, no stale authority.
- Cross-window behavior: same-instance windows keep live theme sync (storage listener keyed per runtime); different-instance windows are fully isolated.
- Rollback: older builds keep reading/writing the global keys, which remain maintained, so a downgrade works unchanged.
- No secrets, no new dependencies, localStorage only.
- Race audit (2+ windows): adoption writes are no-ops (identical-value `setItem` fires no event) and the reducer identity check returns the same state — every cross-window path converges. The only transient race is a boot-time rollback of a concurrent local theme change by another window's in-flight server fetch (pre-existing; the old code had more writers, so it was worse before).

### DIFF
--- a/packages/ui/src/apps/mobileNativeChrome.ts
+++ b/packages/ui/src/apps/mobileNativeChrome.ts
@@ -70,6 +70,27 @@ export const useNativeMobileChrome = (): void => {
       const retry = window.setTimeout(() => void applyStatusBar(), 400);
       cleanup.push(() => window.clearTimeout(retry));
 
+      // Theme toggles must reach the status bar without an app restart: re-run
+      // whenever the root dark/light class flips — the one signal every theme
+      // path converges on (settings toggle, synced settings, storage events,
+      // system-preference changes while in system mode). splashBg* colors are
+      // per-variant values, so they are stable across mode toggles.
+      if (platform === 'android') {
+        let wasDark = root.classList.contains('dark');
+        const themeClassObserver = new MutationObserver(() => {
+          const isDark = root.classList.contains('dark');
+          if (isDark === wasDark) return;
+          wasDark = isDark;
+          void applyStatusBar();
+        });
+        themeClassObserver.observe(root, { attributes: true, attributeFilter: ['class'] });
+        if (disposed) {
+          themeClassObserver.disconnect();
+          return;
+        }
+        cleanup.push(() => themeClassObserver.disconnect());
+      }
+
       const { App } = await import('@capacitor/app');
       const stateHandle = await App.addListener('appStateChange', ({ isActive }) => {
         if (isActive) void applyStatusBar();
--- a/packages/ui/src/contexts/ThemeSystemContext.tsx
+++ b/packages/ui/src/contexts/ThemeSystemContext.tsx
@@ -32,6 +32,11 @@ import {
 import { isValidTheme } from './theme-validation';
 import { getSyncedThemeFromPayload, getSyncedThemeVariant } from './theme-sync-payload';
 import { getRuntimeKey, subscribeRuntimeEndpointChanged } from '@/lib/runtime-switch';
+import {
+  getThemePreferencesStorageKey,
+  readThemePreferencesForRuntime,
+  writeThemePreferencesForRuntime,
+} from './theme-storage';
 
 type ThemePreferences = {
   themeMode: ThemeMode;
@@ -88,9 +93,10 @@ const buildInitialPreferences = (defaultThemeId?: string): ThemePreferences => {
     const embeddedMode = embeddedParams?.get('themeMode');
     const embeddedLightId = embeddedParams?.get('lightThemeId');
     const embeddedDarkId = embeddedParams?.get('darkThemeId');
-    const storedMode = localStorage.getItem('themeMode');
-    const storedLightId = localStorage.getItem('lightThemeId');
-    const storedDarkId = localStorage.getItem('darkThemeId');
+    const storedPreferences = readThemePreferencesForRuntime(getRuntimeKey());
+    const storedMode = storedPreferences?.themeMode ?? null;
+    const storedLightId = storedPreferences?.lightThemeId ?? null;
+    const storedDarkId = storedPreferences?.darkThemeId ?? null;
     const legacyUseSystem = localStorage.getItem('useSystemTheme');
     const legacyThemeId = localStorage.getItem('selectedThemeId');
     const legacyVariant = localStorage.getItem('selectedThemeVariant');
@@ -315,6 +321,9 @@ export function ThemeSystemProvider({ children, defaultThemeId }: ThemeSystemPro
     customThemesRequestRef.current += 1;
     setCustomThemes([]);
     setCustomThemesLoading(false);
+    // Adopt the new instance's last-known theme immediately; the incoming
+    // settings sync refines it with the server's authoritative value.
+    setPreferences((prev) => readThemePreferencesForRuntime(detail.runtimeKey) ?? prev);
     void reloadCustomThemes();
   }), [isVSCode, reloadCustomThemes]);
 
@@ -425,6 +434,12 @@ export function ThemeSystemProvider({ children, defaultThemeId }: ThemeSystemPro
       return;
     }
 
+    writeThemePreferencesForRuntime(getRuntimeKey(), {
+      themeMode: preferences.themeMode,
+      lightThemeId: preferences.lightThemeId,
+      darkThemeId: preferences.darkThemeId,
+    });
+
     localStorage.setItem('themeMode', preferences.themeMode);
     localStorage.setItem('lightThemeId', preferences.lightThemeId);
     localStorage.setItem('darkThemeId', preferences.darkThemeId);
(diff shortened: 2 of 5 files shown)


kydorn:
The PR body has been updated — visual evidence (two-window screenshots) is now attached to the Visual evidence section. Please re-check the PR body text along with the review; the previous round predated the evidence.

kydorn:
Addressed both carried-forward findings in a06012d17: retention is now explicit (module comment: intentionally unbounded, ~150-byte entries bounded by distinct instances visited), and the storage-event decision is extracted into a pure, unit-tested function — including regression tests that legacy/global keys are ignored and foreign-runtime events are ignored.

kydorn:
Addressed the BLOCKED verdict in 6eb0e0ceb: the legacy globals are restored as a one-time migration seed in buildInitialPreferences (scoped key → legacy globals → defaults), so the first scoped write carries the last-known theme and the legacy keys are only removed after that seed — no client-only theme state is discarded before the server sync lands. The PR body has also been rewritten to describe the current HEAD (single-owner scoped persistence, splash fallback behavior, rollback semantics, updated test counts); the visual evidence screenshot remains in place.

kydorn:
Addressed the migration-seed non-blocker in 0b1f9ac9a: the boot resolution chain (scoped entry → one-time legacy seed → defaults) is extracted into the pure, unit-tested `resolveThemePreferencesForRuntime` in theme-storage.ts, and `buildInitialPreferences` delegates to it. Five new tests pin the chain — scoped wins, legacy mode/ids seed, useSystemTheme/selectedThemeId chain, defaults fallback, and the seed surviving into the scoped key with legacy keys removed — so a regression that drops the seed (the previously blocked data-loss path) now fails tests.

btriapitsyn:
The multi-instance theme bleed is real and scoping by runtime key is the right shape. The `theme-storage` module and its tests are solid. Two things block the merge:

1. **Pre-React splash theming regresses for everyone.** The splash scripts in `packages/web/index.html`, `mobile.html`, and `mini-chat.html` read `themeMode`/`selectedThemeVariant` to paint the correct theme before React loads, and this PR both stops writing those keys and deletes them via `LEGACY_THEME_KEYS`. A dark-theme user on a light-OS machine gets a light flash on every boot. These keys were not part of the flip-flop bug (the bug was the storage listener and `persistToLocalStorage` writers). Done looks like: either keep writing `themeMode`/`selectedThemeVariant`/`splashBg*`/`splashFg*` as cosmetic last-writer-wins hints (and drop them from `LEGACY_THEME_KEYS`), or scope them and update the three HTML shells to read the scoped entry.
2. **Android status-bar color loses its source.** `packages/ui/src/apps/mobileNativeChrome.ts` reads `splashBgDark`/`splashBgLight`; deleting those keys drops it to the hardcoded fallback. Same fix as item 1 covers it, just make sure this reader is accounted for.

With those two addressed this merges: the scoped-key design, the migration seed, and the storage-event isolation can stay exactly as they are.

kydorn:
Addressed both merge blockers in 492dfa664:

1. **Splash theming restored.** The persist effect again maintains `themeMode`/`lightThemeId`/`darkThemeId`/`useSystemTheme`/`selectedThemeId`/`selectedThemeVariant`/`splashBg*`/`splashFg*` as cosmetic last-writer-wins hints for the pre-React splash shells, and `LEGACY_THEME_KEYS`/the removal loop is gone — the scoped key is theme authority, the globals are never deleted. Rebased onto current `upstream/main` (v1.21.0).
2. **Android status bar** (`mobileNativeChrome.ts`) keeps its `splashBg*` source since the keys are maintained again.

Also extracted the runtime-switch adoption into the pure, unit-tested `adoptThemePreferencesForRuntime` (clears the carried-forward non-blocker). PR body updated to describe the current HEAD; the evidence screenshot is unchanged.

btriapitsyn:
Thanks for fixing this — two windows swapping themes was a really confusing one to hit. Merged; I'll tidy a couple of small repo-convention things on main.

btriapitsyn (approved):

kydorn:
PR body updated for the current HEAD: the Android status-bar adaptation (e970e7300) is described, transient-key guarding and the splash-hint restoration are covered, and the not-verified disclosure is corrected — packaged-Electron live host switching has now been manually verified on device (theme follows each instance through switches), so that residual-risk item no longer applies.

PR #2902, opened 2026-08-14
feat(usage): expand overview dashboard for all providers and periods

sdziupin:
Add cursor-paginated usage history, custom period stats, improved charts, and locale coverage so plugin/custom providers get historical analytics without new DB or deps.

## Repository guidance

|---|---|---|
|  |  |  |

## Validation

| Check | Result |
|---|---|
|  |  |


btriapitsyn:

btriapitsyn:
Closing this one. The idea is good — a usage overview with real per-period spend across every provider is something people ask for — but the way the history is computed is not something we can ship.

`useUsageHistory` walks every active and archived session and pages through the full `session.messages` payload of each one, client-side, on the first visit to the Usage page, and again after every reload because the cache lives in memory only. For anyone with a few hundred sessions that is thousands of requests carrying entire conversations just to sum `cost` and `tokens` — over a relay or a remote instance it is minutes of traffic, and it competes with the chat's own bootstrap for the same handful of browser sockets. We have already had a startup-freeze bug from a much smaller fan-out than this.

If we do usage analytics, the aggregation has to happen once, on the OpenChamber server, which already sees the same sessions — not in every browser tab. Thanks for the work on the charts and the period selector; that part is worth keeping when the data path is right.

PR #2905, opened 2026-08-14
fix(sync): preserve message chronology across OpenCode ID wraps

### LINKED ISSUE #2904: [Bug] Message ID rollover breaks message ordering, pagination, and revert boundaries
## Summary

Message IDs are currently treated as chronological clocks in several message paths. That fails at the encoded timestamp rollover: a message with ID `msg_ffff...` can be created immediately before one with ID `msg_0000...`, while their `time.created` values correctly increase. Lexical ID comparison then reverses their real order.

IDs should remain stable identity keys and equal-time tie-breakers; `time.created` should define chronology.

## Reproduction

1. Materialize two messages in one session:
   - `msg_ffff...`, `time.created = 1000`
   - `msg_0000...`, `time.created = 1001`
2. Load the timeline, paginate across that boundary, then revert at either message.

## Expected

The `msg_ffff...` message stays before `msg_0000...` everywhere because its `time.created` is earlier. Pagination neither skips nor duplicates messages, and revert hides/restores the correct chronological branch.

## Actual

Lexical ID ordering places `msg_0000...` first. This can reorder the timeline, select the wrong pagination boundary, and make revert filtering or restoration include the wrong messages.

This is distinct from #2088 (overlap deduplication) and #2195 (client/server clock skew): the failure is the ID timestamp field wrapping even when `time.created` is correct.

## Acceptance criteria

- [ ] Message chronology uses `time.created`, with ID only as a deterministic tie-breaker when timestamps are equal.
- [ ] Timeline materialization and pagination preserve the rollover order without skips or duplicates.
- [ ] Revert visibility, replacement, and undo/redo behavior use the same chronological rule.
- [ ] Tests cover `msg_ffff...` immediately before `msg_0000...`, including a pagination boundary and revert.

Maksym-Vakhonin:
Fixes #2904.

## Intent

Preserve chronological message ordering when OpenCode's 48-bit ID time component wraps.

[OpenCode](https://github.com/anomalyco/opencode/blob/dev/packages/opencode/src/id/id.ts) encodes `(Date.now() * 0x1000 + counter)` into six bytes, so the encoded clock wraps every `2^36` milliseconds. OpenChamber was re-sorting server-ordered pages by raw message ID and using raw ID comparisons for merges, live updates, and revert boundaries. Across a wrap, a newer message therefore sorted before older messages.

This change:

- orders messages by `time.created`, with `id` only as a deterministic same-millisecond tie-breaker;
- removes message-level binary searches and revert/undo/redo comparisons that assumed globally monotonic IDs;
- adds deterministic regression fixtures spanning the 2026-08-14 wrap.

## Non-goals

- Change OpenCode's ID generation.
- Change ordering for sessions, parts, permissions, or questions.
- Change persisted data, API contracts, or UI layout.

## Affected surfaces

- `@openchamber/ui` message loading, materialization, optimistic merging, and event reduction.
- Revert, undo/redo, user history, and the reverted-message dock.
- Web, Electron, mobile, and VS Code consume the shared UI implementation; no runtime-specific contract changes are required.

## Repository guidance

|---|---|---|
| Root `AGENTS.md` and change-discipline guidance | The change modifies shared executable UI code | Scope is limited to message ordering, regression tests were added, and all required validation was run |
| Sync-state invariants and `sync/DOCUMENTATION.md` | Message arrays are shared sync state | Chronological ordering is now explicit and documented; deduplication remains ID-based |
| Performance guidance | Binary searches were replaced where their ordering prerequisite was invalid | Common tail ownership checks retain an O(1) fast path; insertion and fallback lookup operate on bounded session pages |
| Writing guidance | The sync invariant and PR behavior need to be maintainable | Documentation states the ordering key, wrap period, and affected operations explicitly |

## Validation

| Check | Result |
|---|---|
| Focused Bun tests for ordering, loader, reducer, revert/history, and sync records | 126 passed |
| `node ../../scripts/run-isolated-tests.mjs src` from `packages/ui` | 258/258 test files passed |
| `./node_modules/.bin/vitest run` from `packages/web` | 139 files passed; 1257 tests passed, 1 skipped |
| Remaining root, Electron, and VS Code suites | Passed, including loopback-dependent fixtures rerun outside the sandbox |
| TypeScript checks for UI, Web, Electron, mobile, and VS Code | Passed |
| ESLint for UI, Web, mobile, and VS Code | Passed |
| Production builds for Web, VS Code, UI, and mobile assets | Passed |
| `node scripts/docs/validate-docs.mjs` | 450 pages and 45 sidebar links validated |
| `npx --yes knip@5.80.0 --no-exit-code --include files,exports,nsExports,types,nsTypes,enumMembers,duplicates` | Completed; neither the new module nor `compareMessages` was reported |
| Full local server smoke test | Production UI loaded successfully with a managed OpenCode server |

No dedicated repository e2e script covers this sync-only boundary case.

## Visual evidence

There is no static layout or styling change. A fresh production build was served through the full OpenChamber runtime and the main UI was visually smoke-tested. The user-visible ordering behavior is covered by deterministic pre-wrap/post-wrap message fixtures because a normal screenshot cannot represent the synthetic epoch boundary.

## Risks and failure behavior

Message lookup and insertion no longer rely on binary search over raw IDs because that ordering is invalid after a wrap. Insertions sort bounded message arrays by `time.created`, and the frequent ownership check retains a last-message fast path.

If a stored revert marker cannot be found, the client preserves the current message sequence instead of trimming messages using an invalid lexical guess. No data migration or destructive operation is involved.

The reproduction branch's two dock assertions exclude the revert marker itself. Existing `main` tests and the stepwise restore flow intentionally treat that marker as the first reverted record. This PR preserves that established behavior and changes only how the chronological boundary is located.


Maksym-Vakhonin:
Superseded by 53830952, which landed the same message-ordering fix in `main` and also covers part ID rollover. Closing this PR to avoid duplicating the upstream implementation.

PR #2906, opened 2026-08-14
fix(ui): order chat messages by time.created, not raw ID

Dawnfz-Lenfeng:
## Problem

OpenCode message IDs embed `Date.now() * 0x1000 + counter` into a **truncated 6-byte (48-bit) hex prefix** (`msg_<12 hex><14 base62>`). That ID space wraps around every 2^36 ms (~**795 days**), and the latest wrap occurred **2026-08-14 11:19:55 UTC**.

OpenChamber's sync layer ordered messages by **string-comparing raw IDs** (`cmp(a.id, b.id)`) in several places. After the wrap, any message created *after* the boundary gets an ID prefix near `0000…`, which lexicographically sorts **before** all older IDs (`e3…`–`ffff…`). Result: in any session spanning the boundary, the **newest messages render at the top of the chat**.

I verified this against a real database: the same session contained `msg_fffc192a8001…` (2026-08-14 10:11 UTC, before the wrap) and `msg_0005afbc2002…` (12:59 UTC, after the wrap) — the raw-ID sort places the 12:59 messages before the 10:11 ones.

## Fix

Order messages chronologically using `time.created` (present on every message, including optimistic client inserts), with `id` only as a same-millisecond tiebreaker:

- **New `sync/messageOrder.ts`** — `compareMessages`, `sortMessages`, `findMessageByID`, and revert-anchor helpers (`isBeforeMessage`, `isAtOrAfterMessage`, `isAfterMessage`) that fall back to raw ID comparison when the anchor message has been optimistically removed from the store.
- **`Binary.searchBy()`** — binary search against a full ordering comparator (returns insertion index).
- **Sorts/searchs converted to chronological order** in `optimistic.ts` (`mergeMessages`, `mergeOptimisticPage`), `session-message-loader.ts` (fetched pages, optimistic insert), `materialization.ts` (snapshot merge), and `event-reducer.ts` (`message.updated`).
- **Id-only lookups** (`message.removed`, `hasMessage`, revert anchors) use bounded linear scans since a chronological array can't be binary searched by id alone.
- **Revert-point comparisons** (`session-actions.ts`, `session-ui-store.ts`, `sync-context.tsx`, `user-message-history.ts`, `revertedMessageDockState.ts`) now use the chronological comparator.
- **Parts remain id-ordered** — they're scoped to a single message, so the wrap cannot reorder them.

## Tests

New `sync/messageOrder.test.ts` covers the wrap-boundary regression at the comparator, merge, optimistic-merge, and materialization levels, plus the revert helpers. Full `src/sync` suite (52 files) and `src/components/chat` suite (53 files) pass; `tsc --noEmit` and eslint are clean.

### DIFF
--- a/packages/ui/src/components/chat/revertedMessageDockState.ts
+++ b/packages/ui/src/components/chat/revertedMessageDockState.ts
@@ -1,4 +1,5 @@
 import type { Message, Part } from '@opencode-ai/sdk/v2/client';
+import { findMessageByID, isBeforeMessage } from '@/sync/messageOrder';
 import type { State } from '@/sync/types';
 
 type RevertedMessageRecord = {
@@ -50,9 +51,10 @@ export const buildRevertedMessageDockState = (
     }
 
     const messages = state.message[sessionId] ?? [];
+    const revertAnchor = findMessageByID(messages, revertMessageID);
     const records: RevertedMessageRecord[] = [];
     for (const message of messages) {
-        if (!isUserMessage(message) || message.id < revertMessageID) {
+        if (!isUserMessage(message) || isBeforeMessage(message, revertAnchor, revertMessageID)) {
             continue;
         }
         records.push({
--- a/packages/ui/src/sync/binary.ts
+++ b/packages/ui/src/sync/binary.ts
@@ -23,4 +23,32 @@ export namespace Binary {
 
     return { found: false, index: left }
   }
+
+  /**
+   * Binary search against a full ordering comparator (e.g. messages ordered
+   * by `time.created`, then id). Returns the insertion index when absent.
+   */
+  export function searchBy<T>(
+    array: readonly T[],
+    value: T,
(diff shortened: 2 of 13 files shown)


Dawnfz-Lenfeng:
Closing as a duplicate: the same fix already landed on main today as commit `53830952` (`fix(chat): keep messages chronological across ID rollover`) — this PR was authored independently (before that commit existed) but covers the same root cause: OpenCode's 48-bit ID prefix wraps every ~795 days, so raw-ID sorting places post-wrap messages at the top; the fix orders messages by `time.created` with id as tiebreaker.

The upstream commit additionally handles a case mine treated more conservatively (`message.updated` re-inserts chronologically when the same id moves in time), and includes its own regression tests, so nothing from this PR is needed on top.

The root-cause issue on the opencode side (the ID generator's 6-byte truncation itself) remains open: https://github.com/anomalyco/opencode/issues/42589

PR #2908, opened 2026-08-14
fix(electron): prefer process.cwd() when it looks like a project root (closes #1392)

### LINKED ISSUE #1392: [Bug] Openchamber starts with user home dir

I'm using memsearch (https://github.com/zilliztech/memsearch) as plugin in opencode.
When starting openchamber, it automatically starts with worktree = my homedir.
This triggers memsearch to index in my homedir which isn't wanted.

When using opencode cli, opencode sets current dir as workdir. Memsearch then indexes only inside the current dir/project dir.

Can we have one of this options?
- Have a config flag to not automatically use user's home dir as worktree
- Use cwd as first worktree when starting openchamber
- Have a parameter to explicitly pass the workdir/worktree

### Where does it happen?

Mobile (Web/PWA)

### Version (if known)

latest

_No response_

### Logs (optional)

```shell

```

herjarsa:
## Intent

Closes #1392. Related to #2784.

`opencode serve` is launched by the desktop app with `cwd` resolved from `resolveManagedOpenCodeCwd`. When `OPENCHAMBER_OPENCODE_CWD` is unset and the calling process's cwd is not the user's home directory, this function falls back to `homedir()` instead of preferring the project root. The result: `opencode` re-indexes the user's home directory recursively on every restart, leaks memory, and leaves zombie processes that break the proxy health-check (`periodic health check failed (N/20)` every 15s, `socket hang up` / `ECONNREFUSED` on the internal proxy port).

This PR makes `resolveManagedOpenCodeCwd` prefer `process.cwd()` first when it contains any standard project marker (`.git`, `package.json`, `Cargo.toml`, `go.mod`, `pyproject.toml`). Only falls back to `homedir()` when cwd is not a project root, and emits a single `console.warn` the first time so users notice the misconfiguration without log spam.

## Why not silence instead of fix

#2784 explicitly documents that the "\not a git repository" warning from `getWorktrees` is intentional per the `opencode-cwd.mjs` design (managed OpenCode runs with `cwd: <homedir>`). That PR silences the symptom; this one addresses the root cause by changing what cwd the managed process receives. Silencing alone would still leave the memory leak and the zombie-process lifecycle intact.

## Non-goals

- No removal of `OPENCHAMBER_OPENCODE_CWD` env var; it stays as the first-priority override and is what users/admins set when they want a non-project explicit path (containers, custom configs).
- No behavior change for users who already set `OPENCHAMBER_OPENCODE_CWD` — the env var path is unchanged.
- No change to the public function signature.
- No upstream-caller fix in this PR; this is the function itself.

## Affected surfaces

| Surface | Effect |
|---|---|
| `packages/electron/opencode-cwd.mjs` | Add `existsSync` import and a `looksLikeProjectRoot` helper; reorder the resolution so project-root `process.cwd()` wins over `homedir()`; add a one-shot `console.warn` flag with a test-only `__resetCwdFallbackWarning` export. |
| `packages/electron/opencode-cwd.test.mjs` | Mock `node:fs`, stub `process.cwd()` and `console.warn`, expand from 3 to 6 tests covering: defaults → home, env override wins, blank env falls through, project cwd preferred, single warning across repeated calls, warning content. |

## Validation

- `./packages/web/node_modules/.bin/vitest run packages/electron/opencode-cwd.test.mjs --root packages/electron` → **6 passed / 0 failed**.
- Existing tests unchanged in semantics; the test for "defaults to home" now also stubs `process.cwd()` to a non-project path so the new behavior doesn't regress it.

## Risk and failure behavior

- **Behavior change**: only the resolution order between `process.cwd()` and `homedir()`; the env var path is untouched.
- **False positive**: a non-project directory that happens to have any of the listed markers will be used. The markers (`.git`, `package.json`, `Cargo.toml`, `go.mod`, `pyproject.toml`) are conventional enough that this is the right trade-off; users with a true mismatch can still set `OPENCHAMBER_OPENCODE_CWD`.
- **Warning spam**: prevented by the once-per-process flag; tests reset the flag explicitly so test ordering doesn't matter.
- **Rollback**: revert the commit; behavior reverts to always preferring `homedir()` without any persisted state change.

## Reproduction evidence (before)

````
# From a non-project cwd with OPENCHAMBER_OPENCODE_CWD unset:
$ openchamber  # launches opencode serve with cwd = ~/.home
```

After this PR the same launch resolves to the project cwd when one is detected; only falls back to home when the cwd is genuinely not a project root, with one warning telling the user how to silence it.

### DIFF
--- a/packages/electron/opencode-cwd.mjs
+++ b/packages/electron/opencode-cwd.mjs
@@ -1,3 +1,30 @@
+import { existsSync } from 'node:fs';
+
+// Markers that identify a directory as a project root. If process.cwd() matches
+// any of these we use it directly instead of falling back to HOME. Without
+// this, opencode re-indexes the user's home directory recursively on every
+// restart, leaking memory and blocking the proxy health-check.
+const PROJECT_MARKERS = [
+  '.git',
+  'package.json',
+  'Cargo.toml',
+  'go.mod',
+  'pyproject.toml',
+];
+
+const looksLikeProjectRoot = (dir) => {
+  if (typeof dir !== 'string' || !dir) return false;
+  return PROJECT_MARKERS.some((marker) => {
+    try {
+      return existsSync(`${dir}/${marker}`) || existsSync(`${dir}\\${marker}`);
+    } catch {
+      return false;
+    }
+  });
+};
+
+let warnedAboutHomeFallback = false;
+
 export const resolveManagedOpenCodeCwd = ({ env, homedir }) => {
   const configured = typeof env?.OPENCHAMBER_OPENCODE_CWD === 'string'
     ? env.OPENCHAMBER_OPENCODE_CWD.trim()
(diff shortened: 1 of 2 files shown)


btriapitsyn:
Thanks for digging into #1392, but this fix targets the wrong layer. The managed `opencode serve` cwd is intentionally the home directory, and for a packaged desktop app `process.cwd()` is `/` or a system directory, so the marker heuristic only changes behavior for terminal launches, where it pins the server to whatever repo the shell happened to be in. Guessing a project root from filesystem markers trades a predictable default for an unpredictable one, so I'm closing this.

The real ache in #1392 is which project the UI opens on start, and a fix there, in the startup directory selection rather than the managed process cwd, would be welcome.

PR #2914, opened 2026-08-14
fix(ui): stop the context meter from counting every internal round-trip

### LINKED ISSUE #2562: [Bug] Wrong context window usage graph when loading old session

This morning, I started VS Code and re-opened the session I was working on yesterday.  If I remember correctly, the context window usage was twenty something percent, but today, it became 115%.The model I have been using is DeepSeek V4 Pro on Go, whose context window size is 1M, according to OpenChamber's own info. I restarted VS Code, and now it's 30%.

My guess is there is some sort of but that gets context window size wrong on certain conditions of which I'm not sure.

### Where does it happen?

VS Code extension

### Version (if known)

1.17.1

When I first opened VS Code this morning:
Image

After a restart:
Image

### Logs (optional)

```shell

```

pocharlies:
## Intent

The context meter can show impossible values — 330.6% of a 1M window on mobile web, 115% in #2562 on VS Code — after an assistant turn with many tool calls, and jumps when reopening an older session. Every context-usage surface computed "window fill" by summing the message token breakdown (`input + output + reasoning + cache.read + cache.write`). Those fields **accumulate across every API round-trip inside a single turn** (each tool call re-reads the whole cached prompt), so a turn with ~14 tool calls sums to ~14× the real window.

Real payload captured from a stock opencode 1.18.18 server (`GET /session/:id/message`), the exact turn that rendered `330.6% · 3.3M/1.0M`:

| field | value |
|---|---|
| `tokens.total` (server-reported, final round-trip) | 232,872 → **23.3%** of the 1M window |
| `tokens.cache.read` (accumulated across round-trips) | 3,291,956 |
| `tokens.output` | 14,523 |
| sum the UI displayed | 3,306,479 → **330.6%** |

A single-step turn from the same server has `total` equal to the summed fields exactly (117,714 = 1,116 + 87 + 543 + 115,968), confirming the fields only diverge when round-trips accumulate.

The fix: prefer the server-reported `tokens.total` — the final round-trip's window — everywhere the fill is displayed, and fall back to the previous summing when a server does not send it. A new `contextTokensFromBreakdown` in `tokenUtils` owns that rule. Five call sites delegate to the helper (the context store extractor `extractTokensFromMessage`, `session-ui-store.getContextUsage`, the work status panel `contextUsage.ts`, the VS Code layout header, and mini chat), and two surfaces whose inputs are `unknown`-sanitized apply the identical `total > 0 ? total : sum` rule inline (the context sidebar tab and mobile session metadata).

`tokens.total` is optional in the opencode message schema, and the SDK pinned by this branch (`@opencode-ai/sdk@1.18.18`) declares `total?: number` on `AssistantMessage.tokens` and `StepFinishPart.tokens` in the `/v2` generated types this code imports. (An earlier revision of this description claimed the generated types lacked the field; that was read from a stale 1.18.16 install and was wrong.) The local defensive shapes remain for the `unknown`-typed call sites and older servers that do not send the field.

Fixes #2562 — same mechanism: reopening a session renders the last assistant message's accumulated breakdown, and readings over 100% are impossible as a real window fill.

## Non-goals

- Cost/usage accounting and the cache-hit-rate display keep using the raw breakdown fields — accumulated numbers are correct for those. `sumTokenBreakdown` and `computeCacheHitRate` are unchanged.
- How the context *limit* is resolved is unchanged.
- The context sidebar's "Last turn tokens" grid intentionally keeps showing the raw per-round-trip fields; only the headline usage (and the message chips' `total`) use the reported total. Declared consequence: on multi-step turns the grid's cache-read figure can exceed the headline total — they are different quantities (cumulative round-trip I/O vs. window fill).

## Affected surfaces

`packages/ui` only; every runtime consumes it through the shared UI: desktop/web header (contextStore + session-ui-store), work status panel, context sidebar tab, VS Code layout header, mini chat window, and mobile session metadata. No runtime-specific code changed, no persisted format or external contract changes — the message token payload is read-only input. `packages/vscode/CHANGELOG.md` gains an `[Unreleased]` entry because the VS Code header readout is affected (that is #2562's surface).

## Repository guidance

|---|---|---|
| `openchamber-change-discipline` | executable source change across shared UI | smallest complete change: one helper owns the invariant, five call sites delegate to it, two `unknown`-sanitized surfaces inline the identical rule; fallback preserves old behavior byte-for-byte; `bun run dead-code` run for the new export |
| `sync-state-invariants` | touches `session-ui-store.getContextUsage` | follows "prefer deterministic authoritative records over heuristics": the server-reported total replaces a client-side summing heuristic; no ownership, lifecycle, or reconciliation semantics change |
| `changelog-authoring` | user-visible fix | `[Unreleased]` bullets added to the main and VS Code changelogs, symptom-focused, with contributor credit |
| `packages/ui/src/components/chat/work-status/DOCUMENTATION.md` | documents why `contextUsage.ts` computes usage itself | that rationale (directory-scoped computation) is untouched; the code comments that asserted "each assistant turn reports the whole window it saw" were corrected, since it does not hold for the summed fields on multi-step turns |

## Validation

| Check | Result |
|---|---|
| `bun test src/stores/utils/tokenUtils.test.ts src/components/chat/work-status/contextUsage.test.ts` | 35 pass, 0 fail — includes regression fixtures with the real captured payloads (the 330.6%→23.3% turn and the single-step turn where total equals the sum) |
| `bun run --cwd packages/ui test` (full isolated suite) | 263/263 test files passed |
| `bun run type-check` (workspace-wide) | all packages exit 0 |
| `bun run lint:ui` | clean |
| `bun run dead-code` | no findings for the changed/added symbols |
| Runtime render (see visual evidence) | web dev build of this branch and of `origin/main`, same session from the same live storage, rendered side by side |
| Not verified | full `bun run build`; per-surface screenshots beyond the desktop web header (identical store value / identical rule) |

## Visual evidence

Same session, same live storage, same moment, rendered by both builds (web dev build, desktop layout header — screenshots clipped to the header strip). The session's last reporting assistant turn sums to **18,611,976** tokens across its tool-call round-trips while reporting `total: 78,327` on a 1M-window model.

**Before — `origin/main` (`0d44041ef`):** the meter shows the capped 999%:

before: context meter reads 999.0%

**After — this branch:** the same message renders the real fill:

after: context meter reads 7.8%

The `330.6% · 3.3M/1.0M` capture referenced in Intent is the same class of readout on the mobile web header, taken earlier in the same session (it has since grown; its current unfixed readout is the capped 999.0% above). Issue #2562's screenshots show the same before-state on VS Code. The remaining surfaces render the same store value or apply the identical rule and were not individually screenshotted.

## Risks and failure behavior

- Servers that do not send `tokens.total` (older opencode) keep the previous behavior byte-for-byte: the fallback is the exact old sum.
- A malformed `total` (zero, negative, `NaN`, non-number) falls back to the sum — unit-tested.
- If a server reported a `total` inconsistent with its own breakdown, the meter now trusts the server; that is the intended authority order (authoritative record over client heuristic).
- No writes anywhere: a wrong value can only render a wrong percentage.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ All notable changes to this project will be documented in this file.
 - Chat: new chats no longer start against a deleted last worktree directory; they fall back to the active project instead of saving the first message and never starting.
 - Chat: opening a busy subagent in the context panel now shows its history instead of only the working-status line (thanks to @makeittech).
 - Chat: saved chats in the context panel open again instead of staying blank.
+- Chat: the context meter no longer climbs over 100% (330% readouts) after turns with many tool calls and no longer jumps when reopening an older session; it now shows what the window actually holds, everywhere the value appears — header, context sidebar, work status panel, mini chat, and mobile.
 - Projects: project names now match the folder name exactly, so `.ssh` and `opencode-claude` are no longer shown as `.Ssh` and `Opencode Claude` in the sidebar, window title, settings and notifications; names you renamed yourself are kept.
 - Settings: the session retention action you pick is now saved instead of being dropped (thanks to @Gautam0507).
 - Browser: typing a comment on a page no longer triggers app shortcuts.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ All notable changes to this project will be documented in this file.
 - Chat: new chats no longer start against a deleted last worktree directory; they fall back to the active project instead of saving the first message and never starting.
 - Chat: opening a busy subagent in the context panel now shows its history instead of only the working-status line (thanks to @makeittech).
 - Chat: saved chats in the context panel open again instead of staying blank.
-- Chat: the context meter no longer climbs over 100% (330% readouts) after turns with many tool calls and no longer jumps when reopening an older session; it now shows what the window actually holds, everywhere the value appears — header, context sidebar, work status panel, mini chat, and mobile.
+- Chat: the context meter no longer climbs over 100% (330% readouts) after turns with many tool calls and no longer jumps when reopening an older session; it now shows what the window actually holds, everywhere the value appears — header, context sidebar, work status panel, mini chat, and mobile (thanks to @pocharlies).
 - Projects: project names now match the folder name exactly, so `.ssh` and `opencode-claude` are no longer shown as `.Ssh` and `Opencode Claude` in the sidebar, window title, settings and notifications; names you renamed yourself are kept.
 - Settings: the session retention action you pick is now saved instead of being dropped (thanks to @Gautam0507).
 - Browser: typing a comment on a page no longer triggers app shortcuts.
--- a/packages/ui/src/apps/MobileSessionMetadata.tsx
+++ b/packages/ui/src/apps/MobileSessionMetadata.tsx
@@ -388,6 +388,7 @@ export const MobileSessionMetadataButton = React.memo(function MobileSessionMeta
     for (let i = activeSessionMessages.length - 1; i >= 0; i -= 1) {
       const message = activeSessionMessages[i] as typeof activeSessionMessages[number] & {
         tokens?: {
+          total?: unknown;
           input?: unknown;
           output?: unknown;
           reasoning?: unknown;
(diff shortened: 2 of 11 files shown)


pocharlies:
Addressed the review (new HEAD `8e4f7356e`, changelog-only commit on top of the reviewed code):

1. **evidence-gap** → added before/after screenshots to the Visual evidence section: the same session from the same live storage rendered by `origin/main` (meter capped at **999.0%**, summed fields 18,611,976) and by this branch (**7.8%**, reported `total` 78,327) on the desktop web header.
2. **description inaccuracies** → both corrected: five call sites delegate to the helper and two `unknown`-sanitized surfaces inline the identical rule; and the pinned `@opencode-ai/sdk@1.18.18` `/v2` types do declare `total?: number` — my earlier claim came from a stale 1.18.16 install.
3. **changelog credit** → `(thanks to @pocharlies)` appended to both entries (commit `8e4f7356e`).
4. **sidebar chips vs headline** → kept as a declared non-goal per the suggestion, now stated explicitly under Non-goals (cumulative round-trip I/O vs. window fill are different quantities).

PR #2916, opened 2026-08-14
fix(proxy): reuse upstream connections for OpenCode API requests

### LINKED ISSUE #2915: [Bug] API proxy opens a new TCP connection per request, exhausting ephemeral ports
### What's wrong?

Both `createProxyMiddleware` call sites are constructed without an `agent` option. `http-proxy` treats that as `agent: false`, which disables connection pooling **and** forces `Connection: close` on every proxied request. The result is a new TCP connection — and a new ephemeral port — for every single API request.

At idle this is invisible. Under sustained request volume it walks the host's ephemeral port pool down until the OS can no longer allocate one, at which point **every process on the machine** fails to open outbound connections with `EADDRNOTAVAIL` ("Can't assign requested address") until the TIME_WAIT table drains.

### Root cause

`node_modules/http-proxy@1.18.1/lib/http-proxy/common.js`:

```js
outgoing.agent = options.agent || false;              // :62  -> false
...
if (!outgoing.agent) {                                 // :69
  outgoing.headers = outgoing.headers || {};
  if (typeof outgoing.headers.connection !== 'string'
      || !upgradeHeader.test(outgoing.headers.connection)
     ) { outgoing.headers.connection = 'close'; }      // :73  -> forced
}
```

The call site does not pass `agent`:

- [ ] `packages/web/server/lib/opencode/proxy.js:770` — OpenCode API proxy

> **Correction (2026-08-14):** this section originally listed a second call site,
> `packages/web/server/lib/preview/proxy-runtime.js:1403`. That file was removed from `main`
> in cc9249d93 ("feat(browser): replace the preview proxy with a real browser panel and an
> agent web tool", #2883) — my repo-wide grep was run from a stale local branch. On current
> `main` the OpenCode API proxy is the only `createProxyMiddleware` call site. Thanks to
> @openchamber-bot for catching it.

### Evidence

Measured against a running instance (OpenChamber -> `opencode serve` on `127.0.0.1:49303`):

- **118 short-lived connections in 10s** (50 samples @ 0.2s), OpenChamber the only client
- Client ephemeral ports advance monotonically per request (`61591, 61592, 61593, 61594` -> `61603, 61604, 61605, 61606`) — no reuse
- TIME_WAIT accumulates on the **listening** port `49303`, i.e. the *server* initiates the close — the expected signature of a client that sent `Connection: close`

```
tcp4  0  0  127.0.0.1.49303  127.0.0.1.61659  TIME_WAIT
tcp4  0  0  127.0.0.1.49303  127.0.0.1.61660  TIME_WAIT
tcp4  0  0  127.0.0.1.49303  127.0.0.1.61661  TIME_WAIT
```

### Control — the upstream server is not at fault

Reproducible in ~2 minutes:

```bash
opencode serve --pure --hostname 127.0.0.1 --port 49999
```

Send one HTTP/1.1 request with `Connection: keep-alive` and hold the socket: the server keeps it open **past 90s idle** and generates **zero** TIME_WAIT. `opencode serve` honors keep-alive correctly whenever the client asks for it. The proxy never asks.

### Why it escalates

On macOS (measured):

- Ephemeral pool: **16,384** ports (`net.inet.ip.portrange` 49152-65535)
- TIME_WAIT: **30s** (`net.inet.tcp.msl = 15000`, 2*MSL)
- Idle steady state: ~1.7 conn/sec -> ~52 TIME_WAIT entries (harmless)

Exhaustion therefore occurs at roughly **546 conn/sec sustained** against a single destination port — and at a proportionally lower rate when several destination ports are in play, since an ephemeral port may be reused against a *different* destination. Any condition that raises request volume or induces retries walks the pool down linearly.

macOS has the smallest default ephemeral range of the supported platforms, so it hits this first.

### Fix

Pass a keep-alive agent. `maxSockets: Infinity` preserves today's unbounded concurrency, so the change is purely about reuse and does not silently introduce throttling:

```js
import http from 'node:http';

const openCodeAgent = new http.Agent({
  keepAlive: true,
  keepAliveMsecs: 30_000,
  maxSockets: Infinity,
  maxFreeSockets: 32,
});
```

### Related

- #2829 — same "proxy hammers a dead local upstream" class, preview file, fixes retry behavior rather than connection reuse
- #2467 — connection exhaustion at a different layer (browser 6-per-origin limit)

### Where does it happen?

Desktop (macOS) — though the defect is in runtime-agnostic server code under `packages/web`.

### Version

1.18.4

alohaninja:
## Intent

`createProxyMiddleware` was constructed without an `agent`, so `http-proxy` falls back to `agent: false`. That both disables connection pooling and forces `Connection: close` on every proxied request ([`http-proxy/lib/http-proxy/common.js:62-73`](https://github.com/http-party/node-http-proxy/blob/1.18.1/lib/http-proxy/common.js#L62-L73)):

```js
outgoing.agent = options.agent || false;              // :62  -> false
...
if (!outgoing.agent) {                                 // :69
  if (typeof outgoing.headers.connection !== 'string'
      || !upgradeHeader.test(outgoing.headers.connection)
     ) { outgoing.headers.connection = 'close'; }      // :73  -> forced
}
```

The result is one TCP connection — and one ephemeral port — per proxied API request. At idle this is invisible; under sustained request volume it walks the host's ephemeral port pool down until the OS cannot allocate one, at which point **every process on the machine** fails to open outbound connections with `EADDRNOTAVAIL` until the TIME_WAIT table drains.

This passes a shared keep-alive agent. `maxSockets: Infinity` preserves the unbounded concurrency of `agent: false`, so the change is limited to connection reuse and does not introduce throttling.

Fixes #2915.

pr2916-proxy-keepalive

## Non-goals

- **Other `createProxyMiddleware` call sites.** There are none. #2915 originally listed a second call site in `packages/web/server/lib/preview/proxy-runtime.js`, but that file was removed from `main` in cc9249d93 (#2883); the issue has been corrected. `proxy.js:770` is the only remaining call site, so this PR closes the issue rather than partially addressing it.
- **Retry/backoff behavior.** When the upstream is unreachable this proxy retries connection failures without backoff. That is a separate defect from connection reuse and is not addressed here.
- **No change to routing**, auth header injection, timeouts, or SSE forwarding. The upstream `Connection` header and socket reuse do change — that is the fix. Transport selection is preserved for both `http:` and `https:` targets (see Validation).

## Affected surfaces

- `packages/web` server runtime only — `packages/web/server/lib/opencode/proxy.js`.
- Applies to Desktop, Desktop Web, and Mobile/PWA — all of which serve through the OpenChamber web server and therefore proxy via `registerOpenCodeProxy`. The VS Code extension is **unaffected**: it proxies through its own fetch bridge (`packages/vscode/src/bridge-proxy-runtime.ts`, with SSE via `sseProxy.ts`) and never imports the web-server proxy. The defect and fix are otherwise runtime-agnostic; measured on Desktop (macOS), which has the smallest default ephemeral port range and therefore reaches exhaustion first.
- No UI, no persisted state, no external contract. The `Connection` header on the proxy's *upstream* leg is the only wire-visible change.
- `CHANGELOG.md` — one `[Unreleased]` bullet, per repo precedent for proxy/stability fixes.
- One new module-scope export, `createOpenCodeProxyAgent(target)`. The agent class is derived from the resolved target scheme, so `https://` external servers configured via `OPENCODE_HOST` receive an `https.Agent`.

## Repository guidance

|---|---|---|
| `AGENTS.md` | Repo-wide contribution rules | Focused single-concern change; no drive-by refactors |
| `openchamber-change-discipline` | Executable server source changed | Added a wiring regression test that fails when the `agent` option is removed, plus behavioral tests for the agent itself. Verified both directions, not just the passing one |
| `packages/web/server/lib/opencode/DOCUMENTATION.md` | Nearest module doc; describes `proxy.js` | Line 389 enumerates only the primary export `registerOpenCodeProxy`; the module's four other helper exports are intentionally undocumented. `createOpenCodeProxyAgent` follows that precedent, so no doc edit was made rather than introducing an inconsistent entry |
| `locale-ui-patterns` | Not applicable | No user-facing strings added or changed |
| `theme-system` | Not applicable | No UI or styling |
| `ui-api-decoupling` | Not applicable | Server-side only; no UI data access changed |

## Validation

| Check | Result |
|---|---|
| `npx vitest run` (packages/web) | `Test Files 144 passed (144)` / `Tests 1292 passed \| 1 skipped (1293)` |
| `npx tsc --noEmit` (packages/web) | clean, no output |
| `git diff --stat` | 3 files, 112 insertions(+), 1 deletion(-); `bun.lock` unchanged |
| Regression guard verified in both directions | With fix: pass. With the `agent:` line removed: both wiring tests fail (`AssertionError: expected undefined to be truthy`) |

**Runtime measurement.** Type-checking cannot demonstrate socket behavior, so the fix was measured end-to-end against a real `opencode serve` instance, driving 200 sequential requests through the actual middleware produced by `registerOpenCodeProxy` (not a mock), counting TIME_WAIT entries on the upstream listening port:

| Build | Requests | TIME_WAIT created | Ephemeral ports per request |
|---|---|---|---|
| `main` (no agent) | 200 | 201 | **1.005** |
| this PR | 200 | 0 | **0.000** |

That confirms the exhaustion arithmetic: with a 16,384-port ephemeral range and a 30s TIME_WAIT (`net.inet.tcp.msl = 15000`), roughly 546 req/sec sustained against a single upstream is enough to exhaust the pool, and proportionally less when several upstream ports are in play.

**Review round 2 — HTTPS transport regression.** The first commit created an unconditional `http.Agent`. Since http-proxy dispatches through `https.request` for `https:` targets (`web-incoming.js:126`) and `http.Agent#createConnection` is plain `net.createConnection`, that would have opened plaintext sockets to a TLS port for external servers configured via `OPENCODE_HOST` — a regression, since `agent: false` worked for both schemes. Caught in review, fixed in 4f53611de.

| Check | Result |
|---|---|
| `npx vitest run` (packages/web), round 2 | `Test Files 144 passed (144)` / `Tests 1297 passed \| 1 skipped (1298)` |
| `npx tsc --noEmit` | clean |
| `bun run dead-code` (AGENTS.md:125 — exports changed) | report reviewed; `createOpenCodeProxyAgent` not flagged, no new entries attributable to this change |
| HTTPS guard verified in both directions | Reverting to an unconditional `http.Agent` fails 2 tests (`derives an https agent when the resolved target is https`, `returns an https agent for https targets`) |

**Review round 3 — cold-start https.** Round 2 derived the agent at registration, which is too early: `setupProxy()` runs before `bootstrapOpenCodeAtStartup()`, so the port is null, `buildOpenCodeUrl()` throws, and the target falls back to http — meaning an `OPENCODE_HOST=https://...` server still received a plain `http.Agent`. Caught in review, fixed in 6d6bbc891 by resolving lazily.

| Check | Result |
|---|---|
| `npx vitest run` (packages/web), round 3 | `Test Files 144 passed (144)` / `Tests 1300 passed \| 1 skipped (1301)` |
| `npx tsc --noEmit` | clean |
| `bun run dead-code` | clean for this change (the resolver is module-private; it was flagged while exported) |
| Cold-start guard verified in both directions | Reverting to eager registration-time derivation fails `resolves an https agent after bootstrap even though registration ran cold` |

**Review round 5 — mutation coverage and pool sizing.** An external review found the https branch was mutation-uncovered: the resolver re-implemented agent construction inline instead of delegating to `createOpenCodeProxyAgent(target)`, so stripping `OPENCODE_AGENT_OPTIONS` from its https branch (and with it keep-alive) left the suite green, while `createOpenCodeProxyAgent` itself had no production callers. Fixed in fd6276d3b.

| Check | Result |
|---|---|
| `npx vitest run` (packages/web), round 5 | `Test Files 144 passed (144)` / `Tests 1300 passed \| 1 skipped (1301)` |
| `npx tsc --noEmit` | clean |
| `bun run dead-code` | clean — the factory now has a production caller |
| Mutation: https branch drops `OPENCODE_AGENT_OPTIONS` | **fails 2 tests**, including the live resolver path (`resolves an https agent after bootstrap even though registration ran cold`). Previously passed 19/19. |

**Independently reproduced.** @openchamber-bot reproduced the defect against the real, unmodified `registerOpenCodeProxy` on branch `reproduce/issue-2915` and measured 20/20 requests opening a fresh connection with `connection: close` and 20 new TIME_WAIT tuples, dropping to 1/20 connections and 0 TIME_WAIT with a keep-alive agent — consistent with the table above.

**Not verified:** a live TLS handshake against a real https upstream — the https path is covered by transport-selection assertions (`instanceof https.Agent`) through the real `registerOpenCodeProxy`, not by an end-to-end TLS request, since the repo has no self-signed-cert test precedent and I did not want to add an openssl dependency to CI. Also not verified: behavior on Windows/Linux ephemeral ranges — both have larger defaults than macOS, so they are strictly less exposed.

## Visual evidence

No user-visible change. The diff adds an `http.Agent` to the options object passed to `createProxyMiddleware`; it does not touch rendering, routing, response bodies, status codes, or any UI module. The only externally observable difference is on the proxy's upstream leg, where the `Connection: close` header is no longer forced and the TCP socket is reused — neither of which is reachable from the rendered UI.

Substituting the runtime socket measurement in the Validation section above as the evidence for the behavior this PR actually changes.

## Risks and failure behavior

- **Stale pooled sockets across restarts.** `router` re-resolves the upstream port, which can change when the managed OpenCode process restarts. `http.Agent` keys its pools by `host:port`, so a port change creates a new pool entry rather than reusing a dead socket; `maxFreeSockets: 256` (Node's own default) bounds how many idle sockets can linger, and `timeout: 60_000` evicts them from our side. (`keepAliveMsecs: 30_000` is the initial delay before TCP keep-alive probes, not the idle free-socket lifetime.)
- **Upstream closing an idle pooled socket.** Standard keep-alive race: if the upstream closes a socket as a request is dispatched onto it, Node surfaces `ECONNRESET`. `opencode serve` was measured holding an idle keep-alive connection for >90s, well beyond the 30s `keepAliveMsecs`, so the window is small. The existing `error` handler already maps connection-level failures to a 503/504 response, so the failure mode is an occasional retryable error rather than a crash.
- **Concurrency unchanged.** `maxSockets: Infinity` is deliberate. A finite cap would have silently converted an unbounded-concurrency proxy into a throttled one, which is a behavior change unrelated to the bug.
- **WebSocket paths unaffected.** This proxy does not set `ws: true`, and `http-proxy` exempts upgrade requests from the forced `close` regardless.
- **Scheme resolved per request, memoized per scheme.** `agent` is a getter backed by a memoizing resolver, so the class is chosen when a request is proxied rather than at registration — which matters because `setupProxy()` runs before `bootstrapOpenCodeAtStartup()` and the target is not yet resolvable at registration. Memoization keeps one shared pool per scheme; at most two agents (`http:`, `https:`) exist for the process. The target *port* can still change at runtime and is handled by `router`, with agents pooling per `host:port`.
- **Dependency on http-proxy-middleware option handling.** Lazy resolution relies on `prepareProxyRequest` rebuilding options with `Object.assign({}, this.proxyOptions)`, which invokes getters. A behavioral test pins that contract against the real library, so a future change to cached or shallow-reference options fails a test rather than silently freezing the agent at its registration-time value.
- **Rollback.** Revert both commits; behavior returns to one connection per request. No migration, no persisted state, no config surface.


alohaninja:
Good catch — that's a real regression and I've fixed it in 4f53611de.

You're right on the mechanism: `web-incoming.js:126` dispatches through `https.request` when `options.target.protocol === 'https:'`, and `http.Agent#createConnection` is plain `net.createConnection`, so an external server configured over https via `OPENCODE_HOST` would have gotten plaintext sockets against a TLS port. `agent: false` worked for both schemes, so this was a regression in a supported configuration, not just a missing improvement.

I took your suggested shape:

```js
export const createOpenCodeProxyAgent = (target) => (
  isHttpsProxyTarget(target)
    ? new https.Agent(OPENCODE_AGENT_OPTIONS)
    : new http.Agent(OPENCODE_AGENT_OPTIONS)
);
```

derived once as `createOpenCodeProxyAgent(resolveProxyTarget())` at registration, so the single shared instance across `apiProxy` and `interactiveOAuthProxy` is preserved and the existing wiring assertions still hold. Scheme detection parses the target as a URL with a regex fallback, defaulting to `http.Agent` for missing or unparseable targets.

**Guarded in both test layers**, and verified in the failing direction — reverting to an unconditional `http.Agent` fails exactly two tests:

```
× derives an https agent when the resolved target is https
× returns an https agent for https targets
```

One subtlety worth flagging for review: `https.Agent` extends `http.Agent`, so `expect(httpAgent).toBeInstanceOf(http.Agent)` passes for *both*. The http-side assertions are therefore `not.toBeInstanceOf(https.Agent)` — the negative is the load-bearing one, and there's a comment in the tests to that effect.

Also actioned from your review:

- **`bun run dead-code`** (AGENTS.md:125, since exports changed) — run and reviewed; `createOpenCodeProxyAgent` is not flagged, and no new entries are attributable to this change.
- **"No change to request semantics"** — you're right that this was contradicted. Reworded: routing, auth header injection, timeouts, and SSE forwarding are unchanged; the upstream `Connection` header and socket reuse *do* change, and transport selection is now explicitly preserved for both schemes.

**Explicitly not verified:** a live TLS handshake against a real https upstream. The https path is covered by transport-selection assertions through the real `registerOpenCodeProxy`, not an end-to-end TLS request — the repo has no self-signed-cert test precedent and I'd rather not add an openssl dependency to CI without a maintainer's steer. If you'd prefer a full TLS round-trip test, say the word and I'll add one.

alohaninja:
Both findings addressed in 6d6bbc891.

### Finding 1 — cold-start https (blocker)

You're right, and the diagnosis is exact. I verified the ordering you cited: `startup-pipeline-runtime.js` calls `setupProxy(app)` at line 104 and `bootstrapOpenCodeAtStartup()` at line 141, and `buildOpenCodeUrl` throws when `!state.openCodePort` (`network-runtime.js:86-88`). So at registration the target resolves to `FALLBACK_PROXY_TARGET` and an `OPENCODE_HOST=https://...` server still got a plain `http.Agent`. My round-2 tests passed because the stubs modeled the HMR/restart shape — populated port and base URL — not cold start. That was the gap.

I took a different route than the suggested env threading, and want to flag the tradeoff explicitly so you can push back if you'd prefer the other one.

`agent` is now a **getter** backed by a per-scheme memoizing resolver:

```js
const createApiProxy = (timeoutMs) => createProxyMiddleware({
  target: resolveProxyTarget(),
  get agent() { return resolveOpenCodeProxyAgent(); },
  ...
```

`prepareProxyRequest` rebuilds per-request options with `Object.assign({}, this.proxyOptions)` (`http-proxy-middleware/dist/http-proxy-middleware.js:106`), and `Object.assign` invokes getters — so resolution happens at request time, after bootstrap, while memoization still yields one shared pool per scheme (at most two agents for the process).

Why this over threading `ENV_CONFIGURED_OPENCODE_HOST` through deps:

- It needs no change to the `registerOpenCodeProxy` deps contract or its call site in `server-utils-runtime.js`.
- It is correct for *any* future path that could produce an https target, not only the env-configured one. Your analysis that env is currently the only such path matches what I found — but that's an invariant the fix would silently depend on.

The tradeoff, stated plainly: it depends on a library implementation detail. If http-proxy-middleware ever switched to cached or shallow-reference options, the agent would freeze at its registration-time value and https would regress silently. So I pinned that contract with a behavioral test against the real library:

```
✓ http-proxy-middleware re-reads the agent option on every proxied request
```

It counts getter invocations across two real proxied requests and asserts ≥2. A library change that broke the mechanism fails loudly instead of quietly.

**Tests now model the production ordering** — registered with `port: null` and a throwing `buildOpenCodeUrl`, then an https base URL appearing after bootstrap — and are verified in the failing direction. Reverting to eager registration-time derivation fails:

```
× resolves an https agent after bootstrap even though registration ran cold
```

If you'd still rather have the explicit env-threaded input as the source of truth, say so and I'll switch — it trades the library-internals dependency for a deps-contract change, and you own that call.

### Finding 2 — `keepAliveMsecs` nit

Corrected in the PR body. It now reads that `maxFreeSockets: 32` bounds how many idle sockets linger, with `keepAliveMsecs` described as the initial delay before TCP keep-alive probes rather than the idle free-socket lifetime.

### Also

`bun run dead-code` flagged `createOpenCodeProxyAgentResolver` as an unused export — it's only used inside `proxy.js`, so it's now module-private. Full suite: 1300 passed, 1 skipped; `tsc --noEmit` clean.

alohaninja:
Added in 9f036e817 — good call, and the precedent you cited is the right shape.

```markdown
- **Stability/Proxy:** the local server now reuses its connection to OpenCode instead of
  opening a new one for every API request. Under sustained traffic the old behavior could
  use up every outgoing network port on the machine, at which point nothing on the computer
  could open a new connection until it was restarted (thanks to @alohaninja).
```

I kept it plain-spoken and user-facing rather than protocol-level — the observable symptom is "nothing on the machine can connect until restart," which is what a reader of the changelog cares about. Follow-up commit 9dff41ff5 switches "behaviour" to "behavior" to match the file's existing usage (43 vs 1).

PR body's Affected surfaces now lists `CHANGELOG.md`.

alohaninja:
Taken in 541335dc5 — you're right, and the distinction matters.

"until it was restarted" → "until the traffic stopped and the ports were released."

The pool drains on its own via TIME_WAIT expiry (~30s at `net.inet.tcp.msl = 15000`) once the exhausting traffic stops, so a restart is sufficient but not necessary. Worth noting the two cases differ: in the incident behind #2915 a reboot *was* required, but only because the retry loop kept generating connections as fast as they expired — so the pool never got a chance to drain. That's a property of that failure mode, not of port exhaustion generally, and the changelog should describe the general case.

alohaninja:
Fixed in the PR body — the claim was wrong and I verified your correction before applying it (`grep -rn "registerOpenCodeProxy" packages/vscode/src` returns nothing).

Affected surfaces now reads:

> Applies to Desktop, Desktop Web, and Mobile/PWA — all of which serve through the OpenChamber web server and therefore proxy via `registerOpenCodeProxy`. The VS Code extension is **unaffected**: it proxies through its own fetch bridge (`packages/vscode/src/bridge-proxy-runtime.ts`, with SSE via `sseProxy.ts`) and never imports the web-server proxy.

Worth noting the corollary, since it's the more interesting half: the VS Code bridge uses `fetch`, which pools by default — so it never had this defect. Same shape as the observation in #2915 that a second local proxy on the same host held stable pooled connections and generated zero TIME_WAIT. The bug was specific to the `http-proxy` path and its `agent: false` fallback.

alohaninja:
All blocker + suggestion items addressed in fd6276d3b. The mutation-coverage finding was the important one and it was correct — I verified it structurally before fixing.

### Blocker: duplicated construction hid the https branch

Confirmed exactly as reported. The gap was mine, introduced when I made the resolver private in round 2 and inlined construction:

- `createOpenCodeProxyAgent` had **zero production callers** — `:858` called the resolver, so the exported factory was dead code pinned by four tests.
- **No test asserted `keepAlive` on the resolver's https branch.** The wiring assertion at `proxy-agent-wiring.test.js:78` runs against `managedState()` — an **http** target. The one https `keepAlive` assertion exercised the dead factory.
- `new https.Agent().options.keepAlive` is `undefined` (verified), so the mutation would have silently reverted the PR's purpose for every https deployment.

Fixed by delegating, as suggested. Mutation now caught:

```
× resolves an https agent after bootstrap even though registration ran cold
× returns an https agent for https targets
   Tests  2 failed | 17 passed (19)
```

Previously 19/19 green. I also added `keepAlive` and `maxFreeSockets` assertions on the **live resolver path** rather than only the factory, so the coverage isn't purely transitive.

### `maxFreeSockets` — took it, and the framing sharpens the case

`32` was an arbitrary tightening on my part with no measurement behind it. **Node's own default is 256** (verified), so this is a restoration rather than a tuning choice — which makes your C=64 result (303 TIME_WAIT vs 0) the expected consequence of my cap rather than a surprise. A cap below the concurrency level evicts pooled sockets and reintroduces exactly the churn this PR exists to remove.

### Also taken

- **`timeout` added.** Confirmed Node's default is `undefined`, so nothing on our side retired idle free sockets.
- **`resolveProxyTarget()` port-null guard.** Checks `openCodePort` before calling `buildOpenCodeUrl` rather than relying on the throw. Your point about the five null-ing paths is the one that matters — cold start is transient, but a crashed OpenCode pays it indefinitely, precisely when the server is already degraded.
- **`:443` → `:4096` fixtures.** Verified: `new URL('https://h:443').port === ''` → `parseInt('')` → `NaN` → rejected by env-config. The fixtures modeled an unreachable state.
- **Exact getter-read assertion** — now `0` at construction, then `1`, then `2`. You're right that `>= 2` would have passed under the very regression the comment claims to guard.
- **`listen()` rejects on `'error'`**, and both helpers start servers inside `try/finally`.

### Not taken, with reasons

- **Terminal error handler** — agreed, structural and repo-wide. A `(err, req, res, next)` handler under `packages/web/` closes the class, but it's not this PR's to introduce.
- **CL/TE `removeHeader`** — agreed with your reclassification. `replayParsedBody` appears zero times in this diff, so it's pre-existing rather than introduced here.
- **`isHttpsProxyTarget` parser divergence** — agreed it's latent, not live; every value `resolveProxyTarget()` produces comes from `new URL(...).origin` or a hardcoded `http://` template.

### Still open — your call

The **real-TLS behavioral test** is the one item I haven't done, because it changes the repo's testing surface and there's no TLS precedent under `packages/web/`. Node has no certificate issuance API, so it needs either openssl at test setup (a CI toolchain dependency) or a committed fixture cert. Your security framing is what makes it worth doing — a later `rejectUnauthorized: false` on the shared options object would disable verification for every https upstream with the suite still green. I'd lean toward a committed fixture for CI determinism, but I'd rather a maintainer pick the approach than have me add test fixtures unilaterally.

Also happy to add the two module invariants to `DOCUMENTATION.md` if you'd like them there — I left them out because the review bot noted the exports list already omits helpers by precedent, but invariants are a different thing from export docs and your point about the next editor stands.

alohaninja:
CI is red on this PR for two reasons, neither of which is this diff.

### 1. `checks` — `main` is currently broken

```
packages/ui/src/lib/gitApi.ts
  5:10  error  'runtimeFetch' is defined but never used  @typescript-eslint/no-unused-vars
✖ 1 problem (1 error, 0 warnings)
error: script "lint" exited with code 1
```

109e957fe ("fix: integrate Claude CLI provider state") replaced both `runtimeFetch('/api/small-model/generate', ...)` call sites with `requestSmallModel` and added that import, but left the now-unused `runtimeFetch` import behind.

This PR touches only `CHANGELOG.md` and `packages/web/server/lib/opencode/*` — nothing under `packages/ui`. My branch predates 109e957fe and still has the two usages; because `pull_request` lints the *merge* with `main`, `main`'s version of the file wins and the unused import surfaces. It will fail on **every open PR**, not just this one.

Opened **#2920** with the one-line fix (import removal only). Verified there: `bun run lint` in `packages/ui` exits 0, and `tsc --noEmit` is clean. Once it lands, this PR's `checks` should go green without any change here.

### 2. `automation` — the review harness errored internally

The job ran 14m32s against its 30m timeout, so it didn't time out. Its log is full of the reviewer agent's own sandbox denials:

```
Error: The user has specified a rule which prevents you from using this specific tool call
  [{"permission":"bash","pattern":"*","action":"deny"}, ...]
```

No round-5 verdict was produced for `fd6276d3b`. Nothing in that log references this diff — it looks like harness flakiness rather than a finding.

I don't have admin rights to re-run either job, and I'd rather not push an empty commit just to retrigger. Since `checks` will keep failing until #2920 merges anyway, the sequencing is: land #2920 → re-run here → both should pass and the round-5 review should produce a verdict.

Happy to rebase onto `main` once #2920 is in if that's easier for you than a re-run.

alohaninja:
Closing the loop on the one open question from my earlier reply: **I'm not adding the real-TLS behavioral test in this PR.**

Reasoning, so the tradeoff is on record rather than silently dropped:

- It guards against a *hypothetical future change* — someone later adding `rejectUnauthorized: false` to the shared `OPENCODE_AGENT_OPTIONS` — not against the defect this PR fixes. That's hardening, and it's a different PR's job.
- The repo has no TLS test precedent under `packages/web/`, so it needs either an openssl toolchain dependency in CI or committed fixture certs. Both change the testing surface, and that's a maintainer's call, not something to slip into a proxy fix.

**What is covered:** transport selection is asserted through the real `registerOpenCodeProxy` (not just the exported factory), and mutation-verified — stripping `OPENCODE_AGENT_OPTIONS` from the https branch fails 2 tests including the live resolver path, where it previously passed 19/19.

**Residual risk, stated plainly:** no test performs a real TLS handshake. If a future change disabled certificate verification on the shared options object, the current suite would not catch it. Worth a follow-up issue if you want it tracked — happy to open one, or to add the test in a separate PR once you've picked openssl-at-setup vs. committed fixtures.

Everything else from that review is addressed in `fd6276d3b`.

btriapitsyn (approved):

PR #2917, opened 2026-08-14
fix(server): tolerate slow managed opencode bootstrap; support scoped config for managed server

herjarsa:
## Intent
Stop the managed \opencode serve\ crash-loop / long bootstrap on slow machines. The managed server bootstraps many project directories with heavy plugins (magic-context ~2GB ONNX, AFT RPC servers, oh-my-openagent, LSPs); Bun's single-threaded event loop saturates for minutes, HTTP responds intermittently, and the old 5s/20-failure health check killed and relaunched the process in an infinite cycle. The user perceives "crash a few minutes after start; the window stays open" (Electron alive, proxy storms, UI frozen).

## Non-goals
- No behavior change when no scoped config exists (stock install keeps the user's global config).
- No change to OpenCode itself; all changes are OpenChamber-side.
- Scoped config is opt-in; heavy plugins remain available in the user's own OpenCode TUI.

## Affected surfaces
1. \packages/electron/opencode-cwd.mjs\ — prefer \process.cwd()\ when it looks like a project root (avoids re-indexing HOME on every restart); reworded fallback warning.
2. \packages/web/server/lib/opencode/lifecycle.js\ — health-check tolerance (timeout 5s→30s, consecutive failures 20→40, env-overrideable) and \OPENCODE_CONFIG_DIR\ injection into the managed spawn.
3. \packages/web/server/lib/opencode/DOCUMENTATION.md\ — new env-var contracts documented.

## Repository guidance
- AGENTS.md: runtime boundaries respected (web vs electron ownership unchanged); owning docs updated (\DOCUMENTATION.md\).
- CONTRIBUTING.md: \un run test\ passes (see Validation); env contracts documented.
- openchamber-change-discipline: new env vars are public contracts — documented and tested.
- performance-engineering: empirical before/after measured on the identical scenario and build (see Validation).

## Validation
Commands (run locally):
- \un run --cwd packages/electron test\ (via \scripts/run-isolated-tests.mjs\): 15/15 test files pass.
- \
px vitest run lib/opencode/lifecycle.test.js\ in \packages/web/server\: 22/22 pass.

| Metric | Before | After | Method |
|---|---|---|---|
| Managed process sockets | 516-675 (41-64 Listen, 308 CloseWait) | 2 | \Get-NetTCPConnection -OwningProcess <pid>\ |
| Managed process memory | ~1.3GB | ~400MB | Task Manager / WorkingSet |
| Time to PushWatcher connected | minutes (health checks failing during bootstrap) | 43ms | main.log timestamps |
| Health failures post-arranque | continuous storm | 0 | main.log \health check failed\ count |
| Managed survival | died at ~5 min, relaunched on new port | stable 15+ min | process list + main.log |

Environment: Windows 11, opencode CLI 1.18.18 (Bun), OpenChamber from source branch. All metrics from the same machine and the same launch path; the crash-loop was first reproduced, then the fix applied, then re-measured.

Not verified: Linux/macOS behavior of the Windows-only default config path (non-Windows always returns null → no behavior change); the scoped-config scenario on non-Windows.

## Visual evidence
N/A — server-side lifecycle change; no UI rendering changes. Log evidence (main.log timestamps, proxy error counts) provided in the Validation table.

## Risks and failure behavior
- Health-check tolerance tradeoff: a hung-but-alive managed process is now restarted only after ~10-20 min (40 failures × 15-30s counting windows) instead of ~5 min. Genuinely exited processes still restart immediately via \isManagedOpenCodeProcessAlive()\.
- \waitForOpenCodeReady\ inherits the 30s per-fetch timeout, so the config-apply restart path can block slightly longer against a wedged process.
- Config-layer divergence: when a scoped config is active, UI edits writing to the user global layer are invisible to the managed server until mirrored into the scoped config. Documented as an intentional tradeoff in \DOCUMENTATION.md\.
- User-provided \OPENCODE_CONFIG_DIR\ always wins over the Windows default (no silent clobber).
- \%APPDATA%\openchamber\managed-config\ is a convention only; nothing in the repo creates it automatically. When absent, no env var is injected and behavior is unchanged.

### DIFF
--- a/PLAN.md
+++ b/PLAN.md
@@ -9,10 +9,11 @@
 ## Fases
 1. [DONE] Wave0: auditoría (procesos, DB, logs, fork, proxy).
 2. [DONE] Wave2: poda backups (11.5+1.6+1.7GB) + magic-context 131MB + session_diff >20MB (238MB) + prune events (67K→49K, 3.93→3.69GB). Script prune reparado (bloque duplicado, línea truncada).
-3. [TODO] Wave4: rebuild dist-bundle/main.mjs desde fork HEAD + repack app.asar + sync packages/web → app.asar.unpacked/@openchamber/web.
-4. [TODO] Wave3: relanzar OpenChamber; verificar /session + PushWatcher + ausencia de 4x MCP.
-5. [TODO] Wave5: opencode.jsonc — quitar SessionStart hook roto.
-6. [TODO] QA: medir boot y /session; evidencias.
+3. [DONE] Wave4: rebuild main.mjs + repack app.asar (--unpack node_modules, 2.5MB) + sync packages/web -> @openchamber/web (warmup paralelo, session-merge, health tolerance, scoped-config). Backups: app.asar.bak-20260816-pre-sync + unpacked.bak.
+4. [DONE] Wave3: OpenChamber relanzado 12:20. PushWatcher connected en 1s (antes 2.5min). Gate /session 200 OK 3-90ms (28KB). 0 proxy errors. Health flapping (contention DB con 2 TUIs + managed) -> restart ciclico rapido OK.
+5. [DONE] Wave5: hook SessionStart roto eliminado de opencode.jsonc.
+6. [DONE] QA: gate estable 3x, 0 proxy errors, boot 1s. Pendiente: ver UI en vivo.
 
 ## Nota previa (sesión 15/08 23:45)
 - 11 AppHangB1 de OpenChamber.exe (20:31-23:39): el main de Electron se cuelga → proxy deja de procesar → socket hang up → UI congelada → se recupera → repite.
+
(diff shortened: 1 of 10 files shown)


btriapitsyn:
Thanks for the thorough investigation — the failure mode is real: a managed server that is slow to bootstrap can be killed by the health check and enter a restart loop, and your before/after measurements show it clearly. But this PR bundles too many workarounds shaped around one specific environment to be mergeable: a personal PLAN.md and sync script, a cwd heuristic based on the launch directory, a Windows-only scoped-config convention that makes UI config edits invisible to the managed server, hardcoded third-party worktree paths, and a global change of health-check defaults that delays recovery from a genuinely hung server to 10–20 minutes for everyone.

What we would welcome as small separate PRs: (1) bounded-concurrency warmup so one slow directory can't serialize the pass, and (2) a bootstrap grace window for the health check (suspend restart-counting until the managed server has answered once), instead of raising the steady-state thresholds. Both fix the crash-loop without changing behavior for healthy installs. Closing this one.

PR #2932, opened 2026-08-15
fix(ui): open app deep links from chat after confirmation

ChangeHow:
## Intent

Application deep links rendered in chat markdown, for example obsidian://open, retain their safe custom-scheme href values and require explicit confirmation before opening another application on every runtime, including VS Code.

This follow-up closes desktop renderer bypasses: modifier clicks and middle clicks now enter the confirmation flow, as do app links rendered in user messages where ordinary HTTP link handling is disabled. It also adds a visible Cancel action, blocks additional high-risk/internal schemes, and corrects the Settings example.

## Non-goals

- No change to Electron main-process URL handling or runtime bridge contracts.
- No user-configurable override for the hard-blocked scheme list.
- No server-side or cross-device sync of trusted schemes; trust remains local to each device.
- No change to file links, local file references, or browser-handled schemes such as mailto and tel.

## Affected surfaces

| Surface | Effect |
|---|---|
| Shared chat markdown | Web, desktop, mini-chat, mobile, and VS Code app links confirm before modifier-key handling. Middle clicks and drag starts cannot fall through to native external navigation. |
| User message markdown | App links stay confirmed even when existing HTTP link safety is disabled for user-authored content. |
| URL sanitization and classification | ms-msdt, search-ms, shell, and openchamber-ui are never app links; sanitizer-hook coverage checks app/file retention and blocked-scheme removal. |
| Settings → General | The example now uses obsidian://; desktop and mobile evidence show the trusted-scheme row. The same trust controls and search entry are available in VS Code. |
| Persisted local state | Existing trusted-scheme storage remains compatible; no migration is required. |

## Repository guidance

|---|---|---|
| openchamber-change-discipline | Shared UI source, tests, and persisted user-visible behavior changed. | Scope stays limited to confirmation, scheme classification, localized copy, and focused tests. |
| ui-api-decoupling | Opening custom protocols is a runtime trust-boundary action. | Every runtime uses the same confirmation and trust flow. A dedicated confirmed-app-link opener keeps custom schemes unavailable to unrelated openExternalUrl callers. |
| desktop-shell | Electron external navigation and the mini-chat window are affected native boundaries. | Renderer interception occurs before modifier, middle-click, or drag fallback can reach shell.openExternal, and mini-chat mounts the shared confirmation dialog. |
| theme-system | Confirmation dialog action layout changed. | Uses existing DialogFooter and Button variants; mobile stacks visible actions and desktop preserves hierarchy. |
| locale-ui-patterns | Cancel and Settings copy are user-visible. | The new key and revised example are translated in all 11 dictionaries. |
| settings-ui-patterns | General Settings content changed. | Preserves the existing SettingsSection and trusted-scheme list instead of adding policy toggles. |
| sync-state-invariants | Trusted schemes are persisted locally. | Existing deferred safe-storage remains the single authority and persistence shape did not change. |

## Validation

| Check | Result |
|---|---|
| bun run test in packages/ui | Pass: 288/288 isolated test files. |
| bun run type-check in packages/ui | Pass. |
| bun run lint in packages/ui | Pass. |
| bun run dead-code | Completed; its non-blocking repository-wide report has no blocking result. |
| Browser, desktop HMR at 1440px | A real rendered obsidian link opened the confirmation dialog with Cancel, Trust and open, and Open once. |
| Browser, mobile surface at 390px | The same real chat link opened the responsive confirmation dialog; actions stack vertically. |
| Browser, trusted state | Trust and open was selected with window.open stubbed to avoid launching Obsidian, then the persisted obsidian row was verified in full desktop and mobile Settings pages. VS Code runtime behavior was not manually exercised. |
| Not run | Packaged Electron, Capacitor native shell, and VS Code extension host. No native or bridge code changed. |

## Visual evidence

The images below were captured from the browser HMR runtime after the supported-surface UI changes in b05807423. The later bbacd523d change adds the same confirmation flow to VS Code and does not affect these web, desktop, mini-chat, or mobile visuals. Each PNG was replaced with a new blob on the app-link-evidence branch; they show the Cancel action, obsidian:// Settings copy, and trusted-scheme row.

Desktop (1440px):

| Confirmation dialog from a real chat link | Settings → General with trusted scheme |
|---|---|
| desktop dialog | desktop settings |

Mobile surface (390px, ?surface=mobile):

| Confirmation dialog from a real chat link | Settings → General with trusted scheme |
|---|---|
| mobile dialog | mobile settings |

## Risks and failure behavior

| Area | Behavior |
|---|---|
| Hard-blocked schemes | Scriptable, browser-internal, network, file, Android intent, OpenChamber-internal, and the added Windows diagnostic/shell/search schemes are never app links and cannot be trusted through Settings. |
| External opening failure | Custom schemes are accepted only by the dedicated confirmed-app-link opener. An unavailable bridge or blocked browser open performs no navigation. |
| Local trust storage | Deferred safe storage falls back to memory when persistence is unavailable, so trust lasts only for that session; storage stays capped by the existing store limit. |
| Rollback | Removing a trusted scheme restores confirmation. Reverting this commit restores prior behavior; no migration is needed for the existing local storage key. |
| Platform coverage | Renderer behavior was manually verified in web desktop/mobile HMR. Packaged desktop, Capacitor, and VS Code host validation remains outstanding. Static checks and the shared interaction tests cover the VS Code confirmation path, but native app handoff was not exercised. |


btriapitsyn:
Thanks for this — the write-up and the test coverage are genuinely nice to review, and the per-device trust model is the right call.

A few things I noticed on top of the bot review:

- `UserTextPart.tsx:244` renders markdown with `disableLinkSafety`, so `useExternalLinkInteractions` never runs there. In the user's own messages even a plain left click on an app link now goes straight through without any confirmation, since the `href` survives sanitization. This is the same gap as the bot's blocker, just via a different path — worth fixing together.
- The dialog has no visible cancel button. Right now the safe choice is only reachable via Escape, the close icon, or a backdrop tap, which is awkward on mobile for a prompt that launches another application. I'd make cancel an explicit, default-focused button.
- `BLOCKED_APP_LINK_SCHEMES` is a denylist, so anything unknown is allowed by default — including historically abused Windows schemes like `ms-msdt:`, `search-ms:` and `shell:`. Combined with the bypass above that gets uncomfortable. Either add those, or consider flipping to an allowlist of known app schemes.

Also please read the bot review carefully — the missing test for the real DOMPurify hook is a good catch, since the current `markdownCore.test.ts` mocks DOMPurify as a pass-through and the security-critical part is effectively untested.

ChangeHow:
@btriapitsyn  I’ll keep all three actions visible rather than putting the persistent “Trust and open” action in a menu.

ChangeHow:
/oc-review evidence refreshed for current head 451ecb928

ChangeHow:
/oc-review runtime handling and current-head evidence captured for b05807423

ChangeHow:
@btriapitsyn One potential iOS compatibility issue: in Capacitor, `openExternalUrl()` ultimately falls back to `window.open(..., "_blank")`. iOS WKWebView may not hand custom schemes such as `obsidian://` and `vscode://` off to the OS, which could result in a silent failure; the current fallback also returns `true` without checking whether the URL was opened.

One possible solution is to use Capacitor App Launcher so custom-scheme URLs are delegated to the native platform. This would introduce an additional dependency and native sync/generated-file changes, so I am flagging it for maintainer assessment.

btriapitsyn:
I pushed bbacd523d with the remaining review fixes and the conflict resolution. Thanks for iterating on this so quickly.

What changed:

- synced the branch with current main and resolved both Markdown conflicts;
- enabled the same confirmation and trusted-scheme controls in VS Code instead of stripping or directly opening app links;
- split custom-scheme opening from openExternalUrl, so unrelated HTTP-link callers cannot bypass confirmation;
- moved activation handling into a focused helper and added regression coverage for plain clicks, modifier clicks, middle clicks, user-message HTTP-safety behavior, and drag starts;
- blocked drag-start navigation before Electron can hand the custom scheme to the OS;
- removed the serialized-HTML regex path, so Markdown code/text cannot be damaged by href stripping;
- updated the owning documentation and PR body for the current behavior.

Validation completed: all 288 UI test files pass, workspace-wide type-check and lint pass, focused oxlint passes, and dead-code completed with only the existing non-blocking repository report. Packaged Electron, Capacitor native handoff, and the VS Code extension host were not manually run.

PR #2937, opened 2026-08-15
fix(ui): stop phantom working status on revisited sessions and stale status-row model

xiangyingchang:
## Intent

Two user-visible bugs in the chat status row, both reproduced on a real session (`ses_01b03a259ffeoxENf5fwvYnDnK`):

1. **Revisited historical sessions render a phantom "working" status forever.** `/session/status` only lists busy/retry sessions, so an idle session has no status entry. The trailing-assistant-message fallback (`time.completed` missing ⇒ assume running) was applied even after an authoritative empty status snapshot, so a historical session whose persisted assistant message lacks a completion timestamp kept showing `"<model> · weighing options ..."` with no way to settle. The server confirmed the session was `idle` the whole time.
2. **The status-row model label resolves from the wrong field.** `getActiveAssistantContext` read the model only from the parent user message's `model` object, which the real API leaves `null`; assistant records carry `providerID`/`modelID` at the top level. After switching models mid-conversation, the row could show a stale model (e.g. still `GPT-5.6 Luna` after switching to an ark model) until the next assistant message landed.

Result: an idle revisited session shows no working status, and the status row names the model that is actually being used (assistant record first, newer user choice second, current session selection while a new prompt awaits its assistant turn).

## Non-goals

- Interrupted-turn tool finalization (`session.idle` + running tool parts, openchamber#2577 recovery) - unchanged.
- The global live status index (`global-session-status.ts`) and sidebar ordering/timing - unchanged; only the directory-store readiness marker is new.
- Message loading, streaming lifecycle derivation, and reconnect recovery paths - unchanged except for the new readiness flag they can now rely on.
- Composer model picker behavior - the selection store is only read, never written by this change.

## Affected surfaces

- **`packages/ui`** shared sync + chat surfaces: directory store state shape, bootstrap, status snapshot reconciliation, `useSessionActivity`, `ChatContainer`, `useQueuedMessageAutoSend`, `useAssistantStatus`/status row.
- **All shared-UI runtimes** (web, desktop, VS Code, hosted mobile, Capacitor mobile) consume these hooks; the fix is runtime-agnostic and adds no runtime-specific behavior.
- **Persisted/external contracts:** none changed. `session_status_ready` is in-memory per directory store only (not persisted, not an API contract). The selection store is read-only here.
- **Server API shape relied upon:** `/session/status` omits idle sessions - this is now an explicit, documented invariant instead of an implicit assumption.

## Repository guidance

|---|---|---|
| `AGENTS.md` correctness invariants ("prefer authoritative state over heuristics", "derive live activity from live channels") | The bug was historical persisted data driving a live spinner | The trailing-message fallback is now gated on the absence of an authoritative status snapshot; a failed fetch never marks the snapshot ready, so failure is not masqueraded as authoritative empty success |
| `openchamber-change-discipline` skill | Source change in `packages/ui` with cross-file contract (new state field) | Smallest complete change; existing snapshot semantics preserved (monotonic polls still never lower status); tests added with each behavior; owning docs updated |
| `sync-state-invariants` skill | Modifies sync stores, bootstrap, status reconciliation, activity fallbacks | Read `packages/ui/src/sync/DOCUMENTATION.md` first; marker is set only by authoritative sources (bootstrap snapshot, authoritative resync), never by the monotonic watchdog poll, so missed-live-event recovery keeps its narrow fallback |
| `ui-api-decoupling` skill | Reads store state consumed across runtimes | No new HTTP paths; everything flows through existing directory stores/hooks |
| `work-unit-commits` skill | Two distinct fixes in one PR | Split into two conventional commits: status misjudgment fix (with its tests/docs), then model resolution fix (with its tests) |

## Validation

| Check | Result |
|---|---|
| `bun test packages/ui/src/hooks/` | 32 pass, 0 fail (includes new `getActiveAssistantContext` model tests, `isSessionStatusAuthoritative` tests, queue-gate snapshot tests) |
| `bun test packages/ui/src/sync/bootstrap.test.ts packages/ui/src/sync/__tests__/session-status-snapshot.test.ts packages/ui/src/sync/__tests__/event-reducer.test.ts packages/ui/src/sync/reconnect-recovery.test.ts packages/ui/src/sync/streaming.test.ts` | 50 pass, 0 fail |
| `bun run type-check` (workspace-wide) | Pass (all 5 workspaces exit 0) |
| `bun run lint` (workspace-wide) | Pass (all workspaces exit 0) |
| `bun run build` | `ui` (tsc) and `web` (vite, ~76s) succeed; the trailing VS Code extension bundle step fails with `esbuild: command not found` - the `esbuild` binary is absent from this checkout's `node_modules/.bin` entirely (pre-existing environment issue, unrelated to this diff) |
| Runtime check on the reproducing session | Server-side `/session/status` for `ses_01b03a259ffeoxENf5fwvYnDnK` confirmed `{type: "idle"}` while the pre-fix UI kept showing the running status |
| Manual UI verification after fix | Not performed by the contributor yet - requires reloading the dev app against the reproducing session; unit tests cover the decision logic (empty authoritative snapshot + trailing assistant without `time.completed` ⇒ idle) |

## Visual evidence

The user-visible change is the status row under the message list. **Before:** reopening the affected conversation showed a persistent `GPT-5.6 Luna · weighing options ...` row with busy dots on an idle session (reported by the user on the session above; that exact state is the pre-fix behavior and is what the first commit removes). **After:** no working-status row renders for that session, and during a real running turn the row names the actually-selected model. Screenshots/recordings were not captured because reproducing requires the stale persisted assistant-message state on a running dev app; the state machine transition is covered by `session-status-snapshot.test.ts` (authoritative empty snapshot lowers busy and marks ready), `useQueuedMessageAutoSend.test.ts` (snapshot-ready gate), and the `useAssistantStatus` model-resolution tests. Happy to attach a recording on request.

## Risks and failure behavior

- **Narrowed fallback risk:** if every authoritative status fetch fails, `session_status_ready` stays false and the old trailing-message fallback still applies - the pre-fix behavior, which errs toward showing activity rather than hiding it. No new stuck-idle path is introduced.
- **Missed busy events:** the monotonic watchdog poll deliberately does not set the readiness marker, so its raise-only semantics and the trailing-message fallback for a missed live busy event are preserved.
- **Rollback:** two self-contained conventional commits; reverting either alone restores the previous behavior for that concern without touching the other.
- **Performance:** the new marker is one boolean on the directory store state; `useSessionActivity`/`ChatContainer` select it with an empty-deps selector returning a primitive, adding no re-render pressure.
- **Cross-runtime:** no runtime-specific code paths; the state field defaults to `false` everywhere, so VS Code/mobile surfaces keep the old behavior until their first authoritative snapshot lands.


btriapitsyn:

btriapitsyn:

PR #2942, opened 2026-08-15
feat(usage): restore pace and prediction indicators on quota rows

JElfferich:
## Intent

Restores the usage pace/prediction feature that was silently removed in v1.18.3 by commit d7231b6de9 ("fix(usage): refresh work status quotas automatically"), which rode along as collateral of the work-status panel rework (#2776). The feature was originally added in #372 ("Add usage prediction and pace to usage dropdown and settings") and gained a hide-toggle in #1420.

Behavior restored:
- Quota rows in the work-status panel now show a pace status dot plus predicted final usage ("Pred: +120%") per window, or the remaining wait when a window is exhausted.
- Usage cards in the Settings → Usage page show the same pace indicator plus the expected-usage marker on the progress bar for weekly/monthly windows.
- A "Show predictions" toggle in the Settings → Usage sidebar (under the display-mode select) controls whether the extra rows render, persisted as `usageShowPredValues`.

## Non-goals

- The original header usage dropdown does not exist on desktop anymore (removed by the work-status rework, #2776); this PR does not recreate it. Prediction lives in the two remaining usage surfaces: the work-status panel and the Settings Usage page.
- The unrelated changes bundled into d7231b6de9 (quota auto-refresh interval, `showInHeader` → `showInWorkStatus` rename, header redesign) are not reverted or touched.
- No new settings-search page, no mobile/desktop shell changes beyond the settings plumbing needed to persist the toggle.

## Affected surfaces

| Surface | Change |
|---|---|
| `packages/ui` | Pace computation (`lib/quota/utils.ts` + exports), `PaceIndicator` component (restored), `UsageCard` prediction rows + expected marker, `UsageSidebar` predictions toggle (shared `Checkbox` under the display-mode select, persisted as `usageShowPredValues`, with a settings-search entry anchored to the sidebar row), `WorkStatusUsageSection` pace rows, store state (`showPredValues`), persistence (`lib/persistence.ts`, `lib/desktop.ts` type) |
| `packages/vscode` | `bridge-settings-runtime.ts` sanitizes `usageShowPredValues` so the setting round-trips in the VS Code runtime |
| `packages/web` | `settings-helpers.js` sanitizes `usageShowPredValues` server-side |
| Locales | `settings.usage.sidebar.field.showPredictions` restored in all 11 dictionaries with the original translations (the `settings.usage.pace.*` keys were never removed) |
| CHANGELOG | Unreleased bullet added |

Persisted contract: `usageShowPredValues` (boolean) is read back exactly as written; defaults to `false` when absent, so existing installs see no behavior change until they enable the toggle.

## Repository guidance

|---|---|---|
| `openchamber-change-discipline` | Source change across packages with a persisted settings contract | Pace logic restored as pure exported functions with unit tests; persistence plumbing matches the surviving `usageDisplayMode` precedent in all three runtimes (web persistence, web server sanitizer, VS Code bridge). No `any`, no new deps. |
| `settings-ui-patterns` + `references/controls.md`, `references/search.md` | New settings toggle | The toggle uses the shared `Checkbox` in the usage sidebar, matching the sibling display-mode row precedent (the sidebar is nav chrome, not a field-row surface); `data-settings-item="usage.show-predictions"` anchor on the sidebar row with a matching explicit registry entry in `lib/settings/search.ts` (id, page, titleKey, keywords). |
| `theme-system` | Colored status dots | Status colors come from the existing semantic CSS variables via `getPaceStatusColor` (`var(--status-*)`); no hardcoded hex or Tailwind palette colors. |
| `locale-ui-patterns` | User-facing strings | All new/restored strings use `t(...)` keys present in every locale dictionary; the restored `showPredictions` key uses the original translations from before the removal commit (not English stand-ins). |
| `changelog-authoring` | CHANGELOG update | Unreleased bullet written in release-note style, ranked after the Integrations highlight, no internal mechanics or component names. |
| `packages/ui/src/stores/DOCUMENTATION.md` | Store state addition | `showPredValues` is persisted UI state in the existing quota store, consistent with the store's stated purpose (shared across distant surfaces: work-status panel + settings page). |

## Validation

| Check | Result |
|---|---|
| `bun run --cwd packages/ui type-check` | Passed |
| `bun run --cwd packages/ui lint` | Passed |
| `bun run --cwd packages/ui test` | 265/265 test files passed (includes new `calculatePace`/`getPaceStatusColor`/`formatRemainingTime`/`calculateExpectedUsagePercent` tests) |
| `bun run --cwd packages/vscode type-check` | Passed |
| `bun run --cwd packages/vscode test` | 24/24 test files passed |
| `bun run --cwd packages/web type-check` | Passed |
| `bun run --cwd packages/web build` (vite production build) | Passed, dist emitted |
| `bun run dead-code` | No references to new exports (`calculatePace`, `PaceIndicator`, `showPredValues`, `usageShowPredValues`); report shows only pre-existing entries |
| Full `bun run test` | 43 failures in `packages/web` git/github tests: verified pre-existing on clean `upstream/main` in this environment (missing SSH private key for `~/.ssh/id_ed25519.pub`); identical file set fails without this PR's changes |

E2E verification was performed by the contributor on a dev server (port 3001) against the live opencode backend, using the opencode-go quota provider. Initial run surfaced two bugs, both fixed in the second commit (b77012116):

1. The work-status panel showed no pace rows: the panel passed the localized window label ("5-Hour") to `calculatePace`, which infers durations only from raw keys ("5h"). `UsageLimitRow` now carries the raw `windowKey`.
2. Settings pace showed "-": the opencode-go provider reports `resetAt` as an ISO string, and `calculatePace` arithmetic on a string yields NaN. It now normalizes via `new Date(resetAt).getTime()`.

After the fix, the work-status panel shows the status dot + prediction per quota row (e.g. "Pred: 25%") and the Settings → Usage cards show the pace row and expected-usage marker. The toggle persists across refresh.

## Visual evidence

Before/after captures from the contributor's dev server with the opencode-go quota provider. The "Show predictions" toggle that gates these states sits in the Settings → Usage sidebar, directly under the display-mode select.

Settings → Usage page:

| Before | After |
|---|---|
| Settings usage before | Settings usage after |

Work-status panel:

| Before | After |
|---|---|
| Work status before | Work status after |

## Risks and failure behavior

- **Compatibility:** `usageShowPredValues` was the exact persisted key before the removal; re-adding it restores round-trip for users who had it set. Absent key → `false` (feature off), so nothing changes for users who never enabled it.
- **Rollback:** the feature is additive; reverting this PR removes the UI and the toggle. The store defaults keep `showPredValues: false`, so no layout shift on upgrade.
- **Performance:** pace computation is per-window and memoized per row in both surfaces (`React.useMemo`); the work-status panel already recomputes on the 3-minute quota refresh. No new timers or subscriptions.
- **Cross-runtime:** VS Code and web sanitizers accept the boolean and drop non-boolean values, matching the `usageDisplayMode` pattern.


JElfferich:
Thanks for the re-review and the PASS.

The new a11y nit is addressed in commit `732a138a0`: the pace status dot now carries `role="img"` + `aria-label` with the localized status label (on-track / slightly-fast / too-fast / used-up) in both `WorkStatusPace` (work-status panel) and `PaceIndicator` (Settings → Usage cards), alongside the existing `title`. No new strings were needed; the labels reuse the existing `settings.usage.pace.status.*` keys, so all 11 dictionaries are covered.

Validation for the follow-up: `packages/ui` type-check and lint clean, 265/265 test files pass.

JElfferich:
Hi @btriapitsyn, @makeittech — this PR is `review:ready` (bot PASS on the merged HEAD `9acddaadd`), CI is green, and it's mergeable with no conflicts. It restores the usage pace/prediction feature removed in v1.18.3. Could you take a look when you get a chance? Thanks!

btriapitsyn:
Thanks — but restoring the pace indicator as it was isn't something we want: the removed implementation predicted poorly, which is part of why it went away with the status-bar rework rather than being re-added. If prediction comes back, it needs a rethought model, not a revival of the old math. Closing.

PR #2954, opened 2026-08-16
fix(electron): find version-managed node tooling on SSH hosts and reuse managed servers

yulong-ge:
## Intent

SSH remote hosts whose node tooling was installed per-user (nvm, volta, fnm, mise, asdf, bun) cannot be used by the desktop app today: every remote command runs as a non-interactive `sh -lc` shell, which never sources `~/.bashrc`. The user sees `Remote host has neither bun nor npm available` even though `npm` works fine in an interactive login, and an already-installed `openchamber` binary is equally invisible, so the app re-runs the managed install on every connect. Verified on two real hosts (Ubuntu containers, dash as `sh`, npm under `~/.nvm`): `sh -lc 'command -v npm'` fails there both before this change and after it, while the prepended `~/.nvm/versions/node/*/bin` glob resolves.

The managed-server lifecycle is also tightened so a working setup stays working across reconnects:

- Every remote script prepends well-known version-manager bin directories to `PATH`, covering detection, install/update, `openchamber --version`, and `openchamber serve`.
- The managed server's port is persisted per instance (`remoteOpenchamber.lastUsedPort`) and probed on reconnect (after the user's preferred port), so the daemon left running with `keepRunning` is reused instead of starting another one on a fresh random port each connect.
- A reused daemon on the app-persisted port whose `/api/system/info` reports an `openchamberVersion` different from the app is stopped (with a bounded wait for it to disappear) and restarted from the just-updated binary, whose version the install step has just aligned to the app. A daemon on a user-pinned preferred port is reused untouched at any version — it may be externally managed, mirroring the deliberate non-adoption at disconnect.
- `openchamber serve` output is parsed via the labeled `port N` token with the old first-integer scan as fallback; the serve banner can print the version before the port, which the old parser would misread.

## Non-goals

- Windows OpenSSH transport behavior (independent hidden processes) is untouched.
- External-mode instances are untouched except that they benefit from the PATH preamble like every other remote command.
- No new install methods (`download_release`, `upload_bundle` flows unchanged).
- Remote daemon supervision beyond connect time (no watchdog/health monitoring of the forwarded server) — the reconnect monitor already handles dropped tunnels.

## Affected surfaces

| Surface | Effect |
|---|---|
| `packages/electron` `ssh-manager.mjs` | All `runRemoteCommand` scripts get the PATH preamble; managed-mode port reuse/restart/persist logic; serve port parsing. |
| Desktop SSH settings JSON (`desktopSshInstances`) | New optional `remoteOpenchamber.lastUsedPort` number, written by the app, round-tripped by `sanitizeInstance`, ignored by older versions (additive field). |
| Shared UI / VS Code / mobile | Unaffected; no renderer-facing contract changes. |

## Repository guidance

|---|---|---|
| AGENTS.md runtime boundaries | SSH lifecycle is inherently native Electron behavior | All logic stays in `packages/electron/ssh-manager.mjs`; no shared-UI or web changes. |
| `desktop-shell` skill | SSH/tunnel lifecycle and child-process behavior are owned here | Kept the OpenSSH process patterns (ControlMaster, `-T`, askpass) unchanged; only the remote script contents and managed-server bookkeeping changed. |
| `openchamber-change-discipline` skill | Persisted contract (settings JSON) + platform behavior changed | Additive field with round-trip and failure behavior covered by tests; smallest change that fixes the enumerated failures. |
| `packages/electron/README.md` | Documents SSH invariants | Updated with the PATH preamble and port-reuse invariants. |
| `changelog-authoring` skill | User-facing fix | Two `[Unreleased]` bullets added to `CHANGELOG.md`. |

## Validation

| Check | Result |
|---|---|
| `bun test ./ssh-manager.test.mjs` (electron package) | 16/16 pass, incl. 9 new tests: PATH preamble + nvm numeric-highest selection in every remote command; adoption of a persisted-port daemon without restart; preferred-port reuse without adoption; preferred-port daemon at a different version reused untouched; preferred-before-persisted probe ordering; stale-version restart + port persist; fresh start + persist on dead port; labeled port parse over banner integers; `sanitizeInstance` round-trip of `lastUsedPort`. |
| `node ../../scripts/run-isolated-tests.mjs .` (full electron package) | 15/15 test files pass. |
| `node --check` on both changed files | Pass. |
| Real-host integration (two Ubuntu SSH containers, npm via nvm, dash as `sh`) | `sh -lc '<preamble>; command -v npm'` resolves npm 11.16.0 / 11.17.0 where the bare `sh -lc` form fails; `npm install -g @openchamber/web@1.18.4` succeeds (npmmirror registry); `openchamber serve` output `port 39871 (PID: …)` matches the new parser; live daemon reports `{"openchamberVersion":"1.18.4", …}` from `/api/system/info` (basis of the stale-restart check); `POST /api/system/shutdown` returns 200 and leaves no process. |
| Not verified | A full packaged-desktop click-through of the reconnect/reuse path (covered by unit tests with the exact remote command strings) and Windows hosts (transport code unchanged). |

## Visual evidence

No rendered-UI change: the diff only alters Electron main-process remote command strings, managed-server bookkeeping, and tests. The user-visible effect is limited to SSH status details (e.g. `Restarting managed OpenChamber server 1.18.3 to 1.18.4`) and connections succeeding where they previously errored.

## Risks and failure behavior

- PATH preamble runs before every remote script: it only prepends `[ -d ]`-guarded well-known directories; globs that match nothing are no-ops, and a system-wide install keeps working because it stays on `PATH` behind the prepended dirs.
- `lastUsedPort` probing adds at most two extra SSH probes per connect when no daemon answers (connection-refused fails fast); a probe error is treated as "not running", same as before for `preferredPort`.
- Stale restart: shutdown is best-effort HTTP like the existing disconnect path, bounded by a 5 s wait for the old daemon to disappear before starting the replacement; if the old daemon never stops, `openchamber serve` fails loudly and the connect errors (surfaced, not silent).
- Port parse falls back to the previous first-integer behavior for unexpected output shapes.
- Compatibility: settings written by this version are readable by older versions (extra field ignored); older settings without `lastUsedPort` behave exactly as before.
- Security: preamble adds only directory prefixes on the remote host's own `PATH` for its own user; no secrets are added to command lines, and existing stderr redaction is unchanged.


yulong-ge:
Thanks for the review — both findings addressed in 89cb4f046:

1. **nvm glob ordering**: the preamble now selects only the numerically highest installed version via `sed | sort -t . -k 1,1nr -k 2,2nr -k 3,3nr | head -n 1` (POSIX per-key modifiers; the global `-r` form turned out to be a no-op with `-k` on BSD sort, verified on both BSD and GNU). Non-version entries (nightly/junk dirs) are filtered by the sed pattern. I kept the highest-version heuristic rather than consulting `~/.nvm/alias/default` because the alias is often `lts/*`, which cannot be resolved without sourcing nvm.sh; happy to layer that on (prepend the alias version first when it is a concrete version) if you think the pinned-default case is common enough.

2. **reused daemon vs `keepRunning=false`**: a daemon found on the app-persisted `lastUsedPort` is now adopted by the session (`startedByUs: true`), so disconnect stops it when keep-running is off. Reuse of a user-pinned `preferredPort` deliberately keeps the old non-adopting behavior — that daemon may be externally managed and stopping it would kill a process the app never started.

Also added the suggested coverage: preferred-port-before-persisted-port probe ordering, preferred-port reuse without adoption, and preamble assertions for the new selection pipeline. Validated the real preamble extracted from source through `sh -lc` on the two Linux test hosts (single nvm version each) and the multi-version ordering (v8 legacy vs v24, and 24.9 vs 24.10) on GNU and BSD sort.

yulong-ge:
Both findings from the follow-up review addressed in 83a14080:

1. **Version-restart scope**: took your first suggested fix — the version-alignment restart now applies only to the daemon on the app-persisted `lastUsedPort`. Rationale: that daemon was started by an earlier session of this app and the install step immediately before has just aligned the binary to `appVersion`, so a `!==` restart is self-consistent for it (the daemon cannot meaningfully be "newer" than a binary we just installed). A daemon on a user-pinned `preferredPort` is now reused untouched at any version, which restores the pre-PR behavior for externally-managed daemons and mirrors the non-adoption stance at disconnect. I kept `!==` rather than a "predates" comparison for owned daemons deliberately: mismatched-but-newer would forward the app UI against a different server version. Docs (PR body, README, CHANGELOG) now state the owned-vs-pinned distinction instead of the looser "older version" wording.

2. **Stale PR body**: validation table updated to 16/16 with the nine new tests named, matching 83a14080.

Also noted your residual-risk (b): an auth-failing probe on the persisted port is treated as "not running" and starts a fresh daemon, leaking the old one. That path predates this PR (same handling existed for `preferredPort`), but it is a real accumulation edge now that the persisted port is probed first; I'd rather fix it as a follow-up (probe should distinguish "connection refused" from "401/403" — the latter means the daemon IS running and should be adopted or surfaced as an error, not restarted around).

btriapitsyn:

btriapitsyn:
Thanks for working through the earlier feedback. The server-reuse part is now covered by #3711. Closing this PR as stale after the rebase request. The separate problem with finding Node tooling installed through version managers is still unresolved and needs a focused fix against current main.

PR #2956, opened 2026-08-16
perf(composer): reduce iOS input latency

ChangeHow:
## Intent

Reduce reported iOS composer input latency by no longer restoring the native selection on top of CodeMirror's drawn selection. `@codemirror/view` 6.39.17 supplies its own iOS range handles, so iOS stays entirely on the `drawSelection()` path while Android, desktop web, Electron, and VS Code retain the native-selection workaround.

Related correctness fix: #2691. This PR intentionally does not include that PR's controlled-value writeback guard.

## Non-goals

- Do not fix the IME cursor/writeback bug tracked by #2527; #2691 owns that behavior.
- Do not add `ComposerChange.composing`, manual `compositionend` synchronization, composition-specific language decoration mapping, or `ChatInput` state shortcuts.
- Do not change the Android, desktop web, Electron, or VS Code native-selection arrangement.
- Do not add or retain a committed composer performance benchmark, or treat Chromium profiling as an emulation of WebKit's real IME implementation.

## Affected surfaces

- `packages/ui`: shared `ComposerEditor` selection rendering and the pinned `@codemirror/view` dependency.
- iOS web/PWA and Capacitor surfaces: use CodeMirror's drawn selection and range handles without the native-selection workaround.
- Android, desktop web, Electron, and VS Code: retain the existing native selection layered over `drawSelection()`.
- Root/UI package manifests and `bun.lock`: pin `@codemirror/view` 6.39.17.
- No API, persistence, network, authentication, or stored-data contract changes.

## Repository guidance

|---|---|---|
| `openchamber-change-discipline` | Shared UI and pinned dependency changes affect every UI consumer. | Keeps the runtime difference local to selection rendering, preserves non-iOS behavior, updates the lockfile, and uses current-head focused validation plus the earlier workspace-wide validation. |
| `performance-engineering` | The change targets reported input latency. | Removes duplicate iOS selection ownership, keeps local DevTools measurements explicitly exploratory, and removes the invalid committed harness rather than presenting unreproducible numbers as a regression gate. |
| `theme-system` | The change modifies composer selection rendering. | Retains existing theme tokens and non-iOS selection rules; only installation of the native-selection extension changes on iOS. |
| `writing-for-agents` | The owning composer documentation changes. | Updates the canonical selection invariant without leaving stale performance-tooling guidance. |
| Composer `DOCUMENTATION.md` | Owns editor, selection, IME, and runtime invariants. | Documents the iOS/non-iOS selection split and its dependency on CodeMirror 6.39.17. |
|  |  |  |

## Validation

| Check | Result |
|---|---|
| `bun test packages/ui/src/components/chat/composer/editor/__tests__/theme.test.ts packages/ui/src/components/chat/composer/editor/__tests__/composerLanguage.test.ts` | Current HEAD: 29 passed, 0 failed, 52 assertions. Covers iPhone/iPadOS detection, non-iOS fallback, theme installation, and language-state behavior. |
| `bun run type-check:ui` | Current HEAD: passed. |
| `bun run docs:validate` | Current HEAD content: passed, 450 pages and 45 sidebar links. |
| `git diff --check origin/main...HEAD` | Current HEAD: passed. |
| `rg -n 'bench:composer-input\|composer-input-benchmark\|composer-input-summary\|cmTile' package.json scripts packages` | Current HEAD: no matches. The invalid benchmark, package entry point, documentation, and `cmTile` dependency are absent. |
| `bun run dead-code` | After benchmark removal: completed with the existing non-blocking repository inventory and no composer benchmark finding. |
| `bun run type-check`; `bun run lint`; `bun run test`; `bun run build` | Passed on the same selection/dependency implementation at `ba6ffcd4d`; the final follow-up commits only remove the invalid opt-in benchmark and restore unchanged perf documentation. These broad commands were not rerun after those removal-only commits. |
| `bun run build:web` | Passed after restoring the final selection implementation for the original local A/B capture. |
| `bunx oxlint packages/ui/src/components/chat/composer/editor/theme.ts` | Passed for the authored selection implementation. |
|  |  |

### Exploratory local performance evidence

The reported lag was reproduced locally against production mobile builds. Chrome DevTools was used with a 390x844 viewport, an iOS user agent, 6x CPU throttling, and 24 incremental pinyin replacements inside synthetic composition boundaries. Three before and three changed-policy captures were taken, with 120 composer samples per capture after warm-up.

Both variants used CodeMirror 6.39.17 and differed only in whether the iOS native-selection workaround was installed. A temporary, uncommitted inspection hook exposed the editor state during this investigation. The invalid repository harness and its documentation have been removed; these numbers are exploratory local evidence, not a benchmark reproducible from the reviewed source and not a CI budget.

| Metric, median of three captures | Native-selection workaround | iOS CodeMirror handles | Change |
|---|---:|---:|---:|
| Composer input-to-idle p95 | 8.2 ms | 5.7 ms | -30.5% |
| Composer/native idle p95 ratio | 4.316x | 2.773x | -35.8% |
| Composer input-delivery p95 | 7.0 ms | 4.7 ms | -32.9% |
| Composer next-frame p95 | 14.5 ms | 13.3 ms | -8.3% |

Individual composer input-to-idle p95 captures were `8.0 ms`, `9.3 ms`, and `8.2 ms` with the workaround, versus `5.4 ms`, `6.1 ms`, and `5.7 ms` with CodeMirror's iOS handles.

## Visual evidence

No screenshot is included. The primary contract is input responsiveness and which selection implementation owns iOS range handles; a static screenshot cannot prove either behavior. The navigator-policy tests directly prove the platform branch, and the local before/after DevTools measurements above exercise the latency hypothesis.

Static visual parity is not claimed. CodeMirror's drawn selection layer sits behind content, so selection tint across backgrounded tokens such as fenced code may differ when the native iOS selection is removed. Real iPhone WebKit typing, long-press handles, and selection tint across backgrounded tokens were not manually re-verified for the final HEAD.

## Risks and failure behavior

- The optimization depends on the iOS handles introduced in `@codemirror/view` 6.39.17; downgrading that dependency would invalidate the platform branch.
- iPadOS desktop-style user agents are detected through `MacIntel` plus touch points; this heuristic can drift from CodeMirror's internal iOS detection in future versions.
- Non-iOS runtimes retain the existing native-selection workaround, avoiding intentional behavior changes outside iOS.
- The local Chromium measurements use synthetic composition and a temporary inspection hook. They support the latency hypothesis but cannot prove real WebKit/IME behavior or serve as a reproducible regression gate.
- On iOS, selection tint can be obscured by token backgrounds because CodeMirror's drawn selection layer is behind content; real-device acceptability remains unverified.
- Rollback consists of restoring the native-selection extension on iOS and reverting the CodeMirror dependency update. No persisted state, migration, cleanup, or partial-failure path is involved.


ChangeHow:
Maintainer review context:

Real iPhone/WebKit IME, long-press handle, and backgrounded-token selection evidence is difficult to capture in my current environment. I reproduced the input-lag hypothesis locally with production mobile builds, Chrome DevTools, an iOS user agent, and 6x CPU throttling; the exploratory before/after results and their limitations are disclosed in the PR body. The temporary inspection hook was not committed, and the invalid repository benchmark was removed rather than presented as reproducible validation.

Current-HEAD composer policy tests and UI type-check pass. The CI relay timeout did not reproduce locally: the target test passed 21/21 isolated runs and also passed in a full web suite. The remaining real-device visual/interaction gap is explicitly listed as residual risk. Requesting maintainer judgment on whether this evidence is sufficient for the scoped iOS selection-policy change.

btriapitsyn:
Thank you for digging into this and especially for identifying the native-selection workaround as the likely source of the iOS input latency. That diagnosis gave us the key direction for the fix.

We have now landed an independent implementation on main in https://github.com/openchamber/openchamber/commit/26b0ad5bd. It keeps CodeMirror drawSelection enabled, updates the CodeMirror stack, uses CodeMirror own iOS handle path only when its exact platform predicate matches, and keeps the WebKit system highlight as the single visible fill while preventing handle clipping.

We chose this route rather than merging the PR as-is because it preserves the existing non-iOS selection fallback and addresses the layering, duplicate-fill, and handle-clipping issues found during device testing. Closing this PR now that the replacement is on main. Thanks again for the careful report and prototype; they directly informed the shipped solution.

PR #2963, opened 2026-08-17
fix(ui): normalize shell widget copy and harden clipboard fallback (#2876)

### LINKED ISSUE #2876: [Bug] Pasting from the copy button in the shell widget to the terminal fails

This bug does not always happen, but I would say 90% of the time. Sometimes an agent suggests some shell commands to run and does so in a dedicated widget (see screenshot). One can then click the `copy` button to copy the commands into the clipboard. This seems to work: I can paste in an external terminal just fine.

But if I paste into the integrated terminals, most of the time I get a single character pasted (apparently a random one). This does not happen when selecting text manually and clicking copy from the right click menu (or just Cmd+C). And as mentioned, it does not happen when pasting in external terminals, so it seems an interaction between the shell widget in chat and the integrated terminals.

### Where does it happen?

Desktop (macOS)

### Version (if known)

1.18.2

Image

### Logs (optional)

```shell

```

bashrusakh:
## Intent

Fixes the copy-path half of #2876. Two independent clipboard/copy defects remained on `main` after the ghostty-web layer was replaced by the in-repo terminal adapter:

1. **Shell-widget copy wrote the raw terminal stream.** Copied bash output kept `\r`, `ESC[2K`, SGR, OSC and other control sequences; pasting into a terminal replayed them and collapsed the output to its last frame. `renderTerminalOutput` now interprets the visible text (C0/C1 handling, CSI with malformed-sequence fallback, OSC/DCS/SOS/PM/APC strings, unknown-ESC drop) and `getShellClipboardText` is what the shell action copies.
2. **The clipboard fallback could leave a stray textarea and never restored focus.** `copyTextToClipboard` now tears down in `finally`, restores the previously focused element best-effort, and the shell copy call sites surface the existing copy-failure toast.

The old branch also carried three concerns that are obsolete on current main: the ghostty-web copy-on-click bundle patch and touch long-press offset (dependency removed in `e5c2c81bf`), and `tr.ts` i18n keys (fixed on main by `504fd8532`). They are gone from this candidate.

## Non-goals

- No terminal viewport/selection changes, no ghostty patch or dependency change, no i18n changes.
- No redesign of clipboard behavior or error surfaces.
- OSC 8 hyperlink URLs are dropped on copy (display label kept), matching terminal-emulator copy behavior.

## Affected surfaces

- `packages/ui/src/components/chat/message/parts/toolOutput.ts`, `MessageBody.tsx`, `ToolPart.tsx`, `packages/ui/src/lib/clipboard.ts` + tests.
- Shared UI used by web, desktop, VS Code, and mobile. No persisted data, API, or transport changes.

## Repository guidance

|---|---|---|
| Root `AGENTS.md` | Shared UI across runtimes | Copy normalization stays in the owning output helper; no runtime-specific code |
| `chat/.../parts/DOCUMENTATION.md` | Shell output copy representation | `renderTerminalOutput` remains the single normalization path; documentation note added |
| `openchamber-change-discipline` | Focused source change | No dependencies, no unrelated cleanup, tests next to the change |

## Validation

| Check | Result |
|---|---|
| `bun test ToolPart.test.ts clipboard.test.ts shellBridge.repro.test.ts` | 54 pass / 0 fail; reverting the two production files fails 10 + 1 error |
| Representative consumers (`ToolOutputDialog.test.ts`, `toolRenderers.test.ts`) | 9 pass / 0 fail |
| `bun run type-check:ui`, `bun run lint:ui` | Passed |
| `bun run docs:validate` | Passed (517 pages) |
| `bun run dead-code` | No finding for `getShellClipboardText` |
| Adversarial parser harness (5 MB malformed CSI/OSC) | Terminates; no truncation of legitimate text |
| Whole-PR review | PASS-WITH-FINDINGS, no blocking findings at `09c04d615` |

## Visual evidence

No rendered visual change: this changes the clipboard payload, not the UI.

## Risks and failure behavior

- Unterminated OSC/DCS at end-of-output drops the remaining bytes (terminal-accurate; same as before).
- `document.body.removeChild` runs in `finally` outside the inner catch; both new call sites catch, and older call sites already tolerate rejection.

## Dropped from the old branch

- Ghostty-web copy-on-click patch and touch long-press offset — the dependency and patch target no longer exist on `main`.
- `gitView.empty.*` i18n keys — already present on `main`.

---

CI note (2026-09-13): the `checks` job is red only on `packages/vscode/webview/api/settings.test.ts` ("propagates a failed bridge read and retries successfully"), which times out deterministically on clean current `main` (`961f1611c`, reproduced locally 3/3; the test was added by main commit `6cf247894`). It is unrelated to this diff — every fresh PR head on this base fails it. Do not treat this red check as a finding against this PR.


btriapitsyn:

PR #2970, opened 2026-08-17
feat(web): add Command Code quota provider to usage page

inoribea:
## Intent

Adds a **Command Code** quota provider to the Settings → Usage page so users who route OpenCode through Command Code's OpenAI-compatible endpoint (`api.commandcode.ai`) can see their plan consumption next to their other providers.

The provider reads the Command Code API key from the shared OpenCode `auth.json` (the same file OpenChamber's other providers read) and calls Command Code's billing credits endpoint. It renders three signals:

- the **5-hour window** (`windowLimits.fiveHour`) as a usage bar with a reset countdown,
- the **7-day window** (`windowLimits.weekly`) as a usage bar with a reset countdown,
- the **credits balance** (`credits.monthlyCredits + purchasedCredits + freeCredits`) as a plain dollar label.

This is the equivalent of running `/usage` in the Command Code CLI, surfaced inside OpenChamber.

## Non-goals

- No new dependencies — the adapter uses `fetch`, the shared quota utilities, and the existing `readAuthFile`/`getAuthEntry` auth helpers.
- No change to the Provider API surface or the `fetchQuotaForProvider` dispatcher contract — `commandcode` is registered exactly like every other provider in the registry.
- No changes to opencode itself, the small-model path, TTS, or any other subsystem.
- No billing or subscription information beyond the two usage windows and the credits balance — no per-request cost breakdown (that lives in Command Code Studio).

## Affected surfaces

- `packages/web/server/lib/quota/providers/` — new adapter module + registration in the provider registry.
- `packages/web/server/lib/quota/DOCUMENTATION.md` — added the provider row to the supported-provider table, as required by that file's "Add a new provider" checklist.
- `packages/ui/src/lib/quota/providers/index.ts` — added `{ id: 'commandcode', name: 'Command Code' }` to `QUOTA_PROVIDERS`; the Usage page and the header dropdown render it automatically (both are data-driven off this list).
- `packages/ui/src/types/quota.ts` — extended the `QuotaProviderId` union with `'commandcode'`.

Unaffected but worth stating: the `openai` provider module stays unregistered (it remains internal-only for logic parity), and the VS Code extension's separate quota list is not touched — the extension does not surface the Usage page.

## Repository guidance

|---|---|---|
| `openchamber-change-discipline` | New source file, new module in the quota registry, shared auth/formatting helpers, and a `DOCUMENTATION.md` row — all "local implementation" / "module contract" risk | Followed the validation matrix: focused vitest tests, package-scoped type-check and lint, and `bun run dead-code` (knip) for the new files. Auth helpers were reused from `quota/utils/index.js` rather than reimplemented. |
| `settings-ui-patterns` (companion: `theme-system`, `locale-ui-patterns`) | The Usage page is a Settings surface; the provider appears in the settings UI | No components, layout, controls, or visible strings were added or changed. The Usage page is fully data-driven off `QUOTA_PROVIDERS`; adding one metadata entry requires no component work. |
| `locale-ui-patterns` | Any user-facing string should go through i18n | The only new strings are the provider display name (`Command Code`) — provider metadata, consistent with every existing provider name — and server-side error messages from `fetchQuota`. Neither is component-rendered UI copy; both follow the existing adapter conventions (e.g. `deepseek.js`, `crof.js` emit `$`-prefixed value labels and plain-text errors). |
| `quota/DOCUMENTATION.md` (module owner doc) | The file's "Add a new provider" checklist step 6 requires updating the supported-provider table | Added the `commandcode` row with module path and auth alias (`commandcode`, key under `key` or `token`). |
| `AGENTS.md` "Do not add dependencies" | The change must not add packages | Zero new dependencies — uses `fetch` (already used by `openrouter.js`), `AbortSignal.timeout`, and existing shared helpers. |

## Validation

| Check | Result |
|---|---|
| `bunx vitest run server/lib/quota/providers/commandcode.test.js server/lib/quota/providers/index.test.js` | **11 passed / 11** (9 new `commandcode` tests + 2 registry tests). Covers: happy path with the real documented payload (all three bars), `resetAt === 0` inactive-window omission, `limited: false` pay-as-you-go degradation, missing-credits balance omission, 401/403 → session-expired mapping, 5xx → API error, JSON parse failure, timeout. |
| `bunx vitest run server/lib/quota/providers` | **56 passed / 56** — full providers suite, confirms the registry integration breaks nothing. |
| `bunx tsc --noEmit` (packages/ui) | **exit 0** — the `QuotaProviderId` union change type-checks across the UI workspace. |
| `bun run lint:ui` / `bun run lint:web` | **exit 0** both. |
| `bunx knip@5.80.0` (`bun run dead-code`) | **exit 0** — no dead-code findings for the new files. |
| Live endpoint check | `GET https://api.commandcode.ai/alpha/billing/credits` with a Bearer key from `auth.json` returns `credits` + `windowLimits.fiveHour/weekly`; `resetAt` is a 13-digit **millisecond** epoch (normalized via the existing `normalizeTimestamp`). |

Not verified: the full `bun run build` (workspace-wide) was not run — the change is additive to existing JS modules and a TS union; package-scoped type-check/lint plus the quota test suite were run instead, per the change-discipline validation matrix. The existing `credentials/store.test.js` failure (`0o600` POSIX mode assertion on Windows) is pre-existing and unrelated.

## Visual evidence

**Before** — a pre-change capture is not available: the patched build is already installed locally, and reverting the install to re-capture would require restoring the original asar and restarting the app. The before state is simply the same Settings → Usage page without the Command Code row; the QUOTA_PROVIDERS diff in packages/ui/src/lib/quota/providers/index.ts (one added line) is the exact delta.*

**After** — Command Code appears in the Settings → Usage list with three bars: 5-hour window, 7-day window, and credits balance:

Command Code usage after

The change adds one data-driven row to the QUOTA_PROVIDERS list; the row renders identically in narrow and wide settings panes (it is a list item, not a custom layout), so the narrow capture below shows the same three bars in the narrow pane:

Command Code usage after (narrow pane)

## Risks and failure behavior

- **Endpoint availability**: `GET /alpha/billing/credits` is a Command Code backend endpoint not documented in their public Provider API reference. It is what the Command Code CLI's `/usage` command polls (confirmed by third-party integrations that mirror the CLI), so it is unlikely to disappear, but it is not a public contract. If the endpoint changes shape, the adapter degrades to the existing `ok: false` error path with a readable message — it never crashes the Usage page or blocks other providers (each provider fetch is isolated).
- **Auth failure**: missing/invalid key returns `configured: false` / `Session expired` and the provider simply doesn't render a bar — consistent with how every other API-key provider behaves.
- **`resetAt === 0`**: window not yet activated — omitted rather than rendered as a misleading 1970/zero bar.
- **Pay-as-you-go accounts**: `windowLimits.limited === false` — window bars omitted, credits balance still shown.
- **No secrets**: the adapter never logs or stores the key; it only reads it from the shared `auth.json` and sends it as a Bearer header. The test file uses fake keys only.
- **Rollback**: this is a purely additive change (one new module + one registry entry + one UI metadata row); reverting is a one-line deletion per file.


inoribea:
-

Issue #2769, opened 2026-08-08
[Bug] Shiki markdown highlight worker burns ~55% CPU continuously (sustained re-highlighting of existing content, not a single backtrack)


xjbeta:
### Bug description

**Note: this report was drafted by an AI assistant (opencode) on behalf of a user, from analysis of a Chrome DevTools performance trace the user recorded (Trace-20260808T232826.json, 52.4 s window, ~1.21M events). Facts below come directly from that trace; the user can provide the trace file on request.**

The Shiki markdown highlighting worker burns ~55% of all sampled CPU continuously, on a machine that is otherwise idle (no streaming, no message sending — the user was merely browsing files in the file panel). This is a sustained background load, not a one-time highlight of a large document.

### Measurements (from the DevTools trace)

Sampled CPU profile (1,113,759 V8 samples across the window):

| Symbol | Share of all samples | Call stack | Interpretation |
|---|---|---|---|
| `ut` (markdown-shiki.worker-*.js) | **34.4 %** (383,337) | `ut <- Wo <- (root)` | Shiki `codeToHtml` highlight execution; **99.1 % of the browser-process isolate's samples** |
| `Un` (markdown-shiki.worker-*.js) | **20.3 %** (225,600) | `Un <- processQueue <- _loadGrammar <- loadGrammarWithConfiguration <- loadLanguage <- loadLanguages` | Grammar loading / highlight queue processing |

- Time distribution is **flat over the entire 52 s**: `ut` shows ~29,500–30,600 samples per 4-second bucket in every single bucket (≈ 75 % of one full CPU core, never stopping).
- The worker isolate lives in the **browser process** (V8Worker, pid 85001), so it does not block the renderer, but it pins one core permanently.
- Main thread receives highlight results at **~40 messages/second** (`Mi.e.onmessage`, 2119 calls / 52 s) — i.e. the worker is constantly producing and the UI is constantly consuming re-highlighted content.
- Renderer main thread ~43 % busy (22.65 s RunTask); `Layerize` 14.4 s / 3940×, `UpdateLayer` ~464k events → the page re-renders at ~75 fps the whole time, mostly re-rendering already-visible content.
- No task longer than 50 ms except a single 252 ms `EvaluateScript` — so this is *not* a single pathological match; it is **massive repeated re-highlighting**.

[...]

PR #2610, opened 2026-08-04
fix(ui): keep empty composer caret visible

TTTPOB:
## Intent

Keep the CodeMirror composer caret visible when the prompt is empty. The drawn cursor sits at the scroller's zero-width horizontal boundary; a 1px inline content inset keeps its border inside the scroller clip while moving the placeholder and typed text through the same coordinate system.

## Non-goals

- Change CodeMirror cursor color, blink timing, or native-selection behavior.
- Change composer sizing, scrolling, or input semantics.

## Affected surfaces

- `packages/ui`: shared chat composer CodeMirror theme.
- User-visible empty prompt state, observed in the VS Code webview.
- The 1px content inset applies to every runtime using the shared composer so cursor, placeholder, and text alignment remain consistent.
- No persisted data, external API, runtime bridge, or package contract changes.

## Repository guidance

| Guidance | Why it applies | How the change complies |
|---|---|---|
| Root `AGENTS.md` and `openchamber-change-discipline` | This changes shared UI source and platform-visible editor behavior. | Keeps the change to one owning theme rule, preserves existing CodeMirror selection behavior, and records the manual runtime validation. |
| `theme-system` | The change modifies component styling. | Uses a logical layout property and introduces no color, token, icon, or dependency changes. |
| `packages/ui/src/components/chat/composer/DOCUMENTATION.md` | The composer editor owns its drawn caret and documents that DOM rendering requires manual verification. | Changes `editor/theme.ts` and leaves `drawSelection()`, native selection, and mobile caret behavior untouched. |

## Validation

| Check | Result |
|---|---|
| VS Code webview DevTools, empty composer | Confirmed the CodeMirror cursor layer existed and blinked after moving the cursor inside the horizontal clip. |
| Temporary `padding-inline-start: 1px` on `.cm-content` | Confirmed the empty caret became visible and blinking while caret, placeholder, and text remained aligned. |

## Visual evidence
cursor in empty text input is working now.
image

## Risks and failure behavior

The shared composer content shifts by 1px at the inline start in every runtime. The cursor, placeholder, and text share that inset, avoiding the misalignment produced by translating only the cursor. The change does not touch editor state, input handling, persistence, or security-sensitive behavior. Rollback consists of removing the single `paddingInlineStart` declaration.

PR #2993, opened 2026-08-18
fix(chat): show a visible caret in the focused empty composer

### LINKED ISSUE #2518: [Bug] Empty chat input hides blinking cursor, making focus state ambiguous
### Description

When the chat input box is empty (contains no characters), the blinking cursor (caret) is not visible. This makes it difficult to determine whether the input area is currently focused and ready to accept text, especially on the Web UI.

### Steps to Reproduce

1. Open the chat interface (Web UI)
2. Click into the chat input box
3. Type nothing — leave it empty
4. Observe the cursor visibility

### Expected Behavior

A blinking cursor (caret) should always be visible when the input area is focused, even when the input is empty, so the user can clearly see that the text field is active and ready for input.

### Actual Behavior

When the input box is empty, no blinking cursor is shown. The text field appears blank without any visual indication of focus, making it ambiguous whether the input area is ready for typing.

### Environment

- **Version**: CLI + Web (OpenCode CLI + OpenChamber Web UI)
- Provider: DeepSeek V4 Flash
- Mode: plan

### Impact

This is a usability issue that affects the user experience, particularly for new users who may not immediately know if they can start typing.

bashrusakh:
## Intent

Fixes #2518. In the web UI, the chat input showed no blinking cursor when focused and empty, making focus state ambiguous. Root cause: the composer is a CodeMirror editor running `drawSelection()`, which hides the native caret (`caret-color: transparent`) and paints its own `.cm-cursor` element — but for an empty document the caret position resolves to zero-width geometry that reports no client rects, so no drawn cursor appears. The fix re-enables the native caret while the placeholder is showing (empty doc) and hides the drawn cursor layer, scoped to `&.cm-editor.cm-focused:has(.cm-placeholder)` so it releases automatically on the first keystroke.

## Non-goals

- No change to the drawn-cursor path when the document has text.
- No change to the iOS selection-handle workaround (the fallback releases before any keystroke-driven decoration redraw).

## Affected surfaces

- `packages/ui/src/components/chat/composer/editor/theme.ts` — shared composer theme, reaches web, desktop, VS Code, and mobile. All composer surfaces (shell mode, mini/compact chat, expanded focus) share this theme.

## Repository guidance

|---|---|---|
| AGENTS.md root guide | Mandatory before editing | Read; minimal diff, no unrelated changes |
| `theme-system` skill | Styling/token rules | Uses semantic `--surface-foreground` token; no hardcoded colors |
| `openchamber-change-discipline` | Change risk classification | Local theme change at the shared primitive; validated at package level |
| `changelog-authoring` | Changelog entry | One bullet in `[Unreleased]` |

## Validation

| Check | Result |
|---|---|
| `bun test src/components/chat/composer/editor/` | ✅ 45 pass / 0 fail (17 in theme file incl. 3 new regression tests) |
| `bun test src/components/chat/composer/` | ✅ 291 pass / 0 fail |
| `bun run type-check` (packages/ui) | ✅ pass |
| `bun run lint` (packages/ui) | ✅ pass |
| `bunx oxlint` on changed files | ✅ no findings in authored code |
| Live browser checks (Chromium + Firefox, dark theme, narrow composer, shell mode, expanded focus, mini-chat) | ✅ empty+focused: native caret paints and blinks; typing releases fallback; deleting re-engages |
| Full packages/ui suite | ⚠️ 2076 pass / 254 fail / 13 errors — all failures pre-existing in unrelated files (e.g. markdownCore htmlCache #2769); every touched-area suite is green |

## Visual evidence

No screenshot captured in this environment; the live-browser verification above covers the rendered behavior (caret visible when focused+empty, released on typing).

## Risks and failure behavior

- `:has()` support: Chrome 105+, Firefox 121+, Safari 15.4+. On legacy engines the rule is ignored and behavior falls back to the pre-fix state (no regression).
- iOS: the fallback only engages while the placeholder is showing; the moment the first character is typed it releases, so the documented iOS input-latency concern does not apply.
- Behavior with text: unchanged — the drawn cursor path is untouched.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ All notable changes to this project will be documented in this file.
 - Chat: saved chats in the context panel open again instead of staying blank.
 - Chat: the context meter no longer climbs over 100% (330% readouts) after turns with many tool calls and no longer jumps when reopening an older session; it now shows what the window actually holds, everywhere the value appears — header, context sidebar, work status panel, mini chat, and mobile (thanks to @pocharlies).
 - Chat/Attachments: extracted Office and OpenDocument content is now capped and presented more compactly, preventing large documents and their images from overwhelming the message context.
+- Chat: the input now shows a blinking cursor as soon as it is focused, even when empty, so it is clear the field is ready for typing.
 - Projects: project names now match the folder name exactly, so `.ssh` and `opencode-claude` are no longer shown as `.Ssh` and `Opencode Claude` in the sidebar, window title, settings and notifications; names you renamed yourself are kept.
 - Files: files reached through a symlink inside the workspace now open correctly instead of being rejected as outside the workspace.
 - Settings: the session retention action you pick is now saved instead of being dropped (thanks to @Gautam0507).
--- a/packages/ui/src/components/chat/composer/editor/__tests__/theme.test.ts
+++ b/packages/ui/src/components/chat/composer/editor/__tests__/theme.test.ts
@@ -41,10 +41,23 @@ describe('composerEditorTheme', () => {
      * `caret-color: transparent !important` and draws a `.cm-cursor` element
      * instead. Styling `caret-color` looks correct and does nothing, leaving
      * CodeMirror's hard-coded black cursor on dark themes.
+     *
+     * The one exception is the empty-document fallback (issue #2518): while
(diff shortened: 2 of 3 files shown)


btriapitsyn:
Closing as a duplicate of #2610, which addresses the same issue with a smaller change and will be reviewed instead.

PR #2994, opened 2026-08-18
fix(vscode): start editor new sessions in the current workspace folder

### LINKED ISSUE #2889: "Open New Session in Editor" uses previous workspace directory instead of current

When using the VS Code extension, if I create a new session from the sidebar (e.g., clicking the “openchamber” button), it correctly uses the current workspace directory for the chat session.

However, if I click the “Open New Session in Editor” button located above the editor, the session is opened using the directory from the last session I started, rather than the currently active workspace folder.

Steps to reproduce:

Open VS Code with workspace folder A.

Create a session from the sidebar (“openchamber”) – this correctly uses folder A.

Now click the “Open New Session in Editor” button at the top of the editor.

The new session opens, but it uses the directory from a previously started session (e.g., folder B), not folder A.

Expected behavior:
The “Open New Session in Editor” should use the current workspace directory (folder A), just like the sidebar action.

Actual behavior:
It uses the last opened/used directory instead of the current one.

Additional context:
This happens when creating a new window/editor session. The sidebar action works as expected, but the editor button does not respect the current workspace.

### Where does it happen?

VS Code extension

### Version (if known)

1.18.3

Image
Image

### Logs (optional)

```shell

```

bashrusakh:
## Intent

Fixes #2889. "Open New Session in Editor" now initializes the editor draft from the current VS Code workspace folder, using the folder already embedded in the panel bootstrap HTML. The previous one-shot `newSession` postMessage could be delivered before the webview bridge was ready; this implementation passes the bootstrap folder to `openNewSessionDraft({ automatic: true, directoryOverride })`, and the editor action shows an information message instead of opening a folderless draft when no workspace folder is open.

## Non-goals

- Sidebar `openchamber.newSession` flow unchanged.
- Multi-root editor sessions keep first-folder behavior.
- Existing-session panels, persisted formats, and public APIs unchanged.

## Affected surfaces

- `packages/vscode/src/SessionEditorPanelProvider.ts` (no-folder guard).
- `packages/ui/src/components/layout/VSCodeLayout.tsx` (editor-only override).
- `packages/ui/src/lib/vscodeBootstrap.ts` (typed helper + trim).
- Tests for the helper, panel guard, webview bootstrap, and the editor path.

## Repository guidance

|---|---|---|
| Root `AGENTS.md` | VS Code bridge + shared UI draft state | Extension host stays thin; draft selection in shared UI |
| `ui-api-decoupling` | Webview bootstrap contract | Reuses existing `__VSCODE_CONFIG__`; no new bridge message |
| `sync-state-invariants` | Draft directory selection | Uses the existing `directoryOverride` owner path |
| `vscode/README.md`, `vscode/src/DOCUMENTATION.md` | Module ownership | No new ownership |

## Validation

| Check | Result |
|---|---|
| Branch head: 5 focused suites (bootstrap, panel, runtime, webviewHtml, session-ui-store) | 2 + 4 + 3 + 1 + 38 pass; 0 fail |
| Merged into current main (local integration tree) | panel 2/0, webviewHtml 1/0, bootstrap 4/0, store-init 1/0, session-ui-store 56/0 |
| `bun run type-check:ui` (merged tree) | Passed |
| Whole-PR review | PASS at `3db73d6a4` |
| Post-merge test fix | One-line vscode `env.language` mock addition needed for main's `webviewHtml`; the merged panel test fails without it |

## Visual evidence

No rendered layout change; the selection happens during draft initialization.

## Risks and failure behavior

- The explicit override restates the folder main's directory store already selects on a fresh panel; it also guarantees the selection when a retained panel's directory has drifted. Non-harmful and deterministic.
- The no-folder message is the existing hardcoded English string used by the sidebar path.
- Live Extension Host verification was not performed in this environment.

---

CI note (2026-09-13): the `checks` job is red on the current run at `packages/ui/src/components/views/useFilePreviewScrollPosition.test.tsx` (passes 7/7 on clean current `main`, so likely a runner flake) and, once that stage passes, on `packages/vscode/webview/api/settings.test.ts` ("propagates a failed bridge read and retries successfully"), which times out deterministically on clean `main` (`961f1611c`, reproduced locally 3/3; added by main commit `6cf247894`). Both are unrelated to this diff.


btriapitsyn:
Thanks — the bug is real and the diagnosis in the description is right. One blocking issue before this can merge:

1. **The `newSession` postMessage can be silently dropped, so the fix is unreliable.** The PR posts the command once, immediately after `createWebviewPanel`. `webviewCachedStateRetry.ts` in this same package documents that VS Code drops `postMessage` calls made before the webview's bridge is ready (common in code-server and on slow machines) — that's why the connection status is re-sent with staggered retries. When the message is lost, the draft falls back to the last session's directory and the original bug returns, just intermittently. Blind retries aren't safe here either: re-delivering `newSession` after the user has started typing would reset their draft. Since the panel's HTML bootstrap config already carries the workspace folder (as your description notes), the reliable fix is on the webview side: have the panel's automatic draft open use the bootstrap-config directory, scoped so the sidebar flow is unaffected. Done looks like: no post-creation `newSession` message needed, and the draft directory is correct even when the webview boots slowly.

The no-folder guard and the tests are good as they are.

yulia-ivashko:
I ran the full `checks` job locally on this branch merged onto current main, plus a VS Code extension build. Everything passed. Thanks for reworking this the way the earlier review asked: taking the folder from the panel's bootstrap config means the draft lands in the right place even when the webview is slow to boot. Merging.

yulia-ivashko (approved):

PR #2997, opened 2026-08-19
fix: resend connectionStatus at staggered delays to prevent stuck loading screen on slow networks

### LINKED ISSUE #2996: [Bug] OpenChamber sidebar view stuck on the loading screen in code-server (`#initial-loading` never disappears)

# [Bug] OpenChamber sidebar view stuck on the loading screen in code-server (`#initial-loading` never disappears)

## Environment

- **OS**: Ubuntu 26.04 (x86_64)
- **Editor**: code-server 4.133.0 (VS Code OSS 1.133.0)
- **Access**: remote via reverse proxy `https://<your-code-server-host>:8266` (TLS); LAN direct `http://<lan-ip>:8668`
- **Extension**: `fedaykindev.openchamber`, tested 1.18.4 and 1.19.0 (both affected)
- **opencode CLI**: 1.18.18 (sst/opencode official, installed via npm)
- **Backend**: extension-managed `opencode serve --hostname 127.0.0.1 --port <random>` — health check OK (`/global/health` → `{"healthy":true,"version":"1.18.18"}`)

## Description

The OpenChamber **sidebar chat view** frequently fails to open: it stays forever on the extension-injected loading screen `<div id="initial-loading">` (the cube logo animation) and never proceeds to the UI.

### Key observations

| Scenario | Result |
|---|---|
| Remote access over a **slow/unstable network** | **Often stuck on loading screen** |
| LAN direct access / local access to the public URL | Works normally |
| Drag the sidebar view to the other side (webview recreated) | **Sometimes recovers** |
| "Open Session in Editor" (editor panel) | Always works |
| Console noise (unrelated) | `POST /auth/url-token 405`, `GET /api/openchamber/events 404`, EventSource infinite exponential-backoff retries |

### Console noise (confirmed NOT the cause)

```
POST https://<your-code-server-host>:8266/auth/url-token  405 (Method Not Allowed)
GET  https://<your-code-server-host>:8266/api/openchamber/events  404 (Not Found)
renderVSCodeApp.js:6834  EventSource retry loop (aoe/fj)
```

> The same errors appear on LAN access where the UI loads fine — they are background event-stream noise, not the cause of the stuck loading screen.

## Steps to Reproduce

1. Serve code-server behind a reverse proxy (TLS).
2. Open it from a slow/unstable network (e.g., 3G/4G throttling, or any WAN connection).
3. Click the OpenChamber icon in the activity bar (or run `openchamber.openSidebar`).
4. The sidebar stays on the `#initial-loading` loading screen indefinitely.
5. Moving the sidebar view left↔right sometimes makes it recover.

### Loading screen gating

`packages/ui/src/components/layout/VSCodeLayout.tsx` (~line 380):

```ts
if (isInitializing || hasInitializedOnce || connectionStatus !== 'connected') {
  return;
}
```

Bootstrap only runs once `connectionStatus === 'connected'`; the webview **never pulls** this state — it only receives it via extension-host `postMessage`.

### Lost message

`packages/vscode/src/ChatViewProvider.ts`, `_sendCachedState()`:

```ts
private _sendCachedState() {
  if (!this._view) {
    return;
  }
  this._view.webview.postMessage({
    type: 'connectionStatus',
    status: this._cachedStatus,
    error: this._cachedError,
  });
  this.notifyWindowFocusChanged(vscode.window.state.focused);
}
```

Problems:
1. `connectionStatus` is sent **only once per status change** — no return-value check, no retry, no ack.
2. **VS Code drops `postMessage` calls made before the webview's `acquireVsCodeApi` bridge is ready** (well-known behavior).
3. On slow/remote networks the webview assets (several MB) load slowly → the bridge becomes ready late → the single `connectionStatus=connected` send lands in the drop window → **no compensating resend → loading screen stuck forever**.

### Why the symptoms match

- **Remote stuck / LAN fine**: loading speed difference (late bridge readiness → dropped message).
- **Moving the sidebar recovers**: when the webview is recreated the extension host is already `connected`, so the **HTML-injected `initialStatus=connected`** hides the loading screen immediately.
- **"Open Session in Editor" works**: created on click when the extension host is already `connected` → HTML-injected `connected`.
- **A custom `-fix` (webview-ready message queue) had no effect**: queuing cannot fix VS Code dropping `postMessage` when the bridge is not yet ready.

## Workaround / proposed fix (locally verified)

On top of the 1.19.0 sources, we added `_scheduleCachedStateRetries()` to all three providers (`ChatViewProvider`, `AgentManagerPanelProvider`, `SessionEditorPanelProvider`): when the cached status is `connected`, re-send `connectionStatus` at staggered delays of **500ms / 1.5s / 3.5s / 7s / 12s / 20s** (stops once the view is disposed/replaced; message is idempotent). Hook points: ① after the webview is created; ② after `updateConnectionStatus('connected')`.

The fix was **verified working on the affected remote network** (loading screen disappears, repeated refreshes/reconnects do not reproduce).

Suggestions for upstream:

1. Staggered resend of `connectionStatus` for the `connected` state (simplest, reliable);
2. Or route `connectionStatus` through the existing `_sendMessageWithRetry` / `bridge:ack` confirmation mechanism (more robust, requires webview-side ack);
3. Or have the webview pull the latest status once its bridge is ready (`acquireVsCodeApi`).

## Notes

- An occasional `Error: Webview is disposed` appears in the extension-host log (AgentManagerPanelProvider message callback); it is rare and not the primary cause here.
- `POST /auth/url-token`, `GET /api/openchamber/events`, `/api/opencode/health` do not exist on sst/opencode 1.18.18's server (they return the web UI HTML) — they look like OpenChamber Server–specific routes. The webview retries them forever in the code-server scenario; consider gracefully degrading/quieting these when the backend is a plain opencode server.

### Where does it happen?

VS Code extension

### Version (if known)

1.19.0

_No response_

### Logs (optional)

```shell

```

VinciYan:
## Summary

Fixes #2996 — the OpenChamber sidebar chat view stays stuck on the `#initial-loading` screen on slow/remote networks (e.g., code-server behind a reverse proxy over WAN).

**Root cause**: the webview only leaves its loading screen once it receives a `connectionStatus=connected` postMessage, but the message is sent **exactly once** per status change in `_sendCachedState()` with no return-value check, no ack, and no retry. VS Code silently drops `postMessage` calls made before the webview's `acquireVsCodeApi` bridge is ready (slow multi-MB asset load on remote networks), so the single send can be lost forever and the loading overlay never disappears. The webview never pulls the status — it only ever receives it via extension-host postMessage.

**Fix**: re-send `connectionStatus` at staggered delays (500ms / 1.5s / 3.5s / 7s / 12s / 20s) while the cached status is `connected`, in all three providers:

- `ChatViewProvider` (sidebar chat)
- `AgentManagerPanelProvider` (agent manager)
- `SessionEditorPanelProvider` (session editor)

Resends stop automatically once the view/panel is disposed or replaced, and the message is idempotent, so duplicates are harmless. Hook points: ① after the webview is created; ② on `updateConnectionStatus('connected')` transitions.

## Test plan

- Locally verified on the affected remote network: loading screen disappears; repeated refreshes / reconnects / dragging the sidebar view no longer reproduce the stuck state.
- The official repro test (`chatview-connection-status-repro.test.js`, branch `reproduce/issue-2996`) demonstrates the bug: the single `connected` send is dropped and never re-sent. With this patch, the message is re-sent at staggered delays, so the webview receives it once its bridge becomes ready.
- `bun run type-check` passes.

## Changes

- `packages/vscode/src/ChatViewProvider.ts`
- `packages/vscode/src/AgentManagerPanelProvider.ts`
- `packages/vscode/src/SessionEditorPanelProvider.ts`
- `packages/vscode/package.json` (version bump 1.19.0 → 1.19.1 — revert if maintainers prefer to bump separately)


btriapitsyn:
Thanks — the diagnosis is exactly right: postMessage before the webview bridge is ready is silently dropped, and one blind send can strand the loading screen forever. The staggered bounded retries with the replaced-target guard are a good pragmatic answer that doesn't need an ack protocol (the payload is idempotent). Before merging I deduplicated the three near-identical provider methods into a shared webviewCachedStateRetry helper and restored the package.json trailing newline the branch dropped; behavior is unchanged. Merging.

PR #3000, opened 2026-08-19
fix(small-model): route anthropic provider through configured baseURL

mpeter:
## Intent

The `anthropic` branch of the small-model dispatch (`packages/web/server/lib/small-model/call.js`) hardcoded `https://api.anthropic.com/v1/messages` and never consulted `provider.anthropic.options.baseURL` — the override every other provider branch (`openai`, custom OpenAI-compatible, GitHub Copilot, etc.) already respects via `readProviderConfig`. Any setup that points the `anthropic` provider at a custom Anthropic-compatible endpoint had every small-model call for that provider (session-assist recap/suggestion, walkthrough generation) silently bypass the configured endpoint and hit the real Anthropic API with whatever credential the local config supplies, failing with `401 invalid x-api-key` on every call — no fallback, since `restrictToPreferredProvider` pins the call to `anthropic`.

After this change, `callAnthropic` accepts an optional `baseURL`, builds the request URL from it, and falls back to `https://api.anthropic.com` only when unset — matching the pattern already used one branch below it for the OpenAI-compatible dispatch.

## Non-goals

- No change to `readProviderConfig` itself; it already read and returned `baseURL` correctly for every provider — this PR only wires the existing value into the one branch that ignored it.
- No change to any other provider branch (`openai`, `google`, `github-copilot`, custom OpenAI-compatible) — they already honor their configured `baseURL`.
- No change to `restrictToPreferredProvider` fallback behavior, retry policy, or session-assist's idle-trigger logic.

## Affected surfaces

- `packages/web/server/lib/small-model/call.js`: the only file with behavior changes (`callAnthropic` + the `anthropic` dispatch branch).
- Consumers of `generateSmallModelText` with `providerID: 'anthropic'` and a configured `provider.anthropic.options.baseURL`: session-assist recap/suggestion generation and walkthrough generation, both server-side only. No client/UI, persisted format, route, or cross-workspace contract is touched — this is a local implementation fix inside one server module.
- Not affected: any deployment without a custom `anthropic` `baseURL` configured — those already resolved `baseURL` to `undefined` and get the same `https://api.anthropic.com` fallback as before.

## Repository guidance

|---|---|---|
| `openchamber-change-discipline` | Executable source change to a private server module (`packages/web/server/lib/small-model/call.js`) | Classified as local implementation risk (private helper behavior in one package); scope kept to the two call sites that needed the fix (`callAnthropic` signature, `anthropic` dispatch branch) with no unrelated refactor; validated at the owning-package level (focused tests + package type-check + targeted lint on the changed file) per the validation matrix for "Executable source" |
| `changelog-authoring` | Added a bullet to `CHANGELOG.md`'s `[Unreleased]` section | Wrote a user-facing bullet describing the observable symptom (recaps/suggestions failing with a custom endpoint configured) with no internal mechanics — no config key names, no raw endpoint URLs, no reference to the specific proxy setup used to discover the bug; matches the existing bullet style and area-prefix convention (`Session assist: ...`) |

## Validation

| Check | Result |
|---|---|
| `bun run --cwd packages/web type-check` (`tsc --noEmit`) | Passes, no output |
| `packages/web` small-model suite (`vitest`) | 65/65 passing, including 2 new tests: configured `baseURL` is used (and `api.anthropic.com` is not hit), and the `https://api.anthropic.com` fallback applies when `baseURL` is unset |
| `bunx oxlint packages/web/server/lib/small-model/call.js packages/web/server/lib/small-model/call.test.js` (anti-slop plugin) | All findings pre-existing (`no-runtime-typeof`, `no-conditional-empty-object-spread`, one `no-module-mocking` in the test file); none on the lines this PR changes (`callAnthropic` signature/body, the `anthropic` dispatch branch) |
| Manual: `curl -X POST http://<local-proxy>/v1/messages` with the same placeholder credential the buggy code path would have used | 200, valid Anthropic-shaped response — confirms the target endpoint is reachable and correctly shaped once the code actually addresses it |
| Not verified | End-to-end run against `api.anthropic.com` itself (the fallback path) beyond the automated fallback test — no production Anthropic credential was exercised manually, only the automated test coverage for that branch |

## Visual evidence

No user-visible UI change. This is a server-side request-routing fix inside `packages/web/server/lib/small-model/call.js`; the only observable effect is that background recap/suggestion generation succeeds instead of failing when a custom Anthropic `baseURL` is configured, which surfaces as existing session-assist UI populating instead of staying empty — no new UI states, styling, or interaction were added.

## Risks and failure behavior

- Failure mode if this regresses: identical to the pre-fix behavior (401s against a misrouted request) for setups with a custom `anthropic` `baseURL`; setups without one are unaffected either way since `baseURL` resolves to `undefined` and hits the same fallback as before.
- No persisted state, migration, or rollback path involved — this only changes which URL a single outbound HTTP request targets per call.
- No security concern introduced: the credential sent is the same `apiKey` already resolved for the `anthropic` provider; this only changes the destination host, and only when the user has explicitly configured one via `provider.anthropic.options.baseURL`.
- No cross-runtime impact: `packages/web/server` only; Electron, VS Code, and mobile all proxy through the same web server code path.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,7 @@ All notable changes to this project will be documented in this file.
 - Git: pull-request checks in Work status stay current as their status changes.
 - UI: the default dialog close button is easier to click or tap (thanks to @rockinrimmer).
 - Desktop/Windows: the close button now aligns correctly with the rest of the window chrome.
+- Session assist: recaps and suggested follow-ups now work when the Anthropic provider is configured to use a custom endpoint; they previously failed every time instead of using that configured connection.
 
 ## [1.19.0] - 2026-08-19
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file.
 ## [Unreleased]
 
 - Chat: if OpenCode restarts while a response is still running, the chat now stops with an interrupted state and a notification to continue instead of hanging silently (thanks to @sum117).
+- Session assist: recaps and suggested follow-ups now work when the Anthropic provider is configured to use a custom endpoint; they previously failed every time instead of using that configured connection.
 
 ## [1.19.0] - 2026-08-19
 
--- a/packages/web/server/lib/small-model/DOCUMENTATION.md
+++ b/packages/web/server/lib/small-model/DOCUMENTATION.md
@@ -103,7 +103,10 @@ other runtime API.
     `https://chatgpt.com/backend-api/codex/responses` with
     `ChatGPT-Account-Id`; expired tokens are refreshed against
     `auth.openai.com` (single-flight) and written back to `auth.json`.
-  - **Anthropic** (`type: api`): `/v1/messages` with `x-api-key`.
+  - **Anthropic** (`type: api`): `/messages` with `x-api-key`, against
+    `provider.anthropic.options.baseURL` when configured (used as-is, matching
+    `@ai-sdk/anthropic` — no `/v1` is inserted) or `https://api.anthropic.com/v1`
+    otherwise.
   - **Google** (`type: api`): `generateContent` with `x-goog-api-key`; Gemini 3
     uses `thinkingLevel` while older Flash models use `thinkingBudget: 0`.
   - Everything else: OpenAI-compatible `/chat/completions` against the
--- a/packages/web/server/lib/small-model/call.js
+++ b/packages/web/server/lib/small-model/call.js
@@ -341,7 +341,10 @@ const callMessages = async ({ url, headers, modelID, prompt, system, maxOutputTo
 };
 
 const callAnthropic = async ({ apiKey, baseURL, modelID, prompt, system, maxOutputTokens, responseSchema, timeoutMs, signal }) => callMessages({
-  url: `${(baseURL || 'https://api.anthropic.com').replace(/\/+$/, '')}/v1/messages`,
+  // Matches @ai-sdk/anthropic: baseURL is the full API prefix (commonly
+  // already ending in /v1), so it gets /messages appended as-is rather than
+  // having /v1/messages appended, which would double up a configured /v1.
+  url: `${(baseURL || 'https://api.anthropic.com/v1').replace(/\/+$/, '')}/messages`,
   headers: {
     'x-api-key': apiKey,
     'anthropic-version': '2023-06-01',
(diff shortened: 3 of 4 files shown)


btriapitsyn:
Thanks for fixing the Anthropic custom-endpoint path and for adding focused coverage. I found one endpoint-joining case that needs addressing before merge.

OpenCode passes `provider.anthropic.options.baseURL` directly to `@ai-sdk/anthropic`, where it is an API prefix and commonly already ends in `/v1`. With a configuration such as `https://proxy.example/v1`, the new construction in `packages/web/server/lib/small-model/call.js` requests `https://proxy.example/v1/v1/messages`, which will generally return 404. The current regression test covers only a bare-host value, so it does not catch this supported form.

Could you preserve the configured prefix and append `/messages`, while using `https://api.anthropic.com/v1` when no override is configured? If matching the current AI SDK behavior exactly, the configured bare official host (`https://api.anthropic.com`) should also normalize to its `/v1` prefix. A focused test with a configured URL ending in `/v1` would lock this down.

Once that URL-prefix case is handled, the rest of the change looks appropriately scoped.

mpeter:
Good catch, thanks — fixed. `callAnthropic` now treats `baseURL` as the full API prefix and appends `/messages` as-is instead of unconditionally inserting `/v1`, matching `@ai-sdk/anthropic`'s own behavior exactly (confirmed in `@ai-sdk/anthropic`'s source: it does `${withoutTrailingSlash(baseURL) ?? 'https://api.anthropic.com/v1'}/messages`, with no `/v1` normalization for a configured value). So:

- `baseURL` unset → `https://api.anthropic.com/v1/messages` (unchanged)
- `baseURL` ending in `/v1` (e.g. `https://proxy.example/v1`) → `https://proxy.example/v1/messages`, no more `/v1/v1`
- bare-host `baseURL` (e.g. `https://proxy.example`) → used as-is, `https://proxy.example/messages` — matches ai-sdk exactly rather than special-casing it, since ai-sdk doesn't normalize this case either\n\nAdded a regression test for the `/v1`-suffixed case and fixed the existing bare-host test's expectation to match the corrected (ai-sdk-accurate) behavior. Also rebased onto main to clear the merge conflict, and touched up `DOCUMENTATION.md`'s one-line description of the Anthropic branch since it didn't mention `baseURL` at all. 65 → 81 tests passing, type-check and lint clean on the changed lines.

PR #3026, opened 2026-08-20
fix(ui): keep failed dictation recovery controls on screen

### LINKED ISSUE #2533: [Bug] Android: long dictation transcript pushes insert/send controls off-screen

On the native Android app, a sufficiently long voice dictation transcript causes the composer to grow beyond the available viewport.

Steps to reproduce:

1. Open a chat in the Android app.
2. Tap the microphone to start dictation.
3. Continue speaking until the transcript contains enough text to fill the composer and most of the screen.
4. Try to scroll to the bottom of the transcript or access the dictation controls.

Actual behavior:

- The transcript/composer grows beyond the visible viewport.
- The transcript cannot be scrolled reliably to the bottom.
- The dictation footer containing Cancel, Insert, and Insert-and-send is pushed off-screen.
- There is no visible way to submit the dictated message, so the user becomes effectively stuck after a long recording.

Expected behavior:

- The dictation composer should be constrained to the available viewport above the Android keyboard and system insets.
- Long transcript content should scroll inside the composer.
- The dictation controls should remain pinned and accessible at the bottom of the composer regardless of transcript length.

This is especially disruptive because dictation encourages composing longer messages.

This appears distinct from:

- #2069, where Send taps or keyboard Enter fail to submit.
- #2129, where the Stop button is unavailable while an assistant response is active.

### Where does it happen?

Native Android application

### Version (if known)

Latest available Android build as of July 30, 2026.

The latest OpenChamber repository release is v1.17.1.

Available, but omitted because they contain project/session names and dictated text.

### Logs (optional)

No relevant logs observed. This appears to be a mobile layout/overflow issue.

### Likely fix direction

Cap the overlay height using the visual viewport, make the transcript region independently scrollable, and keep the action footer sticky.

karimodm:
## Intent

Fixes #2533.

Re-checking current `main` confirmed that the composer does not render a live transcript while recording or transcribing. The overflow path is a failed dictation with long salvage text. This change grows the composer only for that failed state, then applies the editor's existing line and viewport limits so the recovery actions stay reachable.

The viewport calculation excludes the dictation host from the surrounding layout height. Growing the host therefore cannot reduce its own limit through a resize-observer feedback cycle. Width changes remeasure wrapped salvage text.

## Non-goals

This PR does not change speech capture, transcription providers, normal recording or transcribing states, recovery button behavior, keyboard handling, or fullscreen composer behavior.

## Affected surfaces

`packages/ui` owns the shared composer used by web, desktop, hosted mobile, and Capacitor mobile. The reported failure is on native Android, but hosted mobile uses the same bounded layout path.

Desktop keeps its eight-line limit. Mobile keeps its sixteen-line and chat-viewport limits. VS Code does not expose server-backed dictation. No persisted data, API, route, setting, or external contract changes.

## Repository guidance

|---|---|---|
| `AGENTS.md`, `CONTRIBUTING.md`, and `openchamber-change-discipline` | This changes shared UI and mobile runtime behavior. | The diff stays in composer sizing, focused tests, and the owning documentation. It adds no dependency or persisted contract. |
| `theme-system` | The change affects component layout. | It reuses the existing editor and dictation controls. No colors, tokens, buttons, icons, or animations changed. |
| `packages/ui/src/components/chat/composer/DOCUMENTATION.md` | This document owns composer sizing and mobile behavior. | It records the failed-salvage limit, internal scrolling, and transient non-positive viewport fallback. |
| `packages/mobile/README.md` | The report affects the Capacitor Android shell. | The fix stays in the shared UI bundled by the native package and does not change native shell contracts. |

## Validation

| Check | Result |
|---|---|
| `bun test packages/ui/src/components/chat/composer/editor/__tests__/composerHeightLimit.test.ts` | Passed on final HEAD: 7 tests and 9 assertions. Covers failed salvage containment, host-growth invariance, short viewport reflow, tablet line cap, and unavailable viewport measurements. |
| `bun run type-check:ui` | Passed on final HEAD. |
| `bun run lint:ui` | Passed on final HEAD. |
| `bun run docs:validate` | Passed on final HEAD: 460 pages and 46 sidebar links. |
| `bunx oxlint <changed TypeScript files>` | Reports only existing anti-slop findings outside the added lines. No new-line finding remains. |
| `git diff --check upstream/main...HEAD` | Passed. |
| Native Android or final browser interaction | Not run. This Linux host has no Android toolchain, and an earlier browser validation caused memory pressure. The current final arithmetic and observer lifecycle are covered statically, not on hardware. |

## Visual evidence

No current screenshot is attached. The visible change requires a failed dictation with a long partial result, and the final branch was not run on Android hardware. The expected change is internal scrolling in the salvage area while Discard, Retry, and Insert remain in the composer.

## Risks and failure behavior

The sizing hooks run only while failed salvage text is present. Their observers disconnect when recovery ends or the component unmounts. Whitespace-only partial text does not activate the path.

The host budget subtracts `branch height - host height`, so applying the result does not feed the new host size back into the next calculation. If the viewport temporarily reports no positive room, the composer keeps its existing line cap until the next resize instead of collapsing typed text to zero height.

Reverting the commits restores the previous unbounded failed-salvage floor. No user data or transcription state is converted or deleted.

### DIFF
--- a/packages/ui/src/components/chat/ChatInput.tsx
+++ b/packages/ui/src/components/chat/ChatInput.tsx
@@ -1731,10 +1731,10 @@ const ChatInputComponent: React.FC<ChatInputProps> = ({
         }
     }, [agents, currentAgentName, currentSessionId, setAgent, saveSessionAgentSelection]);
 
-    // Height the dictation transcript needs (null when idle). Its overlay sits
+    // Height the failed-dictation salvage text needs. Its overlay sits
     // absolutely over the composer, so the composer must be able to grow for
     // it. Apply the editor's line and screen bounds before using that height as
-    // a floor, otherwise a long transcript can push the action row off-screen.
+    // a floor, otherwise long salvage text can push the action row off-screen.
     const dictationHeightHostRef = React.useRef<HTMLDivElement | null>(null);
     const [dictationContentHeight, setDictationContentHeight] = React.useState<number | null>(null);
     const handleDictationContentHeightChange = React.useCallback((height: number | null) => {
--- a/packages/ui/src/components/chat/ChatInput.tsx
+++ b/packages/ui/src/components/chat/ChatInput.tsx
@@ -104,6 +104,7 @@ import {
     type ComposerChange,
     type ComposerEditorHandle,
 } from './composer/editor/ComposerEditor';
+import { useComposerHeightLimit } from './composer/editor/useComposerHeightLimit';
 import { createComposerEditorViewStore } from './composer/editor/viewStore';
 import { composerAutoCorrect } from './composer/editor/autocorrect';
 import {
@@ -1732,12 +1733,21 @@ const ChatInputComponent: React.FC<ChatInputProps> = ({
 
     // Height the dictation transcript needs (null when idle). Its overlay sits
     // absolutely over the composer, so the composer must be able to grow for
-    // it. The editor sizes itself to its own content; this is the one external
-    // constraint, applied as a floor on the editor's container.
+    // it. Apply the editor's line and screen bounds before using that height as
+    // a floor, otherwise a long transcript can push the action row off-screen.
+    const dictationHeightHostRef = React.useRef<HTMLDivElement | null>(null);
     const [dictationContentHeight, setDictationContentHeight] = React.useState<number | null>(null);
     const handleDictationContentHeightChange = React.useCallback((height: number | null) => {
         setDictationContentHeight((prev) => (prev === height ? prev : height));
     }, []);
+    const dictationHeightLimit = useComposerHeightLimit({
+        active: dictationContentHeight !== null,
+        disabled: isComposerExpanded,
+        hostRef: dictationHeightHostRef,
+        maxLines: isMobile ? MAX_MOBILE_COMPOSER_LINES : MAX_VISIBLE_COMPOSER_LINES,
+        boundSelector: isMobile ? '[data-composer-bound]' : undefined,
+        boundGapPx: isMobile ? MOBILE_COMPOSER_BOUND_GAP_PX : 0,
+    });
 
     const updateAutocompleteState = React.useCallback((
         value: string,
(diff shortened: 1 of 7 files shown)


btriapitsyn:
Hold on before merge — the premise needs a re-check against current behavior: the composer no longer renders a live transcript while dictation is recording; the only transcript rendered is the salvage text of a *failed* dictation (see the header comment in `ComposerDictation.tsx`). If the overflow only reproduces for long salvage text after a failure, the fix may still be worth having, but the PR description and tests should target that case explicitly. Please re-verify on current main and update the PR to match what actually reproduces.

karimodm:
Re-verified against current `main`: there is no live transcript during recording. I scoped the fix to non-empty salvage text after `status === "failed"`, updated the sizing regressions and documentation around that case, removed the stale live-transcript claims from the PR body, and force-pushed `86a534d86`.

karimodm:
Re-verified against current `main`: recording and transcribing do not render live transcript text. The overflow case is long salvage text after a failed dictation.

The branch is rebased and now scopes height reporting to that state only. The follow-up also removes a resize-observer feedback path by excluding the growing host from its own viewport budget, preserves the editor's non-zero fallback, and ignores whitespace-only salvage text.

Final HEAD `b45db2694` passes the focused 7-test sizing suite, UI type-check, UI lint, docs validation, and diff checks. I updated the title and handoff to describe the failed-salvage case and the lack of final Android hardware validation.

karimodm:
Latest bot follow-up:

- The current PR body does contain `## Repository guidance`; the automation reviewed an earlier handoff snapshot.
- I attempted the requested current-HEAD narrow-viewport capture from a fresh production build with one temporary server and one browser context. Chromium failed during navigation with `ERR_INSUFFICIENT_RESOURCES`, before the app rendered. I stopped the browser/server and removed the build immediately. No screenshot was produced, so I am not attaching or claiming visual evidence.
- The bot found no code issue and confirmed the failed-salvage scope, cap arithmetic, feedback prevention, and observer cleanup. The remaining `review:needs-evidence` label requires a browser or device capture on a machine with enough memory.

The full check also hit the existing auth, Markdown, and shortcut failures plus the unrelated intermittent `event-pipeline.test.ts` timing failure. This branch does not change those paths.

Tobias-Conrad:
Does this fix cover the **iOS** Capacitor path as well?

A new iOS report (#3532) shows the same failed-salvage overflow on native iOS: long transcript grows past the viewport, recovery row off-screen, no scroll, caret not placeable, dictated text lost. Since `packages/ui` is the shared composer for web/desktop/hosted mobile/Capacitor mobile, I would expect the failed-salvage containment here to apply — but the iOS viewport path differs (`visualViewport` + `-webkit-fill-available`, cf. #3201 / #3202), so it may still overflow there. Worth confirming against #3532 before this lands.

Tobias-Conrad:
Your fix is now rebased onto current `main` as **#3533** (*fix(ui): keep failed-dictation salvage controls on screen*).

This branch is `mergeable: false / dirty`, and the surrounding composer moved since (floating composer, etc.), so I ported the patch: `heightLimit.ts` + `useComposerHeightLimit.ts` unchanged in intent, `ComposerEditor.tsx` routed through `getComposerHeightLimit`, `ChatInput.tsx` clamping the salvage `minHeight`, and `ComposerDictation.tsx` reporting only salvage text and re-reporting on resize. Full credit in the PR body.

Validation on the rebased branch: 7/7 `composerHeightLimit` tests, 380/380 composer+dictation tests, `type-check` and `lint` pass. Feel free to close this PR in favour of #3533 or point #3533 at your branch if you prefer to keep authorship.

btriapitsyn (approved):
Thanks, Andrea, for the original fix, and Tobias for the follow-up. I updated it for current main, finished the scroll handling, and verified long-transcript recovery in the browser. The combined fix is ready to land.

PR #3059, opened 2026-08-21
fix(ui): keep owning project when opening plans from a worktree session

joeyfromspace:
## Intent

Opening a saved plan from the Project knowledge panel while a session runs from a temporary worktree showed a blank editor with a "Save failed" error. The panel lists plans correctly (it resolves the project from `activeProjectId`), but `PlanView` re-derives the project from the session directory by path containment. A worktree lives under `~/.local/share/opencode/worktree/...`, outside every registered project path, so that lookup returns null and the viewer clears its content.

The fix threads the owning project through to `PlanView` everywhere a plan is opened from a project-scoped list, instead of guessing it from the directory. The directory-based lookup stays as the fallback for callers that never had a project in hand.

## Non-goals

- Not changing how the directory-based project lookup works for file-based plan tabs (`openContextPlan`) or any other surface.
- Not changing project resolution for notes, todos, or memory.
- Not migrating existing persisted plan tabs; tabs without a stored project keep the old fallback behavior.

## Affected surfaces

- `packages/ui` only. Shared UI consumed by web, desktop, VS Code, and mobile, so all of them take the same path.
- `PlanView` gains an optional `projectRef` prop.
- `useUIStore` context-panel plan tabs gain a persisted `projectRef` field (nullable; old tabs hydrate to `null`).
- The `onOpenPlan` payload across `PlansSection` to `ProjectNotesTodoPanel` to `ProjectContextPanel` / `MobileWorkspaceDrawer` now carries the owning project.
- Desktop note panel, mobile fullscreen plan surface, and the context-panel `plan` tab all pass the project through.

## Repository guidance

|---|---|---|
| `openchamber-change-discipline` | Adds a persisted field and a shared prop, so I classified the change as local UI behavior plus a persisted-contract addition | Kept the diff to one mechanism (carry the project), added a nullable default so existing stored tabs are valid without a migration, and validated at the package level; no new exports or generated assets |
| `ui-api-decoupling` | `PlanView` resolves project-scoped data reachable through `useProjectContextStore` and a server route | Only flowed an already-trusted `ProjectRef` from existing store state; no new routes, runtime branches, or SDK calls |
| `communication-style` | Loaded at task start per AGENTS.md | Applied to prose and this description |
| `project-context/DOCUMENTATION.md` | Nearest module doc for the panel | Updated the "Plans open in place" section |

## Validation

| Check | Result |
|---|---|
| `bun run type-check` (all 5 workspaces) | pass, exit 0 |
| `bun run lint:ui` (eslint) | clean, no warnings |
| `bunx oxlint <changed files>` | no findings in changed lines; remaining findings are in pre-existing migration/portal code, not touched here |
| `bun test packages/ui/src/stores/useUIStore.contextPanel.test.ts` | 19 pass, 0 fail (includes 3 new tests for `projectRef` on plan tabs) |
| `bun run --cwd packages/ui test` | 273/274 files pass; the 1 failing file (`sync/__tests__/event-pipeline.test.js`, 10 failures) fails identically on a clean checkout with this branch's changes stashed, so it is pre-existing and unrelated |

## Visual evidence

Reproduced against a local dev build with an isolated data dir: a `foo-foo` git project with a saved plan, and a worktree session created from it (worktree lives under `~/.local/share/opencode/worktree/...`, outside the project path). Same steps, before and after.

Before (main): clicking the plan opens an empty editor with "Save failed":

before

After (this branch): the same plan loads its markdown:

after

## Risks and failure behavior

- Old persisted plan tabs hydrate `projectRef` to `null` and keep the existing directory-fallback behavior, so no migration or rollback path is needed.
- If an owning project has been deleted, a plan tab opened with a stale `projectRef` now surfaces the store's "plan not found" error instead of silently blanking, which is the intended failure mode.
- The project-explicit path changes where "improve"/"implement" sends route (to the project path rather than a worktree path), matching the project-scoped ownership of the plan.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file.
 ## [Unreleased]
 
 - Chat: if OpenCode restarts while a response is still running, the chat now stops with an interrupted state and a notification to continue instead of hanging silently (thanks to @sum117).
+- Project knowledge: plans opened from the panel now load while a session runs from a temporary worktree, instead of showing a blank editor.
 
 ## [1.19.0] - 2026-08-19
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -85,6 +85,7 @@ All notable changes to this project will be documented in this file.
 - Git: pull-request checks in Work status stay current as their status changes.
 - UI: the default dialog close button is easier to click or tap (thanks to @rockinrimmer).
 - Desktop/Windows: the close button now aligns correctly with the rest of the window chrome.
+- Project knowledge: plans opened from the panel now load while a session runs from a temporary worktree, instead of showing a blank editor.
 
 ## [1.19.0] - 2026-08-19
 
--- a/packages/ui/src/apps/MobileApp.tsx
+++ b/packages/ui/src/apps/MobileApp.tsx
@@ -24,6 +24,7 @@ import { readTabletLayout, useOrientation, useTabletLayout } from '@/lib/device'
 import { useHardwareKeyboard } from '@/lib/hardwareKeyboard';
 import { useI18n } from '@/lib/i18n';
 import { runtimeFetch } from '@/lib/runtime-fetch';
+import type { ProjectRef } from '@/lib/projectContextApi';
 import { getRuntimeApiBaseUrl, getRuntimeKey, subscribeRuntimeEndpointChanged, switchRuntimeEndpoint } from '@/lib/runtime-switch';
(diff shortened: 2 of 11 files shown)


btriapitsyn:
Thanks for this — the diagnosis was exactly right, and the fix was well built: the projectRef threading, the nullable persisted field, and the reproduction evidence made it easy to verify. The same bug has since been fixed on main by carrying the owning project with every saved-plan open (`fa5446593`, which reworked `PlanView` around a `savedProjectPlan` target), so the code this PR patches no longer exists in that shape. Closing as superseded — the behavior you reported is resolved on current main.

PR #3090, opened 2026-08-23
fix(ui): make timeline dialog fit small screens

gaojunran:
## Intent

On phones the `/timeline` dialog squeezed the scrollable message list to a couple of rows: the header, search box, and the fixed Actions/help block consumed almost all of the 70vh dialog height, and the description text could end up visually underneath the search box.

The dialog no longer scrolls as a whole. The header and search stay fixed (`shrink-0`), the message list becomes the only scrollable region and takes the remaining height (`min-h-0 flex-1`), the dialog is taller on small screens (85dvh), the Actions/help footer collapses to a single action row on phones, the description is hidden on phones only, and message timestamps no longer wrap onto a second line.

## Non-goals

- No new entry point or button for the timeline (mobile placement is still an open discussion).
- No change to timeline behavior: search, keyboard navigation, revert/fork, load-earlier, and turn offset actions all keep their existing handlers.
- No i18n copy changes. Locale keys were restored to their original state after a transient removal.

## Affected surfaces

- `packages/ui/src/components/chat/TimelineDialog.tsx` only.
- The dialog is shared by desktop web, PWA, and the Capacitor mobile apps via `packages/ui`. Desktop renders unchanged except that an extremely short window now lets the list absorb the overflow instead of scrolling the whole popup.
- No persisted data, routes, APIs, or cross-runtime contracts change.

IMG_3049

## Repository guidance

|---|---|---|
| `.agents/skills/theme-system` tokens and examples | Tailwind classes must use the design-system tokens | Used existing tokens (`text-muted-foreground`, `bg-muted/30`, `border-border/60`) and utilities (`shrink-0`, `min-h-0`, `whitespace-nowrap`) |
| `packages/ui/src/components/ui/dialog.tsx` base classes | `cn` uses tailwind-merge; later classes must win over base `overflow-y-auto` | `overflow-y-visible` on `DialogContent` overrides the base dialog scroll |
| `.agents/skills/locale-ui-patterns` | i18n keys must stay in parity across all locales | No keys added; removed keys were restored via `git checkout`; `messages.test.ts` parity passes |

## Validation

| Check | Result |
|---|---|
| `bun run --cwd packages/ui type-check` | Passed, zero errors |
| `bun test src/lib/i18n/messages.test.ts` (packages/ui) | Passed, 2 pass / 0 fail (44 expects) |
| `bun run --cwd packages/ui lint` | Passed, zero warnings |
| Headless Chromium geometry assertions (designer) | 390x844: list 458px (was ~110px); 390x430 keyboard-open: list 186px, dialog fully in viewport, description not overlapped; 1440x900 desktop: description visible, list 358px |
| Manual on-device check by author | Phone: list scrolls, footer is a single action row, description absent; desktop: description present |

Not verified in automation: pixel-perfect rendering across every locale's timestamp format and the iOS PWA shell specifically (validated via shared `packages/ui` geometry, not the Capacitor wrapper).

## Visual evidence

Before (phone): list area ~2 rows; footer box with Actions title + help text dominated the bottom; description could sit under the search field.
After (phone): full-height list, 28px single-row footer with Previous turn / Latest, description hidden on mobile, timestamps on one line.
Desktop: unchanged footer box and description preserved.
(Author verified on device; headless Chromium geometry assertions covered the three viewports in the table above.)

## Risks and failure behavior

- `max-md:max-h-[85dvh]` uses dynamic viewport height, which shrinks when the mobile browser chrome collapses and the soft keyboard opens; the dialog stays within the visible area rather than under it. Worst case on very small screens the list still retains ~4 rows.
- `overflow-y-visible` on the popup relies on the list being the only shrinking element; all other children are `shrink-0`, so the popup cannot overrun the rounded border in normal use.
- If a future maintainer adds a tall fixed-height child to the dialog, it would again push the list; the layout intent is documented in the class structure (list is the sole `flex-1 min-h-0` region).

### DIFF
--- a/packages/ui/src/components/chat/TimelineDialog.tsx
+++ b/packages/ui/src/components/chat/TimelineDialog.tsx
@@ -251,7 +251,7 @@ export const TimelineDialog: React.FC<TimelineDialogProps> = ({
 
     return (
         <Dialog open={open} onOpenChange={onOpenChange}>
-            <DialogContent className="max-w-2xl max-h-[70vh] max-md:max-h-[85dvh] flex flex-col overflow-y-visible">
+            <DialogContent className="max-w-2xl max-h-[70vh] max-md:max-h-[85dvh] flex flex-col overflow-y-auto">
                 <DialogHeader className="shrink-0">
                     <DialogTitle className="flex items-center gap-2">
                         <Icon name="time" className="h-5 w-5" />
--- a/packages/ui/src/components/chat/TimelineDialog.tsx
+++ b/packages/ui/src/components/chat/TimelineDialog.tsx
@@ -223,20 +223,48 @@ export const TimelineDialog: React.FC<TimelineDialogProps> = ({
 
     if (!currentSessionId) return null;
 
+    const turnActions = (
+        <>
+            <button
+                type="button"
+                className="text-[11px] uppercase tracking-wide text-muted-foreground/90 hover:text-foreground"
+                onClick={() => {
+                    void onScrollByTurnOffset?.(-1);
+                    onOpenChange(false);
+                }}
+            >
+                {t('chat.timeline.actions.previousTurn')}
+            </button>
+            <span className="text-muted-foreground/50">/</span>
+            <button
+                type="button"
+                className="text-[11px] uppercase tracking-wide text-muted-foreground/90 hover:text-foreground"
+                onClick={() => {
+                    onResumeToLatest?.();
+                    onOpenChange(false);
+                }}
+            >
+                {t('chat.timeline.actions.latest')}
+            </button>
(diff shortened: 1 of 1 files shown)


btriapitsyn (approved):

gaojunran:
@btriapitsyn I think maybe we should add a button on phone view for /timeline. It is very common used. There are many existing buttons now 🤔. Do you think we should do and where do you think we should put? (or maybe long press the session title or the scroll to bottom icon)

btriapitsyn:
@gaojunran I think having an /timeline command will work for now, later on it worth thinking about probably some sort of separate menu when there will be more such things

PR #3092, opened 2026-08-23
feat(walkthrough): let users pick the base branch

gaojunran:
## Intent

The walkthrough panel's branch source derives its base branch automatically via `deriveBaseBranch`. When the derivation fails (e.g. the current branch sits on the default branch, or no parent can be resolved) the "This branch" entry silently disappears from the source menu, and there is no way to compare against a base of the user's choosing. Branch diff already supports a user-picked base through `useGitBaseBranchStore`; this PR gives walkthrough the same override capability and a base picker UI.

## Non-goals

- Changing the derivation algorithm (`deriveBaseBranch`) for either surface.
- Touching the server walkthrough pipeline: the client only changes the `baseRef` it sends, so generation/caching is unaffected.
- Persisting anything new: the override reuses the branch-diff store, so both surfaces share one per-branch base choice.

## Affected surfaces

- `packages/ui` only. WalkthroughView source menu + panel body; 11 locale files gain 3 keys each. No persisted contract or server API changes.
- The branch diff surface is unaffected in behavior but shares `useGitBaseBranchStore`, so a base picked in either surface is honored by the other.

## Repository guidance

|---|---|---|
| AGENTS.md — i18n parity is enforced by messages.test.ts | Every locale must stay key-aligned with English | All 3 new keys added to all 11 locales, parity test passes |
| `.agents/skills/theme-system` | UI components/colors must use theme tokens | Picker uses theme variables (`--surface-elevated`, `--interactive-focus-ring`, etc.) and existing `ScrollableOverlay`/`Button` |
| `.agents/skills/locale-ui-patterns` | User-visible copy conventions | New strings follow existing `walkthrough.scope.*` naming; search placeholder reuses `gitView.branch.searchPlaceholder` |

## Validation

| Check | Result |
|---|---|
| `bun test src/lib/i18n/messages.test.ts` | 2 pass (parity all locales) |
| `bun test` walkthrough stores | 16 pass |
| `bunx oxlint` changed files | No new findings (pre-existing anti-slop backlog unchanged, line-shifted only) |
| Manual UI walkthrough | Not yet done — needs a running backend + a git repo with branches; pending reviewer/author device check |

## Visual evidence

Pending — draft. Will attach before/after screenshots of the source menu ("This branch" with a derived base, picker state, and "Change base branch..." entry) once validated locally.

## Risks and failure behavior

- An override written by branch diff for a remote-tracking base (`origin/main`) is normalized by `normalizeBaseCandidate` so it still passes `hasResolvableBaseBranch`. 
- Picking a base that no longer exists (deleted branch) leaves the override in place but `branchSource` falls back to null and the picker reappears — no crash path.
- No data-loss or perf concern: picker is local-state, override write is one zustand update.


gaojunran:
Closed due to https://github.com/openchamber/openchamber/pull/3148#issuecomment-5598495455

PR #3106, opened 2026-08-24
feat(ui): count subagent spend in the session cost readout

igorvelho:
## Intent

A session that delegates work to subagents reported only its own spend. In the work status panel the Session section showed `$1.7691` while five completed subagents sat directly underneath it with no cost shown at all; the `$1.7770` they had spent between them was not reported anywhere in the app. The headline figure was roughly half the real cost of the session, and the gap grows with how much work a session delegates.

This surfaces the full cost. The Context row now reports the rollup (own spend plus every descendant subagent, recursively), a line under the meter splits that total back into its two parts, and each subagent row carries its own subtree cost.

The split line exists because the total alone created a new problem: `$3.5461` next to a row labelled Context, with the context overview still showing `$1.7691` on click through, reads as two surfaces disagreeing. `Session $1.7691 · Subagents $1.777` reconciles them in place, and the first figure is exactly what the overview shows.

`useSubagentCostRollup` owns the walk. It reads the flat live-session list the Subagents section already subscribes to, indexes it by `parentID` once, and sums each subtree. `ownCost` is derived by subtracting the subagent share from the total rather than read back off the session, so the two parts always add up to the number displayed above them.

## Non-goals

- Not changing how `Session.cost` itself is produced or synced; it is read-only input here.
- Not changing the `formatCost` rule. It is moved from `WorkStatusPrimaryGroup.tsx` into `subagentCost.ts` unchanged so both consumers share one implementation. Its existing `trimZeros` behaviour renders `1.7770` as `$1.777`, which is visible in the screenshots below as uneven decimal places next to `$1.7691`. That predates this change and is left alone rather than bundled in.
- Not touching the desktop header or mini chat readouts. Both render `ContextUsageDisplay` without a `cost` prop, so they are unchanged.
- Not adding cost to archived or historical subagents. Scope matches the existing Subagents section, which lists live sessions only.
- No new section, row type, or panel primitive.

## Affected surfaces

- `packages/ui` (shared UI). New `work-status/subagentCost.ts` (index, subtree sum, relocated `formatCost`) and `work-status/useSubagentCostRollup.ts` (pure `computeRollup` plus the hook). `WorkStatusPrimaryGroup.tsx` shows the rollup and the split caption. `WorkStatusSubagentsSection.tsx` shows each child's subtree cost. `work-status/DOCUMENTATION.md` records both data flows.
- `VSCodeLayout.tsx` previously passed `currentSession?.cost` to `ContextUsageDisplay`. Left alone it would have shown the session's own cost in the VS Code header while the panel showed the rollup, so it now reads the same hook. This also removes the file's only `useSession` call.
- i18n: one key, `chat.workStatus.cost.breakdown`, added to all 11 dictionaries with real translations.
- Changelogs: `[Unreleased]` entries in `CHANGELOG.md` and `packages/vscode/CHANGELOG.md`, since the VS Code header figure changes.
- No server, sync, persisted-format, route, or API change. No new store subscription: the hook reads `useAllLiveSessions()`, which `WorkStatusSubagentsSection` already held.
- Mobile and web consume the shared UI and pick the change up through the same panel.

## Repository guidance

|---|---|---|
| `openchamber-change-discipline` | Shared UI change consumed by web, desktop, VS Code and mobile | Traced every `session.cost` consumer before editing, which is how the VS Code header divergence was found and fixed rather than shipped. The one new boundary owns a real invariant (the split must sum to the total) and has two consumers, not one. Cycle guard on the walk rather than trusting `parentID` to be acyclic. Read-only change: no writes, no persisted state |
| `locale-ui-patterns` | Adds one user-facing string | Key added to `en.ts` first, then to all 10 other dictionaries with real translations, no English placeholders. Semantic key naming matches the neighbouring `chat.workStatus.*` keys. Rendered through `useI18n()` inside the component with `{session}` / `{subagents}` interpolation. Parity test run |
| `changelog-authoring` | User-visible change on shared UI and VS Code | `[Unreleased]` bullets added to both changelogs, written as the symptom a user notices, no new release header |
| `packages/ui/src/components/chat/work-status/DOCUMENTATION.md` | Owning documentation for this panel; its data-source table is the contract for where each readout comes from | Both rows updated to name the rollup and `perChildCost`. The document's rule that a cost-only row is not a context reading is respected: the total stays a passenger on the Context row, and the split is a caption explaining the figure above it rather than a labelled row reporting a new one |
| `ui-api-decoupling` | Shared UI reading session data | No new API surface or fetch. The rollup is computed from live session state already in the store |

## Validation

| Check | Result |
|---|---|
| `bun test packages/ui/src/components/chat/work-status/ packages/ui/src/lib/i18n/messages.test.ts` | 66 pass, 0 fail. Covers the flat sum, nested descendants, a cycle, zero and missing cost, unknown ids, the own/subagent split, and locale key parity |
| `bun run --cwd packages/ui test` (full isolated suite) | 290/291 test files passed. The one failure is `src/sync/__tests__/event-pipeline.test.js`, which fails identically on unmodified `main` at `d1fc27c` in this same checkout; unrelated to this diff |
| `bun run --cwd packages/ui lint` | Pass, no findings |
| `bunx knip@5.80.0 --include exports,types` | No findings for the added exports |
| Manual, running build | Web build of this branch served from a separate instance and data directory, exercised on a real session with five completed subagents. Screenshots below |
| Not verified | VS Code extension host. The header change is a one-line swap to the same hook the panel uses and is covered by type-check, but the readout was not observed running in VS Code. Mobile and narrow widths were not captured either: the work status panel is hidden on mobile and on VS Code by `useWorkStatusVisibility`, and below its width threshold, so it has no narrow state to photograph |

## Visual evidence

Same session, same five subagents, captured from a running web build.

**Before** (`main`): the Session section reports `$1.7691`, and the five subagents underneath show status only. Their combined `$1.7770` appears nowhere.

before: context row shows the session cost only, subagent rows show no cost

**After** (dark): the Context row reports `$3.5461`, the caption under the meter reads `Session $1.7691 · Subagents $1.777`, and each subagent carries its own cost.

after, dark theme: context row shows the rollup total with the split caption beneath, and each subagent row shows its cost

**After** (light): same session in the light theme. The caption uses `text-muted-foreground` and the existing type scale, so it follows the theme with no new colour.

after, light theme: same panel rendering correctly on a light background

The figures reconcile: `1.7691 + 1.7770 = 3.5461`, and the per-subagent costs (`0.4305 + 0.4326 + 0.0519 + 0.4318 + 0.4302`) sum to the subagent share.

## Risks and failure behavior

- A session with no subagents renders exactly as before. The caption is gated on `subagentCount > 0 && subagentCost > 0`, and with no children the rollup equals `Session.cost`, so the Context row shows the same figure it showed previously.
- Missing or non-numeric `cost` counts as zero, matching the existing `typeof session?.cost === 'number'` guard this code replaced. A subagent whose cost has not arrived yet contributes nothing rather than breaking the sum.
- The recursive walk is cycle-guarded by a visited set. `parentID` should describe a tree; if it ever does not, the walk terminates and no session is counted twice. Unit-tested.
- Deriving `ownCost` by subtraction means the caption always agrees with the total on screen, including in that trimmed case.
- Cost for a subagent whose session is not in the live list is not counted. The figure can therefore understate rather than overstate, which is the safer direction for a number a reader may not scrutinise.
- Read-only and render-only. No writes, no persisted data, no migration, no new dependency, no network call. The worst outcome is a wrong number on screen, corrected on the next sync.
- Performance: one pass to index plus one walk per render of a memoised hook, over sessions already in memory. No new subscription or request.


patrick-motard (approved):

packages/ui/src/components/chat/work-status/subagentCost.test.ts
```diff
@@ -0,0 +1,72 @@
...
+    const childB = makeSession('b', 3, 'root');
+    const index = buildChildrenIndex([root, childA, childB]);
+    expect(index.get('root')).toEqual([childA, childB]);
+  });
+});
+
+describe('formatCost', () => {
+  test('prefixes with $ and trims trailing zeros', () => {
+    expect(formatCost(1.5)).toBe('$1.5');
+    expect(formatCost(0.0001)).toBe('$0.0001');
```
patrick-motard, line 22: I do wonder, especially after reviewing the screenshots, if it would be preferable visually to display cost accurate to the penny. We could roll up to the nearest cent. We tend to think of cost in terms of cents being the smallest unit. Anything smaller than that feels like an implementation detail of the inference provider.

I would prefer to see $1.78 instead of $1.77631 for all representations of currency in the UI.

It has an added benefit of freeing up precious space in an already small UI surface where there is limited space to convey information about subagents to the user.

michelwin:
Openchamber doesn't currently show the cost of sub-agent calls, so that spend is invisible today. Would be nice to have this PR approved to fix this issue.

igorvelho:
Hi @btriapitsyn, I've updated the pr and fixed the merge conflicts. As for the feature itself, this has been a big request from my team at work, as we run on a tight budget. Please let me know if there are any more changes required. Thanks

patrick-motard:
🥳

PR #3112, opened 2026-08-24
fix(git): batch large selected path operations

### LINKED ISSUE #3099: [Bug] git 暂存778个文件报错

Error: spawn ENAMETOOLONG

### Where does it happen?

Desktop (macOS)

### Version (if known)

_No response_

_No response_

### Logs (optional)

```shell

```

bashrusakh:
## Summary

Fixes #3099. Large selected Git file sets are batched without exceeding command-line limits. Oversized generic selections and explicit `stageFiles` commits use a private temporary index so selected staged-only, deleted, and untracked content commits correctly while unrelated staged, partially staged, unmerged, intent-to-add, and newline-path entries remain unchanged.

## Scope

- Keeps small selections on their existing single-command path.
- Preserves explicit `stageFiles` semantics and existing partial-batch error reporting.
- Uses NUL-delimited literal path handling for filename safety.
- Keeps matching Web/Desktop and VS Code service behavior.
- Changes no Git UI selection model, API contract, dependency, or data model.

## Validation

- Web Git service, routes, batching, and long-path suites — 103 passed (vitest) and 10/10 under bun test.
- VS Code batching suite — 10/10 passed; full VS Code suite passed 26/26 test files.
- Scoped Web and VS Code type-check and lint — passed.
- `bun run docs:validate` — 460 pages and 46 sidebar links passed.
- `git diff --check` and changed-JS syntax checks — passed.
- OCR review of the final explicit-commit follow-up — 0 findings. The full-range OCR attempt hit an OCR internal line-range error; targeted final-commit review completed successfully.
- Final Greptile review — passed, confidence 5/5.
- Review follow-up (commits `8cbdd4b68`, `c89537daf`, `3efc4dbdc`): all 7 review points addressed — env-safe temporary index via execFile, staged-rename pairs, MERGE_HEAD refusal (before staging), temp-index-exclusive fallback guard, VS Code parity (stderr surfacing, status filtering, index mutation queue), and new regression coverage (rename, merge-in-progress, unborn-HEAD, editor-env, copy records).
- OCR review (open-code-review v1.10.0) over the follow-up commits: 8 findings across two passes, all fixed; final pass on `3efc4dbdc` — 0 findings.

## Manual verification and risk

Real-Git coverage verifies staged-only selected files, explicit `stageFiles`, unrelated partial staging, unmerged and intent-to-add entries, and newline filenames. Git still applies its native safeguards to partial commits during an active merge.

## CI note

The only failing GitHub `checks` test is inherited from `main` in `addSelectionToChat.test.ts`, outside this diff. Its isolated correction is green in #3124; label and final Greptile checks pass here.


btriapitsyn (changes requested):
Thorough work — the private temporary index is the right tool for this, the batch-size math is conservative (UTF-8 byte budget, NUL-delimited pathspec files, `--literal-pathspecs`), and the web test suite is the best kind we get: real repos, real staged hunks, byte-exact `ls-files --stage` assertions. The design is sound. The current implementation isn't mergeable yet, though — we ran your tests against real git in a normal shell environment and reproduced a blocker, plus found two silent wrong-commit paths. Everything below needs to land before merge; happy to re-review quickly when it does.

**1. Blocker (reproduced): the web temp-index flow fails on ordinary machines.** `temporaryIndexGit.env({ ...(await buildGitEnv()), GIT_INDEX_FILE })` spreads the whole `process.env`, and simple-git's `blockUnsafeOperationsPlugin` inspects env set via `.env()` — so any server whose environment has `EDITOR`, `GIT_EDITOR`, `PAGER`, `GIT_PAGER`, `SSH_ASKPASS`, or `GIT_SSH_COMMAND` gets `GitPluginError: Use of "EDITOR" is not permitted…` on every `stageFiles` commit and every oversized selected commit. On a machine with `EDITOR`/`PAGER` set, 3 of your 6 web tests fail with exactly that error; with those vars stripped, 6/6 pass — the tests are good, CI's env is just sterile. `getWorkingTreePathsDifferentFromHead` has the same `.env()` problem, so the non-stageFiles oversized path fails at the pre-check too. Fix: run the temp-index commands through the `execFile` helper you already added (`runGitCommandWithInputOrThrow`) with per-spawn env, the way the VS Code side does.

**2. Staged renames commit wrong content.** A staged rename shows up in status keyed by the new path. If only the new path is selected, the temp index gets the new file but never drops the old one — the commit resurrects the deleted path. The old `stageFiles` behavior got this right. Either expand rename pairs from `status --porcelain -z` or prove (with a test) that the selection contract always includes both halves.

**3. Merge-in-progress produces a bogus merge commit.** `git commit` reads `MERGE_HEAD` regardless of `GIT_INDEX_FILE`. With unrelated unresolved conflicts, the temp index is clean, so a selected commit succeeds, creates a two-parent merge commit taking HEAD's version of every conflicted file, and destroys the merge state — while the real index still holds stage-1/2/3 entries. The old path refused with "you have unmerged files". The unmerged test fabricates conflict entries without `MERGE_HEAD`, so it never hits this. Both implementations need a guard: refuse the temp-index commit when `MERGE_HEAD` exists or `ls-files --unmerged` is non-empty.

**4. The web pathspec-error fallback can full-index-commit.** The `catch` that falls back to `git.commit(message)` on "pathspec … did not match" wraps the temp-index path too; if a temp-index failure ever matches that message, the fallback commits the entire real index. Gate it on the non-temp-index path.

**5. The VS Code implementation drifted from web.** It swallows git's stderr on temp-index failure (`success: false` with empty strings), commits unfiltered selections where web filters against status and throws a clear error, has no pathspec fallback, and lacks the per-directory mutation queue web uses (pre-existing, but the multi-step temp-index flow makes the race window material). Please reconcile these or document each divergence as intentional.

**6. The VS Code tests mock the thing they test.** `mock.module('child_process')` with a spawn mock that returns whatever the test seeded proves the implementation mirrors itself, not that it drives git correctly — and `mock.module` is process-global in bun:test, which this repo already treats as a known landmine. Port the web-style real-repo fixtures instead.

**7. Test additions to lock this down:** renames (2), merge-in-progress refusal (3), a commit in an unborn-HEAD repo, and running the suite with `EDITOR`/`PAGER` set so the class of (1) can't regress.

Again — the hard parts here (index-info round-trip, batching, partial-staging preservation) are done well, which is exactly why the remaining gaps are worth closing: 2 and 3 silently commit wrong content in the user-data territory this PR exists to protect.

bashrusakh:
## Review follow-up — all 7 points addressed

Thanks for the thorough review. Every point is fixed in commit `8cbdd4b68`:

1. **Blocker (env):** temporary-index commands now run through `runGitCommandWithEnvOrThrow` (execFile with an explicit env overlay) instead of simple-git `.env()` — `EDITOR`/`PAGER`/`GIT_EDITOR`/`SSH_ASKPASS` no longer trip `blockUnsafeOperationsPlugin`. Covered by a new test that runs the temp-index commit with those vars set.
2. **Staged renames:** `getStagedRenameSourcePaths` parses `status --porcelain -z` and carries the source half of every selected `R` pair into the temporary-index removal list; a new real-git test asserts the old path is not resurrected. (`C` copies are deliberately excluded — the source stays in the index.)
3. **Merge-in-progress:** the selected commit is refused while `MERGE_HEAD` exists, before any temporary-index command; a new test verifies the merge state and unmerged index entries are left untouched.
4. **Pathspec fallback:** now explicitly gated to the non-temporary-index path (`|| commitWithTemporaryIndex`), so a temp-index failure can never degrade into a full-index commit.
5. **VS Code parity:** commit failures now throw with git stderr (the bridge surfaces it in the toast), `explicitStageFiles` is filtered against git status, index mutations are serialized per repo root (`withGitIndexMutationQueue`), and the rename/merge guards are mirrored. The pathspec fallback is intentionally not ported — the VS Code raw commit path never passes pathspecs to `git commit` — documented in code.
6. **VS Code tests:** kept in the existing mock style (real-repo fixtures are a larger follow-up), but the new tests now assert behavior-relevant state (pathspec file contents, command sequences, rejection messages) rather than mirroring the implementation.
7. **New coverage:** staged rename, merge-in-progress refusal, unborn-HEAD first commit, editor-env presence, and copy-record exclusion.

Validation: web git suites 103/103 (vitest) and 10/10 (bun test), VS Code batching 10/10, type-check and lint clean on both packages.

bashrusakh:
## OCR review follow-up

Ran `ocr review` (open-code-review v1.10.0) over the review-follow-up commits:

- `8cbdd4b68` (review points): 5 findings — 2 medium (commit summary regex mis-parses deletions-only output; empty commit/branch when the output header does not match) and 3 low (swallowed `status --porcelain -z` errors on web and VS Code; MERGE_HEAD guard after staging). All fixed in `c89537daf`.
- `c89537daf`: 3 findings — 2 medium (rev-parse fallback carried a misleading `GIT_INDEX_FILE`; summary regexes depend on English git output) and 1 low (missing exit code in the VS Code status-failure message). All fixed in `3efc4dbdc`.
- `3efc4dbdc` (final): **0 findings**.

Final validation on HEAD `3efc4dbdc`: web git suites 103/103 (vitest) and 10/10 (bun test), VS Code batching 10/10, type-check and lint clean on both packages, `git diff --check` clean.

btriapitsyn:
Thanks for the deep work here — the command-line length limit on large selections (#3099) is a real bug, and the test coverage you built is thorough. I'm closing this one on scope, though: the temporary-index commit machinery (HEAD read-tree, porcelain rename parsing, English-pinned commit-output regexes, custom execFile wrappers, duplicated again in the VS Code service) is a parallel git implementation this project would have to maintain and debug for years, and that cost outweighs the bug.

The fix I'd merge is much smaller: pass the selected paths via `--pathspec-from-file=<file> --pathspec-file-nul` (supported by `git add`, `git restore`, `git reset`, and `git commit` since git 2.25/2.26). That removes the argv limit entirely, needs no batching, no temporary index, and no output parsing, and this PR already uses exactly that flag for `git rm`. A PR shaped like that — same tests, a fraction of the mechanism — would be very welcome.

PR #3126, opened 2026-08-25
perf(ui): speed up session switching and scrolling

c-w-xiaohei:
## Intent

Reduce main-thread work during session switching and message scrolling in large workspaces, and prevent stale settings responses from moving the project or session context back to an older selection.

In a workspace with approximately **2,500 sessions and 82 directories loaded in the browser**, while the Sidebar rendered about **18 rows**, rough end-to-end session-switch measurements showed:

| Metric | Average reduction | Maximum reduction |
|---|---:|---:|
| Total interaction time | approximately 45% | approximately 59% |
| Main-thread processing | approximately 57% | approximately 61% |
| Longest task | approximately 57% | approximately 61% |

These measurements cover the complete switch, including Sidebar updates, session-state processing, history mounting, and first-screen message restoration.

#### Issue

The previous `SessionSidebar` owned the shell, session collection, project grouping, row actions, live state, and synchronization in one mounted tree.

That made unrelated state share the same render path. For example, switching from session A to B changes only the current selection, but could still re-enter session collection, grouping, ordering, and row construction. Memoization might prevent the final DOM update, but it could not avoid the selectors, sorting, and allocations performed before that boundary.

The same mount dependency also coupled background session synchronization to Sidebar visibility.

#### Design and why

The Sidebar is split by state ownership and update frequency:

```mermaid
flowchart TD
  Layout["MainLayout / VSCodeLayout"]
  Sync["useSessionListSync<br/>bootstrap, refresh, cleanup"]

  Sidebar["SessionSidebar<br/>shell, search, topology"]
  Collection["SessionProjectCollection<br/>membership, hierarchy, ordering"]
  Scroller["SessionProjectScroller<br/>viewport, sticky headers, DnD"]
  Group["SessionGroupSection<br/>one project/group projection"]
  Tree["SessionTreeItem<br/>recursive structure"]
  Row["SessionNodeItem<br/>per-session live state"]

  Stores["Existing session and UI stores"]

  Layout --> Sync
  Layout --> Sidebar
  Sidebar --> Collection
  Collection --> Scroller
  Scroller --> Group
  Group --> Tree
  Tree --> Row

  Stores -->|"structural inputs"| Collection
  Stores -->|"leaf selectors"| Row
```

Each boundary has a specific reason:

- `useSessionListSync` stays mounted at the layout level because bootstrap, refresh, and cleanup must continue even when the Sidebar or its visible list is hidden.
- `SessionSidebar` owns shell concerns such as search, navigation, and project/worktree topology. It does not derive the session tree.
- `SessionProjectCollection` subscribes only to structural inputs such as membership, parent/child relationships, grouping, and ordering.
- `SessionProjectScroller` owns viewport behavior, sticky headers, and drag-and-drop without becoming another session-data owner.
- `SessionGroupSection` receives one project/group projection, allowing unrelated groups to preserve their references and skip work.
- `SessionNodeItem` subscribes directly to its own active, status, unread, permission, and activity values.

This separation matters because these states change at different rates. Selection and live status are row-level updates; membership and ordering are structural updates; bootstrap and cleanup follow the application lifecycle. Keeping them behind separate boundaries prevents a frequent row update from entering collection-wide work.

Existing stores remain the source of truth. The refactor does not add a Sidebar snapshot, revision counter, structural cache, or second session store.

Prefetch and export also receive the target session's directory explicitly instead of reading the currently active directory during a switch.

#### Result

Switching from session A to B updates the active selectors for the affected rows and the dedicated selection/prefetch effects. It does not rebuild the session collection or unrelated project groups.

Structural changes such as create, archive, delete, reparent, pin, and ordering updates still invalidate the collection at their owning boundary. Background synchronization also continues independently of whether the visible Sidebar list is mounted.

Related commits:

- `bfdfa13c` `perf(ui): streamline session sidebar state`
- `c9baa4a2` `perf(ui): isolate sidebar session selection`
- `0cbaa3bd` `perf(ui): narrow session switch fanout`

#### Issue

Global sessions were primarily stored as arrays. Looking up a session, finding its parent or root, or grouping it by directory could scan those arrays again. A normal workflow may run only 2 to 6 agents in parallel, but each agent produces several state changes. With thousands of loaded sessions, a small event batch could scan the same list and notify the same subscribers several times.

#### Design and why

The global sessions store now maintains indexes for ID lookup, parent/child relationships, root sessions, directory membership, and active/archive membership. A complete snapshot builds the indexes once; incremental events update only changed sessions and related groups.

Session, status, ordering, and activity-timing changes received in one server event batch are merged before each store updates. The existing event batch is the right boundary because it preserves server order without adding a debounce delay. The indexes live with the global store because that store owns every event that can invalidate them.

#### Result

Incremental processing now follows changed sessions and affected groups instead of scanning all loaded sessions for each event. Unchanged groups preserve their object references, so their subscribers do not receive unrelated updates.

The 1,000-event fixtures are complexity regression tests. They guard against a return to `event count x all loaded sessions`; they do not represent normal agent concurrency.

Related commit:

- `7342fb94` `perf(ui): index and batch session updates`

#### Issue

Switching into a long session immediately enabled the normal virtual-list overscan. Off-screen turns mounted in the same task as the first viewport, bringing extra React trees, Markdown DOM creation, row measurement, virtual-list correction, and scroll adjustment into the initial switch.

#### Design and why

History starts with zero overscan, then restores partial and normal overscan across later frames. Pending restoration is canceled if the user switches away first.

Overscan cannot remain low because steady scrolling needs nearby content ready before it reaches the viewport. Splitting activation from steady scrolling keeps the normal scroll buffer without making off-screen content block the first presentation.

#### Result

Off-screen turns no longer all mount in the first activation task. Initial row mounting and measurement are distributed across frames while normal scrolling retains its preload range.

Related commit:

- `94ff85d4` `perf(ui): stage history overscan restoration`

#### Issue

OpenChamber already cached rendered Markdown block HTML, but that cache did not preserve constructed DOM. Returning to a recent session could still parse cached HTML, create DOM nodes, and reapply link and control decorations.

#### Design and why

This PR adds a bounded detached DOM cache. Eligible completed Markdown nodes are retained when leaving a session and restored as the same nodes on return.

Caching DOM is necessary here because enlarging the existing HTML cache would not remove browser-side HTML parsing, node creation, or decoration. Entries are separated by runtime, directory, session, message, part, locale, and rendering mode, and restored nodes leave the cache immediately so two renderers cannot own them.

Streaming content, Mermaid diagrams, temporary interaction state, and oversized content bypass the cache. Mermaid keeps its normal mount path because its controllers, observers, and listeners cannot be restored from DOM nodes alone. The cache keeps up to four blocks for each of eight recent sessions, favoring first-screen restoration across several sessions while bounding memory use.

#### Result

Revisiting a recent session can restore its existing first-screen DOM and skip repeated HTML parsing, node creation, and most decoration. Tests verify original-node reuse, no new `innerHTML` write on a hit, single ownership, and runtime/directory/rendering identity checks.

Related commits:

- `98af147d` `perf(ui): reuse detached markdown DOM`
- `38e95baa` `perf(ui): widen markdown session cache`

#### Issue

A content-addressed Markdown block HTML cache already existed, but initial mount still generated synchronous fallback HTML before reading that cache through the asynchronous rendering path. A fully cached message could therefore repeat synchronous Markdown parsing during remount.

#### Design and why

This PR adds a synchronous full-hit lookup to the existing completed-block cache; it does not add another HTML cache. The cached result is used only when every block is present, so first paint never shows a partial message. Matching block and decoration identities also skip the later DOM update.

#### Result

A complete cache hit skips `renderMarkdownSync()`, matching blocks skip the later morph, and cache misses and streaming content keep the existing rendering behavior.

Related commit:

- `1182b872` `perf(ui): reuse warm markdown blocks`

#### Issue

Markdown controls created fixed icons from SVG strings through `innerHTML`. Mermaid-heavy histories also initialized each diagram separately during mount, concentrating controller setup, size reads, DOM writes, and observer registration in one task.

#### Design and why

Control icons now use DOM APIs and SVG `<use>` nodes. Mermaid setup runs after the initial critical work, batches reads before writes, shares a `ResizeObserver` and resize listener, and reuses observer measurements where possible.

Mermaid remains outside the detached DOM cache because static nodes cannot preserve its controller and observer state.

#### Result

Controls no longer parse fixed icons through `innerHTML`, and Mermaid-heavy histories avoid one synchronous observer and size-measurement chain per diagram.

Related commits:

- `65ba8834` `perf(ui): batch Mermaid viewer initialization`
- `36ad855f` `perf(ui): avoid parsing markdown control icons`

#### Issue

The OverlayScrollbar treated steady scrolling as a React update: read complete geometry, set state, render, and write thumb position. Most frames changed only the scroll offset. ScrollShadow also applied its mask directly to the long overflow element, attaching a viewport-edge effect to the complete message surface.

#### Design and why

OverlayScrollbar now updates the DOM directly. Geometry is recalculated after mount, resize, content changes, or configuration changes; steady scrolling reuses the current ratio and moves the thumb with `transform`. Scroll, resize, and content updates share one animation-frame callback.

Thumb position is local frame-level visual state, so React state added render and commit work without providing shared application state. ScrollShadow now lives on a viewport wrapper because the shadow describes viewport edges, not message content.

#### Result

Steady scrolling no longer requires continuous React renders or complete size reads. Thumb movement uses `transform`, and the shadow mask no longer covers the full overflow content.

Related commits:

- `b6dbd3c1` `perf(ui): streamline overlay scrollbar updates`
- `a84b8fb2` `perf(ui): isolate scroll shadow effects`

#### Issue

Settings requests can finish out of order. An older response containing project A could arrive after the user selected project B and move the UI back to A. Last-response-wins does not help because the last response may belong to the oldest request, while discarding the complete response would also lose valid values for other settings.

#### Design and why

Each local settings change receives an increasing revision number. A request records the current revision when it starts; when it returns, local changes made after that revision are reapplied field by field.

Revisions record user-change order independently of network completion order. Field-level merging keeps valid response fields while protecting values changed during the request. Consecutive changes are coalesced where possible, old history is removed after relevant requests finish, and switching runtimes clears the previous runtime's tracker.

#### Result

An older settings response cannot overwrite a newer project or session selection. Rapid changes preserve the user's latest choice while keeping mutation history bounded.

Related commits:

- `c3411788` `refactor(ui): encapsulate settings mutation tracking`
- `73432eb7` `fix(ui): bound settings mutation history`


btriapitsyn:
Carson, this is excellent work — thank you. The event-batching and index design in the global store came through review cleanly, and the detached DOM cache is more careful than most attempts at this pattern I've seen.

Heads up on what we changed before merging. main moved under this PR last week: the chat timeline scroll engine was replaced (tanstack → @legendapp/list with anchored-turn scrolling), which made a couple of subareas target code that no longer exists. We merged main into your branch and:

- dropped the activation-overscan staging — the tanstack path it tuned is gone, and the new list ships adaptive rendering natively;
- kept the hook/mask-based scroll shadows from main instead of the viewport-wrapper ScrollShadow rewrite (the virtualized list owns its scroll element, so there is no wrapper to hand styling to) — the chat OverlayScrollbar keeps your `disableHorizontal`;
- landed your OverlayScrollbar direct-DOM rewrite, with binding switched to follow the live container node: the chat scroller now remounts on every session switch, and a bind-once contract left the scrollbar attached to a dead element after the first switch;
- merged the markdown DOM cache with the new streaming pipeline (block-commit reveal, streamed-code highlighting), added a content fingerprint to the cache key so an edited part can't restore stale DOM, and moved the key memoization to scalar ids so unrelated store updates don't re-run the render pipeline;
- added an expansion-aware bootstrap-demand owner published from the visible sidebar — the layout-level sync only knows directories, so expanded projects were hydrating one at a time at background priority;
- fixed a settings edge where a change still sitting in the debounce buffer could be briefly reverted by a concurrent settings GET.

One note on the description: the 1000-event fixtures it mentions aren't in the branch. The behavioral tests you did write are the right kind, so nothing is lost — just flagging the mismatch.

The sidebar decomposition, the cleanup lifecycle, and the settings revision tracker merged as designed, and streaming metrics on our harness actually improved on top of the new engine. Thanks again — contributions at this level move the whole project.

btriapitsyn:
Correction to my earlier note: the 1000-event fixture IS in the branch — global-session-status.test.ts exercises a 1000-event batch and asserts both the final state and the single-publication-per-store behavior, exactly as the description says. Our review pass misgrepped it and I repeated that without double-checking; sorry for the noise. Separately, we've landed one follow-up on main (74290852e): the setState monkeypatch on the session-status store is now an explicit replaceGlobalSessionStatusById API — same behavior, one owner for the membership derivation.

PR #3143, opened 2026-08-26
feat(ui): show context usage in subagent panels

alexandrereyes:
## Intent

Show the active subagent session's token count and context usage in the desktop/web context-panel chat header, beside the panel controls. The readout uses the model recorded by that subagent's own latest token-reporting assistant message.

Validation on current `main` is blocked by existing failures fixed separately in #3142. The feature PR remains based directly on upstream `main` and contains none of that prerequisite's files.

## Non-goals

- No changes to the main chat header, Work Status, VS Code, or mobile layouts.
- No new network requests, polling, persisted state, or runtime contract.
- No change to token accounting semantics.

## Affected surfaces

`packages/ui` desktop/web `ContextPanel` chat tabs only. The active session chat gets a compact readout such as `246.3K (24.6%)`; its tooltip includes used tokens, context-window limit, and output limit. Panels without an active chat session render no indicator. VS Code and mobile do not use this context-panel registry.

## Repository guidance

|---|---|---|
| `AGENTS.md`, `openchamber-change-discipline` | Shared executable UI and exports changed | Keeps the feature in existing UI owners, adds focused tests, documents the invariant, and runs repository-wide validation. |
| `sync-state-invariants` and sync documentation | The indicator reads directory-scoped session messages | Subscribes to the active session and explicit directory. It does not infer child usage from the parent or add a second cache. |
| `performance-engineering` and performance tooling documentation | Session messages update while streaming | Only the active chat panel mounts one narrow session subscription; hidden and non-chat panels do no work. The reverse scan is memoized and uses the existing message array. |
| `theme-system` and token reference | A visible header indicator is added | Reuses `ContextUsageDisplay` and existing semantic status, border, typography, and focus styles. |
| `locale-ui-patterns` | Visible numbers, tooltip text, and accessible output are locale-sensitive | Reuses existing tooltip keys and adds a translated full accessible value to every locale. Compact values use `Intl.NumberFormat`. |
| `ui-api-decoupling` implementation map | Shared UI runtime boundaries were checked | No API, bridge, URL, auth, or runtime capability changes. |
| Work-status and context-surface documentation | These modules own usage arithmetic and desktop panel placement | The documented per-session computation now includes coherent model resolution for the panel header. |
| `changelog-authoring` | The change is visible to desktop users | Adds one concrete Desktop bullet to `[Unreleased]`; VS Code changelog is unchanged because that runtime is unaffected. |

## Validation

| Check | Result |
|---|---|
| `bun install --frozen-lockfile` with Bun 1.3.14 and `OPENCHAMBER_ALLOW_UNAUTHENTICATED_LAN` unset | Passed. |
| Focused context tests | Passed: 16 tests, 0 failures. |
| `bun run test` | Feature tests and UI, VS Code, Electron suites pass. Current upstream `main` fails 4 session-goal tests because they read host settings; #3142 fixes those. The complete suite passes with #3142 applied: root 1/1, UI 326/326, VS Code 26/26, Electron 17/17, Web 162/162 files, with 1,627 web tests passed and 2 skipped. |
| `bun run type-check` | Passed with #3142 applied. The feature alone also passed UI type-check. |
| `bun run lint` | Passed with #3142 applied. Current upstream `main` has 9 unrelated unused-code errors fixed by #3142. |
| `bun run build` | Passed for every workspace, both for the feature and with #3142 applied. Existing Vite chunk-size and mixed dynamic/static import warnings remain. |
| `bun run docs:validate` | Passed: 460 pages and 46 sidebar links. |
| `bun run dead-code` | Completed with exit 0. It reports the existing repository backlog; the new formatter is used. |
| Focused `bunx oxlint` | Passed on new and focused files. Existing findings in substantially pre-existing `ContextPanel.tsx` and `ContextUsageDisplay.tsx` were inspected separately; no new line introduces one. |

## Visual evidence

The same implementation was previously exercised in the browser against a real parent and subagent session. The main header remained `37.9K (3.6%)` while the active subagent panel independently showed `246.3K (24.6%)` beside Expand. The current reapplication changes no layout or styling from that implementation, but a fresh screenshot could not be attached from this isolated validation environment.

## Risks and failure behavior

A newer assistant message without token data leaves the latest complete snapshot visible. Live provider limits win over models.dev metadata. If neither source resolves the recorded subagent model, the indicator waits rather than showing the generic 200K fallback as fact; that fallback applies only to a known model with no advertised context limit. The feature performs no writes and starts no network request, so there is no rollback or cleanup path.


btriapitsyn:
Thanks — this is a well-disciplined implementation: single-snapshot model resolution, tests, all locales covered. Two things need to change before merge:

1. **Cold sessions show nothing until they stream.** The indicator reads messages via `useSessionMessages`, which is a passive read of the host directory store. The panel's chat itself runs in an embedded iframe with its own sync store, so the host store only gets this session's messages from live `message.updated` events. After a page reload, an idle subagent panel renders no indicator until the subagent produces new activity — which contradicts the PR's "waits instead of guessing" framing (it waits forever). Use the loader-backed subscription (`useSessionMessageLoadState` in `packages/ui/src/sync/sync-context.tsx`) so the messages are actually fetched when the panel chat tab is active, and keep the mount scoped to the active chat tab as you already do. Done looks like: reload the app with an idle subagent panel open — the indicator appears without new streaming.
2. **`aria-label` regression on existing surfaces.** Before this PR the non-interactive `<div>` variant of `ContextUsageDisplay` always carried `aria-label={t('contextUsage.aria.label')}`. Your change gates it behind `staticProgressbar`, but VS Code (`VSCodeLayout.tsx`), mini-chat (`MiniChatLayout.tsx`), and the mobile tap-to-open variant use that div without the flag — they all lose their accessible name. Restore the label for the non-static div; done looks like: every existing `ContextUsageDisplay` call site renders with an accessible name, covered by tests.

With those two fixed this is mergeable as-is.

PR #3008, opened 2026-08-19
feat(git): add an integrated Git workspace with graph and side-by-side diffs

mattv8:
## Intent

Add an integrated Git review workspace for desktop web and Electron. Reviewers can keep Changes and Commit controls, a paginated branch graph, and side-by-side file diffs visible together, inspect historical commit files, and return to chat with one action. Separate views remains the default. Changed-file tabs show only the selected file; selecting another changed file opens or activates its own tab.

Graph collapse and height are global device preferences rather than per-repository state. Persistence migration drops the ambiguous legacy per-repository values. Git state stays current through a bounded `.git` watcher on web and Electron, authenticated SSE refresh hints, mutation hints through the shared runtime API, and activation or outdated-state revalidation.

Closes #2704.

macOS users can download and install a pre-release with this feature from here: https://github.com/mattv8/openchamber-testing/releases

## Non-goals

- Change staging, commit, sync, revert, or branch-operation semantics.
- Replace the narrow-web, mobile, or VS Code inline composition.
- Add arbitrary branch-to-branch comparison or per-session change attribution.
- Make the integrated workspace the default.

## Affected surfaces

- `packages/ui`: persisted review layout, desktop workspace panes, graph state and cache, global graph collapse and height preferences, historical previews, work-status graph, Git refresh coordination, responsive fallback, localization, theme-aware graph colors, and tests.
- `packages/web`: validated and bounded Git history, refs, merge-base, changed-file, and blob-preview routes. A bounded watcher observes validated repository gitdirs and publishes `openchamber:git-changed` on the existing authenticated event stream.
- `packages/vscode`: bridge parity for the shared Git data contract and mutation refresh hints. VS Code retains its host-native editor composition and relies on mutation hints plus activation revalidation rather than the web server watcher.
- Electron receives the desktop workspace and Git watcher behavior through the shared UI and in-process web runtime. Mobile and narrow web retain the existing inline layout.
- A fresh install defaults `gitReviewLayout` to `separate`. A missing or invalid server value does not overwrite the local `gitReviewLayout`; valid server values continue to synchronize through `settings.json`. VS Code's settings bridge passes the field through, and Electron inherits the web-server behavior in process.
- Global graph collapse and height preferences replace the ambiguous legacy per-repository values without guessing which repository should own them.

## Repository guidance

| Guidance | Why it applies | How the change complies |
|---|---|---|
[...]

PR #3148, opened 2026-08-26
feat(branch-diff): let users pick the base branch

gaojunran:
## Intent

The branch diff compares the current branch against an auto-detected default branch. Once detected, the base was fixed: the base picker only appeared as a fallback when detection failed, and the mobile Changes surface had no branch-diff scope at all. This PR makes the base selectable everywhere and brings the branch diff to mobile:

- **Desktop** (`DiffView.tsx`): a persistent "Base branch: `<name>`" row above the branch file list; clicking it opens the same picker, so the base can be switched (and the list re-fetched) at any time.
- **Mobile** (`MobileChangesSurface.tsx`): the full branch-diff scope — a Changed/Branch split control in the header (Branch appears only when the default branch is known and differs from the current one), base resolution with a searchable base picker, a read-only branch file list with A/M/D/R/C badges, and per-file diffs rendered through the existing `PierreDiffViewer` fileDiff path.
- **Both**: the desktop view now reuses the shared pure helpers (`resolveRepositoryDefaultBranch`, `candidateBranchesForBasePicker`) instead of inline copies, so the base-resolution and candidate-filtering contracts live in one testable place (`branchDiffScope.ts`).

https://github.com/user-attachments/assets/cec0ccc9-746a-42dd-8e40-4424a80acd1c

https://github.com/user-attachments/assets/20b957c6-a533-4e6c-ba35-bf479e29338b

## Non-goals

- No "restore auto-detection" affordance after a manual override (desktop or mobile).
- No server, git-API, or renderer changes — branch diffs reuse the existing `/branch-base`, `/range-files`, and `/range-diff` endpoints.
- No CHANGELOG entry (per PR convention discussed with the author).
- Upstream `main` lint failures are not fixed here (see Risks).

## Affected surfaces

- `packages/ui`: `components/views/DiffView.tsx` (desktop context-panel diff, shared across Web, Desktop, and Remote runtimes); `apps/MobileChangesSurface.tsx` (hosted-mobile and Capacitor Changes surface); new hook `hooks/useMobileBranchDiffScope.ts`; shared helpers and tests in `components/views/branchDiffScope.ts` / `.test.ts`.
- i18n: zero new keys — existing `diffView.scope.*`, `diffView.branch.*`, `gitView.branch.*`, and `gitView.pr.field.baseBranch` keys are reused on both surfaces.
- No persisted or external contract changes; the base override reuses the existing `useGitBaseBranchStore`, keyed per directory and current branch.

## Repository guidance

|---|---|---|
| AGENTS.md — load every matching project skill before editing | UI component styling and behavior change on two surfaces | `theme-system` (shared tokens, layout, touch targets), `locale-ui-patterns` (no new strings, existing keys), `ui-api-decoupling` (git API calls stay runtime-agnostic via shared `gitApi`), `sync-state-invariants` (range-keyed diff cache keeps stale-range completions out), `communication-style` loaded |
| AGENTS.md — keep shared-UI changes in the shared source tree | Change spans desktop and mobile shared UI | All changes inside `packages/ui`; no runtime-specific branches |
| AGENTS.md — keep changes minimal | Simplification pass requested | Desktop inline base/candidate logic replaced by the shared helpers the mobile path already used (net -5 lines) |

## Validation

| Check | Result |
|---|---|
| `bunx tsc --noEmit` (`packages/ui`) | 0 errors |
| `bunx eslint` on changed files | 0 problems |
| `bun test branchDiffScope.test.ts` | 35/35 pass (incl. 12 new mobile helper cases) |
| Product build | 2963 modules; mobile hook strings verified present in the built bundle |
| Manual (desktop) | Preview service, real git repo: open Branch scope -> change base via the "Base branch" row -> list re-fetches for the new base, no stale-list flash |
| Manual (hosted mobile) | Real browser against preview service: Changed/Branch split appears only when a default branch is known and differs; base row and picker work; per-file diffs render |


mattv8:
This overlaps with the Git diff and review workspace work in #3008, so I'll help review and QC it. Give me 24 hours and I'll report back with findings.

mattv8:
Apologies for taking longer than the 24 hrs I promised... I found two mobile failure paths that need handling before merge:

1. `useMobileBranchDiffScope.ts:91-97,118-124`  
   After switching directories with Branch selected, a failed `fetchBranches()` leaves `branches` null. Mobile neither retries nor treats the failure as definite unavailability, so it stays on "Resolving base" indefinitely. Desktop has bounded retries and falls back after exhaustion.

2. `useMobileBranchDiffScope.ts:235-251`, `MobileChangesSurface.tsx:1087-1111`  
   A failed range-diff request removes its cache placeholder, which the detail view interprets as loading again. The file detail therefore spins indefinitely without an error state or retry action. Backing out and reopening retries incidentally, but the current screen gives no recovery path.

Validation: `git diff --check` passed. Focused tests and `oxlint` could not run because this checkout lacks the required installed packages (`react` and `oxlint`).

btriapitsyn:
Thanks @gaojunran for fixing the inaccessible base picker, and for the mobile implementation and tests. I've now implemented and tested a broader Changes/walkthrough update on `bohdan/dev` in https://github.com/openchamber/openchamber/commit/de98eff0c.

Both panels now share the same branch and commit pickers and selection state, so the diff you inspect is the one the walkthrough explains. Branch comparisons include local edits and honor the exact selected ref, and the same flow also supports individual commits. I'm closing this in favor of that shared implementation so source selection and comparison behavior stay consistent across Changes and walkthrough.

The mobile branch-browsing part of your PR isn't replaced by this work. That's still a useful follow-up, and this PR remains the reference for it. Thanks @mattv8 for the review as well.

gaojunran:
Thanks you all!

PR #3152, opened 2026-08-26
Add Debian package builds for Linux

marcrhandalldev:
## Intent

Add Debian package artifacts alongside the existing Linux AppImages so Debian-based distributions can install OpenChamber through standard package tools.

## Non-goals

Linux in-app updates remain AppImage-only. Debian package updates continue to use the system package manager.

## Affected surfaces

- Electron Builder Linux packaging
- Production Linux release workflow
- Desktop release smoke workflow
- Linux installation documentation
- Debian package metadata and application icon

The web, VS Code, mobile, macOS, and Windows runtimes are unaffected.

## Validation

- Built `OpenChamber-1.20.0-linux-amd64.deb` locally.
- Verified Debian package metadata with `dpkg-deb`.
- Verified the 512x512 hicolor application icon is included.
- Verified the AppImage with the existing Linux AppImage verifier.
- Ran Electron architecture and updater tests.
- GitHub Actions will verify native x64 and arm64 Linux builds.

## Risks and failure behavior

The `.deb` package is an additional artifact and does not change existing AppImage packaging or update behavior. Release CI fails if the Debian metadata, desktop entry, or 512x512 icon is missing.


mattv8:
@btriapitsyn, I think this can close in favor of #3232. The Debian packaging and release work overlap heavily, and #3232 has the current Debian verifier, icon generation, and artifact wiring. The only part worth preserving here is the installation documentation if #3232 does not already cover it.

btriapitsyn:
Closing in favor of #3232, which covers the Debian packaging with a verifier and release artifact wiring. We’ll carry the installation docs into that review. Thanks for the contribution, and thanks @mattv8 for pointing us to #3232.

Issue #1051, opened 2026-04-27
[Bug] VS Code Extension Usage Page Missing Multiple Quota Provider Support


sdunfeng:
### Description

The VS Code extension's Usage page shows "Provider not configured" for several providers, even though:
1. The API keys are properly configured in the Providers settings
2. The APIs work correctly (provider functionality is working)
3. The web server version (`packages/web/server/lib/quota/providers/`) has full quota support

### Root Cause

VS Code extension's `quotaProviders.ts` is missing several quota provider implementations compared to the web server.

Two missing pieces for each provider:

1. **`listConfiguredQuotaProviders()`** (line 365) - Does not check for the provider's auth entries

2. **`fetchQuotaForProvider()`** (line 1489) - Missing provider cases, falls through to default:
   ```typescript
   default:
     return buildResult({
       providerId,
       providerName: providerId,
       ok: false,
       configured: false,
       error: 'Unsupported provider',
     });
   ```

### Affected Providers

VS Code extension is missing quota support for these 3 providers:

| Provider ID | Provider Name | Web Server | VS Code |
|-------------|---------------|-----------|---------|
| `minimax-coding-plan` | MiniMax Coding Plan (minimax.io) | ✓ | ❌ |
| `minimax-cn-coding-plan` | MiniMax Coding Plan (minimaxi.com) | ✓ | ❌ |
| `ollama-cloud` | Ollama Cloud | ✓ | ❌ |

### Affected Files

- `packages/vscode/src/quotaProviders.ts`

### Expected Behavior

VS Code extension should support quota tracking for all providers, similar to the web server implementation:

- `packages/web/server/lib/quota/providers/minimax-coding-plan.js`
- `packages/web/server/lib/quota/providers/minimax-cn-coding-plan.js`
- `packages/web/server/lib/quota/providers/ollama-cloud.js`

### Reproduction Steps

1. Configure any of the missing providers (MiniMax or Ollama Cloud) in VS Code extension Providers settings
2. Open Usage page in VS Code extension settings
3. Select the provider
4. Observe "Provider not configured" warning even though API key is set

### Environment

- OpenChamber VS Code Extension
- VS Code
- Platform: Windows/macOS/Linux

### Related Code

Web server has full quota provider support in `packages/web/server/lib/quota/providers/`.

VS Code extension needs corresponding implementations added to `packages/vscode/src/quotaProviders.ts`:
- Add auth checks in `listConfiguredQuotaProviders()`
[...]

PR #3155, opened 2026-08-26
feat(ui): add ClinePass to QUOTA_PROVIDERS frontend whitelist

Ghilteras:
## Problem

The ClinePass server-side quota provider module exists in the backend registry but is invisible in the Usage panel because the frontend `QUOTA_PROVIDERS` array and `QuotaProviderId` type don't include it. The UI never queries `/api/quota/providers` for discovery — it uses a hardcoded compile-time list.

## Changes

Two files, two insertions:

1. `packages/ui/src/types/quota.ts` — add `'cline-pass'` to the `QuotaProviderId` union type
2. `packages/ui/src/lib/quota/providers/index.ts` — add `{ id: 'cline-pass', name: 'ClinePass' }` to `QUOTA_PROVIDERS` array

## Notes

This is the same pattern as [issue #1051](https://github.com/openchamber/openchamber/issues/1051) where MiniMax and Ollama Cloud providers were missing from the VS Code extension frontend. The server-side module at `packages/web/server/lib/quota/providers/cline-pass.js` is already registered in the registry — this PR completes the frontend wiring.

A longer-term improvement would be to have the UI fetch the provider list dynamically from `GET /api/quota/providers` instead of maintaining a separate compile-time whitelist.

### DIFF
--- a/packages/ui/src/lib/quota/providers/index.ts
+++ b/packages/ui/src/lib/quota/providers/index.ts
@@ -7,6 +7,7 @@ export interface QuotaProviderMeta {
 
 export const QUOTA_PROVIDERS: QuotaProviderMeta[] = [
   { id: 'claude', name: 'Claude' },
+  { id: 'cline-pass', name: 'ClinePass' },
   { id: 'codex', name: 'Codex' },
   { id: 'cursor', name: 'Cursor' },
   { id: 'github-copilot', name: 'GitHub Copilot' },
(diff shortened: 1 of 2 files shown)


btriapitsyn:
Thanks for the PR, but the premise doesn't hold: there is no ClinePass provider module in `packages/web/server/lib/quota/providers/` — nothing on main registers `cline-pass` server-side. Adding it to the frontend whitelist alone would show a Usage entry with no backend behind it. If you'd like ClinePass support, it needs the server-side quota provider first; happy to look at a PR that adds both ends together.

PR #3160, opened 2026-08-26
fix(server): support in-app self-update for macOS launchd services (#3159)

### LINKED ISSUE #3159: [Bug] In-app 'Update' fails on macOS launchd service with 'Foreground servers must be updated by their service manager'
### What's wrong?

When running OpenChamber on macOS as a background user service installed via `openchamber startup enable` (`launchd` User LaunchAgent), clicking the "Update" button in the Web / PWA UI fails with HTTP 409 and the error toast:

> *"Foreground servers must be updated by their service manager. Set OPENCHAMBER_SYSTEMD_UNIT when running under systemd, or run openchamber update and restart the service."*

### Environment

- **OpenChamber**: `@openchamber/web@1.20.0` / `@openchamber/web@1.21.0`
- **OS**: macOS (Darwin arm64 / Apple Silicon)
- **Service Manager**: macOS `launchd` (installed via `openchamber startup enable --host 127.0.0.1 --port <port>`)
- **Node**: v22.x
- **Package Manager**: `pnpm` (also applies to `npm`/`bun`)

### Steps to reproduce

1. Install and enable OpenChamber as a native macOS startup service:
   ```bash
   openchamber startup enable --port 14000
   ```
2. Open the OpenChamber Web UI / PWA from a browser or remote mobile device (e.g. over Tailscale).
3. When a new version is available, click **Update** in the UI.
4. The update is rejected with:
   `409 Conflict: Foreground servers must be updated by their service manager. Set OPENCHAMBER_SYSTEMD_UNIT when running under systemd...`

### Root Cause

In `packages/web/server/lib/opencode/openchamber-routes.js` (under `app.post('/api/openchamber/update-install')`):

1. The service mode check identifies that the server is running in foreground mode (`launchMode === 'foreground'`), which is how `launchd` executes the service wrapper.
2. The code assumes any foreground service must be Linux `systemd`:
   ```javascript
   const isForegroundService = launchMode === 'foreground';
   const systemdServiceUnit = isForegroundService ? resolveSystemdServiceUnit(process.env) : null;

   if (isForegroundService) {
     if (!systemdServiceUnit) {
       return res.status(409).json({
         error: 'Foreground servers must be updated by their service manager. Set OPENCHAMBER_SYSTEMD_UNIT when running under systemd, or run openchamber update and restart the service.',
       });
     }
     // ... systemd-run logic ...
   }
   ```
3. Because macOS does not use `systemd`, `systemdServiceUnit` is null, causing all in-app updates to fail on macOS when running under `launchd`.

### Suggested Fix

1. Differentiate between `systemd` on Linux and `launchd` on macOS (`process.platform === 'darwin'`).
2. On macOS, allow the update command (`pnpm add -g @openchamber/web@latest` or equivalent) to run and restart/reload via `launchctl kickstart -k gui/$(id -u)/dev.openchamber.web` (or unload/load `~/Library/LaunchAgents/dev.openchamber.web.plist`).

Jorgeruiz97:
## Intent

Fixes #3159.

When running OpenChamber on macOS via `openchamber startup enable` (native `launchd` User LaunchAgent in foreground mode), attempting to update from the Web / PWA UI fails with HTTP 409:
> *"Foreground servers must be updated by their service manager. Set OPENCHAMBER_SYSTEMD_UNIT when running under systemd, or run openchamber update and restart the service."*

This change allows macOS `launchd` foreground services to update via `pnpm add -g @openchamber/web@latest` (or detected PM) and automatically trigger a service restart via `launchctl`.

## Non-goals

- Altering Linux `systemd` unit resolution or interactive CLI foreground behavior on Linux/Windows.

## Affected surfaces

- `packages/web`: `server/lib/opencode/openchamber-routes.js` (`POST /api/openchamber/update-install`).
- Runtime: Web / PWA on macOS (`darwin`) when managed by `launchd`.

## Repository guidance

|---|---|---|
| `CONTRIBUTING.md` / `packages/web` | Web server update installation routes | Preserves existing systemd behavior on Linux while cleanly handling macOS launchd service reload. |

## Validation

| Check | Result |
|---|---|
| Local manual verification on macOS 15 (Apple Silicon) | Triggered in-app update (`POST /api/openchamber/update-install`) under `launchd` service; successfully upgraded package and reloaded service without 409. |
| Health verification | `GET /health` returned `HTTP 200` with `openchamberVersion: 1.21.0`. |

## Visual evidence

No frontend UI component rendering changed. This resolves a backend API 409 rejection and prevents the error toast from blocking in-app updates on macOS.

## Risks and failure behavior

- **Rollback / Failure**: If `launchctl kickstart` fails, it falls back to `launchctl unload && launchctl load` of `~/Library/LaunchAgents/dev.openchamber.web.plist`.
- **Security & Compatibility**: Retains existing package manager detection and command isolation.

### DIFF
--- a/packages/web/server/lib/opencode/DOCUMENTATION.md
+++ b/packages/web/server/lib/opencode/DOCUMENTATION.md
@@ -365,6 +365,9 @@ an authoritative loopback callback URL even when OpenChamber binds port `0`.
     - Foreground servers running under a systemd user unit queue installation in
       a separate transient unit and restart the configured service afterwards.
       `OPENCHAMBER_SYSTEMD_UNIT` overrides the default `openchamber.service`.
+    - Foreground servers running under a macOS launchd User LaunchAgent update the
+      package and trigger a `launchctl kickstart` (with unload/load fallback) against
+      `~/Library/LaunchAgents/dev.openchamber.web.plist`.
   - `GET /api/openchamber/models-metadata`
   - `GET /api/zen/models`
 
--- a/packages/web/server/lib/opencode/openchamber-routes.js
+++ b/packages/web/server/lib/opencode/openchamber-routes.js
@@ -21,7 +21,6 @@ function quotePosixShell(value) {
   return `'${String(value).replace(/'/g, "'\\''")}'`;
 }
 
-
 export const registerOpenChamberRoutes = (app, dependencies) => {
   const {
     fs,
@@ -37,7 +36,6 @@ export const registerOpenChamberRoutes = (app, dependencies) => {
     fetchFreeZenModels,
     getCachedZenModels,
   } = dependencies;
-
 
   app.get('/api/openchamber/update-check', async (req, res) => {
     try {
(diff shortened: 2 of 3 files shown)


Jorgeruiz97:
Thanks for catching these details! Both items have been addressed in commit `fd1c083`:
1. Restored the explicit `launchMode` declaration so `isForegroundService` evaluates properly.
2. Parenthesized the `launchctl unload && load` fallback so it only evaluates if `launchctl kickstart` fails, preventing redundant load attempts after a successful kickstart.

btriapitsyn:
Thanks — the 409 on launchd-managed macOS servers is a real gap, and the label matches what `openchamber startup enable` installs. Three things before this can merge:

1. **Gate the launchctl path to servers actually managed by launchd.** The `case 'darwin'` branch currently replaces the restart command for every macOS server, including daemon-mode servers started from the CLI with no LaunchAgent installed. For those users the update installs but the restart silently fails (`kickstart` finds no service, the fallback unloads/loads a plist that does not exist), which is a regression from today's working daemon restart. Only use launchctl when the launchd service manages this process — e.g. check that `~/Library/LaunchAgents/dev.openchamber.web.plist` exists and the server runs in foreground mode; keep the existing restart path for daemon mode on macOS.
2. **Assert the restart command in the new test.** It currently only checks the HTTP response, so the exact regression in point 1 would pass. Add a case for macOS daemon mode without the plist, and verify which restart command each mode produces.
3. **Drop the stray blank lines** added throughout `openchamber-routes.js`.

Jorgeruiz97:
Thanks for the feedback @btriapitsyn! I've updated the PR addressing all three points:

1. **Gated `launchctl` restart to verified launchd services**: Now checks that the server is running in foreground mode on macOS and that `~/Library/LaunchAgents/dev.openchamber.web.plist` exists. Unmanaged foreground instances on macOS receive 409, while daemon-mode instances on macOS (with or without the plist) preserve the standard CLI restart command (`node cli.js serve` / `openchamber serve`).
2. **Added restart command assertions in tests**: Expanded unit tests to verify the exact spawned restart command for macOS launchd foreground, macOS unmanaged foreground (409), macOS daemon without plist (CLI restart), and macOS daemon with plist (CLI restart).
3. **Cleaned up formatting**: Removed all stray blank lines in `openchamber-routes.js`.

Jorgeruiz97:
### Real-Life macOS launchd Verification

Validated live service restart on macOS 15 (Apple Silicon) with `dev.openchamber.web` active under `launchd`:

1. **Service Registration & Gating Check**:
   - `~/Library/LaunchAgents/dev.openchamber.web.plist` detected.
   - Verified foreground launch mode gating logic.

2. **Triggered Restart Command**:
   ```bash
   launchctl kickstart -k gui/$(id -u)/dev.openchamber.web
   ```
   *Exit code: `0` (Success)*

3. **Post-Restart Verification**:
   - Process restarted with a new PID under `launchctl list`.
   - Server health endpoint returned `HTTP 200 OK`.

4. **Unit & Documentation Status**:
   - Documentation updated in `packages/web/server/lib/opencode/DOCUMENTATION.md` (commit `a3750a2ac`).
   - All 7 route unit tests in `openchamber-routes.test.js` pass, asserting exact command execution for launchd foreground, unmanaged foreground (409), and daemon mode CLI fallback.

Jorgeruiz97:
### Raw Terminal Output Evidence

```bash
# 1. Verify launchd service status & initial PID
$ launchctl list | grep openchamber
19110   143     dev.openchamber.web

# 2. Execute launchctl kickstart restart command
$ launchctl kickstart -k gui/$(id -u)/dev.openchamber.web
$ echo $?
0

# 3. Verify process restarted with new PID
$ launchctl list | grep openchamber
42539   -9      dev.openchamber.web

# 4. Verify server health endpoint post-restart
$ curl -s -o /dev/null -w "%{http_code}\n" http://localhost:9000/health
200

# 5. Run server route unit tests
$ bun test packages/web/server/lib/opencode/openchamber-routes.test.js
bun test v1.3.5 (1e86cebd)

packages/web/server/lib/opencode/openchamber-routes.test.js:
(pass) OpenChamber foreground update route > rejects a foreground update when the server is not owned by systemd [7.55ms]
(pass) OpenChamber foreground update route > rejects an unsafe systemd unit override before starting an update job [3.03ms]
(pass) OpenChamber foreground update route > queues the install in a transient systemd unit and returns its job identifier [2.71ms]
(pass) OpenChamber foreground update route > rejects foreground update on macOS when launchd plist does not exist [2.47ms]
(pass) OpenChamber foreground update route > allows foreground update on macOS when launchd plist exists and invokes launchd restart command [3.11ms]
(pass) OpenChamber foreground update route > allows daemon update on macOS without launchd plist and invokes CLI restart command [2.52ms]
(pass) OpenChamber foreground update route > allows daemon update on macOS with launchd plist present and still invokes CLI restart command [3.32ms]

 7 pass
 0 fail
 11 expect() calls
Ran 7 tests across 1 file. [143.00ms]
```

PR #3166, opened 2026-08-27
fix(desktop): stop inactive project startup reads

wqpan:
## Intent

Reduce Desktop startup and active-session stalls caused by unnecessary managed OpenCode project/session work. The initial visible directory bootstraps before the full global snapshot is scheduled for idle time; Desktop warms only the last-used directory; exited managed OpenCode children clean up their registry record. Desktop also skips inactive project config prewarming and the tray no longer polls session status for directories the sync layer already owns.

## Non-goals

This does not change OpenCode itself, remove complete global active/archived coverage, or alter Archive, direct-session-route, and retention demand paths.

## Affected surfaces

- `packages/ui`: Desktop, web, and VS Code global session bootstrap scheduling; Desktop config prewarming; Desktop tray status reconciliation.
- `packages/web`: managed OpenCode lifecycle. Desktop historical warmup is reduced and natural child exit cleans its registry entry.
- Desktop runtime: packaged macOS startup and active-session filesystem behavior were sampled against the managed OpenCode PID.
- Persisted registry: stale managed-process entries are removed after natural child exit.

## Repository guidance

|---|---|---|
| `AGENTS.md` and `openchamber-change-discipline` | Shared UI and managed lifecycle behavior changed. | Kept changes in owning UI sync/config/tray and web lifecycle modules with focused tests and docs. |
| `sync-state-invariants` | Global/directory session bootstrap and status ownership changed. | Scoped loads remain non-authoritative; only a complete snapshot sets `ready`; synced directories retain watchdog/event status ownership. |
| `performance-engineering` | Startup CPU, active-session polling, filesystem fanout and antivirus overhead were reported. | Removed unscoped project config prewarm and duplicate tray status polls while retaining fallback coverage for unsynced sessions. |
| `desktop-shell` | Managed Desktop child-process lifecycle changed. | Registry cleanup is idempotent and only removes the child entry after its own exit. |

## Validation

| Check | Result |
|---|---|
| Focused sync/bootstrap/global-polling tests | Passed |
| `bun test src/stores/useConfigStore.prewarm.test.ts` | 3 passed |
| `bun test src/hooks/tray-status-poll.test.ts` | 7 passed |
| `bun test src/hooks/tray-status-poll-unmounted-sync.test.ts` | 2 passed |
| `bun run type-check:ui` and `bun run type-check:web` | Passed |
| `bunx oxlint` on new tray helper/tests | Passed; existing backlog remains in older large source files |
| `bun run electron:build` | Packaged mac-arm64 app and zip successfully; application signature verification passed |
| Active catalog-session runtime sampling | Before: 3,134 cross-project/OMO accesses and 213 `.git`/worktree accesses in 30 seconds; managed OpenCode reached 22%-92% CPU. After the tray fix: 0 cross-project/OMO and 0 `.git`/worktree accesses in 30 seconds; managed OpenCode was 1.5% CPU. Sophos may retain a prior scan backlog but no new matching files were emitted. |

## Visual evidence

No rendered UI structure or styling changes. This PR changes background request scheduling, config prewarm scope, managed-process cleanup, and tray status polling; system-level request/filesystem sampling is the relevant evidence.

## Risks and failure behavior

The complete global snapshot remains eventual and demand-triggerable, so global cache authority and retention behavior are preserved. Desktop inactive-project config is loaded explicitly when selected in Settings or otherwise requested; it is no longer prewarmed. Tray polling continues for unsynced global session directories, while sync-owned directories retain event/watchdog coverage.


mattv8:
Potentially related is https://github.com/openchamber/openchamber/pull/2998 I'll check the implications although my feat only reloads worktrees on mutations.

mattv8:
I checked #2998 after rebasing it onto the latest `main`. Its menu and mutation refreshes stay separate from this PR's background discovery eligibility, so the only overlap is `SessionSidebar.tsx` integration and no behavior needs to be coupled.

btriapitsyn:
Thanks — the session-aware scoping and the render-loop fixes are exactly the right mechanism, and I want both. What I can't take is the settings surface and the default flip. Concretely:

1. **Drop the default flip.** Users who rely on worktree grouping in the sidebar would lose automatic discovery after updating, with no hint why. `selectWorktreeDiscoveryProjects` already limits discovery to the active project and session-owning projects — that alone removes the background cost the issues describe.
2. **Remove the three settings** (`worktreeDiscoveryEnabled`, `worktreeDiscoveryIntervalMs`, `backgroundProjectSessionLoadingEnabled`) and all their locale/persistence/desktop plumbing. The interval knob in particular is a tuning parameter no user should have to understand. If a real need for an off switch shows up later, it can come back as one setting.
3. Keep the referential-stability fixes in `useProjectRepoStatus` and `SessionSidebar` — those stand on their own.
4. Fix the indentation slips in `OpenChamberPage.tsx` and `persistence.ts` (lines with an extra leading space).

With that shape this merges.

wqpan:
Follow-up validation and fix for the remaining background scan path.

The original change correctly scoped worktree discovery, but runtime tracing showed a second path: the root-level global session loader was still issuing an unscoped session-list request on startup and on a blind 45-second interval. That caused the OpenCode server managed by OpenChamber to enumerate many registered project roots.

This follow-up keeps the behavior safe while removing the unnecessary background cost:

- Collapsed projects with only historical sessions are excluded from worktree discovery, repo-status polling, and bootstrap demand.
- The active project is always retained.
- A collapsed project with a genuinely live session is retained.
- Startup and recovery session refreshes now use the eligible directory set and the existing directory-scoped API.
- The blind 45-second global refresh was replaced with recovery signals plus a 10-minute backstop.
- Complete global history is loaded on demand when Archive is opened.
- Auto-cleanup only loads the complete global list when auto-cleanup is actually enabled.
- No new user-facing setting was added; this follows the earlier review direction to prefer a smart default over another tuning toggle.

Verification:

- `worktreeDiscoveryProjects.test.ts`: 12 pass
- `useSessionListSync.test.tsx`: 14 pass
- `useGlobalSessionsPolling.test.ts`: 10 pass
- Session component isolated suite: 28/28 test files passed
- UI type-check: passed
- UI ESLint: passed
- Web assets rebuilt and local `/Applications/OpenChamber.app` restarted
- Clean eslogger sampling after startup showed only 4 project roots accessed in the post-start window, versus approximately 10 roots before the scoped startup/recovery changes

The six follow-up commits are pushed to `wqpan/openchamber`, branch `fix/worktree-discovery-settings`.

wqpan:
Addressed the latest review findings in two follow-up commits:

- `7a1cf8d82 fix(session): restore authoritative startup session load`
  - Restores the one-time `ensureGlobalSessionsLoaded(getAllSyncSessions())` startup load.
  - This restores the `status: 'ready'` / `hasLoaded` contract for authoritative cleanup, archived auto-folders, and global-cache consumers.
  - The blind 45-second full refresh remains removed. Recovery signals and the 10-minute backstop continue to use the eligible-directory scoped refresh path.

- `beef81d3b fix(session): stabilize worktree discovery dependencies`
  - The discovery effect now keys off the existing content-stable `projectWorktreeDiscoveryKey` rather than the volatile eligible-project array.
  - A ref provides the current project entries when the stable-key-driven effect runs, avoiding cancellation/restart of Git/worktree discovery for session ticks that do not change eligible projects.

Verification after the change:

- `worktreeDiscoveryProjects.test.ts`, `useSessionListSync.test.tsx`, and `useGlobalSessionsPolling.test.ts`: 36 pass
- session component isolated suite: 28/28 test files passed
- UI type-check: passed
- UI ESLint: passed

The current `checks` CI failure contains three unrelated test files that this branch does not modify:

- `src/components/auth/SessionAuthGate.behavior.test.tsx`
- `src/components/chat/markdown/markdownCore.test.ts`
- `src/lib/shortcuts.test.ts`

The branch diff does not touch those files or their implementation areas. In particular, the shortcuts failure is the stale expectation for `switch_context_surface` (`mod` expected vs `mod+alt` configured). The review fixes above are pushed to the PR branch for a fresh check run.

wqpan:
Resolved the upstream `main` merge conflict.

- Merged the latest upstream `main` into `fix/worktree-discovery-settings`.
- The only content conflict was in `packages/ui/src/components/session/SessionSidebar.tsx`, where the branch's stable discovery-project reference was combined with upstream's nested-repository/raw-worktree lifecycle changes.
- No conflict markers remain.
- After resolution, full workspace type-check and lint passed.
- Session component isolated tests passed: 31/31 test files.
- Targeted session tests passed: 36 tests.

The merge commit is `644472482` and has been pushed to the PR branch. GitHub now reports the PR as `MERGEABLE`; the new checks are pending.

wqpan:
The failed `checks (pull_request)` run was also fixed.

Root cause: after syncing upstream `main`, the English dictionary contained four new Git empty-state keys that were missing from the Turkish dictionary:

- `gitView.empty.discoveringRepositories`
- `gitView.empty.discoverFailed`
- `gitView.empty.retryDiscovery`
- `gitView.empty.selectRepositoryPlaceholder`

Added the Turkish translations in commit `4a52bab4f`.

Validation:
- i18n dictionary parity: 2 pass
- UI type-check: passed
- UI lint: passed

A new PR checks run is now pending on the updated head `4a52bab4f`.

wqpan:
Follow-up validation on current HEAD `11312acfc`:

The decisive root cause was `useConfigStore.prewarmProjectConfigs`, which ran after every Desktop initialization and iterated every registered project, loading providers and agents once per second. This caused managed OpenCode to read each project's `.opencode`/`.agents` tree and left Sophos with a sustained scan backlog.

`11312acfc` skips inactive-project config prewarm on Desktop while retaining the active project load and all explicit Settings/project-directory loads. A freshly packaged and installed macOS build was sampled for 30 seconds: **0** cross-project/OMO file events from the managed OpenCode PID. The managed process stayed around 10-12% CPU. Sophos remained elevated from the prior backlog, but no new matching files were emitted.

Validation: focused prewarm test 3/3, UI type-check, oxlint for authored test code, LSP diagnostics, packaged app signature verification, and system-level `fs_usage` sampling.

wqpan:
Current HEAD `dcd49fe00` adds tray status-poll ownership deduplication. In the real active `metricflow-catalog` session scenario, the freshly packaged app recorded **0 cross-project/OMO accesses** and **0 `.git`/worktree accesses** over 30 seconds; managed OpenCode was **1.5% CPU**. Before the config-prewarm and tray fixes, the same interval recorded 3,134 cross-project/OMO accesses and 213 `.git`/worktree accesses, with managed OpenCode at 22%-92% CPU. Sophos can still process the earlier scan backlog, but no new matching filesystem reads were emitted.

mattv8:
The core Desktop scoping work is sound, but this needs another pass before merge:

1. **Restore automatic global-session recovery after an in-place runtime switch.** Previously, the 45-second full `refreshGlobalSessions()` restored the store to `ready`. This PR removes that interval; the replacement recovery and backstop paths only call `refreshGlobalSessionsForDirectories()`, which never restores authoritative status. Because the poller does not remount on endpoint changes, web and VS Code can remain non-authoritative indefinitely after switching hosts. Done means an eager-load surface automatically returns to `ready` after a runtime switch while Desktop's demand-only fresh-mount behavior remains unchanged.
2. **Add regression coverage for that switch.** Seed the global store as `ready`, perform the runtime reset to `idle`, trigger recovery, and assert that a full refresh restores `ready`.
3. **Resolve the conflict with current `main`.** `packages/ui/src/lib/i18n/messages/tr.ts` has an ordering conflict around `gitView.empty.retryDiscovery` and `gitView.empty.selectRepositoryPlaceholder`. Preserve upstream placement and run `bun test packages/ui/src/lib/i18n/messages.test.ts`.
4. **Make the lifecycle warmup test runtime-hermetic.** `packages/web/server/lib/opencode/lifecycle.test.js:218-246` expects two warmups without controlling `OPENCHAMBER_RUNTIME`. It passes as web but fails as Desktop because Desktop now warms one directory. Save, set, and restore the intended runtime in `finally`, following the adjacent Desktop test.
5. **Correct `packages/ui/src/components/session/sidebar/DOCUMENTATION.md`.** It still says all known roots remain demanded and collapsing cannot opt a directory out. The implementation now excludes collapsed history-only projects while retaining the active project and live-session owners.
6. **Correct `packages/ui/src/sync/DOCUMENTATION.md`.** Desktop no longer schedules a full snapshot on fresh mount, so Archive is a primary demand edge rather than only cold-start recovery.
7. **Update the PR Intent.** State that Desktop schedules no automatic full global snapshot and completes coverage on demand.

Current validation: the isolated UI suite passes all 374 test files; UI/web type-check and lint pass; the lifecycle suite passes 32/32 as web and fails 31/32 as Desktop at the non-hermetic warmup test above. Packaged Desktop behavior was not independently rerun.

Separately, there is an open product question about whether this scoping should apply to remote Desktop hosts. That is for the maintainer @btriapitsyn to decide, not part of this contributor checklist.

btriapitsyn:
Thanks @wqpan for the Desktop startup investigation and the before/after measurements. Thanks @mattv8 for checking the runtime-switch and test paths. Your latest review remains the next step here; this note does not add another checklist or change who owns the outstanding fixes.

PR #3185, opened 2026-08-27
fix(sessions): new chat opens in the active project by default

lmf-ggal:
Superseded by #3220, rebuilt atomically on the current `main` with the merge conflict removed and updated validation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+- Sessions: **New chat** now opens directly in the active project when one is selected, instead of always dropping into the generic managed `chats` root. Chats opened with no active project (and explicit mini-chat/VS Code targets) still use the managed chats root as before.
+
 ## [1.21.0] - 2026-08-26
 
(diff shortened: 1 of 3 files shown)


btriapitsyn:
Thanks — the premise is real and the shape (reusing the existing `"project"` branch) is right. Two things before this can merge:

1. **Directory can disagree with the active project.** With `target === "project"`, the draft directory resolution prefers `currentDirectory` over the active project's path. If the user is currently inside a managed chat session, `currentDirectory` is that chat's `chats/YYYY-MM-DD/session-…` folder, so pressing "New chat" produces a draft with `target: "project"`, `selectedProjectId: null`, and the old chat folder as its directory — a project-targeted draft bound to a chats-root path. Done looks like: when the target was inferred from the active project (not explicit options), resolve the draft's project and directory from the active project itself, falling back to `currentDirectory` only when it actually belongs to a project (`resolveDraftProjectForDirectory` returns non-null).
2. **Cover that case in the test.** The new test sets `currentDirectoryMock` to the project path, which masks exactly this gap. Add a case where the active project is set but `currentDirectory` points at a `chats/` session folder, asserting the draft ends up on the active project's path and id.

PR #3190, opened 2026-08-27
fix(opencode): prevent OpenCode upgrades from failing with "Bad Request"

### LINKED ISSUE #3121: [Bug]  "Update OpenCode" in the UI always fails with "Bad Request": opencode `/global/upgrade` requires a `target` field, but OpenChamber sends an empty body
## Summary

Clicking "Update OpenCode" in the OpenChamber UI always fails with a toast showing only "Bad Request". The root cause is an API contract mismatch: since opencode 1.18.x, `POST /global/upgrade` requires a `target` field (a valid semver string) in the JSON body, but OpenChamber forwards an empty object `{}` when no explicit version is given. opencode's payload schema rejects it with HTTP 400. Because the rejection response has the shape `{name, data:{message}}` with no `error` field, the server falls back to the raw HTTP status text, so the user sees nothing but the uninformative "Bad Request".

## Environment

- OpenChamber: `@openchamber/web@1.20.0` (latest on npm at time of writing), installed via `bun add -g`, running in `openchamber serve` daemon mode
- OpenCode: standalone install (curl install script) at `~/.opencode/bin/opencode`; reproduced on both 1.18.22 and 1.18.23
- Managed runtime: runtime=web; OpenChamber spawns `opencode serve --hostname 127.0.0.1 --port <random>`
- OS: Debian 13 (Trixie)

## Steps to reproduce

**Via the UI:**
1. Open the OpenChamber web UI and wait for the "new OpenCode version available" prompt (or go to Settings → OpenCode update)
2. Click the update button
3. A toast appears: **"Bad Request"**; no upgrade happens

**Via the API (bypassing UI auth, calling the managed opencode server directly):**

```bash
# Empty body — exactly what OpenChamber forwards when no explicit target is set
curl -i -u "opencode:$OPENCODE_SERVER_PASSWORD" \
  -X POST http://127.0.0.1:<port>/global/upgrade \
  -H 'Content-Type: application/json' -d '{}'
```

## Actual result

```http
HTTP/1.1 400 Bad Request
{"name":"BadRequest","data":{"message":"Missing key\n  at [\"target\"]","kind":"Payload"}}
```

The UI shows only "Bad Request" (the HTTP statusText), with no actionable detail. Behavior is identical on opencode 1.18.22 and 1.18.23. As a control, sending `{"target":"not-a-version"}` returns 400 `Expected a semantic version`, confirming this is strict payload schema validation.

## Expected result

Clicking update without choosing a specific version should upgrade OpenCode to the latest release — or at minimum surface a meaningful error message in the UI.

## Root cause analysis

Four factors combine:

1. **Frontend always sends an empty object** — the update button calls `POST /api/opencode/upgrade` with a hardcoded `body: JSON.stringify({})`
2. **Server forwards it verbatim** — `server/lib/opencode/routes.js` (v1.20.0, `registerOpenCodeRoutes`):
   ```js
   const target = typeof req.body?.target === 'string' && req.body.target.trim().length > 0
     ? req.body.target.trim()
     : undefined;
   // ...
   body: JSON.stringify(target ? { target } : {}),   // ← {} when no explicit target
   ```
3. **opencode 1.18.x requires `target` to be present and a valid semver** — the empty body fails validation and returns 400
4. **The error message is lost** — the server extracts failures via `payload?.error || response.statusText`, but opencode's schema-rejection shape is `{ name, data: { message } }`, so `payload.error` is `undefined` and only "Bad Request" is displayed

Note: the CLI path (`opencode upgrade`) is unaffected — it uses the curl install script and does not go through this HTTP endpoint.

## Workaround

Passing an explicit `target` through OpenChamber's own API works, since it is forwarded verbatim:

```bash
curl -X POST http://127.0.0.1:3000/api/opencode/upgrade \
  -H 'Content-Type: application/json' \
  -d '{"target":"1.18.23"}'
```

Or run `opencode upgrade` in a terminal and restart OpenChamber afterwards.

## Suggested fix

1. When no explicit `target` is provided, resolve the latest available version first (e.g. reuse the existing `/api/opencode/upgrade-status` check) and always send `{ target }` to `/global/upgrade`
2. Improve error propagation: prefer `payload?.data?.message` and `payload?.error` over falling back to bare statusText, so similar mismatches can be diagnosed directly from the UI in the future

Thanks for building OpenChamber — happy to provide more details or test a fix.

### Where does it happen?

Desktop Web

### Version (if known)

_No response_

_No response_

### Logs (optional)

```shell

```

mdatsev:
## Intent

Fixes #3121.

OpenCode now requires `POST /global/upgrade` to include a semantic-version `target`. OpenChamber's existing update action omits that field, causing managed OpenCode updates to fail with HTTP 400 Bad Request.

To preserve the existing omitted-target behavior, the web and VS Code runtime implementations now resolve the latest published OpenCode version using their existing npm/GitHub lookup and include that version in the upstream request. Explicit caller-supplied targets continue through the existing path.

## Non-goals

1. No shared UI or update-toast changes.
2. No change to generic upstream error extraction. Improving diagnostics for other upstream failures can be handled separately.
3. No change to update availability or version comparison.
4. No change to OpenCode installation-method handling.

## Affected surfaces

1. Web resolves an omitted target before calling OpenCode's `/global/upgrade`.
2. Packaged Desktop continues to reject this operation for its bundled OpenCode through the existing capability guard. Desktop configurations that use the supported web-managed path share the web implementation.
3. Hosted mobile and Capacitor clients receive the corrected behavior from the OpenChamber server to which they connect.
4. VS Code applies the same fallback before calling its extension-managed OpenCode process.
5. Externally managed OpenCode remains unsupported through the existing capability guard.
6. No persisted data or rendered UI changes are involved.

## Repository guidance

|---|---|---|
| Root `AGENTS.md` and `openchamber-change-discipline` | The change affects an external API request, multiple runtimes, and managed-process restart behavior. | The patch only resolves and forwards the newly required target. Focused tests cover target selection, request serialization, the upstream endpoint, and successful restart behavior. |
| `ui-api-decoupling` with `references/runtime-parity.md` and `references/implementation-map.md` | The shared update action crosses web, Desktop, hosted-mobile, Capacitor, and VS Code runtime boundaries. | The shared UI contract remains unchanged. Web and VS Code own their OpenCode transport behavior, Desktop reuses the web implementation where applicable, and mobile clients use their connected OpenChamber server. |
| `packages/web/README.md` and `packages/web/server/lib/opencode/DOCUMENTATION.md` | The web server owns `/api/opencode/upgrade` and managed OpenCode lifecycle behavior. | The route documentation records how an omitted target is resolved before forwarding and restarting managed OpenCode. |
| `packages/vscode/README.md` and `packages/vscode/src/DOCUMENTATION.md` | VS Code owns a separate extension-host OpenCode process and restart path. | `opencode-upgrade-runtime.ts` applies the same omitted-target behavior while preserving extension-owned restart handling. |

## Validation

| Check | Result |
|---|---|
| `bun run type-check` | Passed across all workspaces |
| `bun run lint` | Passed across all workspaces |
| `bun run build` | Passed across all workspaces |
| `bun run test` | Passed across all workspaces |
| `bun test packages/vscode/src/opencode-upgrade-runtime.test.ts` | Passed |
| `bun run --cwd packages/web test -- server/lib/opencode/routes-upgrade.test.js` | Passed |
| `bun run build:web` | Passed; existing bundle-size warnings only |
| `bun run vscode:build` | Passed; existing bundle-size warnings only |
| `node --check packages/web/server/lib/opencode/routes.js` | Passed |
| `tool-use docs-list --check` | Passed with 0 warnings and 0 errors |
| `bun run docs:validate` | Passed with 460 pages and 46 sidebar links |
| Raw OpenCode 1.18.23: `curl -X POST http://127.0.0.1:<port>/global/upgrade -H 'Content-Type: application/json' -d '{}'` | Reproduced the original HTTP 400 response with `Missing key at ["target"]` |
| Isolated managed-runtime check: temporary `HOME`, data directory, `PATH`, and OpenCode binary; then `curl -X POST http://127.0.0.1:43824/api/opencode/upgrade -H 'Content-Type: application/json' -d '{}'` | Passed on branch HEAD `957620d4d`. OpenCode upgraded from 1.18.22 to 1.18.23, returned `{"success":true,"version":"1.18.23","restarted":true}`, restarted from port 33951 to 43729, and reported 1.18.23 after restart. |

## Visual evidence

There is a user-visible behavioral correction, but no rendered UI change. The existing Update action, toast, and surrounding UI are unchanged. This patch only corrects the runtime request sent after the existing action is invoked, so screenshots would be visually identical.

## Risks and failure behavior

1. An omitted-target upgrade depends on resolving a version from npm or GitHub.
2. If both lookups fail, OpenChamber returns an error without contacting `/global/upgrade` or restarting OpenCode.
3. If one lookup fails, the existing `Promise.allSettled` behavior uses the successful source.
4. If both succeed, the newer reported version is selected.
5. Explicit caller-supplied targets do not require the metadata lookup.
6. OpenCode remains authoritative for target validation and installation-specific upgrade behavior.
7. OpenCode restarts only after the upstream upgrade request succeeds.
8. Existing capability and in-flight upgrade guards remain in place.
9. No migration, persisted-state change, or rollback path is involved.

### DIFF
--- a/packages/vscode/src/opencode-upgrade-runtime.test.ts
+++ b/packages/vscode/src/opencode-upgrade-runtime.test.ts
@@ -75,13 +75,33 @@ describe('VS Code OpenCode upgrades', () => {
     assert.equal((request?.headers as Record<string, string>).Authorization, 'Basic test');
   });
 
+  test('resolves the latest version when no target is supplied', async () => {
+    const { manager, getRestartCount } = createManager();
+    let upgradeRequest: RequestInit | undefined;
+    globalThis.fetch = async (input, init) => {
+      const url = String(input);
+      if (url.includes('registry.npmjs.org')) return new Response(JSON.stringify({ version: '1.18.9' }));
+      if (url.includes('api.github.com')) return new Response(JSON.stringify({ tag_name: 'v1.18.10' }));
+      assert.equal(url, 'http://127.0.0.1:4096/global/upgrade');
+      upgradeRequest = init;
+      return new Response(JSON.stringify({ success: true, version: '1.18.10' }));
+    };
+
+    assert.deepEqual(await upgradeManagedOpenCode(manager), {
+      status: 200,
+      body: { success: true, version: '1.18.10', restarted: true },
+    });
+    assert.equal(getRestartCount(), 1);
+    assert.deepEqual(JSON.parse(String(upgradeRequest?.body)), { target: '1.18.10' });
+  });
+
   test('serializes concurrent managed upgrades', async () => {
     const { manager } = createManager();
     let release: (response: Response) => void = () => {};
     globalThis.fetch = (() => new Promise<Response>((resolve) => { release = resolve; })) as typeof fetch;
 
-    const first = upgradeManagedOpenCode(manager);
-    const second = await upgradeManagedOpenCode(manager);
+    const first = upgradeManagedOpenCode(manager, '1.18.9');
+    const second = await upgradeManagedOpenCode(manager, '1.18.9');
     assert.equal(second.status, 409);
     assert.equal(second.body.code, 'OPENCODE_UPGRADE_IN_PROGRESS');
 
--- a/packages/vscode/src/opencode-upgrade-runtime.ts
+++ b/packages/vscode/src/opencode-upgrade-runtime.ts
@@ -107,13 +107,14 @@ export const upgradeManagedOpenCode = async (manager: OpenCodeUpgradeManager | u
   if (openCodeUpgradePromise) {
     return { status: 409, body: { success: false, code: 'OPENCODE_UPGRADE_IN_PROGRESS', error: 'An OpenCode upgrade is already in progress.' } };
   }
-  const targetVersion = typeof target === 'string' ? target.trim() : '';
+  const requestedTarget = typeof target === 'string' ? target.trim() : '';
   const operation = (async (): Promise<UpgradeResult> => {
     try {
+      const targetVersion = requestedTarget || await fetchLatestVersion();
       const response = await fetch(new URL('global/upgrade', apiUrl).toString(), {
         method: 'POST',
         headers: { 'Content-Type': 'application/json', Accept: 'application/json', ...manager.getOpenCodeAuthHeaders() },
-        body: JSON.stringify(targetVersion ? { target: targetVersion } : {}),
+        body: JSON.stringify({ target: targetVersion }),
       });
       const payload = await response.json().catch(() => null) as { error?: unknown } | null;
       if (!response.ok) return { status: response.status, body: { success: false, error: typeof payload?.error === 'string' ? payload.error : response.statusText || 'Failed to upgrade OpenCode' } };
--- a/packages/web/server/lib/opencode/DOCUMENTATION.md
+++ b/packages/web/server/lib/opencode/DOCUMENTATION.md
@@ -88,7 +88,7 @@ This module provides OpenCode server integration utilities for the web server ru
   - `GET /api/config/settings`
   - `PUT /api/config/settings`
   - `GET /api/config/opencode-resolution`
-  - `POST /api/opencode/upgrade` (enforces the active runtime's upgrade capability, serializes supported OpenCode upgrades, then restarts managed OpenCode so the new binary is active)
+  - `POST /api/opencode/upgrade` (enforces the active runtime's upgrade capability and serializes upgrades; an omitted `target` resolves the latest published OpenCode version before the required upstream payload is sent, then managed OpenCode restarts so the new binary is active)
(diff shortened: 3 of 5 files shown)


btriapitsyn:
@mdatsev, this one is on us, and I want to explain rather than just close it.

You opened this fix for #3121 on the 27th. The next day #3209 landed with the same approach — resolve the latest published version when no `target` is supplied and send it to `/global/upgrade` — plus a second fix for the error message. It was written without knowing your PR existed: we've been clearing a very large issue backlog with new tooling, and that workflow picked issues off a list without checking whether someone already had a fix open. That's a gap in the process, not anything you did wrong, and it's now closed on our side: nobody takes an issue into work without first checking for an open PR on it.

Since the change is already in `main` and this branch now conflicts with it, I'm closing the PR. You're credited in both changelogs for this fix, because in substance it's yours. Sorry for the wasted effort, and thank you for the fix and the clear write-up.

PR #3195, opened 2026-08-28
fix(chat): queue messages as the only busy-path behavior (#3194)

### LINKED ISSUE #3194: [Bug] Sending a message while a subagent session is running silently drops it (no assistant reply)
### What is wrong?

Sending a chat message while a (foreground) subagent session is executing does not work: the message appears in the chat, but no assistant reply is ever generated. The send is silently dropped.

### Steps to reproduce

1. Start a turn in a session whose agent delegates to a foreground subagent (e.g. a task tool call that blocks the parent loop).
2. While the subagent is still running, type a message and send it (follow-up behavior set to "Steer" or "Send immediately").
3. The message is persisted in the chat, but no assistant response ever comes — even after the subagent finishes.

### Expected behavior

A message sent while the session is busy must not be silently lost. It should either be queued with a visible queue status (and dispatched when the session becomes idle), or rejected with a clear error.

### Actual behavior

The user message is stored, but the OpenCode runner discards the new work: `ensureRunning` in `packages/opencode/src/effect/runner.ts` returns the existing run for `Running`/`ShellThenRun` states and never starts the new prompt. The parent session stays `busy` for the entire foreground subagent duration because the parent loop is blocked in `handleSubtask` on `background.wait`, so the message never gets a turn.

Additionally, `delivery: "steer"` is not supported by the v1 `prompt_async` endpoint (the request body schema has no `delivery` field), so the "Steer" path cannot rescue the send — it hits the same discard path.

### Where does it happen?

Web / Desktop (shared UI), any platform.

### Version

Observed on current main (OpenChamber 1.20.x line, OpenCode 1.18.x embedded).

### Related issues

- #3047 — same silent-drop mechanism when a session is busy (different backend/environment)
- #1451 — queued message flushes too early at subagent delegation handoff (inverse problem)
- #2709 — closed; queued messages flushed when viewing a subagent session
- #1766 — closed; "steer" delivery feature request

### Notes

Root cause is confirmed by code inspection (read-only analysis). The UI mitigation (message queue, default follow-up behavior) works only when follow-up behavior is set to "Queue"; with "Steer"/"Send immediately" the send goes straight into the discard path.

bashrusakh:
## Intent

Fixes #3194. When the target session is busy, Enter, Ctrl+Enter, the send button, presets, dictation, magic prompts, and queued-chip sends keep work in the message queue instead of sending into a runner that may discard a prompt during an active turn.

The queue is now the only busy-path behavior. Idle sends keep the direct path.

## Non-goals

- No native V2 Queue or V2 Steer semantics.
- No new server queue endpoint or public SDK/API contract.
- No second persistence model or manual V1/V2 delivery selector.
- No changes to unrelated session, runtime, or UI behavior.
- No live runtime, browser, Electron, mobile, or VS Code smoke coverage is claimed below.

## Affected surfaces

- `packages/ui`: shared ChatInput routing, queued chips, queue store, auto-send, session sync, directory identity handling, settings normalization, and lifecycle recovery.
- `packages/web`: server-owned queue persistence, FIFO delivery, receipts, lifecycle generations, hold ownership, rollback handling, and shutdown draining.
- Web, desktop, VS Code, hosted mobile, and Capacitor mobile consume the shared queue contract. No runtime-specific queue API was added.
- Rebased onto current `main` (`111c4bf9d`). The queue panel keeps upstream queued-annotation preview (`getQueuedMessagePreview`, bounded `contextPreview`) and its collapsed-by-default behavior.
- Hold ownership is persisted per runtime so a reload keeps the same hold owner and can re-assert or release the previous incarnation hold instead of leaving it to expire.
- Persisted queue state remains backwards compatible. Legacy follow-up values normalize to queue-only behavior.

## Repository guidance

|---|---|---|
| `AGENTS.md` and `CONTRIBUTING.md` | Required repository and PR contract | Scope, ownership, validation, failure behavior, and limitations are documented here. |
| `openchamber-change-discipline` | Shared UI/runtime source, tests, and module contracts changed | Changes stay in owning modules and use focused regression coverage. |
| `sync-state-invariants` | Queue, hydration, reconciliation, lifecycle, runtime switching, and directory identity changed | State remains authoritative, scoped, generation-checked, and recoverable across stale completions. |
| `performance-engineering` | Queue dispatch and session/sidebar sync paths changed | FIFO claims, bounded queue work, and canonical directory lookup avoid duplicate or stale work. |
| `ui-api-decoupling` | Shared UI and server-owned queue transport changed | Runtime ownership and captured transport targets remain explicit. |
| `theme-system` and `locale-ui-patterns` | Queue status and settings UI changed | Existing theme/localization patterns are preserved. |
| `packages/ui/src/sync/DOCUMENTATION.md` | Sync ownership and lifecycle contract changed | Updated with the directory and reconciliation invariants. |
| `packages/ui/src/stores/DOCUMENTATION.md` | Queue store contract changed | Updated with queue ownership, recovery, hold identity, and preview behavior. |
| `packages/web/server/lib/message-queue/DOCUMENTATION.md` | Server queue lifecycle and persistence changed | Updated with FIFO, receipt, hold, rollback, and shutdown behavior. |

## Validation

At candidate `ecad7a4bf` (rebased onto `111c4bf9d`):

| Check | Result |
|---|---|
| UI focused suites (28 files, per-file isolation) | 626 passed, 0 failed |
| Web suites (queue runtime, settings, shutdown) | 164 passed, 1 platform-conditional skip |
| Total focused coverage | 790 passed, 1 skipped, 0 failed |
| `bun run type-check:ui` / `type-check:web` | Passed |
| `bun run lint:ui` / `lint:web` | Passed |
| Node syntax checks for changed server files | Passed |
| `bun run docs:validate` | Passed, 517 pages and 47 sidebar links |
| Rebase integrity | `range-diff` clean; non-delta files byte-equivalent to an ideal 3-way merge |
| Whole-PR review at `ecad7a4bf` | PASS, full PR coverage; OCR unavailable (HTTP 429), native read-only review used |

Hold-identity regression coverage: reload keeps the persisted owner (assert, re-assert, release), sequence continuity from the server-observed value, refused-echo convergence for assert and release, corrupt/blocked storage fallback, runtime switch token reuse, and pruning on delete/forget paths.

## Visual evidence

No live runtime surface was available for before/after screenshots or a recording. The queue status, collapsed/expanded panel, and busy-to-idle paths have focused component and store coverage, but this PR does not claim manual browser, desktop, mobile, or VS Code verification.

## Risks and failure behavior

- FIFO head claims prevent later queued messages from overtaking an unresolved send.
- Taken messages are restored immediately or retained as durable pending restores when runtime, directory, or lifecycle identity changes. Deletion invalidation prevents resurrection.
- Server holds are in-memory, token and sequence ordered, generation and directory scoped, TTL-bound, and preserved across unrelated queue persistence rollback. While a hold is active, only its owning token can release or replace it.
- Hold identity (owner token per runtime, sequence per runtime and session) is persisted in browser storage so reloads and runtime switches back stay the same owner. Storage is failure-safe, bounded, and holds no secrets; if storage is unavailable, identity falls back to memory and an orphaned hold still clears by TTL.
- Enqueue and its compensating cleanup run inside one serialized per-session mutation, so a clear or removal that lands mid-flight cannot deadlock the queue.
- Runtime-switch cleanup uses the captured runtime target, including proxy headers for ngrok-hosted runtimes. Stale runtime, directory, and session operations are rejected without mutating the active runtime.
- Windows drive and UNC aliases use the existing canonical child-store identity. POSIX case variants remain distinct.
- Queue state is bounded and persisted with existing compatibility handling. No migration or dependency change is required.
- Full repository tests, full monorepo build, and live runtime/manual interaction checks were not run in this environment.


btriapitsyn:
Thanks for all the work you've put into this. We tested the reported scenario on the current version by sending a follow-up while a foreground subagent was running, and the message was received and answered. We also checked the OpenCode path: ensureRunning reuses the active run, but the message is saved before that call and the running loop reads updated messages between steps. That doesn't establish the discard described in the issue. We're keeping the Steer/Queue choice, and without reproducing the reported failure, we don't see a reason for a queue rewrite of this size, especially with our v2 work underway. I'm closing this PR and #3194. Thanks again for trying to improve delivery reliability.

PR #3208, opened 2026-08-28
feat(sync): route subagent session.idle/error through notification batcher

herjarsa:
## Intent

Route subagent session lifecycle events (`session.idle` / `session.error`) through a notification batcher so a parent session gains visibility into when its background tasks complete or fail, instead of seeing nothing until its own turn ends.

## Non-goals

- Mid-run visibility (busy / retry / tool-call progress inside a subagent). This PR only changes terminal events (`idle` / `error`). Active descendant tracking is a separate concern (`global-session-status.ts` and the open reproducer tests at `packages/ui/src/sync/reproduce-2521.test.ts` cover that gap).
- Visible aggregate counts or click-to-failure navigation in this PR. The notification body carries no user-facing text; the rendering component is expected to read the notification-store counts and produce its own i18n strings if it wants to surface them. Adding `parentLink` fields to `Notification` and a rendering consumer is intentionally deferred to a follow-up PR.
- No change to top-level session notifications (the direct `appendNotification` path is preserved unchanged).
- No change to permission/question toast handling.

## Affected surfaces

- **New** `packages/ui/src/sync/subagent-notification-batcher.ts` — 128-line module exporting `subagentNotificationBatcher` singleton. Keys batches by `${directory}:${parentID}`, debounces 1.2s, flushes one consolidated notification per key. Per-session dedupe keeps retry cycles (error → idle for the same session) from inflating the batch. Tracks events in a `Map<sessionID, SubagentEvent>` so the latest event per session replaces the prior one.
- `packages/ui/src/sync/sync-context.tsx` — replaced the existing `if (session && ...parentID) { /* skip */ }` branch with a call to `subagentNotificationBatcher.queue(...)`. Top-level sessions keep the original `appendNotification(...)` call. The call site no longer needs `getState()` because the batcher dropped the parent-part walk (which was the only consumer of state).
- **New** `packages/ui/src/sync/subagent-notification-batcher.test.ts` — 8 unit tests. Covers: coalesce to one notification per parent, mixed idle/error produces error, different parents/directories do not coalesce, retry-cycle dedupe (error → idle, idle → error), debounce fixed-deadline semantics, and representative-session selection.
- `packages/ui/src/sync/DOCUMENTATION.md` — added a paragraph under the session.idle/session.error section documenting the routing, the 1.2s debounce, the (directory, parentID) key, and the per-session dedupe.
- `CHANGELOG.md` — `[Unreleased]` bullet crediting @herjarsa. Corrected: subagents previously produced no notifications at all (sync-context skipped them), not one toast per subagent.

## Repository guidance applied

| Source | Why applicable | Rules applied |
|---|---|---|
| `AGENTS.md` runtime boundaries | Sync-layer routing change | "Prefer authoritative state over heuristics" — the batcher reads `session.parentID` from the sync store, not a derived flag; "scope temporary fallbacks narrowly and clear them when authoritative state arrives" — the singleton's pending maps clear on flush. |
| `.agents/skills/sync-state-invariants/SKILL.md` | Live activity routing | The dispatch site is the only path that owns turn-complete/error notifications for subagents; no other reducer is involved. |
| `.agents/skills/openchamber-change-discipline/SKILL.md` | New module + call-site change | Smallest complete change: one new file, one call-site branch, one test file, one docs paragraph, one changelog line. The previous attempt's parent-part walk was removed because no consumer reads those fields. |
| `.agents/skills/locale-ui-patterns/SKILL.md` | User-facing toast text | No user-facing text in the batcher. The notification body carries `error: {}` (empty object) for the error type, so no i18n bypass exists. If counts are surfaced later, the rendering component owns the i18n keys. |

## Validation

- `bun test packages/ui/src/sync/subagent-notification-batcher.test.ts` — 8/8 pass. Covers: single-parent coalesce, mixed idle/error produces error with the representative being the first error, different parents do not coalesce, different directories do not coalesce, retry-cycle dedupe (error then idle keeps one idle entry; idle then error keeps one error entry), debounce fixed-deadline semantics, representative-session selection.
- `bun run type-check` in `packages/ui` — exit 0 for the touched files (`sync-context.tsx`, `subagent-notification-batcher.ts`, `subagent-notification-batcher.test.ts`). Pre-existing errors unrelated to this PR (missing `@types/dom-speech-recognition`, deprecated `baseUrl` option in tsconfig) were not introduced by this change.
- Manual rebase onto current `main` (post #3191, #3205, etc.): clean, zero conflicts.
- `bun run dead-code` — not run by hand; the only added file is `subagent-notification-batcher.test.ts`, and the only removed export is the `getState` parameter on `subagentNotificationBatcher.queue()`, which is a breaking change to an internal singleton (no external consumers).

## Behavior change for the user

Before: a parent session delegating to N background subagents sees zero notifications about the subagents until its own turn completes. The parent cannot tell whether the subagents are still running, have all finished, or have failed.

After: within ~1.2s of the last subagent finishing or failing, the parent sees one consolidated notification: either a `turn-complete` (any session in the batch ended without error) or an `error` (any session in the batch emitted a `session.error`, which wins over completes). The notification is always `viewed: false`. If any subagent in the batch errors, the notification type flips to `error`. Retry cycles that flip the same session's state do not double-count.

The notification body carries no user-facing text. If a downstream component wants to render an aggregate ("N subagents failed, M completed"), it should derive counts from the notification store's existing per-session/per-directory count indices and produce its own i18n strings.

## Risks and failure behavior

- If a subagent fires `session.idle` and then `session.error` within the 1.2s window, both events land in the same batch and the batcher emits one `error` notification (errors take priority over completes). This matches user intent: any failure in the window is surfaced as a failure.
- The batcher uses `setTimeout` from the global pool. If the page is suspended (mobile background tab, frozen window), the timer may be throttled or dropped. Notifications still flush eventually but timing may slip. Pre-existing characteristic of the platform.
- The singleton's internal state is not persisted. A page reload mid-batch drops pending notifications. This matches existing notification behavior and is intentional — the notification store is the system of record.
- The new module exports a process-wide singleton. In Electron, the renderer process has its own singleton (UI module, shared). The main process does not use this module. No cross-process coordination is needed.


herjarsa:
Pushed `3d665e754` addressing all three findings from the BLOCKED review at `8da44f189`:

1. **blocker (dead walk, broken matcher)**: dropped the parent-part walk entirely. The `parentSessionID`/`parentMessageID`/`parentPartID` fields were not part of `Notification`, no consumer read them, and the `<task id="X">` matcher missed the real `<task id="X" state="completed">` format. The batcher now emits a plain notification with no extra fields and no hardcoded text.

2. **non-blocker (hardcoded English)**: removed the `N subagents failed, M completed` string entirely. The notification body carries `error: {}` (empty) for the error type, so no i18n bypass exists. Rendering component owns any future i18n keys.

3. **non-blocker (retry double-count)**: switched the batch's per-session tracking from parallel `errored: SubagentEvent[]` and `completed: string[]` arrays to a single `Map<sessionID, SubagentEvent>` so a session that flips error → idle replaces the prior entry instead of being counted twice.

Also corrected the CHANGELOG claim: subagents previously produced no notifications at all (sync-context skipped them), not one toast per subagent. PR body updated to match the new code (no user-facing text in batcher, no aggregate counts claimed).

Tests: 8/8 pass. Type-check clean for the touched files. Would appreciate a re-review against `3d665e754`.

btriapitsyn:
Closing this PR.

Suppressing notifications for individual subagents is an intentional product decision. Subtasks are internal execution units orchestrated by the main agent — they exist to serve the primary task, not the user directly. Bubbling up notifications or dock badge increments every time a subagent finishes creates unnecessary noise. Users care about the primary agent completing its turn, which is already handled.

Thanks for the effort, but we don't want to change this behavior.

PR #3211, opened 2026-08-28
fix(quota): support Cursor Enterprise usage via auth/usage and team-scoped GetHardLimit

### LINKED ISSUE #3210: Cursor quota provider fails for Enterprise team accounts: false "No active Cursor subscription"
## Description

Cursor Enterprise accounts can't see their usage in OpenChamber. The login works, the Cursor dashboard shows the numbers fine, and Settings → Usage still says "No active Cursor subscription".

The cause lives in `fetchQuota()` in `packages/web/server/lib/quota/providers/cursor.js`. It reads `planUsage` from `GetCurrentPeriodUsage` and treats it missing as "subscription gone". Enterprise team members never get `planUsage` back, so valid accounts fail the check.

## Steps to reproduce

1. Sign in to Cursor with an Enterprise team account.
2. Import the credentials in OpenChamber (Settings → Integrations → Cursor).
3. Open Settings → Usage and select Cursor.

## Expected

Usage that matches the Cursor dashboard: included requests (something like `589 / 1000`) and the on-demand cap (`$0 / $250`).

## Actual

```json
{
  "ok": false,
  "configured": true,
  "error": "No active Cursor subscription"
}
```

## Root cause

```javascript
if (usage?.enabled === false || !usage?.planUsage) {
  return { error: 'No active Cursor subscription' };
}
```

For Enterprise, `GetCurrentPeriodUsage` returns a sparse payload with no `planUsage` key at all:

```json
{
  "billingCycleStart": "1787932328088",
  "billingCycleEnd": "1787932328088",
  "displayThreshold": 100
}
```

Meanwhile `GetPlanInfo` works and happily reports the plan as Enterprise. So the login is fine. The guard is what's wrong.

## Proposed fix

When `planUsage` is missing but the subscription isn't disabled, read the same endpoints the Cursor dashboard uses:

1. `GET /auth/full_stripe_profile` → resolve the team id
2. `GET /auth/usage` → included request counts (`gpt-4.numRequests` / `maxRequestUsage`)
3. `POST GetHardLimit` with the team id → `hardLimitPerUser` for the on-demand cap

Two things worth knowing before implementing:

- `GetHardLimit` without a team id only returns the team pool and omits the per-user cap, so the team id has to go in the request body.
- Skip `GetAggregatedUsageEvents` dollar totals. They reflect internal token cost, not the request/on-demand meters the dashboard shows. We tried that first and got numbers that contradicted the dashboard.

## Verified against a real Enterprise account

Patched build compared to the Cursor dashboard, same account:

| Source | Included | On-demand |
|---|---|---|
| Cursor dashboard | 589 / 1000 | $0 / $250 |
| Patched OpenChamber | 590 / 1000 | $0.00 / $250.00 |

The one-request difference is polling lag.

carvalab:
## Intent

Enterprise team members couldn't see their Cursor usage at all. Valid login, working Cursor dashboard, and OpenChamber still answered "No active Cursor subscription". The reason: `GetCurrentPeriodUsage` returns no `planUsage` for Enterprise accounts, and `fetchQuota()` read that as "subscription gone".

This PR leaves the Pro/Ultra path alone and adds a fallback that reads what the Cursor dashboard itself reads, in both the web server and the VS Code extension:

1. `GET /auth/full_stripe_profile` resolves the team id
2. `GET /auth/usage` gives included request counts, mapped to the `billing_cycle` window. Buckets there use legacy names, so the provider scans them dynamically: skip `startOfMonth`, pick the bucket with the highest `maxRequestUsage`, instead of hardcoding `gpt-4`
3. `POST GetHardLimit` with the team id gives `hardLimitPerUser`, mapped to the `on_demand` window (dollars converted to cents like the rest of the provider)
4. `POST GetMe` resolves the current user's `userId` straight from the access token
5. `POST GetTeamSpend` with the team id returns per-member spend; the member whose `userId` matches supplies the `on_demand` used value (`spendCents`)

**Verified meters only.** A review pass tightened this: a meter renders only when its data is verified. The `on_demand` window needs both the cap and a matched `spendCents`; if `GetMe` or `GetTeamSpend` fails, the window is omitted rather than rendered as `$0`. A plan name alone is not usage — a response with no verified meter returns the `No active Cursor subscription` error instead of `ok` with empty windows. Verified partial results stay (for example `billing_cycle` alone while spend can't be verified).

**Note on on-demand spend:** an earlier version reported the on-demand cap with `$0` used, because `GetHardLimit` exposes the cap but not the spend. Once the account exceeded its included requests, the Cursor dashboard showed real on-demand spend while OpenChamber still showed `$0.00 / $250.00`. There is no per-user spend endpoint, so the provider resolves the caller's `userId` via `GetMe` and scans `GetTeamSpend.teamMemberSpend[]` once for the matching row. `spendCents` matches the dashboard's on-demand meter exactly, and no email configuration, credential field, or cached state is involved.

Fixes #3210

## Non-goals

- No change to how credentials are read, stored, or refreshed in either runtime.

## Affected surfaces

- `packages/web` quota provider `cursor`: Enterprise accounts get usage windows including real on-demand spend; Pro/Ultra results stay byte-identical.
- `packages/vscode` extension `fetchCursorQuota`: same fallback chain, semantics, and failure behavior as the web provider; Pro path unchanged.
- Web/desktop Settings → Usage and the VS Code usage surface for Cursor: the error state becomes real numbers for Enterprise accounts.
- `packages/web/server/lib/quota/DOCUMENTATION.md`: documents the Enterprise semantics, the `GetMe`/`GetTeamSpend` spend chain, verified-meter behavior, and the web/VS Code sync expectation.

## Repository guidance

|---|---|---|
| AGENTS.md instruction order | Source change in the web server quota module and the VS Code extension | Loaded `openchamber-change-discipline`, `communication-style`, and `changelog-authoring`, read `packages/web/server/lib/quota/DOCUMENTATION.md` before editing |
| `openchamber-change-discipline` | Local implementations with an external API contract | Smallest complete change per runtime: guard split plus one fallback branch; every new call degrades to null and never fakes a meter; docs updated because the module doc states provider semantics |
| AGENTS.md correctness invariants | "Never let fetch failure masquerade as authoritative empty success" | Unverified spend omits the window; plan name alone errors; partial results render only verified meters |
| `communication-style` | Issue and PR text | Plain words, concrete numbers, no filler |
| `changelog-authoring` | Review requested a CHANGELOG entry; user-visible fix | Entry deliberately left out of this branch to avoid `[Unreleased]` merge conflicts; add at merge time: `- Usage: Cursor Enterprise accounts now show included requests, the on-demand cap, and the account's on-demand spend instead of a "No active Cursor subscription" error (thanks to @carvalab).` |
| Quota DOCUMENTATION.md precedent (kimi/copilot/zai sections) | Provider quirks get documented in the module doc | "Cursor quota semantics" section covers the team id requirement, the dynamic bucket scan, the `GetMe`/`GetTeamSpend` spend chain, and why `GetAggregatedUsageEvents` totals are not used |
| CONTRIBUTING.md PR contract | PR submission rules | Template completed with concrete evidence below |

## Validation

| Check | Result |
|---|---|
| `bunx vitest run server/lib/quota/providers/cursor.test.js` (packages/web) | 8/8 passed: Pro path regression, Enterprise windows including the team-scoped request bodies, bucket selection picks the highest `maxRequestUsage` and skips `startOfMonth`, on-demand spend matched by `userId`, on-demand window omitted when spend cannot be verified, plan-name-only is not success, disabled subscription error, all-fallbacks-fail error |
| `bunx oxlint` (packages/web) | No new findings; 3 pre-existing `no-runtime-typeof` findings on untouched lines (known backlog) |
| `bun run type-check` (packages/web) | Pass |
| `bun run lint` (packages/web) | Pass |
| `bun run test` (packages/web, 166 files) | 1681 passed, 5 failed. Verified the same 5 fail on clean `main` (git hunk + walkthrough tests, environment-dependent), so not from this change |
| `bun run test` (packages/vscode) | 26/26 test files pass; 6 new focused cursor tests: Pro regression (no fallback endpoints fired), Enterprise chain with team-scoped body, `userId` spend match, bucket selection, plan-name-only error, disabled subscription error |
| `bun run type-check` (packages/vscode) | Pass (both tsconfig projects) |
| `bun run lint` (packages/vscode) | Pass |
| `bunx oxlint` (packages/vscode changed files) | Only the pre-existing `no-runtime-typeof` finding on the untouched openrouter line (known backlog) |
| Live Enterprise account | The web provider path ran against a real Enterprise account on a patched desktop build before this PR; before/after below |

Not verified here: a live Pro-account request and a live VS Code Enterprise request (both covered by mocked tests), and rendered screenshots, since this PR was prepared in a Linux dev environment without the desktop app.

## Visual evidence

No rendered screenshots from this environment, but the user-visible change is exactly this API contract, captured before and after on a real Enterprise account:

Before (current `main` behavior):

```json
{
  "providerId": "cursor",
  "providerName": "Cursor",
  "ok": false,
  "configured": true,
  "error": "No active Cursor subscription"
}
```

After (this patch, with the account over its included requests):

```json
{
  "providerId": "cursor",
  "providerName": "Cursor Enterprise",
  "ok": true,
  "configured": true,
  "usage": {
    "windows": {
      "billing_cycle": { "usedPercent": 100, "valueLabel": "1002 / 1000" },
      "on_demand": { "usedPercent": 3.16, "valueLabel": "$7.90 / $250.00" }
    }
  }
}
```

The Cursor dashboard showed `1000 / 1000` included and `$7.90 / $250` on-demand at the same moment. The small included-requests delta is polling lag.

## Risks and failure behavior

- Every new endpoint call is `.catch(() => null)`. A failed fetch can only remove a window or produce the same "No active Cursor subscription" error as before; it can never render fake values.
- `/auth/usage`, team-scoped `GetHardLimit`, `GetMe`, and `GetTeamSpend` are undocumented endpoints (same stability class as the ones already in use). If their shape changes, Enterprise accounts degrade to fewer windows or the error state, not to wrong numbers.
- `GetTeamSpend` returns all team members and the match is a linear `userId` scan; acceptable at quota-polling frequency, in both runtimes.
- Rollback is reverting the single commit.


btriapitsyn (changes requested):
Thanks for tracing the Cursor Enterprise endpoints and testing them against a real account. Two changes are still required before this can merge:

1. In `packages/web/server/lib/quota/providers/cursor.js`, do not turn missing Enterprise usage into successful empty data or `$0` spend. `planName` alone currently permits `ok: true` with no windows, and failures from `GetMe` or `GetTeamSpend` are displayed as zero spend. Return a distinct unavailable/error result or omit the unverified meter instead. Add cases to `packages/web/server/lib/quota/providers/cursor.test.js` proving that a plan-only response is not successful empty usage and that spend-fetch failure never renders `$0` as authoritative.
2. In `packages/vscode/src/quotaProviders.ts`, add the same Enterprise fallback and focused tests. The extension currently still reads only `GetCurrentPeriodUsage.planUsage`, so #3210 remains live there. Done means Enterprise included requests and per-user on-demand usage follow the same semantics and failure behavior as the web provider.

The required checks are currently green. Please rerun the focused quota tests plus the web and VS Code type-check/lint jobs after these changes.

carvalab:
Both items are in the squashed commit now pushed (`cc197abe8`).

**1. Web provider — no unverified data renders as real.** The success gate is now `!windows.billing_cycle && !windows.on_demand`, so a plan-name-only response returns the `No active Cursor subscription` error instead of `ok` with empty windows. The `on_demand` window renders only when the per-user cap and a matched `spendCents` are both present; a failed `GetMe` or `GetTeamSpend` (or no matching member) omits the window instead of rendering `$0`. Verified partial results still render what they prove — e.g. `billing_cycle` alone while spend can't be verified. New tests: `does not report success from plan name alone` and `omits the on-demand window when spend cannot be verified`; `cursor.test.js` is 8/8.

**2. VS Code extension — same fallback, same semantics.** `fetchCursorQuota` now runs the full chain (`full_stripe_profile` → `auth/usage` → team-scoped `GetHardLimit` → `GetMe` → `GetTeamSpend`) with the same dynamic bucket scan (skip `startOfMonth`, highest `maxRequestUsage`), `userId` matching, verified-meter-only windows, and error strings. Six focused tests cover the Pro path (asserting no fallback endpoints fire), the Enterprise chain including the team-scoped request body, the spend match, bucket selection, plan-only, and the disabled subscription. The module doc's sync note now describes actual parity instead of a declared follow-up.

Validation per your request: focused quota tests (web 8/8, VS Code 26/26 files), `bun run type-check` and `bun run lint` pass in both packages. The web full suite keeps only the 5 failures verified identical on clean `main` (environment-dependent git hunk/walkthrough tests). The CHANGELOG entry stays out of the branch per the conflict concern above; exact merge-time line is in the PR description.

carvalab:
Thanks for the detailed review, both points are addressed in this update.

On the web provider, a plan-only response now returns an error instead of ok with empty windows, and the on_demand window only renders when both the cap and the spend are verified. If GetMe or GetTeamSpend fails the window is left out, so a failure never shows up as $0 spend. cursor.test.js covers both cases.

The VS Code extension got the same fallback in fetchCursorQuota with the same failure behavior, plus focused tests in quotaProviders.test.ts, so #3210 is fixed there too.

I also rebased onto current main and squashed to one commit. Checks I ran: web cursor tests 8/8, VS Code quota tests 131/131, type-check clean in both packages, eslint clean. I verified the Enterprise path against a real account.

Let me know if you want anything else changed, happy to do it.

Issue #3200, opened 2026-08-28
[Bug] Sidebar search: chat sessions disappear when searching, match count always shows 0


Rodneyli:
Thanks for the great work on 1.21.0! I ran into two small issues with the new sidebar session search on Desktop (Windows 11, v1.21.0).

### 1. Chat sessions disappear while searching

When I type anything into the sidebar search box — even an exact substring of a visible chat title — all my managed chats (sessions under `~/.config/openchamber/chats/...`) disappear from the list.

It looks like the `groupSearchDataByGroup` map is only populated for project-section groups, and the managed-chats group (`id: "managed-chats"`) is never added to it. During a search the group renderer falls back to `filteredNodes ?? []`, which is empty for this group, so the whole list renders as empty.

### 2. Match count always shows "0 matches"

The sidebar header seems to receive `searchMatchCount` as a hardcoded `0`, so the label always reads "0 matches" no matter how many sessions actually match.

### Steps to reproduce
1. Create a chat via "New session" so it appears in the sidebar.
2. Search for any text from that chat's title.
3. The chat disappears from the list, and the counter shows "0 matches".

### Expected behavior
- Chats matching the query stay visible.
- The match count reflects the actual results (or is hidden until wired up).

### Extra note
The matcher itself seems fine — I tested the search functions extracted from the shipped bundle against 200 real session titles (Chinese and English substrings, multi-word) and everything matched correctly. So this looks like a data-wiring issue rather than a matching issue.

Thanks for looking into it!

PR #3212, opened 2026-08-28
fix(sidebar): search the managed chats along with project sessions

yulia-ivashko:
## Intent

Typing anything into the sidebar session search made every chat that does not belong to a project disappear from the list, even when the query was an exact substring of a visible chat title.

The sidebar computes per-group search data — filtered nodes, match counts — and while a query is active every group renders `searchData?.filteredNodes ?? []` instead of its own sessions (`SessionGroupSection.tsx:386`). That data was only ever built for groups inside project sections (`useSessionSidebarSections.ts`). The managed chats render outside any project section, were never visited by that pass, and so fell through to the empty fallback. The header count skipped them for the same reason.

The chats group is now passed into the sections hook as a standalone group, so it takes the same search pass and joins the match total. It is built before the hook runs, which keeps search-data ownership in one place rather than having the rendering component write entries into the hook's map.

The report's second half — "0 matches" always shown in the header — is fixed here too. `SessionSidebar` passed the header a literal `0`. The count exists (the sections hook computes it) but only the session list can see it, while the header renders above the list, so the list now reports it upwards and resets it when it unmounts.

## Non-goals

- The matcher is untouched. The reporter had already verified matching works against real session titles, and this change confirms it: the tests run the real `useSessionGrouping` callbacks, not a stub.
- No change to how project sections, worktree grouping, flat display mode, or archived buckets search.
- No CHANGELOG entry. Left out deliberately at the maintainer's request so this branch does not conflict with the other open PRs touching `[Unreleased]`; the entry is to be written at release time. The change is user-facing and would otherwise earn one.

## Affected surfaces

- `packages/ui` — sidebar session search. Reaches Web, Desktop and mobile.
- **Not VS Code.** The extension never renders this group: `chatGroup` returns `null` when `topology.isVSCode` (`SessionProjectCollection.tsx`), because managed chats are not a VS Code surface. No entry in the extension changelog for that reason.
- No persisted data, no API, no stored format involved.

## Repository guidance

|---|---|---|
| `.agents/skills/openchamber-change-discipline` | Shared UI source change | The group already existed and already rendered; the fix adds it to an existing pass rather than introducing a parallel search path. The shared body was extracted into one local function so project and standalone groups cannot drift apart |
| `.agents/skills/performance-engineering` | Sidebar render and search are hot paths | The added work is one group inside a memo that returns early unless a query is active, so idle and typing-free rendering is unchanged. `standaloneGroups` is memoized and falls back to a module-level constant when there is no chats group, so the memo's dependency identity is stable |
| `AGENTS.md` → Correctness Invariants | "Never let a missing input masquerade as an empty result" | The empty list users saw was exactly that: absent search data rendered as "nothing matched". The group now either has data or the search pass is off for everyone |
| `.agents/skills/changelog-authoring` | User-facing fix | No entry in this branch by maintainer decision (conflict avoidance, see Non-goals). The VS Code reachability check above independently rules out an extension entry: that surface never renders this group |
| `packages/ui/src/components/session/sidebar/DOCUMENTATION.md` | Nearest owning doc | Read; it describes grouping and rendering structure, which this change does not alter, so no doc update was needed |

## Validation

| Check | Result |
|---|---|
| New `useSessionSidebarSections.test.tsx` | 4/4 pass; 3 of them fail on the pre-fix code (verified by disabling the new line) |
| `packages/ui` isolated runner, `src/components/session` | 30/30 files pass |
| `bun run type-check` | Pass, all workspaces |
| `bun run lint` | Pass, all workspaces |
| `bunx oxlint` + `eslint` on changed files | Clean |
| `bun run dead-code` | Nothing flagged for the new file or export |
| `bun run test` | Pass: ui, vscode, electron, web 165 files / 1683 tests |
| `bun run build` | Succeeds |

The tests cover a matching chat staying in the list, the match count including chats, a query that matches nothing, and no search pass running when the query is empty.

**Manual, against a production build of this branch (macOS, `packages/web` server serving its own `dist`):**

- `unauthorized` — matches one chat and nothing else: the chat stays visible, header reads `1 match`, no empty state. This is the issue's own reproduction and it failed until `75579ebd`.
- `Restart to update` — matches one chat and one project session: both visible, header reads `2 matches` instead of `0 matches`.
- `qqqqzzz` — matches nothing: empty state, `0 matches`.

Worth knowing for anyone else verifying this by hand: `OPENCHAMBER_DIST_DIR` was set in the shell to the installed desktop app's assets, so the local server served that UI and no local change appeared at all. Unsetting it was required before any of this could be observed.

**Not covered by a test:** the header wiring itself (a prop passed from `SessionSidebar` down to `SidebarHeader`). The count computation is unit-tested; the plumbing is verified manually only.

## Visual evidence

Not attached. This is a user-visible change and the contract asks for screenshots; image capture was unavailable in the environment this was written in, so the three observed states are reported under Validation instead. Reproducing them takes one search box: `unauthorized` before and after `75579ebd`.

## Risks and failure behavior

- **Idle cost unchanged.** The search-data memo still returns immediately when no query is active, so nothing new runs while the user is not searching.
- **One extra group per keystroke while searching**, against the existing per-project-group work. Its cost is the same filter the group would run anyway, over the chats the sidebar already holds in memory.
- **Reference stability preserved.** Without a chats group the hook receives a module-level constant, so the memo does not invalidate on every render.
- **Match count can now be higher than before** for the same query — that is the fix, not a regression: it previously omitted results that were visible on screen.
- **No behavior change outside an active search.** Every code path added is inside a query-active branch.
- **Header trails the list by one frame** per keystroke, because the count is reported through an effect after the list renders. Cosmetic; the alternative — computing it in the header — would duplicate ownership of what matched.

### DIFF
--- a/packages/ui/src/components/session/SessionSidebar.tsx
+++ b/packages/ui/src/components/session/SessionSidebar.tsx
@@ -77,6 +77,8 @@ const SessionSidebarComponent: React.FC<SessionSidebarProps> = ({
   const { t } = useI18n();
   const [isSessionSearchOpen, setIsSessionSearchOpen] = React.useState(false);
   const [sessionSearchQuery, setSessionSearchQuery] = React.useState('');
+  // Reported by the session list below: the header cannot see what matched.
+  const [searchMatchCount, setSearchMatchCount] = React.useState(0);
   const sessionSearchContainerRef = React.useRef<HTMLDivElement | null>(null);
   const sessionSearchInputRef = React.useRef<HTMLInputElement | null>(null);
   const [editingProjectDialogId, setEditingProjectDialogId] = React.useState<string | null>(null);
@@ -631,7 +633,7 @@ const SessionSidebarComponent: React.FC<SessionSidebarProps> = ({
         sessionSearchQuery={sessionSearchQuery}
         setSessionSearchQuery={setSessionSearchQuery}
         hasSessionSearchQuery={hasSessionSearchQuery}
-        searchMatchCount={0}
+        searchMatchCount={searchMatchCount}
         collapseAllProjects={projectView.actions.collapseAllProjects}
         expandAllProjects={projectView.actions.expandAllProjects}
       />
@@ -668,6 +670,7 @@ const SessionSidebarComponent: React.FC<SessionSidebarProps> = ({
           isWorktreeTopologyLoading,
           unresolvedWorktreeProjectPaths,
           projectView: projectView.state,
+          onSearchMatchCountChange: setSearchMatchCount,
         }}
         actions={{
           rowActions: {
--- a/packages/ui/src/components/session/sidebar/list/SessionProjectCollection.tsx
+++ b/packages/ui/src/components/session/sidebar/list/SessionProjectCollection.tsx
@@ -88,6 +88,12 @@ type SessionProjectCollectionProps = {
     isWorktreeTopologyLoading: boolean;
     unresolvedWorktreeProjectPaths: ReadonlySet<string>;
     projectView: ReturnType<typeof useSessionProjectViewState>['state'];
+    /**
+     * The match count belongs in the sidebar header, which renders above this
+     * list, while only the list knows what matched. Reported upwards rather
+     * than recomputed there, so the number and the rows can never disagree.
+     */
+    onSearchMatchCountChange: (count: number) => void;
   };
   actions: {
     rowActions: {
@@ -219,7 +225,7 @@ const VisibleSessionProjects: React.FC<SessionProjectCollectionProps> = ({ topol
     () => chatGroup ? [chatGroup] : EMPTY_STANDALONE_GROUPS,
     [chatGroup],
   );
-  const { projectSections, groupSearchDataByGroup, sectionsForRender, flatSectionsForRender } = useSessionSidebarSections({
+  const { projectSections, groupSearchDataByGroup, sectionsForRender, flatSectionsForRender, searchMatchCount } = useSessionSidebarSections({
     normalizedProjects: topology.projects,
     getSessionsForProject,
     getArchivedSessionsForProject,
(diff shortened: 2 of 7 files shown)


yulia-ivashko:
Both findings are real. Fixed the blocker in `75579ebd`.

**1. Chat-only matches were still hidden.** Confirmed exactly as described: the chats render inside the scroller's `topContent`, and the `sectionsForRender.length === 0` branch replaced the whole list with the empty state. A query matching only a chat showed "No matching sessions" while the header claimed a match. My earlier manual check used `Restart to update`, which also matched a project session, so it never entered that branch — a fair catch.

The scroller now receives whether its top content holds results and keeps it on screen in that branch. The decision lives in `sessionProjectRender.ts` next to the other render-selection helpers.

On the suggested render-path test: `SessionProjectScroller` cannot be imported under `bun test` — its module graph reaches `import.meta.glob`, which only exists under Vite, so a mounting test fails at import. That is why this package tests these helpers as pure functions. The branch decision is now covered there (`SessionProjectScroller.test.ts`), and the scenario is verified by hand below.

Verified against a production build of this branch, with `OPENCHAMBER_DIST_DIR` unset so the local server serves this branch rather than the installed desktop app:

- query `unauthorized`, which matches one chat and nothing else: the chats section shows `Дослідження unauthorized після sync в OpenChamber`, header reads `1 match`, no empty state. Before this commit the same query showed "No matching sessions" with the chat hidden.
- query `qqqqzzz`, matching nothing: `0 matches` and the empty state, unchanged.
- query `Restart to update`, matching one chat and one project session: both visible, header reads `2 matches`.

**2. Visual evidence.** Accepted as a gap. I could not attach screenshots: image capture is unavailable in my environment, and I would rather say that than pass off a text transcript as visual evidence. The three states above are what a reviewer would see; @yulia-ivashko, if you want them in the PR, they are two screenshots away — search `unauthorized` before and after this commit.

On the residual risk you noted: the header count trails the list by one frame per keystroke, since it reports through an effect. Agreed that it is cosmetic, and I left it alone rather than moving the computation up into the header, which would duplicate ownership of what matched.

yulia-ivashko:
Visual proof:
chats   sessions

Before:
before-fix

After fix:
after-fix

btriapitsyn:
Thanks for following the chat-only result through the scroller branch and adding current visual proof. This now fixes both sides of the report without introducing a parallel search path.

Closes #3200.

PR #2825, opened 2026-08-11
fix(ui): reveal overlay scrollbar on hover or active scroll

sergiofspedro:
## Intent

Follow-up to #2581. That PR made the overlay scrollbar thumb permanently visible on desktop runtimes (Electron + VS Code webview) to fix the missing scroll affordance in Settings and elsewhere. It was merged, then reverted (`cc6c5db3`) because — per @makeittech's comment on #2581 — scrollbars shouldn't be unconditionally always-visible. The ask was for a hover/active-scroll-triggered version instead, so this PR implements that.

## Approach

Same shared-component fix location as #2581 (`OverlayScrollbar.tsx` + `index.css`), same two files, but a different trigger model:

- **`OverlayScrollbar.tsx`**: add `mouseenter`/`mouseleave` listeners on the scroll container itself (not just the thumb, which already had hover-to-keep-visible logic for drag interactions). Hovering the container shows the thumb immediately and cancels any pending auto-hide; leaving schedules the existing auto-hide timer. This reuses the existing `isHoveringRef`/`scheduleHide` plumbing that already exists for thumb-hover-during-drag — just extends it to the whole container.
- Not gated by `userIntentOnly` or a `desktop-runtime` class check: hovering directly over a scrollable area is itself deliberate intent, on any platform with a pointer (mouse-based web, Electron, VS Code webview all get it; touch-only mobile simply never fires these events, so it's inert there). This is simpler than #2581's platform-gating because the always-visible behavior that motivated that gating no longer exists.
- **`index.css`**: removed the Settings-specific `display: none` hide rule entirely (#2581 had scoped it to non-desktop; now that the thumb is never permanently visible anywhere, there's no reason to special-case Settings — it gets the same hover/scroll-reveal behavior as every other `ScrollableOverlay` consumer).
- **Fix-up (`8967ccb4`)**: the container is a DOM *sibling* of the thumb (absolutely-positioned overlay), not its parent, so moving the pointer from the container onto the thumb fires the container's `mouseleave` (arming the hide timer) *before* the thumb's `mouseenter` (which would otherwise cancel it). The original hide-timer callback only checked hover state at *schedule* time. Fixed by re-checking `isHoveringRef`/`isHoveringContainerRef` at *fire* time too, inside the existing `setTimeout` callback — so a pointer that has since landed on the thumb cancels the hide instead of the thumb vanishing out from under it.

## Non-goals

[...]

PR #3219, opened 2026-08-28
fix(ui): reveal overlay scrollbar on hover or active scroll (re-port to bindScrollbar)

sergiofspedro:
## Intent

Keep Sergio's hover-to-reveal scrollbar work from #2825 on the current `bindScrollbar` implementation. This follows #2581, whose unconditional desktop visibility was reverted.

Moving a mouse into a scrollable area reveals its thumb without requiring wheel input first. Moving out starts the existing hide delay. Crossing onto the thumb keeps it available for dragging. Touch and pen contact do not count as mouse hover.

The maintainer integrated current `main` in `3176e724c`, preserving the contributor's commits. The device-local **Always show scrollbars** preference already on `main` remains authoritative. Enabled means visible whenever content overflows, including during chat auto-follow. Disabled means hover/active-scroll visibility, with the existing programmatic-follow suppression retained.

Also use `ScrollableOverlay` in the remaining panels and dialogs changed here: sidebar shell, session context, session switcher, help, directory picker, GitHub pickers/integration, worktree form, and Settings content wrappers. Settings navigation already uses it on `main`.

## Non-goals

- No wheel interception, wheel acceleration, or change to chat auto-follow motion.
- No new setting or persistence format in this PR; the always-visible preference comes from `main`.
- No native shell, API, authentication, or dependency changes.
- No per-PR recording infrastructure. The proposed recording workflow, script, and generated GIF were removed during review. The script checked the thumb's geometry flag rather than rendered visibility, altered page content, and could report success without proving hover behavior.

## Affected surfaces

- Shared UI: scrollbar interaction and the listed scroll-container replacements.
- Web and the main Electron window: shared panels, dialogs, Settings, and chat.
- VS Code: shared chat, session switcher, Settings, and reachable session dialogs. It does not mount the desktop `Sidebar`, session context panel, or `HelpDialog`.
- Hosted mobile and Capacitor: shared chat, Settings, and reachable session dialogs. Mouse hover is inert for touch/pen contact; existing scroll-triggered visibility remains. Their dedicated mobile sidebar/switcher components are not replaced.
- Electron mini-chat: shared chat scrollbar and session switcher.

## Repository guidance

| Guidance | Application |
|---|---|
| `AGENTS.md`, `openchamber-change-discipline` | Preserve current-main behavior and the contributor's history; limit the final delta to UI interactions and affected tests. No dependencies or changelog edits. |
| `theme-system` | Reuse `ScrollableOverlay`, existing thumb styles and opacity transitions; no new colors or icons. |
| `settings-ui-patterns` and layout/control/search references | Keep existing Settings controls and search anchors. Preserve vertical-only overflow in Settings wrappers. |
| `performance-engineering`, `scripts/perf/DOCUMENTATION.md` | Keep scroll work coalesced, with no geometry reads or React commits during steady scroll. Browser checks use production-built components and the existing CDP helpers. This is not a claim of application-wide performance improvement. |
| `pr-review`, `CONTRIBUTING.md`, PR template | Review the actual diff and previous requests; distinguish component checks from full native-runtime validation. |
| `locale-ui-patterns`, `communication-style` | No new UI strings. Remove duplicated Turkish keys introduced by the old branch because current `main` already supplies them. |
| Settings/store docs and web/VS Code package docs | Preserve the existing device-local preference and runtime ownership. |

## Validation

Results for `3176e724c`:

| Check | Result |
|---|---|
| `bun run test` | Passed: scripts 3/3 files, UI 414/414, VS Code 40/40, Electron 20/20; web 186 files with 2,081 tests passed and one skipped. |
| `bun test ./packages/ui/src/components/ui/OverlayScrollbar.test.tsx` | 22 passed. Four added regression cases failed before the corresponding integration fixes and pass afterwards. The handoff test now waits past the hide deadline and verifies hiding after thumb exit. |
| `bun run type-check` | Passed across workspaces. UI type-check also rerun after the test-environment fixes. |
| `bun run lint` | Passed with one pre-existing hooks warning in `MobileChangesSurface.tsx`. |
| `bunx oxlint` on `OverlayScrollbar.tsx` and its test | Passed. A broader scan of converted files reports pre-existing findings outside the changed code. The newly introduced double ref assertion was removed. |
| `bun run dead-code` | Completed and reviewed; existing repository findings remain, with no newly introduced public exports. |
| `bun run build:web` | Passed, including web/mobile/mini-chat entries; bundle-size warnings remain. |
| `bun run vscode:build` | Passed; existing bundle-size and mixed-import warnings remain. |
| `git diff origin/main --check` | Passed. |
| Chrome via CDP against an isolated production build of the actual shared components and stylesheet | Passed rendered hover/leave visibility, hide-deadline handoff, programmatic scroll while hovered, always-visible override, suppression, real thumb drag, native wheel movement, and touch-tap inertness. |

The full-suite run initially exposed two test-environment problems: a partial desktop-module replacement omitted exports, and the worktree-dialog DOM setup lacked animation-frame APIs. The replacement was removed and the DOM setup now installs/restores those browser APIs. The full suite passed after both fixes.

Not verified interactively: packaged Electron, VS Code host, iOS/Android devices, or every converted dialog with live data. No benchmark claim is made for those runtimes.

## Visual evidence

The author's earlier recording remains in the PR discussion as historical context, not evidence for the integrated HEAD. The repository no longer carries the generated GIF or its workflow.

Current checks captured the production-built shared scrollbar in Chrome and inspected its actual computed opacity, display, geometry, and scroll position. A first fixture attempt lacked the repository's Tailwind processing; it was corrected before accepting the results. Screenshots were retained locally, not attached here. There is no fresh end-to-end application recording for this HEAD.

## Risks and failure behavior

- The shared component affects every mounted consumer, so hover must remain mouse-only and must not interfere with native wheel/touch scrolling.
- Hover reveal refreshes thumb geometry and position, including after hidden programmatic scrolling. Steady scroll still uses cached geometry.
- Clearing programmatic suppression under a stationary pointer restores hover visibility. Disabling persistent visibility while hovered retains the normal hover behavior; leaving restores auto-hide.
- Always-visible takes precedence over suppression and hide timers. Non-overflowing axes remain hidden.
- Replaced/unmounted containers disconnect listeners, observers, and scheduled work. Current-main cleanup and container-rebinding behavior remain intact.
- No persisted-data migration, remote mutation, or destructive operation is introduced. Reverting the PR restores the previous hover/container behavior without losing the existing scrollbar preference.


sergiofspedro:
Addressed all three findings in the new HEAD (`4f85cd28`):

**1. Evidence gap (visual):** I've updated the PR body with the full template and linked the reference screenshots from the original #2825. The mechanism differs from the old `mouseenter` build, so a fresh recording from this HEAD is the remaining ask — will attach one.

**2. Touch claim (fixed):** The reviewer is right — the Pointer Events spec fires `pointerenter`/`pointerleave` for touch too, so every tap would flash the thumb for `hideDelayMs`. `onContainerPointerEnter`/`Leave` now return early unless `event.pointerType === "mouse"`, making the hover reveal genuinely inert on touch/pen. A regression test asserts a `pointerType: "touch"` enter does not reveal the thumb.

**3. Untested race + leave-to-hide (fixed):** added two tests:
- `keeps the thumb visible while the pointer crosses from container onto the thumb` — container `pointerenter`, container `pointerleave`, then thumb `pointerover`; asserts visibility survives (the fire-time re-check this PR exists for).
- `hides the thumb after the pointer leaves the container` — asserts the auto-hide path fires on `pointerleave`.

btriapitsyn (changes requested):
Thanks for addressing the touch behavior and adding the regression coverage. One item remains before this can merge:

**`packages/ui/src/components/ui/OverlayScrollbar.test.tsx` — fix the failing leave-to-hide test.** The required `checks / Tests` job currently fails because `hides the thumb after the pointer leaves the container` still reads `data-visible="true"` after `pointerleave`. Make the test wait for the actual hide-timer boundary, or fix the implementation if the failure reflects runtime behavior. Done means this regression test passes reliably and the required CI check is green.

sergiofspedro:
Request considered in PR code

> Thanks for addressing the touch behavior and adding the regression coverage. One item remains before this can merge:
> 
> **`packages/ui/src/components/ui/OverlayScrollbar.test.tsx` — fix the failing leave-to-hide test.** The required `checks / Tests` job currently fails because `hides the thumb after the pointer leaves the container` still reads `data-visible="true"` after `pointerleave`. Make the test wait for the actual hide-timer boundary, or fix the implementation if the failure reflects runtime behavior. Done means this regression test passes reliably and the required CI check is green.

sergiofspedro:
GIF of new feature here: 

overlayscrollbarhover

PR #3225, opened 2026-08-29
fix(server): parallelize Windows session merge and raise upstream timeout

herjarsa:
## Intent

The Windows home view fetches a global session list plus a per-project session list for every directory registered in \~/.config/openchamber/settings.json\, then merges the results. Two problems made this surface unreliable under load:

- Each per-project fetch had a 10s upstream timeout. The managed OpenCode backend, when sharing its \opencode.db\ with concurrent TUI instances or under heavy storage contention, can take longer than 10s to answer \/session?directory=...\, so the home view returned \504 OpenCode session list timed out\ even though the server was still alive and serving.
- Project directories were fetched one at a time. With three configured projects and three Windows path variants each (\C:\, \C:/\, \/c/...\-style), a worst-case run is nine sequential round-trips on top of the global fetch.

This PR bounds the merge by the slowest project (parallel fetches across directories) and gives each upstream call enough headroom to finish under real-world load.

## Non-goals

- No change to the underlying OpenCode API or to the global session list path.
- No change to SSE endpoints, the readiness gate, or the lifecycle health-check loop.
- Linux/macOS already skip the SessionMerge handler entirely (it is Windows-only via \process.platform === 'win32'\); they are intentionally unaffected.
- The 30s timeout is the per-fetch cap. It does not extend the broader proxy request deadline or the SSE upstream stall window.

## Second pass — fixes after first review

A first pass of this PR (\5194369bd\) was reviewed by the repo's review bot and flagged as BLOCKED with three correctness findings. This second pass addresses all three:

1. **Failed project reads no longer count as successes.** The earlier refactor returned \{ ok: true, dirSessions: [] }\ on the first non-throwing candidate, which incremented \successfulProjectReads\ for any 5xx/parse-error result. The handler then returned 200 with an empty list instead of 504 — fetch failure masquerading as authoritative empty success, contradicting the AGENTS.md correctness invariant. Now only non-null reads count, and the 504 gate fires whenever the global list is null and no project yielded a non-null read.
2. **Per-directory path-spelling union is preserved.** The earlier refactor returned after the first non-throwing candidate, narrowing the union that the base code produced by iterating all three spellings. This pass iterates all three, collects every non-null response, and unions them through the existing \seen\ set. Path spelling retry survives.
3. **CHANGELOG is dropped from this PR.** AGENTS.md marks both changelogs as maintainer-owned release-time work; the maintainer folds entries in at release. The first pass included a contributor-authored bullet — removed.

The only file changed in this pass is \packages/web/server/lib/opencode/proxy.js\.

## Affected surfaces

- \packages/web/server/lib/opencode/proxy.js\ — Windows-only \/api/session\ handler: per-directory fetches now run in parallel via \Promise.all\ while each directory still iterates its three Windows path spellings sequentially, and only non-null reads count toward the 504 gate.

No other package is touched. The shared UI, runtime API contract, OpenCode SDK calls, and Electron shell are not modified. Web/PWA and VS Code extensions share this proxy code, so they benefit automatically; no per-runtime special-casing is needed.

## Repository guidance

|---|---|---|
| \AGENTS.md\ runtime boundaries | Touches the web server proxy that all runtimes use | Change stays inside \packages/web/server/lib/opencode/proxy.js\; no Electron, mobile, or VS-Code-only paths were touched |
| \packages/web/server/lib/opencode/DOCUMENTATION.md\ (proxy contract) | Defines the SessionMerge invariant on Windows | Maintains the existing return shape (\{ global, extra, merged }\-equivalent JSON array) and the \directory=\ short-circuit to the generic proxy |
| \AGENTS.md\ correctness invariant | Reviewer flagged a fetch-failure-masquerading-as-success bug | Now strictly: a non-null read is required for the success counter; 5xx/parse errors stay non-null-false and the 504 gate still fires when every project errors |
| \AGENTS.md\ CHANGELOG policy | Reviewer flagged the contributor-authored bullet | Removed from this PR; the maintainer writes the entry at release |
| \openchamber-change-discipline\ skill | Cross-workspace contract change candidate | Local implementation risk only; one helper refactored, no exported types changed |

## Validation

| Check | Result |
|---|---|
| \
ode --check packages/web/server/lib/opencode/proxy.js\ | exit 0 |
| Vitest \packages/web/server/lib/opencode/proxy.test.js\ and \opencode-proxy.test.js\ | not run from this environment — neither suite exercises the Windows-only merge handler; both use the Linux branch. No platform stub exists in either suite (the repo uses \setPlatform\ elsewhere but not here), so adding tests would require a small fixture first. |
| \un test\ from repo root with full \
ode_modules\ | not run; this checkout has no installed deps in \packages/web\ |
| Manual Windows reproduction with 3 configured projects and slow OpenCode backend | not run from this Linux checkout; logic verified by code reading |

Static checks alone do not prove runtime behavior. A reviewer should re-run the full vitest suite from an installed workspace before merging, and ideally reproduce the slow-backend scenario on Windows. The correctness invariant the bot flagged cannot be exercised without a Windows platform stub in the test suite, which is a follow-up rather than a blocker for this PR.

## Visual evidence

No rendered output changes. The fix only changes server-side fetch timing, concurrency, and accounting; the UI sees the same JSON shape it already consumed.

## Risks and failure behavior

- **Behavior change in timeout:** when a single project fetch hangs longer than 30s, the merge now waits longer before returning 504. Worst-case wall time per project is unchanged from the base (3 spellings × 30s = 90s sequential within a directory), but cross-directory wall time moves from the sum of all projects to the slowest single project. This is an improvement, not a regression.
- **Concurrency:** \Promise.all\ over N projects with N=3 typical. Node \etch\ budget on the http-proxy-middleware agent (256 free sockets) is sized for burst traffic per \proxy.js\; one extra in-flight per project is well inside headroom.
- **Rollback:** revert the two proxy.js commits. The CHANGELOG bullet is no longer in this PR.
- **Cross-runtime:** VS Code and Web/PWA proxy through the same \
egisterOpenCodeProxy\; the SessionMerge handler only runs on Windows, so non-Windows runtimes see no behavior change at all.
- **Security / data loss:** no input shape change, no new endpoint, no persisted state.

### DIFF
--- a/packages/web/server/lib/opencode/proxy.js
+++ b/packages/web/server/lib/opencode/proxy.js
@@ -782,7 +782,7 @@ export const registerOpenCodeProxy = (app, deps) => {
       if (rawUrl.includes('directory=')) return next();
 
       const fetchWindowsSessionList = async (sessionPath) => {
-        const result = await fetchSessionListPayload(sessionPath, { req, timeoutMs: 10000 });
+        const result = await fetchSessionListPayload(sessionPath, { req, timeoutMs: 30000 });
         if (!result.upstream.ok || !Array.isArray(result.payload)) return null;
         return sanitizeSessionListPayload(result.payload);
       };
@@ -809,9 +809,11 @@ export const registerOpenCodeProxy = (app, deps) => {
             .map((session) => (session && typeof session.id === 'string' ? session.id : null))
             .filter((id) => typeof id === 'string')
         );
-        const extraSessions = [];
-        let successfulProjectReads = 0;
-        for (const dir of projectDirs) {
+        // Fetch every project directory in parallel: the managed server can be
+        // slow under heavy storage contention (large opencode.db, concurrent
+        // TUI instances), so serial fetches multiply the wall time. The total
+        // wait is bounded by the slowest project, not the sum.
+        const projectFetches = projectDirs.map(async (dir) => {
           const candidates = Array.from(new Set([
             dir,
             dir.replace(/\\/g, '/'),
@@ -821,17 +823,23 @@ export const registerOpenCodeProxy = (app, deps) => {
             const encoded = encodeURIComponent(candidateDir);
             try {
               const dirSessions = await fetchWindowsSessionList(`/session?directory=${encoded}`);
-              if (dirSessions) {
-                successfulProjectReads += 1;
-              }
-              for (const session of dirSessions || []) {
-                const id = session && typeof session.id === 'string' ? session.id : null;
-                if (id && !seen.has(id)) {
-                  seen.add(id);
-                  extraSessions.push(session);
-                }
-              }
+              return { ok: true, dirSessions: dirSessions || [] };
             } catch {
+              // try the next path variant
+            }
+          }
+          return { ok: false, dirSessions: [] };
+        });
+        const projectResults = await Promise.all(projectFetches);
+        const extraSessions = [];
+        let successfulProjectReads = 0;
+        for (const { ok, dirSessions } of projectResults) {
+          if (ok) successfulProjectReads += 1;
+          for (const session of dirSessions) {
+            const id = session && typeof session.id === 'string' ? session.id : null;
+            if (id && !seen.has(id)) {
+              seen.add(id);
+              extraSessions.push(session);
             }
           }
         }
--- a/packages/web/server/lib/opencode/proxy.js
+++ b/packages/web/server/lib/opencode/proxy.js
@@ -813,37 +813,55 @@ export const registerOpenCodeProxy = (app, deps) => {
         // slow under heavy storage contention (large opencode.db, concurrent
         // TUI instances), so serial fetches multiply the wall time. The total
         // wait is bounded by the slowest project, not the sum.
+        //
+        // Each project tries ALL three Windows path spellings in turn (the
+        // base behavior — different spellings can return different sessions
+        // because path normalization on the server is not always idempotent
+        // across symlinks and drive casing) and unions the results so a single
+        // spelling returning empty does not silently hide sessions that
+        // another spelling would have surfaced. Only non-null reads count
+        // toward the 504 gate; a 500/upstream error on a single spelling must
+        // not masquerade as authoritative empty success. The successful-read
+        // counter increments up to 3 per directory, matching the base
+        // accounting.
         const projectFetches = projectDirs.map(async (dir) => {
           const candidates = Array.from(new Set([
             dir,
             dir.replace(/\\/g, '/'),
             dir.replace(/\//g, '\\'),
           ]));
+          const collected = [];
+          let nonNullReads = 0;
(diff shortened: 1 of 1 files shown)


herjarsa:
Thanks for the review — all three blockers addressed in the latest push (\72459bfea\):

1. **Failed project reads no longer count as successes.** Replaced \{ ok, dirSessions: [] }\ with \{ nonNullReads, collected }\; only non-null reads increment \successfulProjectReads\, and the 504 gate still fires when the global list is null and no project yielded a non-null read. The AGENTS.md correctness invariant is restored.

2. **Per-directory path-spelling union preserved.** Each \projectFetches\ async fn iterates all three Windows path spellings, collects every non-null response, and unions through the existing \seen\ set. Path-spelling retry is intact; the parallelization is only across directories.

3. **CHANGELOG commit dropped.** AGENTS.md marks both changelogs as maintainer-owned release-time work. The contributor-authored bullet is gone from this PR.

Two non-blockers also addressed:

4. **PR body wording corrected.** The previous claim of \~30s\ worst-case was misleading; the actual worst case is N × (3 spellings × 30s) per directory with parallelism across directories. The updated body describes both the within-directory (sequential, 90s) and cross-directory (parallel, slowest wins) shapes honestly.

5. **No regression test added in this PR.** The repo's proxy test suites (\proxy.test.js\, \opencode-proxy.test.js\) only exercise the Linux branches; the Windows-only \/api/session\ handler isn't reached by either. Adding a fixture would require a \setPlatform\ stub plus platform-conditional loading of the merge handler — that's a follow-up rather than something I can land alongside this targeted bugfix without expanding scope.

\
ode --check\ exits 0 on the changed file. CI is the source of truth for the rest; the existing \utomation\, \checks\, and \label\ checks from the first push should re-run cleanly.

Re-requesting review from the original reviewer.

herjarsa:
Closing as superseded by #3837 (move OpenChamber to OpenCode 2.x).

That migration deleted the entire Windows session-merge handler this PR patches (pp.get('/api/session') win32 branch in \packages/web/server/lib/opencode/proxy.js\): OpenCode 2.x lists sessions across directories natively (\V2 lists sessions across directories on every platform and owns pagination\), and \etchSessionListPayload\ is now called without an explicit timeout, so there is no 10s-timeout serial-fetch loop left to parallelize or raise.

The bot's PASS review (Aug 29) was against the pre-2.x base and is stale. If session-list slowness under storage contention resurfaces on 2.x, it deserves a fresh issue against the new architecture, not a rebase of this diff.

herjarsa:
Closing as superseded by #3837 (move OpenChamber to OpenCode 2.x). That migration deleted the entire Windows session-merge handler this PR patches: OpenCode 2.x lists sessions across directories natively, and fetchSessionListPayload is now called without an explicit timeout, so there is no 10s-timeout serial-fetch loop left to parallelize or raise. The bot PASS review (Aug 29) was against the pre-2.x base and is stale.

PR #3229, opened 2026-08-29
fix(web): scope UI session cookie by request port for LAN multi-instance

### LINKED ISSUE #2377: [Bug] Session cookie collision when multiple instances share the same IP address

jlu-lujing:
## Maintainer follow-up

Completed in [6991c1d08](https://github.com/openchamber/openchamber/commit/6991c1d08): CLI retry coverage now checks both cookie-header APIs with bare and port-scoped names, the auth assertions run under Vitest, and the documentation distinguishes session identity from CSRF. Workspace type-check, lint, build and the full test suite passed on integration HEAD `0667e739d`.

## Intent

Two OpenChamber instances running on the same LAN address but different ports (e.g. `http://192.168.0.1:3000` and `http://192.168.0.1:3001`) share a single `oc_ui_session` cookie, because browsers key cookie jars on the host and ignore the port (RFC 6265). Logging into the second instance silently overwrites the first's cookie, so refreshing **any** tab drops to the login screen and session create/send fail with "Failed to start session" / notification/session identity lookup failure. This folds the request port into the session cookie name (`oc_ui_session_<port>`) so each instance owns its own cookie slot; a host without an explicit port keeps the bare `oc_ui_session` name.

Closes #2377.

## Non-goals

- Same-host loopback instances on different ports receive separate cookie names too. A host without an explicit port keeps the historical bare name.
- Bearer-token validation and pairing are unchanged. Browser passkey login calls `issueSession` and receives the renamed session cookie.
- Does not migrate or honor an existing `oc_ui_session` cookie after upgrade — see Risks.

## Affected surfaces

- `packages/web/server/lib/ui-auth/ui-auth.js` — Set-Cookie / read / clear now derive the name per request.
- `packages/web/server/lib/ui-auth/session-cookie.js` (new) — single resolver shared by the set side, `ui-auth`'s own cookie read, and the notification/session identity read side, so the three can't drift.
- `packages/web/server/lib/security/request-security.js` — `getUiSessionTokenFromRequest` matches the exact slot for the request's own port.
- `packages/web/bin/lib/cli-http.js` — CLI session-cookie extraction accepts the per-port variant.
- Runtimes: Web/Desktop/Mobile browser sessions reached over HTTP. Desktop/VS Code/relay clients that authenticate with a bearer token or `oc_url_token` are unaffected (they don't read this cookie). WebSocket/SSE transports carry the cookie header through unchanged, so origin / URL-token minting / allowlists are untouched.
- Persisted/external contract: cookie name only; nothing on disk or in any stored file changes.

## Repository guidance

|---|---|---|
| `openchamber-change-discipline` | Server-side cookie/session contract + a new source file + notification/session identity consumer. | Smallest complete change; traced every `oc_ui_session` consumer (grep: set side, `ui-auth` read, notification/session identity read, CLI) and updated each to the same resolution; kept set/read symmetric (no divergence); updated the owning `DOCUMENTATION.md`; no changelog touch; removed a helper that became unused rather than leave it as dead surface. |
| `ui-auth/DOCUMENTATION.md` | Owns the cookie/session issuance truth. | Added a "Session cookie scoping" section documenting the port-folding rule, the shared-resolver invariant, and the upgrade re-login compatibility note. |
| `relay-transport` | Session cookie crosses SSE/WS. | Reviewed; the cookie rides existing request headers and the change does not touch origin handling, WS URL-token minting, or allowlists, so relay is unchanged. |

## Validation

| Check | Result |
|---|---|
| `bun test packages/web/server/lib/ui-auth/ packages/web/server/lib/security/ packages/web/bin/cli.test.js` | 111 pass / 0 fail. Includes: new `session-cookie.test.js` resolver cases; a notification/session identity case where a request on `:3001` with both port cookies present reads its **own** slot; a **cross-port isolation** case proving a request on `:3001` with only `oc_ui_session_3000` returns **null** (never borrows another instance's session); and `ui-auth` integration cases driving the real `handleSessionCreate`/`handleSessionStatus`. |
| `bun run type-check:web` (`tsc --noEmit`) | clean |
| `bun run dead-code` | new module is consumed (not reported dead) |
| Real two-instance browser repro over a LAN IP | **Not run** — needs two servers + a real browser on one LAN IP, unavailable here. Covered by integration tests through the real auth/notification/session identity code paths (not mocks). |

## Visual evidence

No rendered UI change: the fix is server-side cookie naming. The observable outcome (both tabs stay signed in) is a browser auth behavior, demonstrated by the regression tests rather than a rendered state.

## Risks and failure behavior

- **Compatibility (intended):** after upgrade, any explicit-port host has a renamed cookie, so currently-signed-in browser sessions must sign in once again. The notification/session identity read is exact-match (no cross-port or bare-name borrowing), so a stale cookie is simply ignored rather than causing a split-brain "logged in on some routes, out on others" state.
- **Security:** when multiple port cookies share a jar, both `ui-auth` and the notification/session identity path read only the slot for the request's own port, so one instance cannot consume another's session. Token verification (`jwtVerify`) is unchanged.
- **Failure behavior:** a missing/mismatched/malformed cookie yields `null` (unauthenticated), exactly as the bare-name path did before.

### DIFF
--- a/packages/web/bin/lib/cli-http.js
+++ b/packages/web/bin/lib/cli-http.js
@@ -3,6 +3,10 @@ import { readDesktopLocalClientTokenFromSettings, readDesktopLocalPortFromSettin
 import { getInstanceFilePath, readInstanceOptions } from './cli-process.js';
 
 const UI_SESSION_COOKIE_NAME = 'oc_ui_session';
+// The server folds the request port into the cookie name (oc_ui_session_<port>)
+// so LAN instances on different ports don't share one jar (issue #2377). Accept
+// both the bare name and any per-port variant when extracting.
+const UI_SESSION_COOKIE_PATTERN = new RegExp(`(?:^|,\\s*)(${UI_SESSION_COOKIE_NAME}(?:_\\d+)?=[^;]+)`);
 
 function extractUiSessionCookie(response) {
   const values = [];
@@ -23,7 +27,7 @@ function extractUiSessionCookie(response) {
   }
 
   for (const setCookie of values) {
-    const match = setCookie.match(new RegExp(`(?:^|,\\s*)(${UI_SESSION_COOKIE_NAME}=[^;]+)`));
+    const match = setCookie.match(UI_SESSION_COOKIE_PATTERN);
     if (match?.[1]) return match[1];
   }
   return null;
--- a/packages/web/server/lib/security/request-security.js
+++ b/packages/web/server/lib/security/request-security.js
@@ -1,3 +1,5 @@
+import { isSessionCookieName, sessionCookieNameForRequest } from '../ui-auth/session-cookie.js';
+
 export const createRequestSecurityRuntime = (deps) => {
   const { readSettingsFromDiskMigrated } = deps;
   // Origins of packaged (non-browser) clients whose WebView origin never
@@ -17,20 +19,27 @@ export const createRequestSecurityRuntime = (deps) => {
     if (!cookieHeader || typeof cookieHeader !== 'string') {
       return null;
     }
-    const segments = cookieHeader.split(';');
-    for (const segment of segments) {
-      const [rawName, ...rest] = segment.split('=');
-      const name = rawName?.trim();
-      if (!name) continue;
-      if (name !== 'oc_ui_session') continue;
-      const value = rest.join('=').trim();
+    // This instance's own cookie name for the host:port the request arrived on.
+    const preferred = sessionCookieNameForRequest(req);
+    const decode = (value) => {
       try {
         return decodeURIComponent(value || '');
       } catch {
         return value || null;
       }
+    };
+    let fallback = null;
+    for (const segment of cookieHeader.split(';')) {
+      const [rawName, ...rest] = segment.split('=');
+      const name = rawName?.trim();
+      if (!name || !isSessionCookieName(name)) continue;
+      const value = decode(rest.join('=').trim());
+      if (name === preferred) return value;
+      // Remember any session cookie as a fallback so single-instance and
+      // port-less (e.g. loopback without an explicit port) flows keep working.
+      fallback ??= value;
     }
-    return null;
+    return fallback;
   };
 
   const rejectWebSocketUpgrade = (socket, statusCode, reason) => {
--- a/packages/web/server/lib/security/request-security.js
+++ b/packages/web/server/lib/security/request-security.js
@@ -1,4 +1,4 @@
-import { isSessionCookieName, sessionCookieNameForRequest } from '../ui-auth/session-cookie.js';
+import { sessionCookieNameForRequest } from '../ui-auth/session-cookie.js';
 
 export const createRequestSecurityRuntime = (deps) => {
   const { readSettingsFromDiskMigrated } = deps;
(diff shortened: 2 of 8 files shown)


mattv8:
Thanks for fixing [#2377](https://github.com/openchamber/openchamber/issues/2377). Keeping issuance and every session read on sessionCookieNameForRequest fixes the collision without letting one port consume another port's session.

@btriapitsyn: Safe to merge

btriapitsyn:
Merged, thanks for fixing the cross-port login collision, and thanks @mattv8 for the review! The [CLI coverage and documentation follow-up](https://github.com/openchamber/openchamber/commit/6991c1d08) is also in main, with the full test suite passing.

PR #3230, opened 2026-08-29
fix(web): canonicalize project path case on case-insensitive filesystems

### LINKED ISSUE #1913: [Bug] Project path case mismatch on macOS causes phantom "main" worktree group and "No sessions in this workspace yet" empty state

On macOS (case-insensitive filesystem) a project whose `path` in `settings.json` differs in case from the actual filesystem path produces a buggy sidebar render:

1. The project header shows **"No sessions in this workspace yet"** even though sessions exist.
2. The sessions appear under a **phantom "main" worktree subgroup** instead of directly under the project.

### Concrete repro on my setup

Two projects, both git repos with `main` upstream, both real path is `/Users/Agent/Desktop/...` (capital D):

```json
// /Users/Agent/.config/openchamber/settings.json
"projects": [
  { "label": "PrdDesignMd", "path": "/Users/Agent/desktop/vcfiles/prdDesignMd" },   // ← lowercase d
  { "label": "Hillm",       "path": "/Users/Agent/Desktop/vcfiles/hillm" }            // ← capital D
]
```

Result in sidebar:
- **PrdDesignMd**: `No sessions in this workspace yet.` → then `main` subgroup → 16 sessions inside  
- **Hillm**: 4 sessions listed directly under project, no subgroup, no empty text

After manually editing the project path to capital `Desktop` and restarting, PrdDesignMd renders correctly.

### Root cause analysis

1. opencode stores session `directory` as the actual filesystem path (capital `Desktop` here, from `git rev-parse` / system APIs).
2. OpenChamber stores `project.path` as whatever case was originally registered.
3. Across the codebase, paths are compared via `===` (case-sensitive string compare), e.g. in `useSessionGrouping.ts`:

```ts
if (normalizedDir !== normalizedProjectRoot && worktreeByPath.has(normalizedDir)) return normalizedDir;
if (normalizedDir === normalizedProjectRoot) return normalizedProjectRoot ?? '__project_root__';
return archivedKey;
```

4. With the case mismatch:
   - `normalizedProjectRoot` = `/.../desktop/.../prdDesignMd`
   - `normalizedDir` (from session) = `/.../Desktop/.../prdDesignMd`
   - They are NOT equal → falls through.
   - The localStorage `oc.worktreeMap` also has a stale entry with the capital-D path, so `worktreeByPath.has(normalizedDir)` is true → session is routed to a phantom "main" worktree group.
   - Root group ends up with 0 sessions → empty-state text shows.

5. `oc.worktreeProjectMap` in localStorage actually contains an explicit cross-case mapping, suggesting the codebase already knows about this mismatch but doesn't fully reconcile it:

```json
"/Users/Agent/Desktop/vcfiles/prdDesignMd": "/Users/Agent/desktop/vcfiles/prdDesignMd"
```

### Where does it happen?

Desktop Web (browser-agnostic — the bug is in server-side `settings.json` so all browsers reproduce it identically).

### Version (if known)

v1.13.8 (latest at time of report). macOS 24.6.0 / x64.

### Suggested fix

On macOS, normalize project paths to the canonical case as reported by the filesystem when registering a project (e.g. via `fs.realpath` or `NSURL`). Either:

1. **Strict canonical**: at project-create time, resolve `realpath` and store that.
2. **Tolerant compare**: use case-insensitive path compare on macOS/Windows wherever paths are matched against session directories.

Option 1 is cleaner and matches user expectation (the path shown to the user matches what they typed/picked).

Will add after reproducing in a fresh environment.

### Logs (optional)

N/A — no errors logged.

jlu-lujing:
## Intent

On macOS/Windows (case-insensitive, case-preserving volumes) a project whose `path` in `settings.json` differs in case from the on-disk path never matches the session `directory` that opencode/git report (always the real case). Because the sidebar compares those paths with case-sensitive `===`/`startsWith`/`Map.has`, the sessions get routed to a phantom worktree group and the project shows "No sessions in this workspace yet" even though sessions exist (#1913).

Persisted path normalization now recovers the true on-disk casing, so `project.path` and `session.directory` agree and grouping is correct.

Closes #1913.

## Non-goals

- **Not** a case-insensitive comparison in the sidebar. A tolerant `===` fallback would wrongly unify `/Foo` and `/foo` on a case-sensitive filesystem (Linux), where those are legitimately different directories. Canonicalizing at persistence keeps the real fix on the one platform that is actually case-insensitive.
- Does not change case-sensitive (Linux) behavior at all — the canonicalization is gated to `win32`/`darwin`.
- The transient optimistic entry created by `useProjectsStore.addProject` before the first server round-trip still carries the typed case; it reconciles to the server-canonical value on the next settings sync (same as today). This PR fixes the persisted/loaded state that produces the reported, restart-stable bug.

## Affected surfaces

- `packages/web/server/lib/opencode/settings-normalization-runtime.js` — `safeRealpathSync` now also recovers casing via `readdirSync`; reached by `normalizePathForPersistence`, `sanitizeProjects`, and the `readSettingsFromDiskMigrated` migration (which `GET /api/config/settings` uses) + `persistSettings`.
- `packages/web/server/index.js` — injects `readdirSync: fs.readdirSync` into the runtime.
- Runtimes: Web/Desktop (macOS/Windows) browser sessions over HTTP. Linux (case-sensitive) is explicitly excluded. VS Code / relay clients that authenticate with bearer/`oc_url_token` are unaffected.
- Persisted/external contract: `settings.json` `project.path` (and `lastDirectory`/`homeDirectory`/`pinnedDirectories`) may be rewritten to on-disk casing on the next read; a `changed` migration already writes those back. Project ids are base64(path) and are reconciled by the existing deterministic-id migration.

## Repository guidance

|---|---|---|
| `openchamber-change-discipline` | Persisted data (`settings.json`) + a shared normalization runtime consumed by read-migration and persist. | Smallest complete change at the existing single chokepoint (`safeRealpathSync`); traced every consumer of `normalizePathForPersistence`/`sanitizeProjects`; prefers exact-name match so case-sensitive volumes are never rewritten; no speculative cache added; updated owning `DOCUMENTATION.md`; no changelog touch (maintainer-owned). |

## Validation

| Check | Result |
|---|---|
| `bun test packages/web/server/lib/opencode/settings-normalization-runtime.test.js` | 16 pass / 0 fail (incl. 5 new #1913 cases: recovers lowercase→disk case; already-correct left unchanged; missing component left unchanged; **linux never rewritten**; `normalizeSettingsPaths` flags `changed`) |
| Real macOS end-to-end (this machine, APFS, `process.platform=darwin`, real `fs`) | `normalizeSettingsPaths` on a project path stored `.../myproject/sub` returns `.../MyProject/Sub` (true FS casing) with `changed=true` — i.e. it would be persisted back and match the real-case session directory. |
| Regression scan `bun test packages/web/server/lib/opencode/` | stable **12 fail, identical to clean-main baseline** (all are pre-existing `vi.stubGlobal/vi.setSystemTime/vi.hoisted` unsupported under `bun test`, unrelated to this change). `core-routes.test.js` (the only flaky-differential name) does not use these functions and does not reproduce. |
| `bun run --cwd packages/web type-check` (`tsc --noEmit`) | clean |
| Real running app + real sessions (sidebar grouping before/after) | **Not run** — needs a full desktop build with seeded sessions; the server-side normalization is exercised directly above against the real filesystem. |

## Visual evidence

No rendered-UI code changed; the fix is server-side path normalization. The user-visible outcome (sessions listed under the project instead of a phantom group) follows from `project.path` matching `session.directory` and is demonstrated by the unit + real-filesystem tests above rather than a screenshot.

## Risks and failure behavior

- **Persisted rewrite:** an existing lower-case `project.path` is corrected to disk casing on next read and written back. This is the intended self-heal; because project id is derived from path, the existing deterministic-id migration handles id/key reconciliation. If a path can't be read (missing/permission), `safeRealpathSync` returns the original unchanged — no data loss.
- **Case-sensitive volumes:** gated to `win32`/`darwin`, and within those the walk prefers an exact name, so a directory that legitimately differs only by case is never rewritten.
- **Security/permissions:** read-only `readdir`; no new credential or network surface.

### DIFF
--- a/packages/web/server/index.js
+++ b/packages/web/server/index.js
@@ -236,6 +236,7 @@ const settingsNormalizationRuntime = createSettingsNormalizationRuntime({
   path,
   processLike: process,
   realpathSync: fs.realpathSync,
+  readdirSync: fs.readdirSync,
   tunnelBootstrapTtlDefaultMs: TUNNEL_BOOTSTRAP_TTL_DEFAULT_MS,
   tunnelBootstrapTtlMinMs: TUNNEL_BOOTSTRAP_TTL_MIN_MS,
   tunnelBootstrapTtlMaxMs: TUNNEL_BOOTSTRAP_TTL_MAX_MS,
--- a/packages/web/server/lib/opencode/DOCUMENTATION.md
+++ b/packages/web/server/lib/opencode/DOCUMENTATION.md
@@ -236,6 +236,8 @@ Managed health failures are classified as `timeout`, `connection_refused`, `conn
   - `sanitizeSkillCatalogs(input)`
   - `sanitizeProjects(input)`
 
+Persistence path normalization (`normalizePathForPersistence` / `normalizeSettingsPaths` / `sanitizeProjects`, reached via the `readSettingsFromDiskMigrated` migration and `persistSettings`) resolves symlinks with `realpathSync` and, on case-insensitive filesystems (`win32`, `darwin`), also recovers the on-disk casing of each path component via `readdirSync` (exact-name match preferred so case-sensitive volumes are never rewritten). `realpathSync` alone does not report on-disk casing on these volumes, so without the readdir step a project stored with the wrong case would never match the real-case `directory` opencode reports for its sessions (issue #1913). A corrected path is flagged `changed` and written back to `settings.json`.
+
 ## Public exports (theme-runtime.js)
 - `createThemeRuntime(dependencies)`: creates custom theme runtime for on-disk theme discovery and JSON normalization/validation.
 - Returned API:
--- a/packages/web/server/lib/opencode/settings-normalization-runtime.js
+++ b/packages/web/server/lib/opencode/settings-normalization-runtime.js
@@ -4,6 +4,7 @@ export const createSettingsNormalizationRuntime = (dependencies) => {
     path,
     processLike,
     realpathSync,
+    readdirSync,
     tunnelBootstrapTtlDefaultMs,
     tunnelBootstrapTtlMinMs,
     tunnelBootstrapTtlMaxMs,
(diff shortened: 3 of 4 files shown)


jlu-lujing:
Note for the reviewer queue: since the reviewed HEAD `15207ab` (the one the `PASS` summary above is pinned to) I pushed one **test-only** commit, `363cafa`, adding a win32 case to the readdir walk (`C:\` drive-root + backslash join + drive-letter uppercasing). It addresses the review's non-blocker "win32 branch ships untested" and touches **no production code** (`+28` in `settings-normalization-runtime.test.js` only). `automation` and `label` checks are green on `363cafa`. The perf non-blocker (re-walking components on each read) was left as-is since it only fires on the case-insensitive gate during settings normalization — happy to add a memoized resolver if a maintainer prefers.

mattv8:
Thanks for tracking this down and adding the win32 coverage. The persisted path now gets the on-disk casing before project IDs are derived, so sessions no longer land in a phantom empty group on macOS and Windows. The change fails safe for unreadable or missing paths and leaves Linux alone.

@btriapitsyn This PR is safe to merge.

btriapitsyn (changes requested):
Thanks for adding the Windows coverage, and thanks @mattv8 for reviewing this. I found two path-identity cases that need fixing before merge:

1. In `packages/web/server/lib/opencode/settings-normalization-runtime.js`, `safeRealpathSync` continues into case recovery after `realpathSync` fails. On a case-sensitive macOS volume, a missing `/volume/foo` can therefore become an unrelated existing `/volume/Foo` and trigger a persisted project-ID migration. Preserve the original path when resolution fails; case recovery must not turn a failed lookup into a different existing directory. Add regressions in `settings-normalization-runtime.test.js` for ENOENT with a case-only sibling and for EACCES, asserting that the stored path and project identity remain unchanged.
2. In the same runtime, `canonicalizePathCase` splits macOS paths on backslashes. Use platform-correct path components so a literal `A\B` directory cannot become `A/B`. Add a regression where both exist and assert that normalization preserves the original directory identity.

Update `packages/web/server/lib/opencode/DOCUMENTATION.md` to describe these failure guarantees, and correct the PR's runtime scope: authenticated bearer and relay clients still consume the normalized server settings. Keep the existing case-recovery and Windows tests passing.

PR #3232, opened 2026-08-29
feat(electron): add Debian package to Linux build

mjuchem:
## Intent

Add a Debian package (`.deb`) to the Linux desktop build so Debian/Ubuntu users can install OpenChamber through their package manager instead of only via AppImage. The deb ships alongside the existing AppImage with the same stable `openchamber` identity: executable, `.desktop` entry, icon, and `StartupWMClass` all match.

The deb is validated by a new verifier in CI, and both x64 and arm64 deb packages are now attached to GitHub releases next to the AppImages.

## Non-goals

- In-app auto-update stays AppImage-only. A deb install updates through `apt`; the existing `updater-capability.mjs` behavior and the AppImage-only requirement are unchanged, and the README calls this out.
- The deb is not yet uploaded to a package repository (no `apt` repo/PPA). It is attached to the GitHub release as a downloadable file.
- Windows/macOS packaging is untouched.

## Affected surfaces

- `packages/electron` packaging config (`package.json`), script list, and README.
- Linux CI: `release.yml` (build, artifact upload, release attachment) and `release-desktop-smoke.yml` (build, artifact upload).
- New scripts: `packages/electron/scripts/verify-linux-deb.mjs` and its unit test.
- `verify-linux-appimage.mjs` exports three helpers (`collectFiles`, `defaultCliVersion`, `REQUIRED_NATIVE_MODULES`) for reuse; behavior unchanged.

The deb and AppImage use the same electron-builder config for Linux identity and arch, so only the packaging step and its verifier differ per artifact. Updater, download, runtime, and native code are unaffected.

## Repository guidance

The Linux desktop build and its verifiers are owned under the Electron package, so `openchamber-change-discipline`, `desktop-shell`, and `locale-ui-patterns` (for the README copy) apply. No changelog entry: both `CHANGELOG.md` files are release-time only and stay untouched.

|---|---|---|
| `openchamber-change-discipline` | Packaging config, build target, exports, and package metadata changed | Kept the change scoped to the packaging/verification path; reused existing verifier helpers instead of duplicating native-module and ELF logic; left unrelated worktree changes alone |
| `desktop-shell` | Electron packaging and native artifact identity | Deb inherits the same stable Linux identity and arch handling as the AppImage; `license`, `homepage`, and a maintainer are set so `apt show` renders correct metadata |
| `locale-ui-patterns` | README copy about install and update behavior | Documented that deb installs update via apt and the in-app updater still requires AppImage, in plain user-facing terms |

## Validation

`test:architecture` (which now includes `verify-linux-deb.test.mjs`) passes 44/44. `test:updater` passes 18/18, re-run after the change. No dead-code findings for the new files.

A real x64 build was run locally: electron-builder produced `dist/openchamber_1.21.1_amd64.deb`, `verify:linux-deb` and `verify:linux-appimage` both passed against it, and the control file carries `Package: openchamber`, `License: MIT`, `Maintainer: OpenChamber <security@openchamber.dev>`, and the homepage. The verifier confirmed OpenCode CLI 1.18.25 and 4 packaged native modules with the expected ELF architecture.

| Check | Result |
|---|---|
| `test:architecture` (includes deb verifier suite) | 44/44 pass |
| `test:updater` | 18/18 pass |
| Real x64 deb build + `verify:linux-deb` | pass (OpenCode CLI 1.18.25, 4 native modules) |
| Real x64 `verify:linux-appimage` | pass |
| `verify:linux-deb` control metadata | `openchamber` pkg, MIT license, correct maintainer/homepage |
| arm64 deb build + verify | not run (no arm64 host); config and arch mapping mirror the existing AppImage path |

## Visual evidence

No rendered UI changes. The diff touches packaging config, CI, verifier scripts, and README prose only, so there is nothing to screenshot.

## Risks and failure behavior

- The deb is a new artifact in the Linux build; if electron-builder or FPM misconfigures it, the Linux build/verify step fails and the release is blocked, which is the intended safety behavior. The deb is verified in CI before upload with `if-no-files-found: error`.
- Keeping the two auto-update channels distinct (deb via apt, in-app via AppImage) is a deliberate behavior split. A user who installs the deb and triggers the in-app updater gets the existing AppImage-only error. This is documented rather than hidden.
- arm64 deb is untested on real hardware; the arch mapping (`x64 -> amd64`, `arm64 -> arm64`) and target selection match the existing AppImage config, so the risk is confined to the untested build path, identified here rather than hidden.
- No secrets or credentials were introduced; the only new metadata is the public `security@openchamber.dev` maintainer contact already documented in the repo.


erfanium (approved):

erfanium (approved):
Approved. the .deb builds and installs cleanly with electron-builder and passes all verifiers (`verify:linux-deb`, `verify:linux-appimage`, `verify:opencode-cli:packaged`).

btriapitsyn (changes requested):
Thanks for the Debian packaging and verifier work, and thanks @mattv8 for identifying this as the PR to keep. There's one release-path blocker to fix.

In `packages/electron/package.json`, set `build.deb.publish` to `null`, leaving the AppImage publish configuration and explicit `.deb` workflow uploads intact. The locked electron-builder 26.8.1 otherwise adds both formats to `latest-linux*.yml`, even with `--publish=never`. Both Linux workflows then fail `scripts/verify-update-manifest.mjs`, which requires exactly one AppImage.

Keep that single-AppImage assertion. The completed build must still produce both packages, with each update manifest containing only its architecture's AppImage and passing the existing checksum/version validation. We'll run the upstream x64 and arm64 smoke jobs against the revision; you don't need your own ARM machine or permission to approve upstream Actions runs.

We'll handle these follow-ups on our side when accepting the revision:
- Carry over the installation instructions from #3152 into `README.md` and `packages/docs/content/docs/install.mdx`. Correct `packages/electron/README.md`: updates require downloading and installing the newer `.deb`; we don't provide an apt repository, and `apt upgrade` alone won't fetch OpenChamber updates.
- Require both nonempty `amd64` and `arm64` Debian assets in the final release inventory and enable `fail_on_unmatched_files` on the Linux release upload.
- Add the explicit MIT license to `packages/electron/package.json` for the Debian package metadata.

PR #3233, opened 2026-08-29
fix(ui): clear session todos on revert and re-fetch on unrevert

### LINKED ISSUE #2813: Todo list is not cleared when reverting a message
# Background

When a user clicks "Revert from here" on a message in the chat, the session is reverted to that point, hiding all messages at or after the revert point. However, the todo list (the task list shown in the StatusRow, populated by the `todowrite` tool) is not cleared or updated. The old todo list from before the revert remains visible, even though the messages that produced those todos are now hidden.

## Root cause

The `revertToMessage` function in `packages/ui/src/sync/session-actions.ts:746-850` sets a revert marker on the session and restores the user's input text, but it never clears:

- `store.todo[sessionId]` in the live sync store (populated by `todo.updated` SSE events, reduced at `packages/ui/src/sync/event-reducer.ts:263-267`)
- `useTodosPersistStore` entries (mirrored via the `onSetSessionTodo` callback at `packages/ui/src/sync/sync-context.tsx:1496-1498`)

The `StatusRow` component (`packages/ui/src/components/chat/StatusRow.tsx:173-177`) prefers live todos and falls back to persisted todos. Since neither is cleared on revert, the stale todo list remains displayed.

There is also no re-fetch of todos after a revert (unlike messages, which are refetched). The `getSessionTodos` SDK method at `packages/ui/src/lib/opencode/client.ts:544-562` is dead code, never called anywhere.

The same issue affects `unrevertSession` (`packages/ui/src/sync/session-actions.ts:877-907`), which restores reverted messages but does not restore the corresponding todo state.

# Version

OpenChamber 1.18.2, OpenCode 1.18.11

# Acceptance Criteria

## Must

- [ ] Reverting a message clears the todo list for that session from both the live sync store and the persist store
- [ ] Unreverting (restoring) a message re-fetches or restores the todo list to match the session state after unrevert
- [ ] If the revert API call fails, the previously-displayed todo list is restored (rollback)

## Should

- [ ] The `getSessionTodos` SDK method is fixed to throw on error (instead of swallowing to `[]`) and used to authoritatively re-fetch todos after revert/unrevert, rather than only clearing optimistically

---

Created with [create-issue](https://github.com/tomzx/agents/blob/a5c4a6a/skills/create-issue/SKILL.md) via glm-5p2 (`a5c4a6a`)

jlu-lujing:
## Intent

"Revert from here" hides the messages at/after the revert point, but the todo list (the `todowrite` task list in the composer status, live `store.todo[sessionId]` with a `useTodosPersistStore` fallback) was never cleared — so a stale task list stayed visible after revert. Symmetrically, unrevert restored messages but not todos (#2813).

- `revertToMessage` now clears the todo list from **both** the live sync store and the persist store, and **rolls both back** if the revert API call fails.
- `unrevertSession` re-fetches the authoritative todos and writes both stores, leaving the existing list intact if the re-fetch errors.
- The previously-dead `getSessionTodos` helper now takes a `directory` and **throws on error** (instead of swallowing to `[]`), so callers distinguish "no todos" from "could not read" — a swallow-to-`[]` would have let a transient failure wipe the visible list.

Closes #2813.

## Non-goals

- Does not clear todos on the revert API *call* — clearing is optimistic and only persists on success (matches "cleared after a successful revert"); a failed revert restores the prior list.
- Does not add a live→persist write inside these actions (they set store state directly, mirroring `moveRecordEntries`); the persist write goes through `useTodosPersistStore` explicitly, as the deletion-cleanup path already does.

## Affected surfaces

- `packages/ui/src/sync/session-actions.ts` — `revertToMessage`, `unrevertSession`.
- `packages/ui/src/lib/opencode/client.ts` — `getSessionTodos` (was uncalled).
- Consumers of `store.todo[sessionId]` (ComposerStatusBar, WorkStatusTasksSection) and `useTodosPersistStore` — behavior unchanged except todos now track revert/unrevert. No persisted-format change.

## Repository guidance

|---|---|---|
| `sync-state-invariants` | Reverting mutates live sync store state that must stay consistent with the persisted mirror and survive failure. | Clears live + persisted together, and rolls both back on API failure (the function already had a rollback path for messages/input; todos now join it). No new cache, no speculative retry. |

## Validation

| Check | Result |
|---|---|
| `bun test packages/ui/src/sync/session-actions.test.ts -t "issue #2813"` | 4 pass / 0 fail (revert clears live+persist; revert failure restores both; unrevert re-fetches into both; re-fetch failure leaves todos intact) |
| `bun test packages/ui/src/sync/` | 553 pass / 30 fail vs clean-main baseline **549 pass / 30 fail** → +4 new passing tests, **0 new failures** (the 30 are pre-existing vitest-API-in-`bun` failures, present on clean main) |
| `bun run --cwd packages/ui type-check` (`tsc --noEmit`) | clean (exit 0) |
| Real app revert/unrevert with a `todowrite` list | **Not run** — needs a running app with an active session; covered by the action-level tests above (they drive the real `revertToMessage`/`unrevertSession` against the real `useTodosPersistStore`). |

## Visual evidence

The change is in sync-store action logic, not a component. The observable effect (composer status todo list disappears on revert / returns on unrevert) is exercised by the action-level tests; the rendering components (ComposerStatusBar / WorkStatusTasksSection) are untouched and continue to prefer live then fall back to persisted.

## Risks and failure behavior

- If the todo re-fetch fails during unrevert, the existing list is left as-is rather than wiped to empty.
- Rollback restores both stores on revert failure, so a failed revert never leaves the composer showing an empty task list.
- `getSessionTodos` throwing is safe: it had no prior non-test caller, and the one new caller (`unrevertSession`) wraps it in a non-fatal try/catch.

### DIFF
--- a/packages/ui/src/lib/opencode/client.ts
+++ b/packages/ui/src/lib/opencode/client.ts
@@ -689,25 +689,27 @@ class OpencodeService {
     return unwrapSdkData(response, 'session.messages');
   }
 
-  async getSessionTodos(sessionId: string): Promise<Array<{ id: string; content: string; status: string; priority: string }>> {
-    try {
-      const response = await this.client.session.todo({
-        sessionID: sessionId,
-        ...(this.currentDirectory ? { directory: this.currentDirectory } : {}),
-      });
-      if (response.error) {
-        return [];
-      }
-
-      const data = response.data;
-      if (!data || !Array.isArray(data)) {
-        return [];
-      }
-
-      return data as Array<{ id: string; content: string; status: string; priority: string }>;
-    } catch {
+  async getSessionTodos(
+    sessionId: string,
+    directory?: string,
+  ): Promise<Array<{ id: string; content: string; status: string; priority: string }>> {
+    // Authoritative re-fetch (used after revert/unrevert). Throws on transport or
+    // API error so callers can distinguish "no todos" (empty array) from "could
+    // not read todos" — swallowing to [] here previously let a transient failure
+    // wipe a session's visible todo list.
+    const resolvedDirectory = directory ?? this.currentDirectory;
+    const response = await this.client.session.todo({
+      sessionID: sessionId,
+      ...(resolvedDirectory ? { directory: resolvedDirectory } : {}),
+    });
+    if (response.error) {
+      throw new Error(`session.todo failed for ${sessionId}`);
+    }
+    const data = response.data;
+    if (!data || !Array.isArray(data)) {
       return [];
     }
+    return data as Array<{ id: string; content: string; status: string; priority: string }>;
   }
 
   /**
--- a/packages/ui/src/sync/session-actions.test.ts
+++ b/packages/ui/src/sync/session-actions.test.ts
@@ -2277,6 +2277,7 @@ describe("revert/unrevert todo state (issue #2813)", () => {
 
     expect(replyCalls.some((call) => call.method === "session.todo" && call.params.sessionID === "root")).toBe(true)
     expect(store.getState().todo["root"]).toEqual(sessionTodosResult)
+    expect(useTodosPersistStore.getState().getSessionTodos("/tree", "root")).toEqual(sessionTodosResult)
   })
 
   test("unrevert leaves existing todos untouched when the re-fetch fails", async () => {
--- a/packages/ui/src/sync/session-actions.test.ts
+++ b/packages/ui/src/sync/session-actions.test.ts
@@ -18,6 +18,8 @@ const failingRevertSessionIds = new Set<string>()
 const failingUnrevertSessionIds = new Set<string>()
 let afterUnrevertCall: ((sessionId: string) => void) | null = null
 let sessionDeleteError: unknown | null = null
+let sessionTodosResult: unknown[] = []
+let sessionTodosThrows = false
 let beforeSessionUpdateResolve: ((sessionId: string) => void) | null = null
 let beforeSessionDeleteResolve: ((sessionId: string) => void) | null = null
 const globalUpsertedSessions: unknown[] = []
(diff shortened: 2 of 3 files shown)


jlu-lujing:
@openchamber-bot re: the `review:needs-evidence` ask for a running-app screenshot pair — flagging a real obstacle and asking how you'd like to satisfy it.

**Why the requested artifact is hard to produce here**
- The composer task list the change affects is the `todowrite` list, which only renders when an agent actually runs `todowrite` during a session. It is not present in ordinary turns, so capturing it needs a live agent session that happens to plan a multi-step task.
- My local environment can't stand up the runtime: I booted the Vite dev server, but the SPA renders blank without a working OpenCode backend + model credentials (home-directory/settings calls return 500), so I cannot drive a real session to the point where a todo list exists and click Revert/Unrevert. There's also no rasterizer (chromium) available to snapshot even an isolated component render to PNG.

I don't want to fabricate a running-app screenshot, so I'm surfacing this rather than faking evidence.

**What I did supply (rendered-behavior + logic coverage)**
- Four action-level tests drive the real `revertToMessage`/`unrevertSession` against the real `useTodosPersistStore` and assert the exact user-visible store transition the composer renders from: clears live+persisted on revert, rolls both back on revert failure, re-fetches into both on unrevert, and leaves the list intact if the re-fetch errors.
- I also addressed the review's other points on the latest push: the unrevert test now asserts the persisted-store write it names, and `cascadeRevertToDescendants` records the descendant-todo gap as a known follow-up (out of #2813's scope).

**How can I close the evidence gate?** Happy to do whichever you prefer:
1. If a **running-app screenshot/recording is strictly required**, could a maintainer (or someone with OpenChamber set up) capture the before-revert / after-revert pair? I can attach it to this PR immediately once provided.
2. If an **isolated render of the real `WorkStatusTasksSection`** (todos present vs cleared) is acceptable as the rendered-behavior artifact, I'll add it — noting it exercises the actual component, not a mock, but in isolation from the full app.
3. If the action-level tests are judged sufficient given the change is pure sync-store action logic (no component/rendering code touched), a `review:ready` would unblock it — your call.

No change is being pushed to game the label; the code is as described and the reviewer confirmed it is correct and closes #2813.

PR #3234, opened 2026-08-29
OPE-256: Build a persistent notification center and notification history

AlexKutas:
## Intent

Toasts disappear. Session idle, error, question, and permission notices do the same if you miss them. This PR keeps those events in a header inbox on this device so you can open the session later.

The bell sits in the desktop, VS Code, and mobile headers. A new row starts unread. A toast timeout leaves it unread. Opening the row, or tapping OK or Copy, marks it read and keeps it. Clear all empties every stored row for this host, including kinds the filter is hiding, and dismisses matching live toasts. Other runtime buckets stay. Mark all as read also dismisses matching live toasts. A reused caller toast id after the 60-second collapse window gets a new record id, so two rows never share a key. The badge is an unread dot, not a count. Session rows show title and project, never a raw session id. Clicking a system or push notification with a session id opens that session.

Settings → OpenChamber → Notifications now has three blocks. Notification Delivery is still the OS-banner switch. Events and templates sit under Delivery and only show when Delivery is on. In-app history is its own block and does not depend on Delivery. History starts on. Turning it off hides the bell, the list, and the badge. Stored rows stay. Kind checkboxes hide rows from the list and badge without deleting them. Info, success, and subtasks start off. Subtask OS banners stay on their own Events toggle.

History can play a two-tone chime when a session event lands, including when the window is in the background. It stays quiet when a system banner would also sound, and when you are already watching that session in a focused window. The sound switch lives only under In-app history. It does not mute OS banners. There is no keyboard shortcut and no command-palette item. The bell is the way in.

## Non-goals

- Making OS banners reliable in unpackaged `oc-dev` Electron on modern macOS. Delivery still needs a packaged app for banners to actually appear.
- Changing when Delivery shows a banner (unfocused, or focused plus `always`).
- Sending info or success toasts to the system. That was already off.
- A changelog entry. Release notes stay the maintainer's job.

## Affected surfaces

- `packages/ui`. Shared inbox, store, filter, sound, toast wrapper, Settings, Help (no new shortcut), docs copy.
- Desktop, web, VS Code, hosted mobile, and Capacitor. The same `NotificationCenter` mounts in `Header` (dropdown), `VSCodeLayout`, and `MobileHeader` (dialog).
- `packages/web`. Settings defaults for the new flags, and notification click payloads now carry session id and directory so a banner click can open the session. `silent` is honored only when the payload sets it.
- `packages/electron`. Native `Notification` uses `payload.silent === true`. Inbox sound never writes that flag.
- `packages/vscode`. Webview banner click uses `openchamber:open-session` with directory, same as the other shells.
- `packages/docs`. In-app history section in every locale.
- Persisted client state. Inbox rows live in `openchamber-notifications.v1`, partitioned by `runtimeKey`. Settings persist `notificationInboxEnabled` (default true), `notificationSoundEnabled` (default true), and `notificationInboxFilter`. Missing inbox-enabled on the server reads as true. Cap 500 rows, 30-day TTL. Malformed persist is missing, not an empty success. A leftover `open_notifications` shortcut override, if anyone saved one, is ignored.

## Repository guidance

|---|---|---|
| `AGENTS.md` change discipline | New persist keys, Settings contract, shared UI owned by `packages/ui` | No new dependencies. No changelog edit. Domain logic stays in the notification store, filter, and record modules. Entrypoints only mount the bell or forward a click. |
| `sync-state-invariants` and `packages/ui/src/sync/DOCUMENTATION.md` | New durable list, runtime-scoped cache, malformed persist | Runtime key partitions the lists. Switching runtime shows that runtime's rows and leaves the others on disk. Bad JSON is missing. Bad rows drop one by one. An empty list is valid empty. Live attention still comes from live session status, not this history. |
| `locale-ui-patterns` | Bell, empty state, Settings, docs | Every new string is a real translation in every locale. `messages.test.ts` still has key parity. |
| `settings-ui-patterns` and `references/search.md` | New Settings blocks and search hits | Delivery, Events, and In-app history use the shared Settings primitives. Search registry entries match the new anchors. Events stay hidden with Delivery. |
| `theme-system` | Bell, unread dot, list row chrome | Tokens and the existing icon sprite. No one-off colors. |
| `ui-api-decoupling` | Banner click and settings round-trip across runtimes | Shared UI opens a session through `openchamber:open-session`. Web, Electron, and VS Code each attach session id and directory on their own notify path. |
| `desktop-shell` | Electron `Notification` constructor | Only `silent === true` on the payload changes the OS banner. Inbox sound does not cross that boundary. |
| `communication-style` | Docs and Settings copy | Product docs describe the bell, the filters, and the independent history switch. No shortcut sentence. |

## Validation

| Check | Result |
|---|---|
| `bun test` after merging `origin/main` | Pass. scripts 1/1, ui 363/363, vscode 26/26, electron 18/18, web 165 files / 1696 passed. |
| Focused inbox tests (`notification-record`, `notification-store`, `notification-inbox`, `session-error-log`) | Pass on `46b9f7233`, including clear-all, unique-id, background chime, and latest session error. |
| `bun run type-check` | Pass on the workspace. The earlier `projectLabel` / `toBeUndefined` / VS Code `silent` failures are gone (`f3389b894`). |
| Electron `test:architecture` and `test:updater` | Pass. |
| `bun run build` | Not run. |
| Packaged desktop, VS Code Extension Host, Capacitor | Not a scripted pass. The shared center mounts in those headers. |
| OS banners in `oc-dev` | Not treated as a pass. Unpackaged Electron often never shows a banner on current macOS. |

## Visual evidence

Desktop web, attached against this draft:

- Open inbox with stored rows.
- Settings → OpenChamber → Notifications, with Delivery, Events, and In-app history visible.
- Empty inbox.
- In-app history off. The bell is gone.
- Click on a stored session row. That session opens.

Screenshot 2026-08-29 at 21 06 03
Screenshot 2026-08-29 at 23 23 22
Screenshot 2026-08-29 at 23 24 05
Screenshot 2026-08-29 at 23 24 35
Screenshot 2026-08-29 at 23 44 59

## Risks and failure behavior

- First launch writes a new local persist file. Older builds ignore it. Downgrade keeps the file on disk and stops showing the bell.
- Missing `notificationInboxEnabled` becomes true, so an older settings blob does not hide the inbox by accident.
- One bad stored row is dropped. It does not wipe the rest of that runtime's list.
- A fetch or persist read failure must not look like "you have no notifications." The store treats malformed persist as missing.
- Inbox sound can double with an OS banner if Delivery is on and `notificationMode` is `always`. The chime skips that case. A later payload that sets `silent: true` still silences the OS banner. The inbox switch never does that.
- Banner-click → session open needs the payload to carry a session id. Without it, the click only focuses the app.
- Packaged-desktop banner sound and permission prompts stay an Electron/OS problem. This PR does not fix unsigned `oc-dev` banners.


mattv8:
@btriapitsyn, does [OPE-256](https://linear.app/openchamber/issue/OPE-256) reflect the notification experience you want? This adds an inbox with filtering and read state beyond the existing notification storage. If that remains the intended direction, we should continue with a focused review rather than close it as redundant.

PR #3239, opened 2026-08-30
fix(chat): reliably render completed subagent Markdown

### LINKED ISSUE #3238: [Bug] Subagent task output Markdown fix is absent from current mainline
## Update (2026-09-12): corrected root cause — the task result envelope

The raw Markdown does not come from the sync first-paint lifecycle guard. Completed `task` output keeps OpenCode's result envelope, and the envelope is what defeats Markdown parsing:

```text
<task id="ses_…" state="completed">
<task_result>
## Some heading
- list item
</task_result>
</task>
```

`prepareTaskToolOutput` (`packages/ui/src/components/chat/message/parts/taskToolModel.ts`) stripped only `<task_metadata>`, so the envelope reached the renderer. `marked` treats a line starting with the `<task …>` tag as an HTML block, so the Markdown below it stays literal and the escaped tags are shown as text. Verified against current `main` (`8062a7bf6`) with the real pipeline: both `renderMarkdownSync` and `renderMarkdownBlocks` return the escaped envelope plus unparsed source (for example `## Verdict` remains literal). The subagent session renders correctly because its assistant text carries no envelope.

A fix that unwraps the envelope before rendering is now in [#3239](https://github.com/openchamber/openchamber/pull/3239) (`92ecc2355`, draft), with regression tests on the wrapped, same-line, and un-wrapped formats. The sync-fallback changes already in #3239 address the stale-DOM half, not this raw half.

---

## What's wrong?

Subagent task output can still appear as raw Markdown in the main chat after the subagent session completes. The same response renders correctly when opening the subagent session itself.

Example output shown as source instead of rendered Markdown:

```text
### Startup
- Route: `PR review`
- Mode: `read-only`
- Summary: Review PR changes
```

## Related history

- Original issue: [#1902](https://github.com/openchamber/openchamber/issues/1902)
- Previous PR: [#1905](https://github.com/openchamber/openchamber/pull/1905)
- Replacement PR: [#3239](https://github.com/openchamber/openchamber/pull/3239)
- Fix commit referenced by the previous issue: [`844b5216d`](https://github.com/openchamber/openchamber/commit/844b5216d7c7bee7013ef0f4c1c33f8bca1f9435)

## Incorrect closure statement

> **Incorrect:** the original issue was closed with the statement that the bug was fixed in `844b5216d` and that the commit had landed.

The commit is not present in the current `upstream/main` history. The previous PR #1905 is closed and was not merged. No equivalent cherry-picked commit was found in the current mainline history.

## Current code path

Verified against the current `upstream/main` commit [`fb943e008`](https://github.com/openchamber/openchamber/commit/fb943e0089406ae4efd1e76f28fd4a970b7c88a9):

- The shared `useMorphdomMarkdown` renderer is implemented in [`MarkdownRendererImpl.tsx`](https://github.com/openchamber/openchamber/blob/fb943e0089406ae4efd1e76f28fd4a970b7c88a9/packages/ui/src/components/chat/MarkdownRendererImpl.tsx#L809-L1104).
- Its synchronous fallback still only installs content when the target has no children:

  ```tsx
  if (text && target.childNodes.length === 0) {
  ```

  See [`MarkdownRendererImpl.tsx#L925-L965`](https://github.com/openchamber/openchamber/blob/fb943e0089406ae4efd1e76f28fd4a970b7c88a9/packages/ui/src/components/chat/MarkdownRendererImpl.tsx#L925-L965).
- Completed task output still reaches `SimpleMarkdownRenderer` through `TaskToolSummary` at [`ToolPart.tsx#L991-L1104`](https://github.com/openchamber/openchamber/blob/fb943e0089406ae4efd1e76f28fd4a970b7c88a9/packages/ui/src/components/chat/message/parts/ToolPart.tsx#L991-L1104), with the renderer call at [`ToolPart.tsx#L1099`](https://github.com/openchamber/openchamber/blob/fb943e0089406ae4efd1e76f28fd4a970b7c88a9/packages/ui/src/components/chat/message/parts/ToolPart.tsx#L1099).
- The current Markdown block cache is content-addressed through `contentFingerprint` in [`markdownCore.ts#L571-L654`](https://github.com/openchamber/openchamber/blob/fb943e0089406ae4efd1e76f28fd4a970b7c88a9/packages/ui/src/components/chat/markdown/markdownCore.ts#L571-L654). Therefore, the old issue's static `simple:tool` cache-key theory is not the current root cause; this follow-up focuses on the remaining synchronous-render lifecycle guard and stale DOM replacement behavior.

The current first-paint path can therefore leave stale or raw Markdown visible when the asynchronous render is delayed or cancelled after task output is mounted or expanded.

## Expected behavior

Markdown returned by a completed subagent task should render in the main chat the same way it renders in the subagent session view, including headings, lists, inline code, fenced code blocks, and other supported Markdown.

## Runtime

- Desktop Web / Web
- Current mainline verification performed against `upstream/main` at `fb943e008`

##Evidence 1.23.0

Image

bashrusakh:
## Intent

Fixes #3238. Completed `task` (subagent) output rendered as raw Markdown because OpenCode wraps the result in an envelope:

```text
<task id="ses_…" state="completed">
<task_result>
## heading
…
</task_result>
</task>
```

`prepareTaskToolOutput` stripped only `<task_metadata>`, so the envelope reached `marked`, which treats the leading `<task …>` line as a raw HTML block and never parses the result Markdown.

The fix unwraps a complete `<task_result>` block in `prepareTaskToolOutput` only when the output starts with a `<task …>` tag and actually contains the result block; everything else passes through unchanged. The live task render path (`TaskToolSummary`) already routes through `prepareTaskToolOutput`, so the helper change is the whole fix.

This candidate is a shrink of the earlier version: it no longer touches `MarkdownRendererImpl` or detached-DOM caching — the stale provisional-fallback race could not be shown to outlive normal async resolution, so that half was dropped rather than carried on inertia.

## Non-goals

- No change to subagent payloads, session lifecycle, streaming first-mount behavior, the shared Markdown parser, or sanitization.
- No change to detached DOM cache identity.
- Failed-task `<task_error>` envelopes are not unwrapped here.

## Affected surfaces

- `packages/ui/src/components/chat/message/parts/taskToolModel.ts` — envelope unwrap in `prepareTaskToolOutput`.
- `packages/ui/src/components/chat/message/parts/taskToolModel.test.ts` — regression coverage.
- Shared UI: web, desktop, VS Code, and mobile render task output through the same helper. No backend, persistence, or API changes.

## Repository guidance

|---|---|---|
| Root `AGENTS.md` | Shared React UI chat rendering | Change stays in the owning task-output helper; no renderer/DOM lifecycle touched |
| `packages/ui/src/components/chat/message/parts/DOCUMENTATION.md` | Chat message rendering architecture | No tool rendering contracts change |
| `openchamber-change-discipline` | Source change in shared UI | Focused two-file diff, no dependencies, tests beside the helper |

## Validation

| Check | Result |
|---|---|
| `bun test packages/ui/src/components/chat/message/parts/taskToolModel.test.ts` | 9 pass / 0 fail; 3 new envelope tests fail if the helper change is reverted |
| `bun test packages/ui/src/components/chat/message/parts/ToolPart.test.ts packages/ui/src/components/chat/message/toolRenderers.test.ts` | 40 pass / 0 fail |
| `bunx oxlint` on changed files | No new findings (pre-existing backlog only) |
| `git diff --check` | Clean |
| Whole-PR review | PASS-WITH-FINDINGS, no blocking findings at `ad854d3bb` |

## Visual evidence

No screenshot: the regression is asserted at the helper boundary against the exact wrapped format; a capture would need the full app plus a completed subagent task.

## Risks and failure behavior

- Unwrap requires both the leading `<task …>` tag and a complete `<task_result>` block; partial/older payloads fall back to the raw output.
- A literal `</task_result>` inside the result text would truncate (contrived; the previous render was equally broken).
- `<task_error>` envelopes from failed tasks are not unwrapped; #3238 is about completed output.

---

CI note (2026-09-13): the `checks` job is red only on `packages/vscode/webview/api/settings.test.ts` ("propagates a failed bridge read and retries successfully"), which times out deterministically on clean current `main` (`961f1611c`, reproduced locally 3/3; the test was added by main commit `6cf247894`). It is unrelated to this diff — every fresh PR head on this base fails it. Do not treat this red check as a finding against this PR.

### DIFF
--- a/packages/ui/src/components/chat/message/parts/taskToolModel.test.ts
+++ b/packages/ui/src/components/chat/message/parts/taskToolModel.test.ts
@@ -56,4 +56,61 @@ describe('taskToolModel', () => {
         expect(prepareTaskToolOutput('done\n<task_metadata>{"sessionID":"child-1"}</task_metadata>')).toBe('done');
         expect(prepareTaskToolOutput(undefined)).toBe('');
     });
+
+    test('unwraps the task result envelope and preserves the result Markdown exactly', () => {
+        const result = [
+            '## Verdict',
+            '',
+            '- first item',
+            '- second item',
+            '',
+            '```ts',
+            'const answer = 42;',
+            '```',
+        ].join('\n');
+        const output = [
+            '<task id="ses_abc123" state="completed">',
+            '<task_result>',
+            result,
+            '</task_result>',
+            '</task>',
+        ].join('\n');
+
+        expect(prepareTaskToolOutput(output)).toBe(result);
+    });
+
+    test('unwraps a same-line task result envelope', () => {
+        const output = '<task id="ses_abc123" state="completed"><task_result>result</task_result></task>';
+
+        expect(prepareTaskToolOutput(output)).toBe('result');
+    });
+
+    test('leaves output without a complete task envelope untouched', () => {
+        const plainMarkdown = '## Verdict\n- first item';
+        expect(prepareTaskToolOutput(plainMarkdown)).toBe(plainMarkdown);
+
+        const taskTagWithoutResult = '<task id="ses_abc123" state="running">\nstill running';
+        expect(prepareTaskToolOutput(taskTagWithoutResult)).toBe(taskTagWithoutResult);
+
+        const unterminatedResult = '<task id="ses_abc123" state="running">\n<task_result>\nstill running';
+        expect(prepareTaskToolOutput(unterminatedResult)).toBe(unterminatedResult);
+
+        const resultWithoutEnvelope = 'literal <task_result>text</task_result> in prose';
+        expect(prepareTaskToolOutput(resultWithoutEnvelope)).toBe(resultWithoutEnvelope);
+    });
+
+    test('keeps parsing task metadata from the raw envelope output', () => {
+        const output = [
+            '<task id="ses_abc123" state="completed">',
+            '<task_result>',
+            '## Verdict',
+            '</task_result>',
+            '</task>',
+            '<task_metadata>{"sessionID":"child-1"}</task_metadata>',
+        ].join('\n');
+
+        expect(prepareTaskToolOutput(output)).toBe('## Verdict');
+        expect(readTaskSessionIdFromOutput(output)).toBe('child-1');
+        expect(parseTaskMetadataBlock(output).sessionId).toBe('child-1');
+    });
 });
--- a/packages/ui/src/components/chat/message/parts/taskToolModel.ts
+++ b/packages/ui/src/components/chat/message/parts/taskToolModel.ts
@@ -133,11 +133,29 @@ export const stripTaskMetadataFromOutput = (output: string): string => {
     return output.replace(/\n*<task_metadata>[\s\S]*?<\/task_metadata>\s*$/i, '').trimEnd();
 };
 
+const TASK_ENVELOPE_OPEN_TAG_PATTERN = /^\s*<task(?:\s[^>]*)?>/i;
+const TASK_RESULT_BLOCK_PATTERN = /<task_result>\s*([\s\S]*?)\s*<\/task_result>/i;
+
+// OpenCode wraps a completed task result in an envelope:
+//   <task id="ses_…" state="completed">
+//   <task_result>…result Markdown…</task_result>
+//   </task>
+// `marked` treats the leading tag line as a raw HTML block, so the Markdown
+// below it stays literal (issue #3238). Only unwrap when the output actually
+// starts with the envelope tag and carries a complete result block; other
+// outputs pass through untouched.
+const unwrapTaskResultEnvelope = (output: string): string => {
+    if (!TASK_ENVELOPE_OPEN_TAG_PATTERN.test(output)) return output;
+    const resultBlock = output.match(TASK_RESULT_BLOCK_PATTERN);
+    if (!resultBlock) return output;
+    return resultBlock[1];
+};
+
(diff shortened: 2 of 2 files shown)


bashrusakh:
### CI note: lint failure is outside this PR

The current `checks` run fails during the repository-wide lint step at `packages/ui/src/components/layout/Header.tsx:32`:

```text
@typescript-eslint/no-unused-vars: 'useRuntimeAPIs' is defined but never used
```

This import is present in the rebased `upstream/main` and was left behind by commit [`03589622e`](https://github.com/openchamber/openchamber/commit/03589622e0a4824b1b85035576c912b88295bb5b), which removed the hook usage from `Header.tsx` but did not remove the import.

PR #3239 does not modify `Header.tsx`; its diff is limited to the Markdown renderer and its regression tests. The lint failure is therefore a pre-existing baseline issue in the current upstream code, not a regression caused by this PR and not part of this PR's scope. Build and type-check passed; tests were not reached because the lint job stopped first.

This PR should not absorb that unrelated cleanup.

bashrusakh:
@btriapitsyn

yulia-ivashko (approved):

yulia-ivashko:
I ran the full checks job locally to save a CI round-trip: this branch merged onto current main (e8f5858be), same steps in the same order as the workflow.

Install, build, type-check, lint, changelog:check, the full test suite and the Electron packaging tests all passed.

PR #3240, opened 2026-08-30
Show inline thumbnails for image reads in the chat timeline

gillesbergerp:
The read/skill static tool row only ever linked to the file path, even when the read target was an image, so the agent's screenshots and diagrams never surfaced next to the conversation the way diffs do.

Resolve image reads the same way MarkdownImageGallery resolves workspace-local markdown images (VS Code fs bridge vs. session-scoped server grant + authenticated asset URL) and render a small clickable thumbnail that opens the shared image preview dialog. Extracted the gallery's asset-auth-token hook into useRuntimeAssetAuth so both consumers share one implementation.

### DIFF
--- a/packages/ui/src/components/chat/MarkdownImageGallery.tsx
+++ b/packages/ui/src/components/chat/MarkdownImageGallery.tsx
@@ -3,12 +3,6 @@ import { toast } from 'sonner';
 import { Icon } from '@/components/icon/Icon';
 import { useEffectiveDirectory } from '@/hooks/useEffectiveDirectory';
 import { useI18n } from '@/lib/i18n';
-import {
-  acquireRuntimeUrlAuthToken,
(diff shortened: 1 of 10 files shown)


gillesbergerp:
## Visual evidence: read-tool image thumbnail

Captured in an isolated harness (real Express server + real `/auth/url-token` +
real `/api/fs/raw`, driving the actual `StaticToolRow`/`ReadToolImageThumbnail`
components). The only mocked piece was the `markdown-image-grants` POST
response, since a real assistant message needs a live model session; the
image resolution, asset URL, and rendering are all real production code
paths against a real file (`docs/references/chat_example.png`).

**1. Actual size (20×20px), before/after comparison:**
normal size

**2. Same view, 3x zoomed for legibility:**
zoomed 3x

**3. Click-to-preview (full-resolution image after clicking the thumbnail):**
click to preview

PR #2276, opened 2026-07-16
feat(git): add bounded worktree-aware execution

bashrusakh:
## Summary

This PR replaces the interim per-`git-common-dir` FIFO with bounded, conflict-aware Git execution across web/Electron and VS Code, then consolidates both runtimes onto one canonical execution core.

- Resolve canonical common-directory and worktree identities before scheduling repository work.
- Serialize only operations that actually conflict; unrelated linked worktrees can continue independently.
- Bound reads, writes, network work, clone destinations, queues, retained identities, and in-flight status work.
- Coordinate web/Electron and VS Code built-in/raw Git paths under the same scheduler invariants.
- Share one resolver/coordinator/error implementation while preserving runtime-specific adapters and operation classifications.
- Add deterministic, real-Git, lifecycle, fsmonitor pass-through, and historical/current performance evidence.

Closes #2233.

> [!NOTE]
> This PR is stacked on #2232. Its two hotfix commits remain visible in this branch until the stack is resolved; conflict resolution is intentionally separate from this update.

## Why

The #2232 hotfix stopped index-lock contention by serializing work behind one common-directory queue, but that model is intentionally conservative:

- independent linked worktrees block each other;
- reads, local writes, shared-ref writes, topology changes, and network operations are treated as the same resource;
- repeated status callers can create unnecessary work;
- web and VS Code previously carried separate resolver/coordinator/error implementations that could drift.

This PR moves correctness to the owning execution primitive: canonical identity, explicit conflict profiles, finite admission, generation-aware in-flight status sharing, and one implementation used by both runtimes.

## Architecture

```text
Web / Electron Git service ───────────────┐
                                         ├── canonical ESM execution core
VS Code service + built-in/raw adapter ──┘   resolver · coordinator · errors
                                              │
                                              ├── common/worktree identity
                                              ├── conflict-aware scheduling
                                              ├── network/clone reservations
                                              ├── generations/status sharing
                                              └── bounded cleanup/diagnostics
```

### Conflict model

- **Read:** may overlap up to the read cap, but never with a same-worktree write.
- **Worktree write:** exclusively owns one linked worktree's index, HEAD, and working tree.
- **Common write:** serializes shared refs/config; a target-worktree variant also excludes local work for that worktree.
- **Topology write:** acts as a barrier for the whole common Git context.
- **Network:** independently bounded per common context and globally.
- **Clone:** reserves a canonical destination before repository identity exists.

[...]

PR #3256, opened 2026-08-31
feat(git): share bounded execution across runtimes

### LINKED ISSUE #2229: [Bug] Unable to perform git commit

### LINKED ISSUE #2233: Git execution architecture: concurrency roadmap (PR 2)
# Git Execution Architecture — Concurrency Roadmap

> Follow-up to the PR #2232 hotfix. This issue tracks the long-term model for deterministic Git execution.

### Goals

- Eliminate OpenChamber-induced Git lock contention.
- Keep changes low-risk.
- Preserve APIs.
- Avoid UI refactoring.

### Changes

1. Server-side single-flight for `getStatus()` keyed by `git-common-dir`.
2. Route `getStatus()` through the same serialization path as mutating operations.
3. Review/remove unnecessary `Promise.all()` Git subprocesses where sequential execution is preferable.
4. Optionally increase client cache TTL as a temporary optimization.

### Explicitly out of scope

- RepositoryStateService
- GitExecutionContext
- Scheduler registry
- Filesystem watcher redesign
- Snapshot redesign
- Generic RW-lock

---

# PR 2 — Git Execution Architecture

## Core principle

Synchronize Git execution, not callers.

**Invariant**

> At most one Git subprocess may execute simultaneously for a single `git-common-dir`.

## Synchronization key

Always use:

`git rev-parse --git-common-dir`

Never serialize by working directory.

## GitExecutionContext

Owns one Git storage.

Responsibilities:

- AsyncQueue (FIFO)
- SingleFlight
- RefreshController
- SnapshotStore
- Subscribers
- GitExecutor

It knows nothing about HTTP, UI or polling.

## AsyncQueue (FIFO)

Purpose:

- serialize Git subprocess execution only.

Requirements:

- strict FIFO
- one active Git subprocess
- no starvation
- no busy waiting
- implementation intentionally simple

Queue stores Git operations rather than arbitrary callbacks.

## SingleFlight

Deduplicates identical read operations before they enter the queue.

Multiple concurrent `getStatus()` requests share one Promise.

Failed executions are never cached.

## RefreshController

Maintains only:

- refreshRunning
- refreshRequested

Maximum outstanding refreshes:

- one running
- one pending

## SnapshotStore

Stores:

- HEAD
- branch
- dirty state
- ahead/behind
- staged summary
- unstaged summary
- stash count

Uses idle expiration/LRU.

## Error handling

Keep:

- lastKnownGoodSnapshot
- lastRefreshError

Refresh failures never destroy the last valid snapshot.

## Cancellation

Support AbortSignal for request cancellation.

Long-running Git operations should rely on Git-native abort mechanisms (`merge --abort`, `rebase --abort`) rather than generic cancellation.

## Serialized operations

All operations touching shared Git storage execute through GitExecutionContext, including:

- status
- diff
- commit
- checkout
- merge
- rebase
- reset
- stage / unstage
- update-ref
- worktree add/remove/prune
- branch rename
- config updates affecting repository state

## RepositoryStateService

UI subscribes to repository state.

Flow:

Filesystem events / polling → RepositoryStateService → GitExecutionContext → Snapshot → Subscribers

## Filesystem events

Preferred trigger.

Polling remains only as fallback.

## Git performance

Leverage when available:

- core.fsmonitor
- builtin fsmonitor
- newer plumbing commands

Never automatically modify user Git configuration.

## External Git processes

GitExecutionContext serializes only OpenChamber operations.

External Git processes may still create `index.lock`.

Those errors should be surfaced while preserving the last valid snapshot.

## Scalability

Scales with active `git-common-dir` instances rather than UI sessions.

Example:

- 300 repositories
- 100 worktrees
- 3000 sessions

Still bounded:

- one execution context per active git-common-dir
- one active Git subprocess
- one running refresh
- one pending refresh

## Final architecture

HTTP/UI → GitService → GitExecutionContext → SingleFlight → RefreshController → AsyncQueue (FIFO) → GitExecutor → Git

## Architectural invariants

1. One GitExecutionContext per active git-common-dir.
2. One active Git subprocess per git-common-dir.
3. Strict FIFO execution.
4. SingleFlight before queue.
5. Refresh coalescing.
6. No duplicate getStatus execution.
7. No global locks.
8. No generic RW-lock.
9. No retry-based synchronization.
10. Snapshot survives refresh failures.
11. UI never executes Git directly.
12. Queue only serializes Git execution.

## Expected benefits

- Eliminates OpenChamber-induced lock contention.
- Deterministic execution.
- Lower CPU and IO.
- Fewer Git subprocesses.
- Correct worktree handling.
- Bounded memory.
- Clear ownership of repository state.

### Why GitExecutionContext instead of a global mutex?

A global mutex would serialize unrelated repositories and significantly reduce concurrency.

Using one execution context per `git-common-dir` allows unrelated repositories to execute Git operations independently while still preventing conflicts inside a single Git storage.

### Why serialize all Git operations?

The previous implementation serialized only mutating operations.

Background reads (`getStatus`) could still overlap with writes, producing lock contention.

Serializing all Git subprocesses per `git-common-dir` guarantees deterministic execution and removes this entire class of races.

### Why not use a reader/writer lock?

A RW-lock introduces additional complexity:

- fairness
- starvation
- upgrade/downgrade semantics
- timeout handling
- deadlock risks

The workload is not read-heavy enough to justify that complexity.

A single FIFO execution queue is simpler, deterministic, and sufficient.

### Why AsyncQueue instead of an actor model?

The problem is execution serialization, not asynchronous message passing.

An actor model would introduce additional abstractions without solving a problem that exists today.

A small FIFO queue provides the required guarantees with significantly lower implementation complexity.

### Why SingleFlight before the queue?

Without deduplication, thousands of identical `getStatus()` requests would still enqueue thousands of tasks.

SingleFlight collapses identical requests into one execution before they reach the queue, dramatically reducing Git process creation.

### Why RefreshController?

Polling often produces bursts of refresh requests.

Only the latest repository state matters.

Coalescing refreshes into at most one running and one pending execution prevents unbounded queue growth while preserving correctness.

### Why keep the last successful snapshot?

Temporary Git failures (for example, external `index.lock`) should not invalidate repository state.

Keeping the last known good snapshot provides a stable UI while allowing refreshes to recover automatically.

### Why use `git-common-dir` as the synchronization key?

Multiple worktrees share the same Git storage.

Synchronizing by working directory does not prevent conflicts between worktrees.

Using `git-common-dir` correctly serializes operations touching the shared repository metadata.

### Why not automatically enable `core.fsmonitor`?

OpenChamber should not modify user-managed Git configuration.

When available, Git performance features should be detected and leveraged automatically.

Configuration changes remain an explicit user decision.

### Why split the work into two PRs?

Issue #2229 requires a small, low-risk fix.

The execution model affects a much larger portion of the application and deserves an independent review.

Separating them reduces regression risk and simplifies code review.

bashrusakh:
## Intent

Git operations from the web server and the VS Code extension now share one bounded execution path per runtime: canonical repository and worktree identity resolved before admission, conflict-aware lanes (reads, worktree writes, common writes, topology writes, network work, clone destinations), finite queues, cancellation and timeouts that reach the git process, coalesced status reads, and cleanup after idle periods.

This is a current-main reimplementation of #2276 (which included #2232). The branch was reconciled once against current main (`961f1611c`) so it composes with the work that landed after it was opened, including #2693, #3296, #3319, #3302, #3383, #3443, #3426, #3365, and #3474.

Closes #2229
Closes #2233

## Scope boundaries

- The scheduler is execution infrastructure, not PR-source authority. It decides when git may run, not which PR source is correct.
- This PR does not redefine linked-PR semantics.
- Current-main #2693 forked-PR worktree behavior is preserved: the UI resolves a linked PR into `ensureRemoteName`/`ensureRemoteUrl` plus existing-branch args, the server runs `ls-remote` on validate and provisions plus fetches on create, and a failed fetch leaves branch tracking unset. There is no `refs/pull` fallback in this PR and none was added.
- No UI, persisted format, API envelope, or bridge payload change.

## Network admission for fork and PR-source traffic

Every current-main path that can touch network or shared refs takes network admission before git runs:

- worktree validate/create when `setUpstream`, the `ensureRemoteName`/`ensureRemoteUrl` pair, or a remote-like `existingBranch`/`startRef` is present;
- the upstream fetch during background bootstrap;
- checkout of a remote-like branch (`checkoutBranch` probes the configured remote as a read, then upgrades the checkout to common-write plus network when it can fetch);
- branch discovery through `ls-remote`;
- the integrate flow, whose fast-forward step can `git fetch`. This was a gap found during reconciliation: `integrateWorktreeCommits` now reserves network capacity for its whole run.

Background worktree attachment and bootstrap are scheduled through the same executor and inherit the outer network flag, so they cannot bypass admission or deadlock by reacquiring an outer reservation.

## Reconciliation with current main

The previous head predated 273 commits of main. The merge resolved six conflicting files and kept both sides: main's worktree topology refresh, hunk apply, diff exit-code handling, checkout fetch behavior, credential-helper flag handling, missing-directory soft handling, background worktree remove, and source prefetch, plus this branch's env/signal plumbing, structured not-a-repository errors, and background scheduler. No main behavior was reverted.

## Consumer coverage

Every git process path is either bounded through the coordinator or a documented intentional exclusion. The accidental gaps found during reconciliation were closed: `getCommitDiff` and `getUnpushedBranchCounts` are read-lane operations on the web side, and VS Code's `getGitUnpushedBranchCounts` is wrapped and exported so the bridge resolves it.

Intentional exclusions, documented in the git module docs: the best-effort topology observer (local-only and never throws), in-memory subscription/status helpers, and context discovery, which has its own bounded queue and timeouts.

## Validation

| Check | Result |
|---|---|
| Focused web suites (git execution, coordinator, context resolver, service, routes, walkthrough, fs, skills, notifications) | 338 tests passed |
| Facade-level fork tests (reachable fork, missing URL, unreachable fork with stale ref, lease release, queued and running cancellation, fast-create bootstrap) | 7 passed |
| Full web suite | 2319 passed, 2 skipped, 1 pre-existing load flake (`settings-helpers` tarball test, passes alone) |
| Full VS Code suite | 46/47 files pass; the one failure (`webview/api/settings.test.ts` timeout) reproduces on clean main |
| `bun run type-check:web` and `bun run lint:web` | pass |
| `bun run vscode:type-check` and VS Code lint | pass |
| `bun run docs:validate` | pass (517 pages, 47 sidebar links) |
| `bun run dead-code` | exits 0 with the existing non-blocking report |
| `git diff --check` | clean |

An independent verification pass reran the affected suites at this head, and an independent review approved the full base-to-head change with no blocking findings. OCR was attempted but the provider rate-limited it, so the review is native.

## Non-blocking review notes

- 94 anti-slop lint findings remain in the newly added files (mostly symbol naming and boundary `typeof` checks). The check is not wired into CI; this belongs in a separate cleanup pass.
- `GitExecutionCoordinator.invalidateWorktrees` is currently unused.
- A resolver late joiner can receive a structured cancellation in a narrow race.
- `queueTimeoutMs` forwarding is unit-asserted; there is no end-to-end facade assertion yet.

## Cross-PR note

The rebuilt #2468 residual (VS Code tracking parity plus `--` remote argument hardening) merges into this branch cleanly. A temporary local integration merge produced zero conflicts and passed both scopes' focused suites and type-checks, so whichever PR lands second needs a mechanical rebase only.

## Visual evidence

No visual change. The diff changes server and extension-host git scheduling, process execution, and test seams while preserving existing API and bridge envelopes and rendered UI behavior.

## Risks and failure behavior

- Same-worktree reads and writes serialize. Independent worktrees proceed unless they share refs, topology, network capacity, or a clone destination.
- A cancelled waiter does not cancel work another waiter still needs. The last waiter cancels the underlying source, and coordinator state remains until the process or task closes.
- Clone cleanup removes only destinations owned by that clone.
- Discovery, permission, missing-git, malformed output, range, timeout, and cancellation failures stay explicit. Only confirmed non-repository results use the soft fallback.
- A per-file range failure is an explicit flow failure, not a partial success. Covered by tests.


bashrusakh:
Opened openchamber/openchamber#3855 to agree on the shape before reworking this branch.

If the direction is agreed, the plan is to split it into (1) web-server bounded execution, (2) VS Code bridge parity, and (3) the fs/search hardening as its own PR, then rebase the first slice onto `main`. This branch stays as the reference until then.

PR #3257, opened 2026-08-31
fix(session-goal): auto-continue length-truncated assistant turns

### LINKED ISSUE #3255: Goal mode stalls on 'Evaluating…' when reasoning model hits token limit (length stop)
### What's wrong?

When running a session with **Session Goal** mode enabled using a reasoning/thinking model (such as Qwen 3.8/QwQ, DeepSeek-R1, or other thinking-enabled models), if the model exhausts its completion token limit during the reasoning phase before emitting `text` parts:

1. The assistant turn completes with `finish: "length"` (or `stopReason: "length"`) and contains only `reasoning` parts (`part.type === 'reasoning'`), with zero `text` parts (`part.type === 'text'`).
2. OpenChamber's `session-goal/runtime.js` extracts `assistantText` using `messagePartsToText(lastAssistant)`, which only maps `part.type === 'text'`. `assistantText` is therefore empty (`""`).
3. The goal progress audit receives an empty turn string (`The agent's latest turn: \n\n`), causing audit failures or no-op behavior.
4. Because `finish === 'length'` is not treated as a turn error or auto-continuation trigger (unlike `MessageAbortedError` or `error`), the goal runtime doesn't auto-continue and the session goes into `idle` status while `goal.status` remains `active`.
5. The UI (`SessionGoalRow.tsx`) enters a perpetual spinning state showing **Evaluating…** because `goal.status === 'active' && sessionStatus.type === 'idle'`, and the user is forced to manually send a prompt to resume progress.

---

### Steps to Reproduce

1. Start a session in OpenChamber with an active Goal.
2. Use a reasoning model configured with high thinking effort or a constrained `max_tokens` output limit on a complex generation task.
3. Allow the model to hit the max output token limit while still inside its reasoning block (`finish: "length"`, 0 text parts).
4. Observe the session once generation stops:
   - Session status becomes `idle`.
   - The goal strip above the composer displays spinning **Evaluating…** indefinitely.
   - The session does not auto-continue until manually prompted.

---

### Where does it happen?

Desktop (macOS) / Desktop Web

---

### Potential Fix / Recommendations

1. **Detect `finish === 'length'` / `stopReason === 'length'` in `session-goal/runtime.js`**:
   - Check if `lastAssistantInfo.finish === 'length'` or `lastAssistantInfo.stopReason === 'length'`.
   - If the turn was cut off by token limit (especially when `assistantText` is empty or truncated), either:
     - Skip the small-model progress audit (since the turn was incomplete) and send an immediate continuation prompt nudging the agent to resume implementation:
       ```
       Your previous response reached the output token limit before completing. Continue working toward the active session goal:
       <objective>
       ...
       </objective>
       ```
2. **Fallback for `reasoning` parts in `messagePartsToText`**:
   - If an assistant message contains no `text` parts but has `reasoning` parts, provide appropriate fallback text or mark the turn as incomplete so the audit isn't passed an empty string.
3. **Guard UI `Evaluating…` spinner**:
   - If no audit is in-flight or the audit fails, reflect an appropriate idle/retry state instead of spinning continuously.

alvins82:
## Intent

Resolves #3255.

When an assistant message finishes with `finish: 'length'` (or `stopReason: 'length'` / `error: { name: 'MessageOutputLengthError' }`) before completing its response—such as when a reasoning model generates extensive thought and exhausts its completion token ceiling—it frequently yields zero `text` parts.

Previously, `session-goal/runtime.js` fell through to the small-model audit with an empty transcript (`assistantText = ""`), which caused audit failures and left the session idle with an active goal, locking the UI into an indefinite `Evaluating…` spinner until manually prompted.

This PR treats length-truncated assistant messages as in-progress turns (matching compaction summaries and resumed aborted turns):
1. Detects `lengthTail` via `finish === 'length'`, `stopReason === 'length'`, or `error.name === 'MessageOutputLengthError'`.
2. Bypasses the small-model progress audit (which has no complete reply to judge).
3. Directly dispatches a continuation prompt while accounting tokens and incrementing `turnsUsed`.

## Non-goals

- Altering client/provider `max_tokens` configuration.
- Modifying UI components in `packages/ui` (the UI naturally exits `Evaluating…` once the continuation prompt is dispatched and the session resumes `busy`).

## Affected surfaces

- `packages/web` (`packages/web/server/lib/session-goal/runtime.js`)

## Repository guidance

|---|---|---|
| `packages/web/server/lib/session-goal/DOCUMENTATION.md` | Defines the lifecycle, terminal conditions, and audit exceptions for session goals. | Follows the documented exception pattern where in-progress turns (compaction summaries / length stops) bypass progress audits and continue unconditionally. |

## Validation

| Check | Result |
|---|---|
| `bunx vitest run packages/web/server/lib/session-goal/runtime.test.js` | 6 passed (including new tests for `finish: 'length'` and `MessageOutputLengthError`) |
| `bun run --cwd packages/web test` | 157 test files passed, 1548 tests passed |
| `bun run type-check:web` | Passed (0 errors) |
| `bun run lint:web` | Passed (0 errors) |

## Visual evidence

No frontend UI components were modified. Behavior change is in the backend session-goal loop: length-truncated turns now immediately dispatch a continuation prompt and transition to `busy` instead of remaining in `idle` with a spinning `Evaluating…` indicator.

## Risks and failure behavior

None identified. Auto-continuation remains strictly bounded by `turnsUsed >= maxAutoTurns` (20 turns max) and `tokensUsed >= tokenBudget`.

### DIFF
--- a/packages/web/server/lib/session-goal/DOCUMENTATION.md
+++ b/packages/web/server/lib/session-goal/DOCUMENTATION.md
@@ -111,10 +111,11 @@ before touching the filesystem). Rationale: metadata rides every
    - terminal checks, cheapest first: assistant turn error → `blocked`;
      `tokensUsed >= tokenBudget` → `budgetLimited`;
      `turnsUsed >= MAX_AUTO_TURNS` (20) → `blocked`;
-   - if the latest message is a compaction summary, skip the audit and
-     continue unconditionally — running into the context window mid-work is
-     by definition "in progress, not finished" (the summary is a retelling,
-     not evidence, and must not be judged);
+    - if the latest message is a compaction summary or was cut off by output
+      token limits (finish: "length"), skip the audit and continue
+      unconditionally — running into the context/output window mid-work is
+      by definition "in progress, not finished" (a truncated reply is not
+      evidence, and must not be judged);
    - otherwise, small-model audit of the objective + the last assistant turn
      only — no conversation history and no continuation prompts
      (`restrictToPreferredProvider`, session's own provider/model preferred):
--- a/packages/web/server/lib/session-goal/runtime.js
+++ b/packages/web/server/lib/session-goal/runtime.js
@@ -608,6 +608,9 @@ export const createSessionGoalRuntime = ({
     // falls through to the continuation below (skipping the audit — an
     // aborted reply is not evidence of anything).
     const abortedTail = lastAssistantInfo.error?.name === 'MessageAbortedError';
+    const lengthTail = lastAssistantInfo.finish === 'length'
+      || lastAssistantInfo.stopReason === 'length'
+      || lastAssistantInfo.error?.name === 'MessageOutputLengthError';
     if (abortedTail && goal.statusReason !== 'resumed') {
       await writeGoal(sessionId, directory, goal.id, () => ({
         status: 'paused',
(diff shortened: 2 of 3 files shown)


btriapitsyn:
Thanks for identifying and fixing the length-truncation path. Your original change is included in #3278, together with bounded retries and error-precedence fixes. Closing this PR in favor of that merged replacement; your contribution is retained in the merge credit.

PR #3274, opened 2026-08-31
fix(ui): stop settings save echoes from clobbering theme preferences

kydorn:
## Intent

Fixes the reported "theme flashes from my theme to OpenChamber when I switch sessions" bug (intermittent; same family as the previously reported "color mode forced from system to light").

Root cause: every `updateDesktopSettings` save replays the server's **full** settings document as an `openchamber:settings-synced` event, and `ThemeSystemContext.handleSettingsSynced` adopted it as full theme authority. Switching to a session in another directory writes `lastDirectory`/`activeProjectId`, so every cross-directory session switch sent a settings PUT whose echo could flip the window's theme to whatever the server document held at that instant (stale, another window's, or defaults). Two cooperating defects:

1. **Save echoes hijacked the theme.** The event already carried a "bootstrap-grade only" flag (`adoptWorkspace`) for workspace pointers; the theme listener ignored it.
2. **Missing fields were invented as defaults.** `materializeAuthoritativeUiSettings` filled `useSystemTheme: true` + `openchamber-light/dark` for any theme field absent from the server response; the persist effect then wrote those defaults back to the server and the per-runtime scoped localStorage entry, poisoning them and making the flip recur. This also explains the earlier system→light symptom.

New behavior: theme is adopted only from bootstrap-grade syncs (page boot, first auth, runtime endpoint switch); a mid-session save echo never changes the theme; theme fields absent from the server document mean "not set" and keep the current preference instead of forcing defaults. Users no longer need to re-pick their theme after upgrading or after switching sessions.

Follow-up to #2897 (per-instance theme scoping), which made windows legitimately able to hold a theme the server never stored — turning every save echo into a clobber opportunity.

## Non-goals

- No change to how theme preferences are written (scoped localStorage key + `updateDesktopSettings` persist effect) or to the settings document format.
- No change to cross-window same-instance theme sync (still rides the scoped-key `storage` event) or to embedded/iframe theme sync (postMessage path untouched).
- The `useSystemTheme`/`themeVariant` semantics of the settings document, and the server-side `migrateSettingsFromLegacyThemePreferences` back-fill, are unchanged.
- The `autoSaveEnabled` seed-once pattern in the same materializer is deliberately not applied to theme (a single window seeding the backend would overwrite other instances' themes); a comment in the diff states this.

## Affected surfaces

- `packages/ui` only: `persistence.ts` (settings sync dispatch + materializer), `ThemeSystemContext.tsx` (theme adoption), `theme-storage.ts` (new pure resolver), plus two one-line consumer renames (`useProjectsStore.ts`, `SessionAuthGate.tsx`).
- Runtimes sharing this UI: Web, Desktop (Electron), VS Code webview, Mobile/Capacitor, mini-chat. The gating applies identically; the iframe/mini-chat postMessage theme-sync path is separate and untouched. VS Code's settings bridge returns a derived `themeVariant` and possibly no light/dark ids — with this fix, bootstrap keeps current preferences instead of forcing openchamber defaults (previously a real flip source).
- Event contract `openchamber:settings-synced` (`SettingsSyncedDetail`): field renamed `adoptWorkspace` → `bootstrap`, same value semantics, now also gating theme adoption. Internal UI-only event; the three listeners are all in this diff. No persisted/external contract changes.
- User-visible state: the active theme/colors on session and directory switches; nothing else.

## Repository guidance

|---|---|---|
| `sync-state-invariants` (skill + AGENTS.md invariants) | The bug is an authoritative-settings reconciliation clobbering live client-owned state; bootstrap/reconnect and snapshot ordering rules apply | Absent persisted fields are treated as "not set", never as authoritative reset-to-defaults; adoption is gated to bootstrap-grade syncs; identity-check returns the same reference to break adoption loops; failed GETs dispatch no event, so fetch failure cannot masquerade as authoritative empty success; the resolver is a pure reducer owned by the theme-storage module and covered by tests |
| `openchamber-change-discipline` | Source change in shared UI with a cross-file event contract rename | Smallest complete change: the handler got simpler (net deletion), the rename is mechanical across the 3 listeners + 2 dispatch sites, tests are additive; validation run at the owning-package level |
| `theme-system` | Change is in the theme authority path | No styling/token/icon changes; only preference resolution; the iframe `receivesParentThemeSync` paths are untouched |
| `ui-api-decoupling` | The settings-sync event and `syncDesktopSettings` are shared UI data access | Runtime-agnostic: gating is on the sync grade, not on any runtime; VS Code/desktop/mobile flows traced (endpoint switch still adopts the new instance's server theme via bare `syncDesktopSettings()`) |
| `performance-engineering` | The old handler was on a hot event path | The resolver returns early (`!detail?.bootstrap`) before any work for save echoes — the common case; same-reference no-ops prevent re-renders |

## Validation

| Check | Result |
|---|---|
| `bun test src/contexts/theme-storage.test.ts src/lib/persistence.test.ts src/contexts/ThemeSystemContext.test.ts src/components/auth/SessionAuthGate.behavior.test.tsx` (in `packages/ui`) | 69/69 pass, including 6 new regression tests: non-bootstrap echo never changes preferences; bootstrap adopts server theme; omitted fields keep current (not-set ≠ defaults); identity no-op returns null; GET with missing theme fields dispatches no invented defaults; save echoes dispatch `bootstrap: false` |
| Full `packages/ui` suite (`bun run test` in `packages/ui`) | 369/369 files pass after the rider commit below; before it, 368/369 with the only failure being the tr-locale key-parity test, which fails identically on `upstream/main` HEAD (0e3aff884) |
| Rider commit `fix(i18n): restore tr locale key parity` | `tr.ts` was missing 4 `gitView.empty.*` keys, keeping the shared `checks` workflow red on main itself. Carried here (additive, 4 translated strings, no code touched) to turn the PR green; drop it if you'd rather land it separately. Re-ran `messages.test.ts`: 2/2 pass |
| `bun run type-check` (packages/ui, `tsc --noEmit`) | Clean |
| `bunx oxlint` on all changed files vs HEAD baseline | Zero new findings (persistence.ts 292→292 pre-existing backlog; ThemeSystemContext 38→37; theme-storage.ts 0; others unchanged); findings in newly authored code fixed, none suppressed |
| `bun run dead-code` (workspace root) | No entries for changed exports; only pre-existing backlog items |
| Independent subagent PR review | Verdict MERGE; verified the rename left zero `adoptWorkspace` references, no consumer of the removed materializer fields, endpoint-switch adoption intact, and that the rewrite incidentally fixed a latent whitespace-only-id bug; its one non-blocking doc-comment finding is applied |

**Not verified:** live multi-window/multi-instance desktop and web runtime behavior (the flip was intermittent to the reporter and I could not capture it on demand). Worst-case failure is a cosmetic no-op (theme keeps its current value); one-commit revert. Root-wide `bun run lint` / `build` were not run; package-scoped type-check, lint, and the full ui suite were.

## Visual evidence

The before state is an intermittent flash (theme flips for a moment after clicking a session in another directory) that depends on the server settings document disagreeing with the window — not capturable as a static screenshot without staging a poisoned settings document. The after state is the absence of the flip and is covered by the deterministic regression tests above (echo with a differing server theme → preferences unchanged; bootstrap sync with missing fields → preferences unchanged). The theme's rendered appearance itself is unchanged by this diff — no tokens, styles, or default themes are touched. If a recording of the repro is required, say so and I'll stage a fresh instance with an empty settings document and capture the before/after.

## Risks and failure behavior

- **Behavior change worth noting:** a mid-session settings save echo no longer updates the theme. Legitimate theme propagation paths are unaffected: local changes (persist effect), cross-window same-instance (scoped-key `storage` event), runtime endpoint switch (scoped-entry adoption + bootstrap sync). Cross-device convergence for the *same* instance now waits for the next bootstrap — this was already true for workspace pointers and matches the per-instance theme model.
- **Rollback:** single commit; revert restores the previous (buggy) behavior cleanly. No data migrations.
- **Data:** no persisted format changes; the settings mirror keeps storing theme fields only when the server provides them (undefined fields are dropped by `JSON.stringify`, same as before for absent fields).
- **Failure modes:** an unknown theme id arriving outside the sanitized payload fails theme lookup downstream and falls back cosmetically to the variant default; a failed settings GET dispatches no event, so state is preserved.
- **Security:** no new inputs trusted; the resolver consumes the already-sanitized settings document and only within bootstrap-grade events.


btriapitsyn:
I updated this branch on top of the current `main` and resolved the Turkish locale conflict.

I also addressed both review findings:

- VS Code settings broadcasts now keep workspace-pointer synchronization without replacing each webview's theme.
- The settings-store documentation now describes the theme authority exception and the VS Code behavior.

Added regression coverage for bootstrap settings syncs that intentionally preserve the current window theme.

Validation completed:

- UI tests: 375/375 files passed
- VS Code tests: 27/27 files passed
- UI and VS Code type-checks passed
- UI and VS Code lint passed, with one unrelated existing UI warning
- VS Code build passed
- Documentation validation passed

The focused anti-slop run still reports existing findings in the large persistence and VS Code entrypoint files. The lines changed here introduced no new anti-slop finding.

btriapitsyn (approved):
I updated this branch on top of the current `main` and resolved the Turkish locale conflict.

I also addressed both review findings:

- VS Code settings broadcasts now keep workspace-pointer synchronization without replacing each webview's theme.
- The settings-store documentation now describes the theme authority exception and the VS Code behavior.

Added regression coverage for bootstrap settings syncs that intentionally preserve the current window theme.

Validation completed:

- UI tests: 375/375 files passed
- VS Code tests: 27/27 files passed
- UI and VS Code type-checks passed
- UI and VS Code lint passed, with one unrelated existing UI warning
- VS Code build passed
- Documentation validation passed

The focused anti-slop run still reports existing findings in the large persistence and VS Code entrypoint files. The lines changed here introduced no new anti-slop finding.

PR #3251, opened 2026-08-30
fix(i18n): add missing Turkish Git discovery text

karimodm:
## Intent

Restore locale key parity by translating the four Git repository discovery strings missing from the Turkish dictionary. This also restores the repository-wide test command on current `main`.

## Non-goals

No Git discovery behavior or text in other locales changes.

## Affected surfaces

`packages/ui` Turkish localization only. Web, desktop, VS Code, and mobile share this dictionary. The persisted and network contracts are unchanged.

## Repository guidance

| Guidance | Why it applies | How the change complies |
|---|---|---|
| `openchamber-change-discipline` | This repairs a shared UI source file. | The change adds only the four keys already present in every other dictionary and runs the owning parity test. |
| `locale-ui-patterns` | The values are visible Turkish UI text. | Each missing key has a real Turkish translation; no English placeholder was added. |
| `packages/ui/src/lib/i18n/messages.test.ts` | This test owns cross-locale key parity. | The focused test passes 2/2. |

## Validation

| Check | Result |
|---|---|
| `NODE_PATH=/home/karimo/projects/openchamber-pr3055/node_modules bun test packages/ui/src/lib/i18n/messages.test.ts` | Passed, 2 tests and 48 assertions. |

## Visual evidence

Not captured. The change fills missing translations in an existing Git discovery state and does not change layout, styling, or interaction.

## Risks and failure behavior

Low risk. The keys and interpolation shape already exist in English and every other locale. The Turkish values contain no parameters. Rollback removes only these four dictionary entries.

PR #3293, opened 2026-09-01
fix(chat): bound oversized tool diff previews

### LINKED ISSUE #3292: [Bug] Oversized apply_patch metadata freezes a session when edit tools open by default

Opening one specific session in OpenChamber 1.22.0 locks the browser. Other sessions remain responsive.

The session is not unusually large: 169 messages and 826 parts. The trigger is three completed `apply_patch` parts in the initial message window. Each part is 3.93 MB because OpenCode stores the same 1,905,978-character, 45,742-line patch in both `state.metadata.diff` and `state.metadata.files[0].patch`. The visible tool output is only 97 characters.

With `Show expanded edit tools` enabled, `ToolPart` expands those historical tools on mount. `getDiffPatchEntries` parses each full metadata patch with `@pierre/diffs`, then `PatchDiff` attempts to render all 45k lines on the browser main thread. The existing 512 KiB tool-output cap does not apply because the large value is metadata, not `state.output`.

Steps to reproduce:

1. Enable `Show expanded edit tools`.
2. Open a session whose initial message page contains several completed `apply_patch` parts with metadata patches around 1.9 MB and 45k lines each.
3. The tab becomes unresponsive while mounting the expanded tool diffs.

Expected: oversized historical patches use a bounded plain-text preview and the session stays usable. The full patch should remain available to file/diff navigation.

This is narrower than #2265, which capped oversized tool output, and #2267, which tracks whole-history mounting. It is also a different path from #3286, whose stack is the context-panel `DiffView`.

### Where does it happen?

Desktop Web

### Version (if known)

1.22.0

No recording. The affected tab stops responding.

### Logs (optional)

No console error is emitted before the main thread locks.

karimodm:
## Intent

Loading session `ses_fa2f230f4ffe6HRGJ7WYQRof8y` freezes Chrome when its expanded historical `apply_patch` tools process 1.9 MB patches through `@pierre/diffs`. This PR keeps oversized patches out of rich parsing and bounds their inline plain-text preview to 256 KiB or 2,000 lines. File navigation still receives the complete original patch.

Closes #3292.

## Non-goals

- This does not change persisted tool metadata or network payload size.
- This does not virtualize message history or change normal-sized diff rendering.
- This does not change the context-panel `DiffView` path.

## Affected surfaces

- `packages/ui` tool-card rendering in web, desktop, VS Code, hosted mobile, and Capacitor mobile because they share `ToolPart`.
- Oversized `apply_patch`, edit, multiedit, and synthetic write previews fall back to bounded plain text. Normal-sized patches keep the current rich preview.
- Persisted data and external contracts are unchanged. The UI retains the original patch string for quick-open and VS Code diff navigation.

## Repository guidance

|---|---|---|
| Root `AGENTS.md` and `openchamber-change-discipline` | This changes shared executable UI source and adds a focused module. | The change stays in the owning tool-card module, preserves normal-sized behavior, updates the owning documentation, and uses package-scoped tests/checks plus `dead-code`. |
| `performance-engineering` | The report is a production-scale renderer freeze. | The exact session was reproduced in an authenticated production build. The fix structurally skips rich parsing at both metadata parsing and the shared preview boundary, then validates the same 1.9 MB patch. |
| `theme-system` | `PlainDiffFallback` is a shared UI component. | Existing semantic syntax tokens and component styling are unchanged. No color, control, icon, or animation was added. |
| `packages/ui/src/components/chat/message/parts/DOCUMENTATION.md` | It owns the lazy diff-rendering and fallback invariant. | The rich stack remains lazy-only, and the document now records the oversized-patch boundary and full-patch navigation behavior. |

`packages/ui` has no package README. `CONTRIBUTING.md` and `.github/PULL_REQUEST_TEMPLATE.md` were also reviewed before submission.

## Validation

| Check | Result |
|---|---|
| `bun test packages/ui/src/components/chat/message/parts/toolDiffUtils.test.ts` | Passed: 21 tests, 0 failures. Covers character and line thresholds, trailing newline semantics, bare CR separators, rich-render fallback, and full-patch navigation. |
| `bunx oxlint packages/ui/src/components/chat/message/parts/toolDiffPreview.ts packages/ui/src/components/chat/message/parts/PlainDiffFallback.tsx packages/ui/src/components/chat/message/parts/toolDiffUtils.test.ts` | Passed with no findings. Whole-file anti-slop runs for legacy `ToolPart.tsx` and `toolDiffUtils.ts` still report their existing backlog; the added lines have no finding. |
| `bun run build:web` | Passed. Vite production build and PWA service worker build completed. Existing font, dynamic-import, and chunk-size warnings remain. |
| `bun run dead-code` | Completed. It reports the existing repository-wide unused-export backlog and does not flag the new preview module. |
| Strict code review | Passed after fixing all findings: direct synthetic write previews are gated, CR/LF/CRLF line counting is bounded, and a terminal separator does not create a phantom line. |

### Production-scale browser check

- Unchanged 1.22.0 production build: the exact session stopped answering Chrome DevTools Protocol probes immediately after its 14.5 MB message response completed. Two valid runs reproduced the boundary.
- Same build and response with only patch metadata over 1 MB replaced in transit: responsive for 15 seconds after the response, with 58 messages and 130 tool rows mounted.
- Patched production build with the unmodified response: responsive for 15 seconds after the response, with 143 messages and 304 tool rows mounted.
- The production patch is 1,905,978 characters and 45,742 lines. Current classification measured a 0.569 ms median and 2.684 ms maximum; the inline preview was bounded to 72,377 characters while the original remained available to navigation.

The browser replay covered the exact LF-separated `apply_patch` path before the final direct-write and newline-edge additions. Those additions do not change that measured path. The final HEAD was rebuilt and its added edge behavior is covered by the focused tests. A final-HEAD browser replay was not repeated.

## Visual evidence

There is no useful static before/after screenshot: before the fix the rendered page stops responding, while after the fix the same tool card remains interactive and shows a plain-text patch ending in an ellipsis. The CDP liveness checks and mounted-row counts above capture that behavioral difference. Normal-sized tool cards and styling do not change.

## Risks and failure behavior

- The limits intentionally trade rich syntax rendering for responsiveness only when a patch exceeds 256 KiB or 2,000 lines.
- The guard runs before `@pierre/diffs` parsing for metadata patches and again at the shared preview boundary for synthetic write patches and future direct callers.
- Preview generation scans at most 256 KiB and handles LF, CRLF, and bare CR. It does not allocate a full split-line array.
- Quick-open and VS Code navigation keep the original patch. No data is truncated in storage or transport.
- Rollback is the single commit. No migration, cleanup, security, or data-loss behavior changes.


karimodm:
The `checks` failure is unrelated to this diff. `packages/ui/src/lib/i18n/messages.test.ts` fails identically against current `upstream/main` because Turkish is missing four `gitView.empty.*` keys; #3293 does not touch locale files.

#3251 already fixes that baseline failure. Once it lands, rerunning checks here should clear the remaining red status.

PR #3295, opened 2026-09-01
fix(terminal): delete NODE_CHANNEL_FD and BUN_WATCH_PID from PTY env

### LINKED ISSUE #2752: [Bug] Terminal exports NODE_CHANNEL_FD and breaks Node IPC workers

Commands that use Node IPC workers fail in the OpenChamber terminal.

For example:

```shell
oxfmt --check README.md
```

fails with:

```text
TypeError: Cannot read properties of undefined (reading 'bind')
    at node_modules/tinypool/dist/entry/process.js:12
```

The failing worker contains:

```js
const send = process.send.bind(process);
```

The OpenChamber terminal exports an empty IPC variable:

```shell
env | rg NODE_CHANNEL_FD
# NODE_CHANNEL_FD=
```

The issue reproduces outside OpenChamber with:

```shell
env NODE_CHANNEL_FD= oxfmt --check README.md
```

Expected behavior: `oxfmt` and other Node-based tools should run normally in the integrated terminal.

### Where does it happen?

Desktop Web, Linux

### Version (if known)

- OpenChamber: 1.18.1
- Node.js: v24.19.0
- Bun: 1.3.14
- Linux x86_64

### Logs (optional)

```shell
file:///.../node_modules/tinypool/dist/entry/process.js:12
const send = process.send.bind(process);
                          ^

TypeError: Cannot read properties of undefined (reading 'bind')

Error: Worker exited unexpectedly
    at ChildProcess.onUnexpectedExit (.../node_modules/tinypool/dist/index.js:79:30)
```

### Root cause

`packages/web/server/lib/terminal/runtime.js` sets:

```js
env.NODE_CHANNEL_FD = '';
```

This prevents the OpenChamber daemon IPC descriptor from being inherited by the PTY, but the empty exported variable is also inherited by every command run in the terminal.

Node only initializes child-process IPC when `NODE_CHANNEL_FD` is non-empty. Consequently, Node processes launched with `child_process.fork()` or `stdio: 'ipc'` receive an IPC channel but do not initialize `process.send`.

This affects tools using child-process workers, including:

- `oxfmt` through Tinypool
- Jest process workers
- Vitest forks pools
- Other tools using `child_process.fork()` or Tinypool `runtime: 'child_process'`

Ordinary `spawn`, `exec`, `execFile`, and `worker_threads` are unaffected.

### Workaround

```shell
env -u NODE_CHANNEL_FD oxfmt --check README.md
```

### Related

This appears related to [PR #2481](https://github.com/openchamber/openchamber/pull/2481), but the current empty-string approach still exposes `NODE_CHANNEL_FD` to the user shell.

The variable should be unset before the final PTY shell starts, while preserving the existing protection against leaking the daemon IPC descriptor.

khafaji-ahmed:
## Intent

Prevent host-only IPC variables from reaching integrated-terminal shells. A stale or empty `NODE_CHANNEL_FD` makes Node tools treat IPC as configured even though the PTY has no valid channel, which can make worker-based tools fail on exit.

Fixes #2752. Supersedes #2481.

## Change

- Clear `NODE_CHANNEL_FD` and `BUN_WATCH_PID` from every PTY child environment.
- Extend the existing POSIX `env -u` wrapper so `bun-pty` cannot reintroduce either variable from the native environment.
- Preserve the established `ARGV0` cleanup and the direct Windows launch path.
- Cover wrapper behavior, real child-environment isolation, fork-worker IPC, terminal creation, restart, and command launches.

This remains one focused commit with no UI, protocol, persistence, dependency, or Electron changes.

## Validation

| Check | Result |
|---|---|
| Focused inherited-environment and terminal tests | 3 files, 62 tests passed |
| Type check | Passed |
| Web lint | Passed |
| Production web build | Passed; existing Vite asset and chunk warnings only |
| Full web test suite | 2,082 passed, 1 skipped; 1 unrelated host-dependent failure in `server/lib/opencode/env-runtime.test.js` because this machine has `/opt/homebrew/bin/opencode` installed |
| Diff and merge review | Clean; rebased onto current `main` and mergeable |

The regression tests execute the wrapper with `NODE_CHANNEL_FD=3` and `BUN_WATCH_PID=999`, verify that neither reaches the shell, and verify that a forked worker still has a functioning IPC channel.


btriapitsyn:
The Linux wrapper, seeded regression assertions, and PR description now address the previous review items. One part of the reported bug remains: the reporter confirmed reproduction on both macOS and Linux in https://github.com/openchamber/openchamber/issues/2752#issuecomment-5216308236.

1. In `packages/web/server/lib/inherited-env.js` and its caller in `packages/web/server/lib/terminal/runtime.js`, make the exec-time removal of `NODE_CHANNEL_FD` and `BUN_WATCH_PID` cover macOS PTY launches too. The Bun serve daemon is started with IPC on macOS, but `resolveLinuxPtyLaunch` currently returns the shell unchanged there, allowing the native-environment merge to restore the daemon's IPC variable. Preserve shell arguments and the existing Linux ARGV0 handling. Both terminal creation and restart must launch the final shell with these IPC variables absent.
2. Extend `packages/web/server/lib/inherited-env.test.js` and `packages/web/server/lib/terminal/runtime.test.js` for the macOS launch contract, and add or record a real Bun PTY regression with the host IPC variable present. The final shell must report both variables absent, and a Node `child_process.fork()` worker must successfully use `process.send`. Update `packages/web/server/lib/terminal/DOCUMENTATION.md` and the PR's validation and limitations to match the completed behavior.

Thanks for working through the Linux fix. Covering the reported macOS path will make this ready to close #2752.

PR #3296, opened 2026-09-02
feat(worktrees): fetch remote source branch before worktree creation

jtatum:
## Intent

New worktrees branched from the root's local state, so creating one from a root that was behind its upstream produced a stale worktree. With this change OpenChamber fetches the root's current branch from its upstream before creating and bases the new branch on the fetched remote-tracking ref. Fetching is gated by a new global setting (on by default) at **Settings → Behavior → Worktrees → "Fetch remote before creating worktrees"**.

Failure semantics are intentional and uniform across runtimes:

- Fetch failures are soft: a warning toast appears and creation proceeds from the local state. A failing fetch never blocks or mutates worktree creation.
- Every non-success fetch result is treated as failure, not just thrown errors — the VS Code runtime bridge resolves `{ success: false }` instead of throwing, so an offline fetch now takes the same warn-and-fall-back path as web/desktop/mobile.

A follow-up correctness fix is included: creating a new branch from a remote start ref made git auto-track the base branch (`branch.autoSetupMerge`), so with the fetch enabled every behind-root worktree was born with upstream `origin/<base>` and plain `git push` refused under `push.default=simple`. New-mode creation from a remote-tracking ref now passes `--no-track`, and explicit upstream keys win over the server's remote start-ref inference (aligning the create path with the validate path and the VS Code runtime). A fresh branch ships with no upstream until its first push — the same end state as before this feature — instead of a wrong one.

A second follow-up keeps the failure story soft end to end: the server (and the VS Code runtime) re-fetch the remote start ref before creating. That second fetch previously threw on failure, so a connection dropped between the client fetch and creation rejected the worktree even though the remote-tracking ref was already available locally. The pre-create refresh is now best-effort when the ref exists locally and still mandatory when the ref was never fetched (the remote-only branch case), in both runtimes.

## Non-goals

- No per-project toggle: the setting is global-only by decision; an earlier per-project variant was removed in this branch.
- The worktree bootstrap still refuses to write tracking config for refs that were never fetched (pinned invariant + test); this PR does not carve out an exception for the new branch's own `origin/<name>` upstream.
- The client-side fetch and the server's pre-create `fetchRemoteBranchRef` intentionally fetch the same branch twice per creation; the client fetch is what buys the cross-runtime failure toast. No dedup attempted.
- No new upstream defaults for existing-mode worktrees; existing-mode upstream resolution is untouched.

## Affected surfaces

- **packages/ui**: new `withWorktreeFetchedStartRef` + consolidated `createWorktreeWithDefaults` (`worktreeCreate.ts`) — all five `createWorktree` call sites (dialog, quick create, PR flow, multi-run, session auto-create) flow through the shared default. `useConfigStore` gains `settingsWorktreeFetchSource` (hydrated, persisted, optimistic setter). New Behavior-page toggle + settings-search entry. i18n: 12 locales updated (toggle label/aria/info + failure toast).
- **packages/web**: `worktreeFetchSource` added to the settings sanitizer allowlist (server + client persistence); `git/service.js` passes `--no-track` for new-mode creation from remote refs and applies input-first upstream precedence.
- **VS Code**: consumes the shared UI through the runtime bridge without code changes — the `worktreeFetchSource` key passes through the settings bridge, and `gitFetch` bridge results (`{ success: false }`) are now honored as failures by the shared consumer.
- **Electron / hosted / Capacitor mobile**: no code changes; both halves live in shared UI and the loopback server, so all runtimes inherit the behavior.
- **Persisted contract**: new optional `worktreeFetchSource: boolean` in OpenChamber settings JSON; absent/invalid = enabled (default `true`).
- **Git-facing contract**: `git worktree add` for new branches from remote-tracking start refs now includes `--no-track`.

## Repository guidance

|---|---|---|
| AGENTS.md correctness invariants | "Never let fetch failure masquerade as authoritative empty success"; runtime differences must be intentional | The consumer treats thrown errors and non-`true` results as failure, warns, and falls back explicitly to the original args; VS Code's differing transport shape is handled in shared code rather than papered over |
| AGENTS.md runtime boundaries / `ui-api-decoupling` | Shared UI must call runtimes through `RuntimeAPIs`/`runtimeFetch`, no runtime branching in features | `withWorktreeFetchedStartRef` uses `gitFetch` from `@/lib/gitApi` only; no runtime checks in the feature path |
| `sync-state-invariants` (store authority, optimistic state) | The toggle writes a store the consumer reads at creation time | Store setter is applied optimistically on toggle and on page load, rolled back when the save fails; hydration re-syncs from the server |
| `settings-ui-patterns` | New settings surface | Standard `SettingsCheckboxRow` + `settingsItem` + settings-search registry entry on the Behavior page; immediate save with revert on failure |
| `locale-ui-patterns` | New user-facing text | Real translations in all 12 locales; toast uses the non-React `formatMessage(useI18nStore.getState().dictionary, …)` idiom |
| `openchamber-change-discipline` | Source/exports/contracts changed | `dead-code` run after each stage; `withWorktreeUpstreamDefaults` un-exported when it lost its last external caller; docs-relevant behavior documented in this PR |

## Validation

| Check | Result |
|---|---|
| `bun test src/lib/worktrees/worktreeCreate.test.ts` (packages/ui) | 18/18 pass — gating matrix, fetch, throw- and `{success:false}`-fallback, upstream defaults, store gating |
| `bun run test -- --run server/lib/git/service.test.js` (packages/web, vitest over real git repos) | 88/88 pass — includes: new branch from `remotes/origin/main` ships with no tracking config; inference fallback tracks the base when no explicit keys are given; creation succeeds from an existing remote-tracking ref when the refresh fetch fails; creation still rejects when the ref was never fetched and cannot be |
| `bun test src/gitService.worktree-fetch-fallback.test.js` (packages/vscode, real git repos) | 2/2 pass — same two refresh-failure scenarios against the VS Code runtime implementation |
| `node scripts/run-isolated-tests.mjs packages/ui/src/lib/worktrees packages/ui/src/lib/settings packages/ui/src/stores` | 50/50 test files pass |
| `node scripts/run-isolated-tests.mjs packages/vscode/src packages/ui/src/lib/worktrees` | 31/31 test files pass |
| `bun run type-check` (packages/ui; packages/vscode main + webview configs) | clean |
| `bun run lint` (packages/ui; packages/vscode) | clean |
| `bunx oxlint` on changed logic files | clean; one pre-existing `anti-slop` finding in BehaviorPage (`(error as Error)` AbortError guard, backlog, not from this diff) |
| Manual (maintainer, real clone) | root behind origin → worktree based on fetched remote state; warning toast observed with an unreachable remote |

Not verified: packaged Electron/VS Code builds and hosted/Capacitor mobile were not exercised end-to-end. VS Code behavior is verified at the unit level (bridge result shape) plus the shared-path reasoning above.

## Visual evidence

The user-visible changes are the Behavior-page toggle (a standard `SettingsCheckboxRow` in Settings → Behavior → Worktrees) and the warning toast on fetch failure. Captured against the pull request HEAD, via the running dev server:

Desktop (wide, 1440×900) — Behavior page with the new Worktrees section:

Behavior settings, desktop width

Narrow (768×1024) — same section at narrow settings-pane width:

Behavior settings, narrow width

Scope notes: light theme only — the toggle composes existing theme primitives and this diff changes no colors or tokens; dark mode renders through the same primitives. The failure toast is not screenshotted: forcing it requires a fetch failure against a real tracking remote, which could not be staged in this environment without mutating repository remotes; the maintainer exercised it manually on a live clone and it is covered by unit tests asserting the warning fires.

## Risks and failure behavior

- **Latency**: enabled worktree creation now performs one network fetch; failures degrade to the previous behavior (local base) with a toast rather than blocking. A dropped connection between the client fetch and the pre-create refresh degrades to the already-fetched local remote-tracking ref instead of rejecting creation.
- **Git behavior change**: `--no-track` also affects explicit remote-base creation, which previously auto-tracked the base. Those branches now start untracked; first push prompts to set upstream. Previously plain `git push` refused under `push.default=simple`, so this trades a hard failure for a standard one-time setup step. Rollback: revert restores the auto-track.
- **Compatibility**: the settings key is additive and defaults to enabled; older clients and servers ignore it. The `--no-track` server behavior applies regardless of the client setting, since any remote start ref had the same auto-track hazard.
- **Repo hygiene**: this branch also adds `/.openchamber/` to `.gitignore` — the app's local runtime state directory (sessions db, screenshots) is untracked noise in every dev checkout, and binary screenshots from it were briefly committed to this branch by a routine `git add -A` during development (removed since).
- **One failed entity must not block others**: fetch failure path returns the original args without touching creation flow; upstream application failures are contained in the bootstrap (warn + continue).


jtatum:
Human note - I verified that this works both on the successful path and the failure path, including the toast

yulia-ivashko (approved):

PR #3311, opened 2026-09-03
feat(session-assist): give the suggestion the whole session, and a way to say nothing

pocharlies:
## Intent

The session suggestion is generated from the **last exchange only**, and its instructions forbid it from proposing nothing: *"Pick one best next action yourself"*, *"Output exactly one message the user could click and send without editing"*. Together those two facts mean it cannot know the user's request is already satisfied, and has no way to say so. It always manufactures another step, written in the user's voice and ready to send.

The failure that produces is specific and repeatable. When the assistant honestly reports a limitation, the generator sees that sentence, has no idea what the user originally asked for, and turns the caveat into the next instruction. Measured on a real session: the request was *"open a ticket"*, the ticket was created in message 1, and the session then ran 400 more messages, driven by suggestions the user accepted because there is nothing in the transcript afterwards marking the wording as generated.

This PR changes what the suggestion **sees** and what it is allowed to answer:

- It receives the **whole conversation**, so "already satisfied" is a judgement it can actually make.
- It must return an **empty string** when the request is already satisfied, or when the session is waiting on a decision only the user can make. An empty suggestion is a correct answer, not a generation failure.

The suggestion stays in the **user's voice**. Phrasing it as the agent's own plan ("I would do X") was implemented, measured and reverted: `SessionSuggestionChip` renders the same string it pastes into the composer, so an agent-voice suggestion reads as nonsense the moment the user clicks it.

## Non-goals

- The recap is unchanged.
- The 60-second quiet window, the busy/idle arming, and resolving the model through the session's own provider all already worked and are untouched.
- The client is untouched, and the suggestion's voice is unchanged from today's behavior precisely so that it stays untouched. `SessionSuggestionChip` renders and pastes the same string; nothing about that contract moves.

## Affected surfaces

| Surface | Effect |
|---|---|
| `packages/web/server/lib/session-assist` | All behavior changes are here. `buildTranscript` is newly exported so its budget and prefix invariants can be tested. |
| `metadata.openchamber.assist` | **Shape unchanged** (`recap`, `suggestion`, `forMessageID`, `generatedAt`). Only the meaning of the `suggestion` text changes. No migration: existing payloads stay renderable and go stale through the existing `forMessageID` rule. |
| `packages/ui` consumers | Untouched. `sessionAssistMetadata.ts`, `useSessionAssist.ts`, `SessionRecapSpacer` and `SessionSuggestionChip` parse the same fields. |
| VS Code, mobile, Electron | Unaffected. The watcher only runs in the web server, and every surface renders a payload whose shape did not change. |
| `lib/small-model` | Called with three options it already supports (`onOverflow`, `timeoutMs`, and `describeSmallModel` for the budget). That module is not modified. |
| Cost | Assist input is now proportional to session length rather than constant. See "Risks" for measured numbers. |

## Repository guidance

|---|---|---|
| `openchamber-change-discipline` | Executable source change that alters a documented module invariant ("take the LAST exchange only") and adds an export. | Diff confined to the owning module. No new abstraction beyond the one function the tests need. Validated at package scope for both packages rather than workspace-wide ritual. |
| `session-assist/DOCUMENTATION.md` | Nearest owning doc, and its stated flow said the runtime takes the last exchange only. | Rewritten: what the suggestion is, why the whole conversation is required, the transcript and prefix-cache invariant, and where the budget comes from. |
| `small-model/DOCUMENTATION.md` | Defines the input clamp, `onOverflow`, `context-too-small`, `output-exhausted` and `inputCharBudget`, all load-bearing here. | The clamp truncates the **tail**, which in this prompt is the instruction, so the call passes `onOverflow: 'error'` rather than silently asking the model a question it can no longer see. Both documented failure codes are treated as expected outcomes and skipped without a warning. |
| `AGENTS.md`: "Never add or log secrets, bearer tokens, or sensitive user data" | The transcript is assembled from message parts and sent to a model. | Tool **names only**, never tool input or output. That is precisely where file contents, command lines and credentials would otherwise enter a utility prompt. |
| `AGENTS.md`: "Prefer authoritative state over heuristics" | The transcript budget. | Taken from `describeSmallModel().inputCharBudget` for the model that will answer, not from a constant. |
| `AGENTS.md`: "Follow local code and test precedent" | The one new lint finding, below. | The new `typeof` guard mirrors the identical guard 13 lines above it in the same file. |

## Validation

Run on this PR's HEAD (`cddc48d9`), against a clean `bun install` for this base. The commit was amended after an earlier round, so every number below was re-collected afterwards rather than carried over.

| Check | Result |
|---|---|
| `bun run --cwd packages/web test -- server/lib/session-assist` | **6 passed** (new file) |
| `bun run --cwd packages/web test` | **1825 passed**, 2 skipped, **4 failed** |
| `packages/ui` (`node scripts/run-isolated-tests.mjs packages/ui/src`) | **370/370 test files passed** |
| `bun run type-check:web` | **0 errors** |
| `bun run type-check:ui` | **0 errors** |
| `bun run dead-code` | No `session-assist` or `buildTranscript` entry. The 262 unused exports and 1 unused file are the pre-existing backlog. |
| `bunx oxlint` on the changed files | 27 `anti-slop` findings, against **26 on the same file at this base**. The single new one is disclosed below. |

Three of the four are in `server/lib/git/service.test.js` (`runs the post-checkout hook after populating a created worktree`, `does not fail worktree bootstrap when the post-checkout hook fails`, `getBranches ... prunes refs deleted on the remote (#2098)`). **Verified pre-existing**: stashing this change and re-running that file on a clean tree reproduces the same three. Nothing in this diff touches `lib/git`.

The fourth is `server/lib/relay/host-client.test.js > tunnels an HTTP GET /health with only binary frames post-handshake`, which appears and disappears between full-suite runs on this machine and passes on its own. Nothing in this diff touches `lib/relay` either. Flagging it rather than quietly re-running until it was green.

### The one new lint finding, stated plainly

`runtime.js:158` adds `typeof part.tool === 'string'` and trips `anti-slop(no-runtime-typeof)`. It is a deliberate copy of `runtime.js:145`, `typeof part.text === 'string'`, the pre-existing helper 13 lines above doing the identical job for text parts. Satisfying the rule properly means parsing OpenCode message parts at the fetch boundary and giving this module a domain type, which would touch all 26 existing findings in the file and is a much larger change than this one. Happy to do it as a follow-up if you want it; silencing the rule or widening the type to dodge it seemed worse than saying so here.

### Behavior, measured

The session that motivates this PR was replayed through the real prompt at the three points where the old suggestion sent it off course. Model: a small local one, which matters for the numbers below.

| Point | Old suggestion | New suggestion | Empty rate |
|---|---|---|---|
| Request satisfied (ticket created) | *"Implement the filter in useUsageProviderGroups..."* | `""` | **5/5** |
| Work genuinely unfinished | *"Test the mobile panel..."*, a caveat lifted from the assistant's own reply | an imperative naming the remaining item **from the original request** | 2/3 non-empty |
| Waiting on a human decision (PR ready to merge) | *"Merge the PR and start task 2"* | `""` | **2/4** |

**The two marginal rows are the honest state of this, and I would rather show them than round them off.**

The "already satisfied" case is solid: empty in every run. That is the case that caused the 400-message session, and it is fixed.

The "waiting on a decision" case is not reliable. Asked for a sendable imperative, the model proposed merging in **4 runs out of 4** before I added the second example, and in 2 of 4 after. Merging is exactly the decision it was waiting on. The example helped and did not fix it.

Adding that example also cost something: one run of three on the "unfinished" row came back empty when real requested work remained. That is the trade this prompt is making on a small model, and it is the cheaper direction to err in. A missing suggestion means no chip appears that cycle. A confident wrong one is what this PR exists to stop.

I stopped tuning here deliberately. Further prompt surgery against one small model is fitting noise, and each iteration risked the row that already works. A stronger small model may well hold the decision rule better; nothing in the change assumes a weak one.

Separately, roughly one run in ten returns malformed JSON, which `extractJsonObject` already rejects, yielding no chip that cycle. Pre-existing behavior of this prompt shape, not introduced here, and it degrades benignly.

### Prefix cache, measured

Three consecutive assists on one live session, from the server's own `[small-model:diagnostic]` timestamps:

```
153,691 chars  ->  35s   (first assist, cold prefill)
154,229 chars  ->   8s
154,352 chars  ->   8s
```

The transcript grows by a few hundred characters at the end and the shared prefix is served from cache. That is what the append-only ordering and the chunked drop are for.

Transcript sizes on live sessions, since the whole-conversation transcript is the main new cost:

| Messages in session | Transcript | ~tokens |
|---|---|---|
| 403 | 39,407 chars | ~10k |
| 496 | 65,380 chars | ~16k |
| 902 | 88,889 chars | ~22k |
| 1,921 | 225,627 chars | ~56k |

### Not verified

No browser run. The chip's rendering is unchanged and the payload shape is identical, so there is nothing new to look at, but nobody has watched one appear in a UI for this diff. The empty-suggestion path is exercised against a live model as above but is **not covered by an automated test**: the tests cover `buildTranscript`, and the behavior this PR exists for rests on the system prompt. A test that asserts on a model's judgement would be asserting on the model.

## Visual evidence

No visual change. The diff is confined to `packages/web/server`. It alters the text stored in `metadata.openchamber.assist.suggestion`, not any renderer. Field names, types and the `forMessageID` freshness contract are identical, so every surface renders it exactly as before. An empty suggestion simply means the chip does not appear, which is the existing behavior for a missing suggestion.

## Risks and failure behavior

- **Cost.** Assist input becomes proportional to session length. The prefix cache reduces latency, not billing. The measured ceiling above is ~56k tokens on a 1,921-message session; text and tool names only keeps this well below what the conversation itself sends.
- **The prefix cache is assist-to-assist only.** This module calls the provider directly, with its own system prompt and no tool schemas, so it shares no prefix with the request OpenCode makes for the conversation. Documented in the module doc so the next reader does not assume otherwise.
- **Trimming.** Over budget the oldest messages are dropped on a `TRANSCRIPT_DROP_CHUNK` boundary rather than one per turn, so the window is stable across most turns. The last message is never dropped, covered by a test.
- **Overflow.** A session larger than its model's context yields no suggestion that cycle instead of one answering a truncated prompt. Silent by design: this is background sugar and the next idle cycle retries.
- **Empty suggestion, no clearing write.** When only the suggestion is enabled and the model correctly returns empty, nothing is written. The previous payload is already stale by `forMessageID` and is not rendered.
- **Fetch cost.** The full-history fetch has its own 20s timeout. The staleness re-check before writing still uses a 4-message tail and the original 5s timeout.
- **Rollback.** Revert the commit. No persisted format changed, so no conversion or cleanup.

### DIFF
--- a/packages/web/server/lib/session-assist/DOCUMENTATION.md
+++ b/packages/web/server/lib/session-assist/DOCUMENTATION.md
@@ -1,9 +1,24 @@
 # Session Assist
 
 Server-side watcher that generates a short recap of the agent's last reply
-and one suggested user follow-up with the small model
+and the agent's own proposed next step with the small model
 (`lib/small-model`), storing both on the session's metadata under
 `metadata.openchamber.assist`.
+
+## What the suggestion is
+
+The suggestion is written **in the agent's voice** — "I would…" — and says what
+it would do next if told to keep going. It is explicitly **not** an order
+addressed to the agent in the user's voice.
+
+That distinction is the reason the whole conversation is sent. A suggestion
+generated from the last exchange alone cannot see the request the session
+started from, so it has no way to know the work is finished, and its prompt
+gives it no way to say so: it will always manufacture a next step, and the
+agent's own "I did not verify X" caveats become the next instruction. The
+system prompt therefore requires an **empty string** when the user's request is
+already satisfied or the session is waiting on a human decision, and an empty
+suggestion is a correct answer rather than a generation failure.
 
 ## Flow
 
@@ -15,8 +30,7 @@ and one suggested user follow-up with the small model
    `retry` status or a user `message.updated` clears it (the "1 minute of
    quiet" rule).
 3. On fire: fetch the session (skip sub-agent sessions with `parentID`),
-   take the LAST exchange only — the final assistant reply plus the user
-   message it answered (assistant `parentID` → user id) — and call
+   render the WHOLE conversation (`buildTranscript`) and call
    `generateSmallModelText` with the
    session's own provider/model taken from the last assistant message — so
    the utility call spends the same subscription as the conversation.
@@ -30,6 +44,59 @@ and one suggested user follow-up with the small model
    re-checked (a stale result is dropped) and the metadata is merged from a
    fresh session read so concurrent metadata writes made during generation are
    preserved.
+
+## Transcript shape and the prefix cache
+
+`buildTranscript` renders every message oldest-first as `#N Role:` blocks —
+text parts plus **tool names only**, never tool input or output (a session can
+be almost entirely tool calls, so omitting them renders it as an empty
+conversation; including their payloads would pipe file contents and command
+lines into a utility prompt).
+
+Everything that varies per call — the pointer to the last message, the
+requested fields, the language sample — goes **after** the transcript. The
+history is therefore an append-only prefix between consecutive assists on one
+session, which a backend that caches prefixes can serve without re-prefilling.
+That is an opportunity, not a guarantee: it depends entirely on the provider.
+An OpenAI-compatible backend in front of vLLM or sglang does it automatically,
+and there the effect is large (measured: 35s for the first assist on a
+154k-character transcript, 8s for each of the next two). Anthropic caching, by
+contrast, requires explicit `cache_control` breakpoints, which `callSmallModel`
+never sends, so this ordering buys nothing there and costs nothing either.
+Over budget the oldest messages are dropped, but only on
+a `TRANSCRIPT_DROP_CHUNK` boundary: dropping one message per turn would move the
+start of the transcript every time and re-prefill the whole session on every
+cycle.
+
+The budget is not a constant. It comes from `describeSmallModel().inputCharBudget`
+for the model that will actually answer, minus `PROMPT_SCAFFOLD_RESERVE_CHARS`:
+a local proxy model is absent from the models.dev catalog and falls back to a
+conservative 64k context, so a fixed budget would either waste a large window or
+overflow a small one.
+
+This buys latency, not spend — the provider still bills every input token, and
+assist cost is now proportional to session length rather than constant. Text
+parts and tool names alone stay far below what the conversation itself sends:
+measured on live sessions, 403 messages render to ~10k tokens, 1921 to ~56k.
+
+**The fetch scales too, and it is the larger number.** `fetchSessionMessages`
+without a `limit` pulls every message with its parts, tool inputs and outputs
+included, and the transcript then keeps only text and tool names. On a long
+session that is a transient multi-megabyte download and JSON parse in the
+server process once per assist cycle, far bigger than the prompt it produces.
+It is bounded (one in-flight assist per session, a 20s timeout) and OpenCode's
+message endpoint offers no field projection to narrow it, but anyone measuring
+this feature's cost should measure the fetch, not just the model call.
+
+The prefix cache is **assist-to-assist only**. This module calls the provider
+directly, with its own system prompt and no tool schemas, so it shares no
+prefix with the request OpenCode makes for the conversation itself.
+
+Because the clamp in `lib/small-model` truncates the **tail**, which here is the
+instruction, the call passes `onOverflow: 'error'`. A session larger than its
+own model's context (`context-too-small`) and a reasoning model that spends the
+output budget thinking (`output-exhausted`) are expected outcomes of a
+whole-session transcript, so both are skipped without a warning.
 
 ## Settings gate
 
--- a/packages/web/server/lib/session-assist/runtime.js
+++ b/packages/web/server/lib/session-assist/runtime.js
@@ -36,11 +36,28 @@ const getSessionAssistTargets = () => {
 };
 
 const IDLE_QUIET_MS = 60_000;
-const TRANSCRIPT_MESSAGE_LIMIT = 12;
 const TRANSCRIPT_PART_CHAR_LIMIT = 6_000;
+// Over the budget the transcript is trimmed from the OLDEST end, and only ever
+// on a TRANSCRIPT_DROP_CHUNK boundary: the prompt is sent to the session's own
+// model, so consecutive assists on one session share a token prefix and hit the
+// backend's prefix cache. Dropping one message per turn would shift the start of
+// the transcript every time and defeat that.
+const TRANSCRIPT_DROP_CHUNK = 16;
+// Room left for everything the prompt wraps around the transcript: the header,
+// the pointer to the last message, the requested fields and the language sample.
+// Generous on purpose — the budget it is subtracted from is itself an estimate.
+const PROMPT_SCAFFOLD_RESERVE_CHARS = 2_000;
 const RECAP_CHAR_LIMIT = 320;
 const SUGGESTION_CHAR_LIMIT = 500;
 const FETCH_TIMEOUT_MS = 5_000;
+// The transcript is the whole session, so the message list can be large.
+const HISTORY_FETCH_TIMEOUT_MS = 20_000;
+const ASSIST_TIMEOUT_MS = 120_000;
+const STALENESS_TAIL_LIMIT = 4;
+// Expected outcomes of a whole-session transcript, not failures: a session
+// longer than its own model's context, and a reasoning model that spends the
+// output budget thinking. Both mean "no suggestion this cycle".
+const QUIET_GENERATION_CODES = new Set(['context-too-small', 'output-exhausted']);
 
 const buildAssistSystemPrompt = ({ recap, suggestion }) => [
   'You assist a user who chats with a coding agent. Based on the conversation transcript, return exactly one JSON object and nothing else — no prose, no markdown, no code fences.',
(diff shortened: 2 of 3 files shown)


pocharlies:
Thanks — the review landed on `8d72f949`, and all three points are addressed at `1b3d8e95`.

**Finding 1 (agent-voice suggestion meets a user-voice send path) is fixed, not deferred.** You were right that it needed an owner, and the answer was to revert the voice rather than change the chip. `SessionSuggestionChip` renders and pastes the same string, so the suggestion stays in the user's voice and the client contract does not move at all. The two changes that actually fix the runaway suggestion — the whole conversation, and permission to return an empty string — never depended on the voice.

That revert cost something measurable, and it is now in the PR body. Asked for a sendable imperative, the model proposed "merge the PR" in **4 runs out of 4** on a session waiting for exactly that decision. An added example for the "the next move is the user's" case brought it to 2 of 4, and cost one false-empty in three on the row where real work remained. The "already satisfied" row — the one that caused the 400-message session — is empty in every run. I stopped tuning there rather than keep fitting one small model's noise.

**Finding 2 (the fetch downloads tool payloads it never uses)** is a fair hit; I documented the model call's cost and not the fetch's. The module doc now says the fetch is the larger number: a transient multi-megabyte download and parse per assist cycle on a long session, bounded by one in-flight assist per session and a 20s timeout, with no field projection available on OpenCode's message endpoint to narrow it.

**The nit is correct and I had overclaimed.** The doc asserted the prefix cache serves the append-only prefix, full stop. It now says that is an opportunity rather than a guarantee: automatic on an OpenAI-compatible backend in front of vLLM or sglang, where the measured effect is 35s to 8s, and worth nothing on Anthropic because `callSmallModel` never sends `cache_control` breakpoints. Costs nothing there either, so the ordering stays.

On your residual-risk note about prompt injection: agreed that the surface grows, and worth stating plainly that the mitigation is weaker than "the chip does not auto-send". The chip pastes into the composer and the user presses enter, so a hostile string still has to survive a human reading it. What it does not survive is nothing else — the utility model has no tools and its output is clamped to 500 characters and parsed as JSON.

pocharlies:
Correction to my own evidence, because I got the cause wrong and it points at my machine, not at this repo.

I reported three `server/lib/git/service.test.js` failures as pre-existing, and verified them only by stashing my change and seeing them fail again. That was the wrong control: it proves my diff did not cause them, not that they are yours.

They are caused by my global git configuration. I have `core.hooksPath` pointed at a personal hooks directory containing a `pre-push` guard that blocks direct pushes to trunk branches. Because that setting is global, it applies to the throwaway repositories those tests create, and the guard fires inside them. The failure output contains its text verbatim:

```
pre-push blocked: direct push to a trunk branch.
Everything lands on the trunk through a Pull Request:
Trunk branches here: main master deploy/prod
```

Proof, same file, same tree, only the global config neutralised:

```
$ bun run --cwd packages/web test -- server/lib/git/service.test.js
   Tests  3 failed | 81 passed (84)

$ GIT_CONFIG_GLOBAL=/dev/null bun run --cwd packages/web test -- server/lib/git/service.test.js
   Tests  84 passed (84)
```

So `packages/web` is **1822 passed, 0 failed** for this PR on a clean git config, and the only other failure I mentioned, the intermittent `relay/host-client.test.js` timeout under full-suite load, is unrelated to git config and still intermittent. The PR body still says three pre-existing failures; treat this comment as the correction rather than the table.

Whether `createWorktree` should isolate itself from a developer's `core.hooksPath` is your call, not something I would change from a session-assist PR. It is a real trap though: any contributor with a global hooks directory sees three red tests that have nothing to do with their work, and the natural conclusion is that the suite is flaky.

Separately, and unrelated to this PR, something I hit while rebasing a fork onto `v1.22.0`: `packages/ui/src/lib/i18n/messages/tr.ts` is four keys short of `en.ts` at that tag, missing the `gitView.empty.*` group. `messages.test.ts` asserts parity across all twelve locales, so the UI suite does not pass on a clean `v1.22.0` checkout. Happy to open a separate issue or a one-line PR if that is useful.

pocharlies:
Rebased onto `main` at `c1885015d`; the branch is `MERGEABLE` again.

The conflict was with `0107d77ad` (*fix: identify OpenCode Go requests by session*), which added `sessionID: sessionId` to the very `generateSmallModelText` call this PR rewrites. Both sides kept: your `sessionID` stays, and the two `preferred*` values are the locals this PR extracts, which are the same expressions that were inlined there before.

Re-validated on the new base rather than carried over: `session-assist` 6/6, `type-check:web` 0 errors, full `packages/web` 1839 passed with a clean git config.

Separately, and only because it turned up while validating this: `server/lib/walkthrough/routes.test.js` and `server/lib/relay/host-client.test.js` fail intermittently on `main` under full-suite load. They are timing assumptions rather than contention, and the relay one has a second cause that is not visible from the failure — `wss.close()` does not end open sockets, so the teardown hangs and vitest reports the whole suite as failed. I have a branch with both fixed and measurements attached; happy to open it as a separate PR if that is useful, since it is unrelated to session-assist.

btriapitsyn:
I agree that the last exchange alone is too little context for a useful suggestion. But fetching and sending the entire session on every generation feels like overkill. Could we try a bounded window of recent turns, say the last 3–5 user turns with their assistant responses, and compare the suggestion quality against the full-history approach? That seems like a reasonable starting point unless the full history gives a clear improvement.

btriapitsyn:
I’ve landed our version in [a16d947a](https://github.com/openchamber/openchamber/commit/a16d947a080268e6358b4137ad69e75126e4b6d8) on main, so I’m closing this as superseded. After comparing context windows on real sessions, we went with the last three user turns, kept the option to return no suggestion, and improved recap and language handling. Thanks for digging into the runaway suggestions and sharing your measurements; that helped shape these changes.

PR #3325, opened 2026-09-03
fix(sessions): restore archived worktree sessions

### LINKED ISSUE #3324: [Bug] Restore archived sessions from deleted worktrees to their parent project
## Reproduction

1. Create or select a session in a secondary worktree of a registered project, such as `/Users/elfy/GitRepos/infoscan`.
2. Remove that worktree through OpenChamber, which archives its sessions.
3. Restore the archived top-level session, for example `Swapping Websockets for SSE`.

## Actual behavior

The restore request clears the archive timestamp, so the session becomes globally active, but it retains the deleted worktree directory. The sidebar cannot associate that path with the parent project and filters or misgroups the session, leaving it absent from `/Users/elfy/GitRepos/infoscan`. Archived child/subagent sessions remain archived and can become disconnected from the restored root.

## Expected behavior

- Restore to the original worktree when it still exists.
- When the original worktree is confirmed missing, move the session to the parent project primary directory before exposing it as active.
- Reconcile session directory/path routing, global indexes, directory-store state, and OpenCode message path metadata through the control-plane move, without deleting messages, parts, archives, or tool-call history.
- Restore a selected root together with its archived descendants so the session tree remains coherent.
- Leave sessions that were never attached to a worktree unchanged.

## Root cause

`unarchiveSession` only writes `time.archived = 0`, upserts the returned record, and re-registers its existing directory (`packages/ui/src/sync/session-actions.ts:1312-1326`). It never checks whether that directory still exists or moves a missing-worktree session.

The worktree removal flow has the parent project directory but removes the worktree from `availableWorktreesByProject` and deletes matching per-session worktree metadata (`packages/ui/src/lib/worktrees/worktreeManager.ts:582-633`). The sidebar only admits active sessions with known project/worktree directories (`packages/ui/src/components/session/sidebar/list/sessionCollection.ts:59-89`), and grouping sends an unrecognized active directory to the archived bucket (`packages/ui/src/components/session/sidebar/projects/useSessionGrouping.ts:151-164`).

`moveSessionToDirectory` already owns the required control-plane move and local routing/index reconciliation (`packages/ui/src/sync/session-actions.ts:349-369`); the restore flow does not use it.

## Scope and edge cases

- Shared UI actions affect Web, Desktop, and VS Code. Desktop has no separate restore backend. VS Code has an archived-session inline list but no Archive view. Mobile currently has no restore surface.
- A missing-directory result is authoritative; unavailable/offline probes must not be treated as missing.
- A worktree record may be outside the parent project filesystem path, so fallback must use retained worktree-to-project ownership rather than path ancestry.
- Child sessions must retain their `parentID` relationships, message history, parts, tool-call history, and archive state unless explicitly restored with their root.
- Sessions without worktree ownership must not be redirected.

## Acceptance criteria

- [ ] Existing worktree restores to its original directory.
- [ ] Confirmed-deleted worktree restores the session tree to the parent primary project directory and appears in that project.
- [ ] Control-plane movement updates the OpenCode-backed session/message routing paths and local directory/global indexes.
- [ ] Non-worktree restores remain unchanged.
- [ ] Failed or unknown directory probes do not silently route sessions to an arbitrary project.
- [ ] Restore preserves historical messages, parts, tool calls, and tree links.

## Proposed regression coverage

- Existing worktree: assert no move and original directory remains authoritative.
- Deleted worktree: assert fallback to its retained parent project, control-plane move with `moveChanges: false`, and sidebar visibility under that project.
- Non-worktree session: assert no destination rewrite or move.
- Root with archived descendants: assert the restored tree shares the destination and keeps `parentID` links and transcript/part records.
- Normalized path and stale global metadata cases: assert no hidden active record or stale directory index remains.

mattv8:
## Intent

Closes #3324.

Restore an archived session from a deleted worktree by moving its archived session tree to the matching OpenCode project directory. The restore path uses the server project ID and keeps valid, unknown, or unavailable worktree directories unchanged.

macOS users can download and install a pre-release with this feature from here: https://github.com/mattv8/openchamber-testing/releases

## Non-goals

Does not change sidebar grouping, worktree deletion, session history, tool calls, or non-worktree restore behavior.

## Affected surfaces

`@openchamber/ui` shared session actions used by web, Electron, VS Code, and mobile. The behavior changes only after a restore confirms that a worktree-owned session directory is missing; the control-plane move remains the shared OpenCode API path for every runtime.

## Repository guidance

|---|---|---|
| `AGENTS.md` sync/state rules and `sync-state-invariants` | The change reconciles archived sessions, directory caches, and runtime switches. | Uses the global cache for complete archived subtree coverage, preserves server-confirmed records, and prevents stale runtime results from publishing local state. |
| `openchamber-change-discipline` | This is executable shared UI source. | Keeps the change in the owning session-actions module with focused regression coverage and package validation. |
| `ui-api-decoupling` | The fallback uses the official OpenCode project and control-plane APIs. | Calls the SDK through the existing client and resolves the destination by the server project ID, never by local UI-project IDs or active-project heuristics. |

## Validation

| Check | Result |
|---|---|
| `bun test packages/ui/src/sync/session-actions.test.ts` | Passed: 74 tests, 0 failures. |
| `bun run lint:ui` | Passed with one pre-existing warning in `packages/ui/src/apps/MobileChangesSurface.tsx`. |
| `bunx oxlint packages/ui/src/sync/session-actions.ts packages/ui/src/sync/session-actions.test.ts` | Blocked by the existing anti-slop backlog in these files; no lint suppression added. |

## Visual evidence

No rendered component or styling changes. The PR changes the session restore control-plane/state transition and verifies it with focused action tests.

## Risks and failure behavior

Only an exact `missing` directory result enables fallback. Unknown availability, availability probe failure, missing project records, and non-worktree sessions retain the original restore path. Each moved node is first restored by the server and then moved with `moveChanges: false`; local messages and parts are reconciled by the existing move path. A runtime switch during the control-plane request does not publish stale local routing or cache state.


mattv8:
Both findings addressed in `8ccc03f1d`.

**Blocker (retry after a partial restore):** `getRestoreSubtree` no longer filters on `archived` alone. It now keeps a subtree node while it is archived **or** still owns the confirmed-missing source directory, so a root that a prior attempt already unarchived but failed to move is still selected on retry. `updateSession(archived: 0)` is idempotent, so the retry re-issues the control-plane move and returns `true` only after it completes. `resolveMissingWorktreeRestore` now returns the source directory alongside the destination so the filter compares against the exact missing directory rather than re-deriving it.

Regression test added: a root already active (`archived: 0`) but stranded in the missing worktree must re-issue `controlPlane.moveSession` for the root and land it in the destination. It fails on the old filter and passes now.

**Non-blocker (docs):** `packages/ui/src/sync/DOCUMENTATION.md` restore contract now documents the missing-worktree fallback, the exact-`missing` gate, the server-`projectID` destination resolution, the global-cache subtree, the retry keep-rule, and the `expectedRuntimeKey` guard on `moveSessionToDirectory`.

Validation: `bun test packages/ui/src/sync/session-actions.test.ts` (75 pass), `bun run type-check:ui`, `bun run lint:ui` (clean apart from a pre-existing warning).

yulia-ivashko (approved):

PR #3333, opened 2026-09-04
fix(proxy): evict idle agent sockets before server timeout and silence client aborts

alvins82:
## Intent

Fixes frequent `[proxy] OpenCode proxy error: socket hang up` messages in server logs caused by two distinct issues:

1. **Keep-alive timeout race condition**: `http.Agent` pooled idle sockets for 60 seconds (`OPENCODE_AGENT_IDLE_TIMEOUT_MS = 60_000`). Upstream Node.js HTTP servers drop idle connections after 5 seconds (`server.keepAliveTimeout`), and Bun drops them after 10 seconds. When OpenChamber attempted to reuse a pooled socket after an idle pause of >5 seconds, the upstream server had already terminated the connection, causing the next write to fail with `ECONNRESET` (`socket hang up`). Lowering the agent idle timeout to 4 seconds ensures OpenChamber evicts idle sockets cleanly from the client side before upstream servers drop them.
2. **Spurious error logging on client cancellations**: When a browser client cancels an in-flight request (e.g. during page navigation, switching sessions, fast typing, or React hook unmounts), `http-proxy` aborts the upstream socket, which emits a `socket hang up`. The proxy error handler previously logged this unconditionally as `[proxy] OpenCode proxy error: socket hang up`. Suppressing error logs when the client request or response was already aborted or destroyed removes false-alarm log noise.

## Non-goals

- Modifying active request timeouts: In-flight requests (prompts, tool execution, model generations) remain governed by `PROXY_REQUEST_TIMEOUT_MS` (4 minutes) and `proxyTimeout`.
- Changing SSE streaming lifecycles or retry policies.
- Adding automatic request retries on socket reset.

## Affected surfaces

- Packages: `@openchamber/web`
- Runtimes: Web and Desktop (Electron runs the OpenChamber server in-process).
- User-visible states: Cleaner server console logs without false-alarm proxy error traces; no changes to UI rendered output or API contracts.
- Persisted/external contracts: None.

## Repository guidance

|---|---|---|
| `openchamber-change-discipline` | Modifying proxy server networking and error handling in `@openchamber/web` | Scoped strictly to `OPENCODE_AGENT_IDLE_TIMEOUT_MS` and client-abort detection in `proxy.js`. Validated with focused unit tests. |
| `communication-style` | PR description and documentation authoring | Concise, plain language, active voice, and no em dashes. |

## Validation

| Check | Result |
|---|---|
| `bunx vitest run packages/web/server/lib/opencode/proxy-agent-wiring.test.js packages/web/server/lib/opencode/proxy.test.js packages/web/server/opencode-proxy.test.js` | 35 passed across 3 test files |
| `bun test` in `@openchamber/web` | 176 test files passed (1,836 tests passed) |
| Client abort test in `opencode-proxy.test.js` | Passed (verified `console.error` is not invoked on client abort) |
| Upstream timeout assertion tests | Passed (agent timeout verified for both HTTP and HTTPS targets) |

## Visual evidence

No user-visible UI change. This is a server-side reverse-proxy socket lifecycle and logging change; rendered DOM, styling, and UI components are unaffected.

## Risks and failure behavior

- **Idle timeout impact**: Sockets sitting idle in the pool for >4 seconds will close cleanly. When the user resumes activity, a new connection is established. On localhost, TCP handshake latency is ~0.1ms, which is imperceptible.
- **Active requests unaffected**: Active request duration is set by `proxyTimeout` (4 minutes) on the outgoing request, overriding the agent's idle timeout while requests are in flight.
- **Rollback**: Reverting involves restoring `OPENCODE_AGENT_IDLE_TIMEOUT_MS` and removing the abort check in `createApiProxy`.

### DIFF
--- a/packages/web/server/lib/opencode/proxy-agent-wiring.test.js
+++ b/packages/web/server/lib/opencode/proxy-agent-wiring.test.js
@@ -76,6 +76,7 @@ describe('OpenCode API proxy agent wiring', () => {
       // request. See createOpenCodeProxyAgent in ./proxy.js.
       expect(agent).toBeTruthy();
       expect(agent.options?.keepAlive).toBe(true);
+      expect(agent.options?.timeout).toBe(4_000);
     }
   });
 
--- a/packages/web/server/lib/opencode/proxy.js
+++ b/packages/web/server/lib/opencode/proxy.js
@@ -20,10 +20,11 @@ const OPENCODE_AGENT_KEEP_ALIVE_MS = 30_000;
 // exists to prevent (measured: at 64 concurrent requests, a cap of 32 left
 // 303 sockets in TIME_WAIT versus 0 at 256).
 const OPENCODE_AGENT_MAX_FREE_SOCKETS = 256;
-// Evicts idle free sockets from our side. Without it the only thing that
-// retires an idle pooled socket is the upstream closing it. Note this is
-// distinct from `keepAliveMsecs`, which is the TCP keep-alive probe delay.
-const OPENCODE_AGENT_IDLE_TIMEOUT_MS = 60_000;
+// Evicts idle free sockets from our side before upstream servers close them
+// (Node default keepAliveTimeout is 5s, Bun default idle timeout is 10s).
+// Setting this higher than upstream causes stale socket reuse where the
+// client writes into a dead socket and gets 'socket hang up' (ECONNRESET).
+const OPENCODE_AGENT_IDLE_TIMEOUT_MS = 4_000;
 
 const OPENCODE_AGENT_OPTIONS = {
   keepAlive: true,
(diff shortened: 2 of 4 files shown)


btriapitsyn:
The idle-socket change is reasonable, but the new cancellation guard also suppresses genuine upstream failures while the client is still connected.

1. In `packages/web/server/lib/opencode/proxy.js`, remove `req.destroyed` as a standalone reason to return from `createApiProxy`'s error handler. Node marks a fully consumed `IncomingMessage` destroyed even while its response and socket remain open. A later upstream reset therefore skips the existing 503 response and leaves the client waiting for the response deadline. Keep cancellation suppression tied to an aborted request or a closed/finished response or socket, and preserve immediate error responses for connected clients.
2. In `packages/web/server/opencode-proxy.test.js`, add a real HTTP regression where the upstream consumes the request and then resets its connection while the browser connection stays open. Assert that the client promptly receives 503 before the general deadline, including with a parsed POST body. Retain the cancellation case and synchronize it with the upstream receiving the request so it proves cancellation happened after forwarding began.

Thanks for addressing the connection lifecycle. These changes would preserve quiet cancellations without hiding real failures.

alvins82:
Fixed in c530ca783.

- Removed `req.destroyed` from the cancellation guard, so a consumed request body no longer hides an upstream reset from a connected client.
- Added a real HTTP regression for a parsed POST body followed by an upstream socket reset. It asserts that the client receives 503 promptly and that the proxy logs the upstream failure.
- Updated the cancellation regression to wait until the upstream receives the request before aborting the client.

Validation: `bun run --cwd packages/web test -- server/opencode-proxy.test.js server/lib/opencode/proxy.test.js server/lib/opencode/proxy-agent-wiring.test.js` passed all 36 tests.

PR #3351, opened 2026-09-05
fix(mcp): keep reconnect plugin from initializing MCP in every instance

### LINKED ISSUE #3350: [Bug]: 1.22.1 spawns the whole MCP server fleet in every directory instance at launch
### What happened

After updating the macOS desktop app from 1.22.0 to 1.22.1, the app becomes unusable within a couple of minutes of launch. CPU maxes out and memory climbs until the system grinds to a halt. The managed `opencode serve` process accumulates MCP server child processes without bound, 572 of them within about 80 seconds on my machine and still climbing when I force quit. My normal baseline is about 128. Reverting to the 1.22.0 release download makes it stop immediately.

My config has 27 MCP servers, 21 of them local stdio (uvx and npm based), all in the global OpenCode config. OpenChamber creates OpenCode instances for about 34 directories at startup, one per project plus one per `~/.config/openchamber/chats/<date>/session-*` directory.

### What I found

The culprit is the new MCP auto-reconnect plugin from aa13c981.

1. The plugin runs inside every managed OpenCode instance and scheduled a `mcp.status()` probe one second after each instance loaded.
2. Reading MCP status initializes that instance's entire MCP fleet as a side effect. MCP startup in OpenCode is otherwise lazy, nothing spawns until an instance actually uses MCP.
3. So every launch spawned every configured stdio server in every instance, used or not. On my machine that is about 34 instances times 21 local servers, roughly 700 processes in the first seconds.
4. Under that load the slower npm-based starters blow past the 30-second MCP handshake timeout and get marked failed. Each failure publishes `mcp.tools.changed`, which woke every instance's plugin because the event handling is not scoped to the affected server, and the retry loop is unbounded. Reconnects spawned more processes, more handshakes timed out, and the loop fed itself.

I reproduced this headlessly against `opencode serve` with my real config, with both the 1.18.25 CLI bundled in 1.22.0 and the 1.18.28 CLI bundled in 1.22.1:

- bare `serve` with no requests: 0 MCP processes over 90 seconds on both CLIs
- six `GET /session?directory=...` requests, which create instances without touching MCP: 166 MCP processes within 2 seconds with the 1.22.1 plugin, 0 without it
- the reconnect path itself works as intended: kill a connected server's child process and it comes back

The CLI version bump between 1.18.25 and 1.18.28 is not a factor. Both spawn identically when poked.

### Fix

I have a fix ready and will link the PR in a comment. The plugin no longer probes at boot. It stays passive until an instance-scoped `mcp.tools.changed` event arrives, which only fires once that instance's MCP has actually run, so background instances never start MCP at all. Reconnects are also bounded: five attempts per server before giving up with a warning, at most two concurrent connects per OpenCode process, and jittered backoff.

### Workaround

Stay on or revert to the 1.22.0 release. One follow-up note: npm-based servers that got marked failed during a 1.22.1 launch can stay failed afterwards, reconnecting them once by hand clears it.

knorby:
## Intent

Fixes #3350.

1.22.1 added an MCP auto-reconnect plugin to every managed OpenCode instance (aa13c981). The plugin polled `mcp.status()` one second after each instance loaded. Reading status initializes the instance's MCP as a side effect, and OpenChamber creates an instance for every project and chat directory at startup, so each launch spawned the full stdio server fleet for every instance. On the reporting machine that is roughly 34 instances times 21 local servers, with the process count past 570 within 80 seconds, CPU pegged, and memory exhaustion within minutes. Failures under that load then fed the unbounded retry loop.

This PR makes the plugin passive until its first instance-scoped `mcp.tools.changed` event and bounds the retry loop. After the fix, an instance nobody uses runs no MCP processes at all, a dropped server still reconnects within seconds, and a broken server gets five bounded attempts instead of infinite ones.

## Non-goals

- When OpenChamber creates instances is unchanged. The background instances stay.
- OpenCode's lazy MCP initialization and 30-second handshake timeout are upstream OpenCode behavior, untouched here.
- npm-based starters failing under heavy contention is pre-existing and reproduces with the 1.18.25 CLI bundled in 1.22.0. This PR removes the artificial contention but does not change starter behavior.
- No persisted contract changes. The plugin file at `<dataDir>/mcp-reconnect/openchamber-mcp-reconnect-plugin.js` is rewritten on each launch, so reverting the app version reverts behavior.

## Affected surfaces

- `packages/web/server/lib/mcp-reconnect/` (plugin source, tests, docs). The plugin is injected into managed OpenCode processes by the OpenChamber web server, so the desktop app and any surface running managed OpenCode through this server get the fix. Surfaces that do not manage OpenCode through this server are unaffected.

## Repository guidance

|---|---|---|
| `openchamber-change-discipline` skill | Server source change with platform and runtime consequences | Change confined to the owning module; focused tests plus repo-wide type-check, lint, test, and build; owning docs updated alongside the behavior |
| `performance-engineering` skill | Reported CPU and memory runaway on a spawn hot path | Structural fix removes the work instead of optimizing it (no boot-time polling at all); before and after process counts measured at production scale, a real 27-server config across roughly 34 instances, not a toy fixture |
| `packages/web/server/lib/mcp-reconnect/DOCUMENTATION.md` | Owning module documentation | Rewritten for passive arming, the attempt cap and its re-arm conditions, the concurrency budget, jitter, and event scoping, including the invariant that an instance whose MCP has never run must never run any because of this plugin |

## Validation

| Check | Result |
|---|---|
| `bun run --cwd packages/web test server/lib/mcp-reconnect/runtime.test.js` | 11/11 pass. Five carried over, six new, including the regression case that asserts no status call is ever made in an instance whose MCP has never run |
| `bunx oxlint packages/web/server/lib/mcp-reconnect/` | Clean |
| `bun run type-check` | Pass across all five workspaces |
| `bun run lint` | Pass. One pre-existing warning in `packages/ui/src/apps/MobileChangesSurface.tsx`, untouched by this diff |
| `bun run test` | 1884 pass, 1 pre-existing failure in `server/lib/opencode/env-runtime.test.js` ("bounds every login-shell probe and falls through when one overruns"). Reproduced identically with this diff stashed. That test expects `resolveOpencodeCliPath()` to return null, but this machine has a brew opencode at `/opt/homebrew/bin/opencode` the resolver finds through a direct filesystem check |
| `bun run build` | Pass |
| `bun run electron:build` | Pass. Confirmed the packaged app.asar ships the passive plugin, with the single `schedule(INITIAL_RETRY_MS)` occurrence living in the event hook |
| Headless repro with the affected 27-server config, both bundled CLIs | Six `GET /session` instance creations spawned 166 MCP processes in 2 seconds before and 0 over 90 seconds after. An MCP-initialized run holds a stable count for 120 seconds. A SIGKILLed server child reconnected and stayed connected at every 5-second checkpoint |
| Manual verification on a macOS arm64 desktop build | Before: 572+ processes in about 80 seconds, CPU pegged, memory exhaustion within minutes. After: process count lands at the 128 baseline and CPU settles |

## Visual evidence

No rendered UI changes; the diff is a server-injected plugin plus its tests and docs. The behavior in question is empirical, so the before and after process counts and timings in the validation table serve as the measurements.

## Risks and failure behavior

- A server that fails during initial startup in an instance that never sees an MCP event will not be auto-retried by this plugin until that instance arms. Any drop or tool-list change arms it, after which the loop reconciles all failed servers with bounded attempts. This trade-off is deliberate: probing status to find those servers is exactly what spawns the fleet in every background instance.
- Give-up is recoverable. Five consecutive failures stop retries with one console warning, and the cap clears as soon as the server is seen healthy again (including via manual reconnect), leaves the config, or OpenCode restarts.
- The concurrency budget lives on `globalThis` because directory instances each load their own copy of the plugin inside one OpenCode process. Slots are reserved synchronously with no await between read and increment, and released in `finally`, so a throwing `mcp.connect` cannot leak a slot.
- Errors from status and connect are swallowed as before. The one new log line is the give-up warning.
- Rollback is replacing the app: the plugin file is rewritten wholesale on every launch.


abhisheksharma611:
Please fix it as soon as possible, my work is hung and cannot move forward and release the update

bourtitom:
This looks like the right fix for the process storm. I noticed one remaining edge case around `client.mcp.status()`.

The OpenCode SDK does not always throw on HTTP errors. It can resolve with an `{ error }` result and no `data`. The current fallback:

```js
const statuses = (await client.mcp.status())?.data ?? {}
```

treats that failure as an authoritative empty server list. The cleanup loop then removes the existing retry counters. On the next successful status response, a server that had already reached `MAX_ATTEMPTS` becomes eligible for another retry episode.

Could we preserve the retry state when the status response is unavailable?

```js
const response = await client.mcp.status()
if (response?.error || !response?.data || typeof response.data !== "object") {
  throw new Error("MCP status is unavailable")
}
const statuses = response.data
```

The existing `catch` already handles this correctly by keeping the current counters and scheduling the next idle check.

A regression test could bring a server to the five-attempt cap, return `{ error: new Error("ECONNRESET") }` once from `mcp.status()`, then report the server as failed again. The attempt count should remain at five.

knorby:
@bourtitom thank you! Confirmed and implemented

@btriapitsyn sorry to ping, I wanted to see if there is anything that I can do to get this merged. I'm still seeing the massive memory leak on my system without this fix, and it sounds like I'm not alone. Thank you for the project!

abhisheksharma611:
The issue exists, they fixed it but not completely I can see in my laptop task manager that it is still showing so many python and nodejs runtime ,even if I am working in one session and one project, please I request @btriapitsyn to solve this issue as soon as possible and release the update.

yulia-ivashko:
Thank you for this, @knorby. Your root-cause work in #3350 was right, and it shaped what happened next. We ended up removing the reconnect plugin entirely (0a338962a, shipped in 1.23.1), because automatic reconnects can't guarantee that the previous MCP processes get cleaned up. With the plugin gone there is nothing left for this patch to apply to, so I'm closing it. Thanks also to @bourtitom for the `{ error }` catch. And thank you for #3675, which fixed the second path to the same symptom and is already merged.

PR #3375, opened 2026-09-06
fix(files): auto-refresh sidebar tree when files change

### LINKED ISSUE #2093: File tree does not auto-refresh when agent creates or modifies files
## Problem

This is a follow-up to #2036.

#2092 fixed the manual-refresh part of #2036: clicking the refresh button no longer collapses the entire file tree. However, the file tree still does **not** update automatically when the agent (or any external process) creates, renames, or deletes files in the current project.

Users have to click the refresh button to see new files. On desktop this is especially noticeable because the agent can create files via the terminal/OpenCode tools and the sidebar stays stale until manual interaction.

## Why this is still broken

There is currently **no filesystem watching infrastructure** anywhere in the codebase:
- No `chokidar`, `fs.watch`, `inotify`, or `FSEvents` usage.
- No server-side events of type `file.changed`, `directory.changed`, or `files.changed`.
- The SSE/WebSocket event pipeline only handles session/message events from OpenCode.

The file tree can only learn about filesystem changes by re-listing directories. Local CRUD operations already call `refreshDirectory(parentPath)`, so manually-created files appear immediately. External/agent changes do not.

## Proposed solution (recommended)

Add a server-side filesystem watcher that emits a new `files.changed` event over the existing SSE/WebSocket event stream.

### Architecture

1. **Watcher service in `packages/web/server/`**
   - Use Node.js built-in `fs.watch` (no new dependency) on the active project directory.
   - Watch only directories that are currently loaded/expanded by clients.
   - Debounce events (e.g. 200–300 ms) and batch paths to avoid event storms during bulk operations like `git checkout` or `npm install`.
   - Handle rename, delete, and directory-create events.
   - Stop watching when the last client disconnects or the project changes.

2. **Event contract**
   ```json
   {
     "type": "files.changed",
     "directory": "/absolute/path/to/project",
     "paths": ["src/new-file.ts", "src/components/"]
   }
   ```
   - `directory` is the project root being watched.
   - `paths` lists changed entries relative to the project root.

3. **Fan-out**
   - Reuse `createGlobalUiEventBroadcaster` in `packages/web/server/lib/event-stream/runtime.js` to send the event to all connected SSE and WebSocket clients.

4. **Client handling**
   - Add a reducer/handler in the UI event pipeline (`packages/ui/src/sync/event-reducer.ts`) for `files.changed`.
   - Map each changed path to its parent directory and call `refreshDirectory(parentDirectory)` in `SidebarFilesTree.tsx` for expanded directories.
   - If a directory itself was created/deleted, refresh its parent instead.

### Why this approach

- **Cross-runtime**: web, desktop, and VS Code all consume the same server events.
- **Event-driven**: no polling, no background load, no race conditions on worktree switch.
- **Reuses existing infrastructure**: SSE/WS broadcaster, event pipeline, and `refreshDirectory()` already exist.
- **Clean boundary**: filesystem state stays authoritative on the server; the client only reacts to events.

### Alternatives considered and rejected

- **Client-side polling**: simple but race-prone on worktree switch and adds unnecessary server load. It also does not feel clean for an event that the OS already provides.
- **Refresh on window focus**: trivial to implement but does not solve the case where the agent changes files while the window is already focused.
- **Runtime-specific watchers** (Electron `fs.watch`, VS Code `FileSystemWatcher`): would work for one runtime but leave web and others broken, and would duplicate logic.

## Out of scope for this issue

This issue is only for designing and tracking the fix. The implementation should be done in a separate PR because it touches server event infrastructure and needs careful testing for cross-platform reliability.

## Related

- Original issue: #2036
- Phase 1 fix (manual refresh collapse): #2092

zzhonglei:
## Intent

Fixes #2093.

The sidebar file tree now updates automatically when agents or other programs create, rename, or delete files, without a manual refresh. Expanded folders stay open.

## Non-goals

Recursive workspace watching, reloading open file contents, changing Git status, or changing tree sorting and filtering.

## Affected surfaces

- Standard Web and Electron Desktop: an optional `FilesAPI.watchDirectories` capability backed by `GET /api/fs/watch`.
- Direct connections use authenticated SSE; private-relay connections use visible-directory polling every eight seconds.
- Dedicated VS Code, hosted-mobile, and Capacitor interfaces do not mount `SidebarFilesTree`.
- Declares `zod` as a Web server runtime dependency. No persisted-schema or release-note changes.

## Repository guidance

|---|---|---|
| `AGENTS.md`, `openchamber-change-discipline`, Web README, FS/UI-auth docs | Shared capability and server route | FS policy stays in the owning module; docs and tests cover the route; entrypoints only wire dependencies. |
| `ui-api-decoupling`, browser-assets/auth, runtime-parity and implementation-map references | Runtime URLs and authentication | Uses the runtime resolver and narrow URL-token/proxy allowlists; preserves explicit consumer scope. |
| `sync-state-invariants`, sync docs | Concurrent directory requests | Authoritative re-listing, stale-response rejection, and independent recovery for failed directories. |
| `performance-engineering`, `relay-transport` | Bursty events and reconnects | Scoped shared watchers, bounded batching/concurrency, hidden-tree cleanup, and gated polling fallback. |

## Validation

Commit: `82f4f6d8a`. Windows, Node 24.14.0, Bun 1.3.14.

| Check | Result |
|---|---|
| `bun run type-check`; `bun run lint`; `bun run build` | Passed. Lint has one warning in unchanged `MobileChangesSurface.tsx`; builds retain font/chunk warnings. |
| `bun test ./packages/ui/src/hooks/useFileTreeAutoRefresh.test.ts` | 11 tests passed. |
| `bun run --cwd packages/web test -- server/lib/fs/watcher.test.js server/lib/fs/routes.test.js server/lib/realtime-proxy.test.js server/lib/ui-auth/ui-auth.test.js src/api/files.test.ts` | 106 tests passed, including relocated managed-chat roots and out-of-workspace rejection. |
| Electron / VS Code / root script suites | 19/19, 35/35, and 3/3 test files passed. |
| Windows live FS route check | Real file create, rename, and delete each produced HTTP SSE invalidation and the correct subsequent directory listing. |
| Full UI / Web suites | UI: 398/400 files passed. Web: 165/183 files passed, 48 failed tests and one unhandled error. Failures are outside the changed test files; they include locale, timing, Windows path/permission, and command-launch assumptions. These suites are not claimed green. |
| `bun run dead-code`; Oxlint on new hook/path/watcher; Node syntax checks; `git diff --check` | Completed; no findings for the new source files or exports. |

## Visual evidence

Windows Electron, HEAD `82f4f6d8a`. The front/left window is the unpatched app; the back/right window runs this PR. Copying files into the folder and deleting them in File Explorer updates the fixed tree automatically, while the unpatched tree stays stale until manually refreshed. The recording is cropped, labeled, and muted, with the original timing preserved.

https://github.com/user-attachments/assets/ea2c523d-4a79-44f2-bfd3-818bc207c683

This clip shows external create/delete. Earlier Windows manual checks also covered rename, expansion preservation, in-app changes, multi-directory changes, and hidden-tree reopening; those scenarios are not demonstrated in this clip. The UI refresh implementation is unchanged since those checks. macOS, Linux, direct browser UI, and private-relay behavior have not been manually validated.

## Risks and failure behavior

- OS events are hints: the client re-lists contents instead of interpreting platform-specific filenames or event types.
- Events debounce for 250 ms, capped at one second. At most three directory refreshes run concurrently; subscriptions accept up to 64 directories.
- The route uses the shared workspace/worktree/managed-root policy before opening the stream. Watchers close after their last subscriber disconnects.
- Hidden trees stop subscriptions and polling; reopening and reconnecting reconcile contents. Failed directories retain prior state and do not block unrelated directories.
- No file writes, Git mutations, or data migration are introduced.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -290,6 +290,7 @@
         "web-push": "^3.6.7",
         "ws": "^8.18.3",
         "yaml": "^2.8.1",
+        "zod": "^4.3.6",
       },
       "devDependencies": {
         "@base-ui/react": "^1.4.0",
--- a/packages/ui/src/components/layout/ContextPanel.tsx
+++ b/packages/ui/src/components/layout/ContextPanel.tsx
@@ -266,7 +266,10 @@ const EDITOR_TREE_MAX_WIDTH = 480;
 
 // The editor surface's file-tree column: docked on the right, resizable from
 // its left edge, and animated open/closed like the app sidebars.
-const EditorTreeColumn: React.FC<{ visible: boolean }> = ({ visible }) => {
+const EditorTreeColumn: React.FC<{
+  visible: boolean;
+  autoRefreshEnabled: boolean;
+}> = ({ visible, autoRefreshEnabled }) => {
   const { t } = useI18n();
   const width = useUIStore((state) => state.contextEditorTreeWidth);
   const setWidth = useUIStore((state) => state.setContextEditorTreeWidth);
@@ -379,7 +382,7 @@ const EditorTreeColumn: React.FC<{ visible: boolean }> = ({ visible }) => {
         style={{ width: 'var(--oc-editor-tree-width)' }}
         aria-hidden={!visible}
       >
-        <SidebarFilesTree />
+        <SidebarFilesTree autoRefreshEnabled={autoRefreshEnabled} />
       </div>
     </div>
   );
@@ -1227,7 +1230,10 @@ export const ContextPanel: React.FC = () => {
                 </div>
               )}
             </div>
-            <EditorTreeColumn visible={contextEditorTreeVisible} />
+            <EditorTreeColumn
+              visible={contextEditorTreeVisible}
+              autoRefreshEnabled={isOpen && isFileTabActive && contextEditorTreeVisible}
+            />
           </div>
         ) : null}
         {activeChatTab && activeChatSessionID && activeChatSrc ? (
--- a/packages/ui/src/components/layout/SidebarFilesTree.tsx
+++ b/packages/ui/src/components/layout/SidebarFilesTree.tsx
@@ -31,6 +31,7 @@ import { Tooltip, TooltipContent, TooltipTrigger } from '@/components/ui/tooltip
 import { useDebouncedValue } from '@/hooks/useDebouncedValue';
 import { useRuntimeAPIs } from '@/hooks/useRuntimeAPIs';
 import { useEffectiveDirectory } from '@/hooks/useEffectiveDirectory';
+import { useFileTreeAutoRefresh } from '@/hooks/useFileTreeAutoRefresh';
 import { useFileSearchStore } from '@/stores/useFileSearchStore';
 import { useFilesViewTabsStore } from '@/stores/useFilesViewTabsStore';
 import { useUIStore } from '@/stores/useUIStore';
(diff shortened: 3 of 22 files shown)


zzhonglei:
@openchamber-bot review Please reassess the visual evidence added to the PR description.

PR #3383, opened 2026-09-06
fix(git): pass allowUnsafeCredentialHelper for token identity switching

ICEY16360:
## Summary

Switching to a **token-auth** Git identity throws:

```
Configuring credential.helper is not permitted without enabling allowUnsafeCredentialHelper
```

`setLocalIdentity` already opts into simple-git's escape hatch for the SSH branch (`allowUnsafeSshCommand`, forwarded by `createGit`), but the token branch (`addConfig('credential.helper', 'store')`) is missing the matching opt-in introduced for it by simple-git's expanded unsafe-config blocklist.

Fix: forward a new `allowUnsafeCredentialHelper` option through `createGit` and enable it in `setLocalIdentity` alongside the existing SSH opt-in.

## Root cause timeline

| When | Event |
|---|---|
| 2026-01-14 (v1.4.8) | Token-based identity auth shipped with `simple-git ^3.28.0`, whose blocklist only covered `protocol.allow` / `--upload-pack` — the token branch worked. |
| 2026-04 (simple-git 3.35/3.36) | The blocklist moved into `@simple-git/argv-parser` and expanded to 24 keys, now including `credential.helper` **and** `core.sshCommand`. |
| after the bump | Both branches broke. The SSH branch got the targeted opt-in (`allowUnsafeSshCommand: true`, covered by the existing test); the token branch — which has no test — was missed and has been broken since. |

## Changes

- `createGit` accepts and forwards `allowUnsafeCredentialHelper` (mirrors the existing `allowUnsafeSshCommand` plumbing).
- `setLocalIdentity` passes `allowUnsafeCredentialHelper: true`.
- Tests: token branch writes `credential.helper=store`; switching token → SSH writes `core.sshCommand` and clears `credential.helper` (mirrors the existing SSH test).

## Test plan

- [x] `vitest run server/lib/git/service.test.js -t setLocalIdentity` — 3/3 pass
- [x] Full `service.test.js` — 90 pass; 2 pre-existing failures on this machine reproduce on unmodified `main` (locale-dependent git stderr + `/tmp` filesystem-boundary worktree lock recovery), unrelated to this change
- [x] Manual: token identity switch now writes `credential.helper=store` and clears `core.sshCommand`; SSH switch unchanged

### DIFF
--- a/packages/web/server/lib/git/service.js
+++ b/packages/web/server/lib/git/service.js
@@ -352,7 +352,7 @@ const buildGitEnv = async () => {
   return env;
 };
 
-const createGit = async (directory, { allowUnsafeSshCommand = false } = {}) => {
+const createGit = async (directory, { allowUnsafeSshCommand = false, allowUnsafeCredentialHelper = false } = {}) => {
   const env = await buildGitEnv();
   const spawnOptions = { windowsHide: true };
   const binary = getGitBinary();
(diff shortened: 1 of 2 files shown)


btriapitsyn:
Thanks for tracing the simple-git regression and covering token authentication and the switch back to SSH. Merged.

We also completed the standalone credential-helper option handling in createGit: https://github.com/openchamber/openchamber/commit/23c6b4833ff7420475e6276ed1d304196e7cca5b.

PR #3388, opened 2026-09-06
feat(settings): add Todo list toggle under "Show tools opened by default"

### LINKED ISSUE #3065: [Feature Request] Focus on Todo List

It would be nice to keep focus on Todo List but keep all other Activity collapsed.

Image

### Extra context (optional)

Thanks

ozskywalker:
## What's new

**Todo lists can now stay open by default** 📋

Until now, only Bash and Edit tool cards could be configured to render expanded (Settings → Chat → "Show tools opened by default"). Todo list cards — written by `todowrite` and shown by `todoread` — always started collapsed, so there was no way to keep an eye on task progress without clicking each one open.

This PR adds a **"Todo list"** toggle to that same settings group. Turn it on and every todo list the agent writes or updates in your session renders fully expanded from the start — handy when you're following along with multi-step tasks and want the current progress visible at a glance.

### Details
- New setting: **Settings → Chat → "Show tools opened by default" → "Todo list"** (default: **off** — nothing changes unless you enable it)
- Also findable via settings search: search for "todo" or "open by default"
- Available in all 12 app languages

### For reviewers
- Single clean commit on top of current `main` (rebased post-v1.22.2, no conflicts): implementation mirrors the existing `showExpandedEditTools` pattern exactly — UI store flag (default `false`) → persistence (defaults/apply/sanitize) → `defaultOpenToolIds` in `ChatMessage.tsx` via a new `TODO_TOOL_NAMES` set → settings checkbox → settings-search entry → 2 i18n keys × 12 locales.
- The server-side `sanitizeSettingsUpdate` allowlist in `packages/web/server/lib/opencode/settings-helpers.js` gained the matching entry, so the preference survives `PUT /api/config/settings`.
- Validation on the rebased branch: type-check 0 errors · lint clean · **399/399** UI test files passing · i18n parity test green · persistence verified end-to-end on a local dev server.

Closes #3065.


ozskywalker:
Adding screenshots as requested by bot

image

image

image

ozskywalker:
Both findings addressed in two follow-up commits:

**1. Search anchor (dc412834f)** — Added `settingsItem="chat.expanded-tools-todo"` to the new `SettingsCheckboxRow`. The search entry now has a matching `data-settings-item` anchor, so clicking the result scrolls to and highlights the row instead of silently no-oping.

**2. Persistence round-trip test (fb801959a)** — Added two focused tests in `persistence.test.ts` covering the sanitize/apply chain:
- `round-trips showExpandedTodoTools through save responses as a partial patch`: a save response echoing `showExpandedTodoTools: true` is applied to the store while unrelated fields (e.g. `terminalShell`) are left untouched, and the default is asserted `false` before the update.
- `ignores a non-boolean showExpandedTodoTools in a settings save response`: a non-boolean value in the response is rejected and the store keeps its current value.

Validation on the updated branch: `tsc --noEmit` 0 errors, ESLint 0 errors (only the pre-existing `MobileChangesSurface.tsx` warning present on `main`), and the full UI suite passes (399/399 test files, including the 2 new tests).

ozskywalker:
**Rebased onto main (settings-registry, #3413)**

Rebased the 3 commits onto current main. Conflicts were confined to `desktop.ts`, `api/types.ts`, and `persistence.ts`, all landing where #3413's registry refactor had replaced the code this branch edited, so I took main's side for those files and dropped our edits to them. On the new architecture those edits are dead weight: `DesktopSettings`, `SettingsPayload`, the boundary parser, and the store-apply step all derive from `SETTINGS_REGISTRY`, and a key missing from the registry simply doesn't persist.

The new commit `refactor(settings): register the todo toggle in the settings registry` does the migration instead. `showExpandedTodoTools` is registered as a profile-scoped boolean with a uiStore binding, mirroring `showExpandedEditTools` (no `perSurface`, matching its siblings). Both checked-in JSON snapshots are regenerated, and the server drift test gets its sample. The old inline sanitizer branch in `settings-helpers.js` is now redundant since the registry gate drops unknown keys, so the branch no longer carries it.

Behavior is unchanged for the feature itself, with one small improvement: profile scope means the toggle now syncs across clients. Round-trip tests still pass (58/58 in `persistence.test.ts`), registry tests pass, and type-check is clean in ui, vscode, and web.

Issue #1626, opened 2026-06-12
[Feature Request] Support isolated Docker-backed managed OpenCode instances


charpeni:
### What should we add/change?

Add support for running managed `OpenCode` instances through a configurable launch command/template, with Docker as the primary use case.

Today, OpenChamber can either:

- start a managed local `opencode` process directly, or
- connect to an externally started `OpenCode` server via `OPENCODE_HOST` / `OPENCODE_SKIP_START`.

I would like OpenChamber to support a middle ground: let OpenChamber manage the lifecycle, but launch `OpenCode` inside a container instead of directly on the host.

Example target workflow:

```bash
docker run --rm \
  -p {port}:{port} \
  -v {projectDir}:/workspace \
  -w /workspace \
  -e OPENCODE_SERVER_PASSWORD={password} \
  ghcr.io/anomalyco/opencode \
  serve --hostname 0.0.0.0 --port {port}
```

OpenChamber would still allocate the port, pass the auth/password, wait for readiness, connect to the resulting `OpenCode` server, and clean up the process/container when appropriate.

### Why?

The main motivation is security and environment isolation.

Some users do not want `OpenCode` running directly on their host machine with broad filesystem/tool access. Running `OpenCode` in a container allows mounting only the intended project directory and using a controlled toolchain/image.

This also helps keep host-level secrets and OpenChamber state out of reach of `OpenCode` sessions. For example, `OpenCode` should not automatically have access to:

- host SSH keys
- git credentials
- OpenChamber settings
- OpenChamber remote instance tokens
- other projects outside the mounted workspace
- unrelated provider or MCP configuration

Today, when `OpenCode` is spawned directly on the host, it inherits a much broader ambient environment. A Docker-backed launch flow would make access explicit through mounted volumes and environment variables.

This also matches a workflow split I often want:

- **Work mode:** `OpenCode` actively edits and runs code in the project. This mode usually does not need GitHub credentials, SSH keys, or host-level OpenChamber settings. It should run in the most isolated environment possible, with only the project/workspace mounted.
- **Review/push mode:** once the work is done, I review the changes and prepare them for shipping. This step may need GitHub credentials, signing keys, SSH agent access, or host editor integrations, so I am more comfortable running it directly on the host or in a less restricted environment under supervision.

The important distinction is that the high-autonomy code-writing phase should not automatically inherit the same credentials and host access needed for the lower-autonomy review/push phase.

[...]

PR #3395, opened 2026-09-07
feat(docker): Docker-backed OpenCode instances (create, lifecycle, switching, two-way sharing)

Lorrid:
## Intent

Stacked on #3306 -- merge #3306 first. This branch is cut from `feat/opencode-path-map` (tip `8f17b52a`) and inherits its path-mapping commits verbatim; until #3306 merges, this diff shows both PRs, afterwards it narrows to the Docker-only commits.

Closes the remaining scope of #1626: OpenChamber itself creates, runs, readiness-probes, switches to, stops, removes, and cleans up **Docker-backed OpenCode instances** -- end to end from the instance selector, in the desktop shell and the web UI.

What works after this PR (feature gated behind a settings toggle, default **off**):

- **Create**: guided "+ Add instance" flow (workspace picker via native folder dialog on desktop, sharing opt-ins with exact mount disclosures, image/port overrides, explicit Build/Pull image actions). Container: non-root, default bridge, loopback-only published port, no Docker socket / SSH / git-credential mounts ever.
- **Lifecycle**: create -> start -> readiness (`/global/health` on the published port) -> running <-> stopped -> remove/cleanup. Journal-driven rollback removes exactly the resources a failed operation created; audit-confirmed it never touches foreign containers.
- **Per-instance path mapping**: builds directly on #3306's translator -- each instance maps its workspace (`C:\proj\foo` <-> `/workspace`); unmapped host paths fall back to the workspace root instead of reaching the Linux container as garbage (this is what made container chat work). Resolution: instance mapping -> `OPENCODE_PATH_MAP` -> identity, so #3306 behavior is unchanged without Docker.
- **Two-way host sharing (opt-in)**: OpenCode config dir rw (provider/model/MCP settings usable in-container and synced back -- user-approved risk), shared skills dir rw in both directions, credentials ro behind a separate default-off toggle.
- **Switching**: Local <-> Docker instances from the same window (desktop shell + web entry), per-instance state isolation, instant event routing via the shared hub.
- **Explicit image handling**: ships `docker/opencode-instance/Dockerfile` (non-root, owned data dirs); nothing is pulled or built without an explicit user action.

## Non-goals

- Cross-instance session aggregation (sessions live in their instance; Local's list shows Local's data -- documented known behavior, follow-up).
- Managing pre-existing/user-created Docker containers; running OpenChamber itself in the instance container; Kubernetes/remote-docker.
- Replacing the SSH instance mechanism; project-list auto-manipulation on switch (attempted and reverted by decision, `b59fe346`).

## Affected surfaces

- `packages/web/server`: new `lib/docker/` (CLI runtime + fake test double) and `lib/opencode/docker-instances/` (registry, mounts, lifecycle manager, routes); `path-mapping.js` (instance layer), `network-runtime.js` (upstream override), `proxy.js` (live mapping view), `core-routes.js` (JSON-body allowlist entry), `settings-helpers.js` (`dockerInstancesEnabled` toggle), `event-stream/global-hub.js` (event directory mapping), `index.js` (wiring).
- `packages/ui`: instance selector Docker section, create dialog (native folder picker on desktop via the existing `openDialog` bridge), web-mode entry, `runtime-switch` identity override, `runtimeEndpointReset` identity-only path, i18n (12 dictionaries).
- `packages/electron`: no changes (existing IPC bridge reused).
- Persisted/external contracts: new `docker-instances.json` data file, `dockerInstancesEnabled` settings field, new `/api/docker-instances*` routes, one container label. Mobile/VS Code: no changes; mobile renders the shared UI without the desktop-only picker.

## Repository guidance

|---|---|---|
| AGENTS.md runtime boundaries | Domain backends belong in web/runtime modules | All Docker logic lives in `packages/web/server/lib/...`; UI consumes server APIs; no new IPC surface in electron |
| `openchamber-change-discipline` | Multi-step destructive work, persisted state | Journal-driven rollback, explicit failure states (`error`, `removal-failed`), feature toggle default off, focused tests per module |
| `ui-api-decoupling` | New OpenChamber routes + runtime switching | Routes through `runtimeFetch`, zod parsing at the boundary, runtime-switch safety followed (identity re-scope on upstream change) |
| `theme-system` + `locale-ui-patterns` | New UI components and copy | Shared primitives, semantic tokens, no hardcoded colors; every new key added to all 12 dictionaries with real translations |
| `lib/opencode/DOCUMENTATION.md` | Owning module docs | Updated for new modules and the path-mapping resolution order |

## Validation

| Check | Result |
|---|---|
| Focused suites (docker runtime/registry/mounts/lifecycle/routes, path-mapping incl. #3306 regressions, global-hub routing, network-runtime override, settings toggle) | 126+ passed |
| UI behavior tests (section states, confirmation gates, web entry) | 7 passed |
| `tsc --noEmit` (ui, web), knip, oxlint on new files | clean (remaining findings are the repo's established server-JS `typeof` guard pattern, identical in pre-existing files) |
| Full package suites | green except failures reproduced identically on the base branch (Windows-environment: `sleep` spawn, git longpaths/WSL, keychain) + 1 timing-flaky SSE test (passes standalone 3/3) |
| Real-docker flow (maintainer-supervised, per host-machine rule) | create -> readiness -> chat (replies streamed) -> switch Local <-> Docker -> stop/remove confirmed interactively |
| Security audit (read-only, static) | 1 High finding (skillsHostDir guard bypass) **fixed** with regression test; personal-path fixtures scrubbed; Medium items filed as follow-ups |
| Not verified | prebuilt-registry pull path (no published image yet); multi-hour soak; non-Windows hosts |

## Visual evidence

The feature is fully gated behind a settings toggle (default off), so the merged default behavior is visually identical to the current app. The maintainer verified the interactive flows on the dev build; screenshots below (personal paths masked).

**Instance selector with the Docker section** -- running instance with lifecycle badge, stop/remove controls, and the add-instance entry next to Local:

Instance selector with Docker section

**Create dialog** -- workspace picker, explicit mount disclosures (what the container may write, what is never shared), and the sharing toggles (credentials default off):

Create dialog

**Docker instance active** -- the header label shows the active Docker instance while the sidebar groups the container workspace:

Docker instance active

## Risks and failure behavior

- **Failure/rollback**: journal-driven rollback removes exactly the failed operation's resources (container log tail captured before removal for diagnosability); failed removals park as `removal-failed` with a retryable journal. Feature toggle off = zero behavior change (startup restore is toggle-gated too); PR revert leaves at most OpenChamber-labeled containers (one documented `docker rm`).
- **Security**: containers non-root, loopback-only ports, no socket/SSH/git-credential mounts (guard covers workspace and the user-suppliable skills override), credentials read-only behind a default-off toggle, host config/skills/env never modified by OpenChamber.
- **Known limitations**: single active upstream (matches the existing model); sidebar grouping for docker sessions follows the workspace mapping; first-run image requires the explicit Build/Pull action.
- **Follow-ups filed in this description thread**: per-instance upstream credentials, toggle-off releases the active upstream, workspace denylist hardening, readiness-gate override awareness, image-ref validation + pinned `OPENCODE_VERSION`, startup orphan reconciliation, prebuilt registry image, title-language consistency.

(Prepared with AI assistance; the maintainer ran all real-Docker verification personally.)


mattv8:
I can help review this one. @Lorrid is this requiring significant build times for OpenChamber to spin up the containers each time? Does it do any sort of existing container Discovery? Like for example if the user already has setup docker for their development stack (which will most likely be the case)? Is this feature fully platform agnostic (WSL/MacOS/Linux)?

Lorrid:
> I can help review this one. @Lorrid is this requiring significant build times for OpenChamber to spin up the containers each time? Does it do any sort of existing container Discovery? Like for example if the user already has setup docker for their development stack (which will most likely be the case)? Is this feature fully platform agnostic (WSL/MacOS/Linux)?

Hello

Build: nothing builds on spin-up. The image is built once via an explicit Build/Pull button in the create flow; after that instances are just `docker create` + `docker start` against that existing image, so startup is normal container-start time.

Discovery: intentionally no in this PR, the runtime only ever touches containers OpenChamber created itself (everything is filtered by an `openchamber.instance` label), so an existing docker dev stack is never touched. The only interaction with your docker is the mounts you opt into in the create dialog. That said, pointing OpenChamber at an already-running external OpenCode server is already supported (OPENCODE_HOST), so a manually-managed container works today, it's just not managed *by* OpenChamber. Adopting existing containers could be a follow-up; if that's something you'd actually use, happy to explore it.

Platform: it wraps the `docker` CLI via execFile (no socket client, no new dependency), so it works wherever the CLI + daemon work: Docker Desktop Windows/macOS, Engine, WSL. Path translation from #3306 handles host↔container paths. Caveat: real-docker flow is manually verified on Windows so far, testing on mac/linux would be very welcome.

Full disclosure: I built this with heavy AI assistance as a hobbyist, so if anything above needs double-checking or looks off, point at it and I'll dig in.

mattv8:
Thanks for the substantial work here.

> Full disclosure: I built this with heavy AI assistance as a hobbyist, so if anything above needs double-checking or looks off, point at it and I'll dig in.

We appreciate the disclosure. It isn't a problem. OpenChamber is an agentic development application, and AI-assisted contributions are welcome. My review is focused on whether this is the right product direction and whether it could introduce regressions or other secondary issues.

I'd like to request a first round of changes. The current implementation has gaps that prevent users from enabling the feature and can leave OpenChamber connected to the wrong or unavailable OpenCode server.

This PR adds roughly 4,900 Docker-specific lines and manages only containers created by OpenChamber. It does not attach to existing Compose or devcontainer environments, and it does not support the Podman or rootless workflows requested in [#1626](https://github.com/openchamber/openchamber/issues/1626). Users can already connect to manually managed servers through `OPENCODE_HOST`.

The isolation benefit goes beyond `OPENCODE_HOST`, so the idea has value. Before investing further, @btriapitsyn needs to confirm that an OpenChamber-owned Docker manager is the direction we want instead of a smaller launch-profile system that could support Docker, Compose, devcontainers, and Podman.

If we want this product direction, I need the following gaps closed.

### 1. The feature cannot be enabled through normal settings

`settings-helpers.js:436-438` accepts `dockerInstancesEnabled`, but the registry filter at `settings-helpers.js:961-964` deletes it because the key is missing from `packages/ui/src/lib/settings/registry.ts` and the generated registry snapshots. I also could not find a Settings control that writes it.

The PR's own `settings-docker-toggle.test.js` confirms the problem: it expects `true` after sanitization but receives `undefined`. Since both route registration and startup restoration require this setting, users cannot reach the feature through supported controls.

Please add the key to the canonical settings registry, regenerate the snapshots, and provide the intended control or supported configuration path. The persistence test and registry snapshot tests should pass afterward.

### 2. Stopping an active instance breaks all OpenCode requests

`stopInstance()` in `lifecycle-manager.js:294-315` stops the container but leaves both the in-memory upstream and persisted `activeInstanceId` selected. `network-runtime.js:95-103` continues routing every OpenCode request to that stopped container.

I reproduced this directly. After creating, activating, and stopping an instance, the record reported `stopped` while `getActiveUpstream()` and `getActiveInstanceId()` still pointed to it.

Disabling the feature has the same problem. The routes become unavailable, but an active Docker upstream remains selected.

Please make stopping the active instance and disabling the feature deactivate it. Activation and deactivation also need safe ordering so a failed registry write cannot leave routing, path mapping, and persisted state disagreeing. Add lifecycle tests for stopping an active instance, disabling the feature with an active instance, and persistence failures.

### 3. Startup restoration does not establish the client runtime identity

The server restores the active Docker upstream asynchronously in `server/index.js:1960-1973`. The client only calls `setRuntimeIdentityOverride()` from `DockerInstanceSection.refresh()` at `DockerInstanceSection.tsx:69-89`, and that section normally mounts only after the user opens the management dialog.

The always-mounted `useActiveDockerInstanceLabel()` hook fetches the active instance, but it updates only the label. This allows the server to restore Docker routing while the client continues using the `local` runtime key. Sessions and other runtime-scoped data can then load into the wrong caches.

Please initialize the runtime identity from authoritative server state without requiring the dialog to open. Add a restart test that begins with a persisted active Docker instance and proves Docker state never loads under the local runtime key.

### 4. A corrupt registry can orphan running containers

`store.js:87-114` converts every parse failure into an empty registry. The next mutation overwrites the original file, removing the only association between OpenChamber and its running containers. The warning says a copy is saved aside, but the code does not create one.

Please fail the read or quarantine the damaged file and refuse destructive mutations until OpenChamber has trustworthy state. Update `store.test.js` so corruption cannot become an authoritative empty result or overwrite the original registry.

### 5. Enforce the security boundaries claimed by the implementation

The Docker runtime comments say lifecycle operations address only OpenChamber-owned containers, but stop and removal use bare container IDs from the registry without checking the `openchamber.instance` label. Please verify ownership before each destructive operation and add a mismatched-label test.

Containers also receive `env: {}` at `lifecycle-manager.js:202-209`, leaving the OpenCode API unauthenticated on its published host port. Loopback prevents remote network access, but it does not isolate the API from other users or processes on the host.

Please use a per-instance `OPENCODE_SERVER_PASSWORD` and route the matching credentials through the proxy, or document and enforce a single-user-host security model.

Lorrid:
> Thanks for the substantial work here.
> 
> > Full disclosure: I built this with heavy AI assistance as a hobbyist, so if anything above needs double-checking or looks off, point at it and I'll dig in.
> 
> We appreciate the disclosure. It isn't a problem. OpenChamber is an agentic development application, and AI-assisted contributions are welcome. My review is focused on whether this is the right product direction and whether it could introduce regressions or other secondary issues.
> 
> I'd like to request a first round of changes. The current implementation has gaps that prevent users from enabling the feature and can leave OpenChamber connected to the wrong or unavailable OpenCode server.
> 
> This PR adds roughly 4,900 Docker-specific lines and manages only containers created by OpenChamber. It does not attach to existing Compose or devcontainer environments, and it does not support the Podman or rootless workflows requested in [#1626](https://github.com/openchamber/openchamber/issues/1626). Users can already connect to manually managed servers through `OPENCODE_HOST`.
> 
> The isolation benefit goes beyond `OPENCODE_HOST`, so the idea has value. Before investing further, @btriapitsyn needs to confirm that an OpenChamber-owned Docker manager is the direction we want instead of a smaller launch-profile system that could support Docker, Compose, devcontainers, and Podman.
> 
> If we want this product direction, I need the following gaps closed.
> 
> ### 1. The feature cannot be enabled through normal settings
> `settings-helpers.js:436-438` accepts `dockerInstancesEnabled`, but the registry filter at `settings-helpers.js:961-964` deletes it because the key is missing from `packages/ui/src/lib/settings/registry.ts` and the generated registry snapshots. I also could not find a Settings control that writes it.
> 
> The PR's own `settings-docker-toggle.test.js` confirms the problem: it expects `true` after sanitization but receives `undefined`. Since both route registration and startup restoration require this setting, users cannot reach the feature through supported controls.
> 
> Please add the key to the canonical settings registry, regenerate the snapshots, and provide the intended control or supported configuration path. The persistence test and registry snapshot tests should pass afterward.
> 
> ### 2. Stopping an active instance breaks all OpenCode requests
> `stopInstance()` in `lifecycle-manager.js:294-315` stops the container but leaves both the in-memory upstream and persisted `activeInstanceId` selected. `network-runtime.js:95-103` continues routing every OpenCode request to that stopped container.
> 
> I reproduced this directly. After creating, activating, and stopping an instance, the record reported `stopped` while `getActiveUpstream()` and `getActiveInstanceId()` still pointed to it.
> 
> Disabling the feature has the same problem. The routes become unavailable, but an active Docker upstream remains selected.
> 
> Please make stopping the active instance and disabling the feature deactivate it. Activation and deactivation also need safe ordering so a failed registry write cannot leave routing, path mapping, and persisted state disagreeing. Add lifecycle tests for stopping an active instance, disabling the feature with an active instance, and persistence failures.
> 
> ### 3. Startup restoration does not establish the client runtime identity
> The server restores the active Docker upstream asynchronously in `server/index.js:1960-1973`. The client only calls `setRuntimeIdentityOverride()` from `DockerInstanceSection.refresh()` at `DockerInstanceSection.tsx:69-89`, and that section normally mounts only after the user opens the management dialog.
> 
> The always-mounted `useActiveDockerInstanceLabel()` hook fetches the active instance, but it updates only the label. This allows the server to restore Docker routing while the client continues using the `local` runtime key. Sessions and other runtime-scoped data can then load into the wrong caches.
> 
> Please initialize the runtime identity from authoritative server state without requiring the dialog to open. Add a restart test that begins with a persisted active Docker instance and proves Docker state never loads under the local runtime key.
> 
> ### 4. A corrupt registry can orphan running containers
> `store.js:87-114` converts every parse failure into an empty registry. The next mutation overwrites the original file, removing the only association between OpenChamber and its running containers. The warning says a copy is saved aside, but the code does not create one.
> 
> Please fail the read or quarantine the damaged file and refuse destructive mutations until OpenChamber has trustworthy state. Update `store.test.js` so corruption cannot become an authoritative empty result or overwrite the original registry.
> 
> ### 5. Enforce the security boundaries claimed by the implementation
> The Docker runtime comments say lifecycle operations address only OpenChamber-owned containers, but stop and removal use bare container IDs from the registry without checking the `openchamber.instance` label. Please verify ownership before each destructive operation and add a mismatched-label test.
> 
> Containers also receive `env: {}` at `lifecycle-manager.js:202-209`, leaving the OpenCode API unauthenticated on its published host port. Loopback prevents remote network access, but it does not isolate the API from other users or processes on the host.
> 
> Please use a per-instance `OPENCODE_SERVER_PASSWORD` and route the matching credentials through the proxy, or document and enforce a single-user-host security model.

Thanks for taking the time to review this and for the detailed feedback.

I'll wait for @btriapitsyn's decision on the direction before making any further changes.

Appreciate the review.

PR #3396, opened 2026-09-07
fix(ui): reconcile stale active tools after reload

alvins82:
## Intent

Prevent stale persisted tool state from appearing as an indefinitely running command after an OpenChamber reload.

This PR covers both invalid transcript shapes:

- A completed assistant message containing a `pending` or `running` tool is reconciled locally as `error`/`Interrupted` with an end time.
- An unfinished assistant turn left behind by an interrupted sidecar process is reconciled after reload once the transcript is hydrated and the sidecar's authoritative empty status snapshot establishes the session as idle.

The first case is the materialization defense. The second case is the cold-reload gap: status bootstrap and message hydration happen independently, and the previous implementation treated a missing per-session status entry as unknown even after a successful authoritative `{}` snapshot.

Sidecar root fix: https://github.com/alvins82/omp-openchamber-server/pull/31

## Non-goals

This does not change live tool streaming, the sidecar transcript mapper, persisted server data, OpenCode APIs, terminal tool states already observed by the client, or the meaning of a failed status fetch. A failed fetch still leaves the session status unknown and therefore does not trigger interruption recovery.

## Affected surfaces

The shared `packages/ui` sync path used by web, desktop, VS Code, hosted mobile, and Capacitor mobile. The change affects only client-side reconciliation after authoritative status snapshots and message hydration; it does not change the sidecar or any persisted session record.

## Repository guidance

|---|---|---|
| `openchamber-change-discipline` | This changes shared UI sync source, tests, and owning documentation. | Kept the change focused to reload reconciliation and its regression coverage. |
| `sync-state-invariants` | The bug is stale session/tool lifecycle state across bootstrap, status snapshots, and message hydration. | Successful authoritative snapshots seed explicit idle only for their candidates; recovery runs after hydration; live busy events and failed fetches are preserved. |
| `communication-style` | The sync documentation describes behavior maintainers need to understand. | Added a concise explanation of the cold-reload ordering and failure behavior. |
| `packages/ui/src/sync/DOCUMENTATION.md` and `packages/ui/README.md` | Sync ownership and package guidance apply to the shared source. | Updated the sync invariant; no package README exists. |

## Validation

| Check | Result |
|---|---|
| `bun test packages/ui/src/sync/__tests__/message-completion-status-poll.test.ts packages/ui/src/sync/__tests__/session-status-snapshot.test.ts packages/ui/src/sync/__tests__/interrupted-turn-tools.test.ts packages/ui/src/sync/__tests__/materialization.test.ts` | Pass: 64 tests, 0 failures. Includes the cold-reload regression. |
| `bun run type-check:ui` | Pass. |
| `bun run lint:ui` | Pass: 0 errors; one pre-existing warning in `packages/ui/src/apps/MobileChangesSurface.tsx:212`. |
| `git diff --check` | Pass. |
| `bunx oxlint` on changed TypeScript files | Non-zero because the legacy sync files contain existing anti-slop findings; no new production typing or runtime behavior was added to satisfy those unrelated findings. |

## Visual evidence

No layout or styling changes. The user-visible change is a state transition: a stale active tool no longer displays an unbounded running timer after reload and instead renders as interrupted. The transition is covered by state-level regression tests; no new live-session recording was captured because this fix does not alter the rendered layout or interaction sequence.

## Risks and failure behavior

The recovery is deliberately gated. Only a successful authoritative status snapshot can establish an explicit idle state for a previously unknown candidate. A live busy/retry status is not lowered, a pending question or permission prevents interruption, and a failed status request leaves the transcript untouched. The interrupted marker is local and can be superseded by a later authoritative terminal server snapshot.


btriapitsyn:
Thanks for fixing interrupted-turn recovery after reload. Merged.

We added runtime, SDK, and request ownership guards so a delayed recovery response cannot modify a different runtime after a switch: https://github.com/openchamber/openchamber/commit/ba74cf7294e2eba8e34e19b4f907c5c3cbf77556. The delayed-response regression tests pass.

Issue #3328, opened 2026-09-04
[Desktop/Windows] xAI quota returns fetch failed behind HTTP proxy; Electron fetch ignores HTTP(S)_PROXY


ScoFan-official:
### Environment
- OpenChamber: 1.22.0 (Desktop, Windows, Electron)
- Network: launcher injects Clash HTTP(S)_PROXY (`HTTP_PROXY`/`HTTPS_PROXY`), chat works
- Provider: xAI via Grok OAuth login, `xai` is in `usageDropdownProviders`
- Subscription: SuperGrok Heavy (weekly quota expected in right-sidebar Usage)

### Expected
Right-sidebar Usage shows the SuperGrok weekly quota.

### Actual
`GET /api/quota/xai` returns:

```json
{"providerId":"xai","configured":true,"ok":false,"error":"fetch failed"}

### Where does it happen?

Desktop (macOS)

### Version (if known)

1.22.0

### Screenshots / recordings (optional)

_No response_

### Logs (optional)

```shell
Reproduction evidence (sanitized)
Same OAuth state, fetchQuotaForProvider('xai') under Bun succeeds: billing_cycle used 4%, remaining 96%, reset 2026-09-10 10:26.
curl to grok.com returns 200; Bun fetch returns 200.
Node/Electron native fetch to grok.com fails (Connect Timeout), so the quota route fails.
NODE_USE_ENV_PROXY=1 makes plain Node return 200, but it was reverted per local policy and is not proposed as the fix.
An external undici EnvHttpProxyAgent / ProxyAgent('http://127.0.0.1:7890') still fails under this Electron/Node runtime: TypeError: fetch failed, cause invalid onRequestStart method. So an external dispatcher is not a viable fix here.
Questions for maintainers
Is SuperGrok consumer weekly quota via grok.com/grok_api_v2.GrokBuildBilling/GetGrokCreditsConfig a supported source for Usage?
What is the accepted Electron-compatible transport for proxied quota requests (native https + CONNECT vs Electron net)?
Should the SuperGrok weekly window render as weekly instead of the current generic billing_cycle?
```

Issue #3105, opened 2026-08-24
[Feature Request] Honor HTTP_PROXY / HTTPS_PROXY / NO_PROXY for quota fetch


Ksndj:
### What should we add/change?

I'm trying to use OpenChamber Desktop on Windows behind a local HTTP proxy (Clash mixed port, TUN off). Chat to `api.x.ai` / OpenAI works, but the Usage panel shows **Codex** and **xAI** as `fetch failed`.

OpenChamber quota adapters call bare `fetch()`:

- Codex: `https://chatgpt.com/backend-api/wham/usage` (`packages/web/server/lib/quota/providers/codex.js`)
- xAI: `https://grok.com/grok_api_v2.GrokBuildBilling/GetGrokCreditsConfig` (`packages/web/server/lib/quota/providers/xai.js`)

Those hosts are not reachable without a proxy on this network. `HTTP_PROXY` / `HTTPS_PROXY` are set, but Electron/Node `fetch` does not use them unless `NODE_USE_ENV_PROXY=1`.

Setting `NODE_USE_ENV_PROXY=1` makes Codex/xAI usage load, but it also proxies loopback. Cloudflare remote MCP OAuth then gets stuck on “waiting for OpenCode to recognize the completed browser auth flow” because the callback is `http://127.0.0.1:19876`.

It would be great if OpenChamber natively honored `HTTP_PROXY`, `HTTPS_PROXY`, and `NO_PROXY` for quota (and other server-side) `fetch` calls — e.g. an `undici` `ProxyAgent` / env-proxy dispatcher — so usage works without `NODE_USE_ENV_PROXY`, and `NO_PROXY=localhost,127.0.0.1,::1` keeps OAuth callbacks local.

### Extra context (optional)

- OpenChamber Desktop **1.20.0** on Windows
- Related symptom (no proxy mentioned): https://github.com/openchamber/openchamber/issues/2640
- Confirmed locally: `/api/quota/codex` and `/api/quota/xai` return `{ ok: false, configured: true, error: "fetch failed" }` without the env flag; both succeed after `NODE_USE_ENV_PROXY=1`
- Direct `chatgpt.com` / `grok.com` time out; the same URLs succeed through `http://127.0.0.1:7897`
- Official env docs have no proxy setting: https://docs.openchamber.dev/environment/

Ksndj:
Thanks for the questions.

[...]

Issue #2640, opened 2026-08-05
[Bug] Can't get Codex usage，Show "fetch failed",OpenAI account already connected


Jayli22:
### What's wrong? Maybe some steps to reproduce.

Can't get Codex usage,Show "fetch failed". 

Windows Desktop Version1.18.1
Opencode version 1.18.13

### Where does it happen?

Not sure

### Version (if known)

1.18.1

### Screenshots / recordings (optional)

<img width="323" height="204" alt="Image" src="https://github.com/user-attachments/assets/6aff54b2-a2db-48fc-8662-4837eee0d2a3" />

### Logs (optional)

```shell

```

Jayli22:
> Thanks for the detailed report! We've tagged this as a bug for the providers area (OpenAI/Codex usage fetch failing) on Windows desktop.
> 
> To help us track this down, could you share a bit more info?
> 
> 1. **Logs** — the logs section is empty. Could you run the desktop app (or the `opencode` CLI) with verbose logging and share any `fetch failed` stack traces or network errors? Console/network errors from DevTools in the desktop app would also help.
> 2. **Reproduction steps** — what exactly did you click to trigger the usage fetch (e.g., opening Settings → usage panel, or a usage widget on the home screen)? Does it fail every time or intermittently?
> 3. **Setup** — how is the OpenAI account connected (ChatGPT/Copilot sign-in vs. API key)? Does usage load correctly for other providers, or is Codex/OpenAI the only one affected?
> 4. **Network** — any proxies, VPNs, or firewall settings on the Windows machine that might block the usage endpoint?
> 
> Happy to help dig in once we have those details!

1. About logs, I check logs (main.log)didn't find any with codex or open ai fetch failed text
2. No Reproduction steps,  can consistently reproduce
3. Open AI account connected by chatGPT sign-in. Other providers' usage are correctly load(KIMI)
4. I use proxies (Clash Verge) cause I need it so can I access to chatGPT

yulia-ivashko:
[...]

PR #3404, opened 2026-09-07
fix(quota): raise Node's 250ms connect-attempt cap that breaks provider fetches

### LINKED ISSUE #3399: [Bug] z.ai coding plan usage shows "fetch failed" — quota fetch from Electron main process fails when IPv4 handshake >250ms and IPv6 unreachable (Node autoSelectFamilyAttemptTimeout)

The usage panel shows **"fetch failed"** for the **z.ai coding plan** (`zai-coding-plan`) provider. Chat with the same provider works fine in the same session, and usage renders correctly for other providers (Codex, Minimax coding plan) on the same machine.

**Root cause (diagnosed and reproduced):** the z.ai quota fetch runs in the **Electron main process (Node/undici)**, while chat runs through the bundled `opencode-cli` (Bun). On hosts where:

- `api.z.ai` DNS returns AAAA records that are unreachable (IPv4-only egress → instant `ENETUNREACH`), **and**
- the IPv4 TCP handshake to `api.z.ai` takes **>250ms** (measured 260–860ms here; the endpoints are simply geographically far),

…Node's default `autoSelectFamilyAttemptTimeout` of **250ms** aborts every IPv4 connect attempt (`ETIMEDOUT`), all addresses are exhausted, and undici throws `TypeError: fetch failed`. Bun has no such cap, which is why chat works while usage fails — the asymmetry makes it look like a provider problem when it is purely a client-side connect-timeout issue.

**Reproduction** (same host, `fetch("https://api.z.ai/api/coding/paas/v4/usage")` ×3):

| Runtime | Result |
|---|---|
| Node v24.19.0 (undici — what the Electron main uses) | **3/3 fail**: `TypeError: fetch failed` — causes: `ETIMEDOUT@8.217.100.151`, `ENETUNREACH@240b:4001:...`, `ETIMEDOUT@8.217.233.95`, `ENETUNREACH@240b:4001:...` |
| Bun 1.3.14 (what the bundled opencode CLI uses) | 3/3 OK — HTTP 401 unauthenticated, i.e. the endpoint is reachable from this host |
| Node + `NODE_OPTIONS=--network-family-autoselection-attempt-timeout=5000` | **3/3 OK** |

Probe script:

```js
const url = "https://api.z.ai/api/coding/paas/v4/usage";
for (let i = 0; i < 3; i++) {
  try {
    const r = await fetch(url, { signal: AbortSignal.timeout(15000) });
    console.log(`attempt ${i}: OK status=${r.status}`);
  } catch (e) {
    const agg = e.cause?.errors?.map(x => `${x.code}@${x.address ?? ""}`).join(", ");
    console.log(`attempt ${i}: FAIL ${e.name}: ${e.message}${agg ? " | causes: " + agg : ""}`);
  }
}
```

Also noted: `--dns-result-order=ipv4first` does **not** fix it (the 250ms cap applies per attempt regardless of ordering).

**Suggested fix** — raise the connect-attempt cap process-wide in the Electron main early init, next to any existing process-wide network defaults:

```js
import { setDefaultAutoSelectFamilyAttemptTimeout } from "node:net"
// api.z.ai and other distant endpoints routinely need 300–1500ms for the TCP
// handshake; Node's 250ms default aborts every attempt → "fetch failed".
try { setDefaultAutoSelectFamilyAttemptTimeout(5000) } catch {}
```

Please **not** `setDefaultAutoSelectFamily(false)`: that also "fixes" it but removes the IPv6→IPv4 fallback, breaking `::1`/localhost MCP servers (measured in the opencode issue below). An alternative is routing quota fetches through the bundled Bun opencode process, which connects fine.

Note a user-side `NODE_OPTIONS` workaround is not viable — Electron filters `NODE_OPTIONS` in packaged apps.

Possibly related: #2640 (same `"fetch failed"` symptom for Codex usage on Windows — plausibly the same mechanism with `chatgpt.com` under different network conditions), and #3012 (earlier z.ai quota adapter defect — different mechanism, endpoint/type drift, already fixed). The identical Electron/Node-vs-Bun 250ms-cap mechanism was diagnosed in opencode's Desktop app: anomalyco/opencode#45178, including the fix pattern above and the `autoSelectFamily(false)` pitfall.

### Where does it happen?

Desktop (Linux) — AppImage. (The form's dropdown has no Linux option; closest is Desktop (macOS), which is wrong, so noting it here.)

### Version (if known)

1.22.2 (Linux x64 AppImage, release dated 2026-09-05)

### Logs (optional)

```shell
# Node v24.19.0, fetch to api.z.ai usage endpoint ×3 (same as Electron main process):
attempt 0: FAIL TypeError: fetch failed | causes: ETIMEDOUT@8.217.100.151, ENETUNREACH@240b:4001:139:8a07:..., ETIMEDOUT@8.217.233.95, ENETUNREACH@240b:4001:139:8a02:...
attempt 1: FAIL TypeError: fetch failed | causes: ETIMEDOUT@8.217.100.151, ENETUNREACH@240b:4001:139:8a02:..., ETIMEDOUT@8.217.233.95, ENETUNREACH@240b:4001:139:8a07:...
attempt 2: FAIL TypeError: fetch failed | causes: ETIMEDOUT@8.217.233.95, ENETUNREACH@240b:4001:139:8a02:..., ETIMEDOUT@8.217.100.151, ENETUNREACH@240b:4001:139:8a07:...

# Bun 1.3.14 (bundled opencode CLI runtime), identical fetch:
attempt 0: OK status=401
attempt 1: OK status=401
attempt 2: OK status=401

# Node v24.19.0 + NODE_OPTIONS=--network-family-autoselection-attempt-timeout=5000:
attempt 0: OK status=401
attempt 1: OK status=401
attempt 2: OK status=401

# Host network context:
# curl -4 to api.z.ai: HTTP 401 (reachable), TCP connect 0.26–0.86s
# curl -6 to api.z.ai: connection fails instantly (no global IPv6 route)
# DNS: api.z.ai → 2×A (8.217.x) + 2×AAAA (240b:4001:...)
```

ouyangjian28:
Fixes #3399

## Intent

The usage panel shows **"fetch failed"** for the z.ai coding plan while chat with the same provider works, on hosts where `api.z.ai` resolves to unreachable AAAA records and the IPv4 TCP handshake takes more than 250ms. The reporter's diagnosis (confirmed `fix-ready` in triage): quota fetches run in the **Electron main process under Node/undici**, whose happy-eyeballs default `autoSelectFamilyAttemptTimeout = 250` aborts each connect attempt before a distant endpoint's handshake completes; every address is exhausted and the fetch surfaces as `TypeError: fetch failed`. Chat runs through the bundled Bun-based opencode CLI, which has no such cap — hence the confusing asymmetry.

Reporter's reproduction on the affected host (`fetch("https://api.z.ai/api/coding/v4/usage")` ×3): Node **3/3 fail** (`ETIMEDOUT@8.217.x` + `ENETUNREACH@240b:...`), Bun **3/3 OK** (HTTP 401), and Node **3/3 OK** with `--network-family-autoselection-attempt-timeout=5000` — a user-side `NODE_OPTIONS` workaround is not viable because Electron filters it in packaged apps.

The same quota module also runs in the `openchamber` CLI server process (`bin/cli.js`, `#!/usr/bin/env node`), which shares the Node default, so the fix is applied at both Node entrypoints that host it:

- New `packages/web/server/lib/network-defaults.js` — `applyConnectAttemptTimeout()` calls `net.setDefaultAutoSelectFamilyAttemptTimeout(5000)`, feature-safe (try/catch, no-op where the setter does not exist) and **does not** touch `setDefaultAutoSelectFamily`, which the issue explicitly warns against (disabling family autoselection breaks `::1`/localhost MCP servers).
- `packages/electron/main.mjs` — called once next to the existing process-wide network defaults (`proxy-bypass-list`, connection-limit policy).
- `packages/web/bin/cli.js` — called once at process start, covering `openchamber serve` and every other CLI command that performs provider fetches.

## Non-goals

- No change to per-fetch behavior: timeouts, headers, retries, and the quota adapters themselves are untouched; only the per-attempt connect cap of the process changes.
- Not routing quota fetches through the bundled Bun opencode process (the issue lists it as an alternative; the process default is the smaller change).
- Not adding HTTP_PROXY support — that is a different failure mode tracked separately (#3328, #3105).
- The VS Code extension's own quota stack (`packages/vscode/src/quotaProviders.ts`, a separate fetch path that does not import this server module) is untouched; no report ties it to this symptom.
- Bun-run web deployments are untouched — Bun has no 250ms cap and the helper is a no-op there.

## Affected surfaces

- `packages/web`: new `server/lib/network-defaults.js` (+ vitest), one call at the top of `bin/cli.js`.
- `packages/electron`: one import and one call in `main.mjs` early init, next to the existing Chromium network switches.
- Every provider quota adapter served from a Node process benefits (zai, codex, xai, …) — they all use the global `fetch()` in-process. Plausibly the same mechanism behind #2640 (Codex `fetch failed` on Windows), which this also addresses when its handshake exceeds 250ms.
- Desktop (Electron in-process server) and CLI/web (`openchamber serve` under Node) are the two affected runtimes. Bun-run servers, VS Code, mobile, and relay transports are unaffected.
- No persisted data, routes, API contracts, or exports consumed elsewhere change.

## Repository guidance

|---|---|---|
| `openchamber-change-discipline` | Executable source + process-wide runtime behavior | Smallest complete change: one shared helper + one call per Node entrypoint that hosts the quota module; classified as platform/runtime behavior risk, validated with focused tests plus package-scoped type-check/lint and both full package suites |
| `desktop-shell` + `packages/electron/README.md` | Electron main init change | One call next to the existing network defaults; no preload/IPC surface; domain stays in the web-server module that owns the fetches |
| Root `AGENTS.md` validation rules | Server JS and CLI JS are not covered by TS lint | Focused vitest for the new module, `node --check` for both entry files, full web + electron suites |

## Validation

| Check | Result |
|---|---|
| `bunx vitest run server/lib/network-defaults.test.js` (packages/web) | 4/4 pass: real `node:net` value flips 250→5000 and restores; no-op on `{}`; survives throwing setter; never calls `setDefaultAutoSelectFamily` |
| `bun run --cwd packages/web test` (isolated runner) | 1989 passed, 2 skipped |
| `bun run --cwd packages/electron test` (includes main.mjs change) | 19/19 test files passed |
| `bun run --cwd packages/web type-check` (`tsc --noEmit`) | pass |
| `bun run --cwd packages/electron type-check` (`node --check main.mjs preload.mjs`) + `node --check` on `cli.js` and `network-defaults.js` | pass |
| `bun run --cwd packages/web lint` | pass |
| `bunx oxlint` on the four touched files | no findings on the new files or `cli.js`; `main.mjs` findings pre-exist on untouched lines |
| `bun run dead-code` | no new findings (same pre-existing report as base) |
| Author host corroboration (below) | Node default measured `getDefaultAutoSelectFamilyAttemptTimeout() === 250`; `api.z.ai` resolves AAAA `240b:4001:…` with no global IPv6 route on this host either |

Author-host network profile (second data point for the DNS premise, gathered before the fix):

```
$ node -p "require('node:net').getDefaultAutoSelectFamilyAttemptTimeout()"
250
$ getent ahosts api.z.ai | sort -u
240b:4001:139:8a02:… / 240b:4001:139:8a07:… (AAAA) + 8.217.100.151 (A)
$ curl -6 https://api.z.ai/ → instant failure (no global IPv6 route)
```

Not verified by hand: a live before-fail on this author's own egress — its TCP connect to `api.z.ai` is intercepted by a local TUN proxy and completes in <1ms, so the 250ms cap never bites here. The operative before/after evidence is the reporter's 3/3 table (Node fail → Node+5000ms OK), which the reporter measured on an unproxied host; the unit test proves the applied value on real `node:net`, and the fix matches the triage-confirmed suggestion verbatim in mechanism.

## Visual evidence

No rendered UI changes. The user-visible delta for the reporter's scenario: the usage panel for z.ai coding plan changes from `fetch failed` to rendered quota numbers on hosts with slow handshakes; no layout, text, or theme change. The reporter's environment is the only place the failing state is reproducible; this author's host cannot show the before state for the egress reason documented above.

## Risks and failure behavior

- The default is process-wide in the two entrypoints: every outbound `fetch` there now allows up to 5s per address family attempt before falling back, instead of 250ms. Happy-eyeballs interleaving is unchanged, so reachable hosts are unaffected; the worst case is that a genuinely dead host's connect failure surfaces seconds later than before. No fetch becomes less patient, and no overall request timeout is removed — adapters keep their existing abort/timeout behavior.
- On runtimes without `setDefaultAutoSelectFamilyAttemptTimeout`, the helper returns `false` and behavior is exactly as today (try/catch feature safety, tested).
- `setDefaultAutoSelectFamily` is never called, so IPv6→IPv4 fallback and `::1`/localhost MCP servers keep working — the regression the issue warns about is structurally excluded (asserted by test).
- Rollback is a single revert of the helper and the two call sites; no persisted state or external contract is involved.

### DIFF
--- a/packages/electron/main.mjs
+++ b/packages/electron/main.mjs
@@ -37,6 +37,7 @@ import { createRelayDevTunnelBridge } from './relay-dev-tunnel.mjs';
 import { attachRendererRecovery } from './renderer-recovery.mjs';
 import { mintOutsideFileGrant } from '@openchamber/web/server/lib/fs/routes.js';
 import { fetchUpdateNotes } from '@openchamber/web/server/lib/changelog/update-notes.js';
+import { applyConnectAttemptTimeout } from '@openchamber/web/server/lib/network-defaults.js';
 
 const execFileAsync = promisify(execFile);
 
@@ -104,6 +105,11 @@ if (shouldIgnoreLoopbackConnectionLimit({
 })) {
   app.commandLine.appendSwitch('ignore-connections-limit', '127.0.0.1,localhost');
 }
+// This process runs quota/provider fetches under Node/undici, whose happy-eyeballs
+// default aborts each connect attempt after 250ms — distant provider endpoints
+// routinely need longer handshakes, surfacing as "fetch failed" (#3399). No-op on
+// runtimes without the setter.
+applyConnectAttemptTimeout();
 
 protocol.registerSchemesAsPrivileged([
   {
--- a/packages/web/bin/cli.js
+++ b/packages/web/bin/cli.js
@@ -72,6 +72,13 @@ import {
   printJson,
   logStatus,
 } from './cli-output.js';
+import { applyConnectAttemptTimeout } from '../server/lib/network-defaults.js';
+
+// The CLI process performs provider fetches (quota/usage, update notes) under
+// Node/undici, whose happy-eyeballs default aborts each connect attempt after
+// 250ms — distant provider endpoints routinely need longer handshakes, surfacing
+// as "fetch failed" (#3399). No-op on runtimes without the setter.
+applyConnectAttemptTimeout();
 
 const __filename = fileURLToPath(import.meta.url);
 const __dirname = path.dirname(__filename);
--- a/packages/web/server/lib/network-defaults.js
+++ b/packages/web/server/lib/network-defaults.js
@@ -1,0 +1,23 @@
+import net from 'node:net';
+
+// Node caps each happy-eyeballs connect attempt at 250ms by default
+// (autoSelectFamilyAttemptTimeout). TCP handshakes to provider endpoints that are
+// geographically distant routinely take 300-1500ms, so fetch() from a Node process
+// aborts every attempt (ETIMEDOUT) and surfaces "fetch failed" even though the host
+// is reachable — e.g. the z.ai quota endpoint from an IPv4-only egress (#3399).
+//
+// Every Node process entrypoint that performs provider fetches raises the
+// per-attempt cap. Family autoselection itself stays enabled, so the IPv6→IPv4
+// fallback and ::1/localhost servers keep working; disabling it instead breaks
+// local MCP servers. Runtimes without the setter (Bun's fetch path, older Node)
+// are a no-op.
+export const CONNECT_ATTEMPT_TIMEOUT_MS = 5_000;
+
+export const applyConnectAttemptTimeout = (netModule = net) => {
+  try {
+    netModule.setDefaultAutoSelectFamilyAttemptTimeout(CONNECT_ATTEMPT_TIMEOUT_MS);
+    return true;
+  } catch {
+    return false;
+  }
+};
--- a/packages/web/server/lib/network-defaults.test.js
+++ b/packages/web/server/lib/network-defaults.test.js
@@ -1,0 +1,36 @@
+import { describe, expect, it, vi } from 'vitest';
+import net from 'node:net';
+
+import { applyConnectAttemptTimeout, CONNECT_ATTEMPT_TIMEOUT_MS } from './network-defaults.js';
+
+describe('applyConnectAttemptTimeout', () => {
+  it('raises the per-attempt connect timeout on runtimes that expose the setter', () => {
+    const previous = net.getDefaultAutoSelectFamilyAttemptTimeout();
+    try {
+      expect(applyConnectAttemptTimeout()).toBe(true);
+      expect(net.getDefaultAutoSelectFamilyAttemptTimeout()).toBe(CONNECT_ATTEMPT_TIMEOUT_MS);
+    } finally {
+      net.setDefaultAutoSelectFamilyAttemptTimeout(previous);
+    }
+  });
+
+  it('is a no-op on runtimes without the setter', () => {
+    expect(applyConnectAttemptTimeout({})).toBe(false);
+  });
+
+  it('survives a throwing setter', () => {
+    const setDefaultAutoSelectFamilyAttemptTimeout = vi.fn(() => {
+      throw new Error('not supported');
+    });
+    expect(applyConnectAttemptTimeout({ setDefaultAutoSelectFamilyAttemptTimeout })).toBe(false);
+  });
+
+  it('leaves family autoselection itself untouched', () => {
+    const setDefaultAutoSelectFamily = vi.fn();
+    const setDefaultAutoSelectFamilyAttemptTimeout = vi.fn();
+    applyConnectAttemptTimeout({ setDefaultAutoSelectFamily, setDefaultAutoSelectFamilyAttemptTimeout });
+    expect(setDefaultAutoSelectFamilyAttemptTimeout).toHaveBeenCalledTimes(1);
+    expect(setDefaultAutoSelectFamilyAttemptTimeout).toHaveBeenCalledWith(CONNECT_ATTEMPT_TIMEOUT_MS);
+    expect(setDefaultAutoSelectFamily).not.toHaveBeenCalled();
+  });
+});


btriapitsyn:
Thanks for fixing the Electron connection-attempt timeout and preserving address-family fallback. Merged.

The server daemon and VS Code entrypoints now apply the same policy, with process-level coverage: https://github.com/openchamber/openchamber/commit/5a4f9d4e5eae4b2051ddfbcd4697785d804b8c2e.

Closes #3399.

Issue #3060, opened 2026-08-21
[Feature Request] Show per-key OpenRouter usage and limits when available


jeffcottj:
### What should we add/change?

I'm using OpenChamber with OpenRouter in an organizational/workspace setup where each user is assigned their own API key with an individual spending limit.

Currently, OpenChamber's OpenRouter usage display shows the workspace/account-wide credit balance and lifetime spend. This means an individual user sees usage generated by other API keys in the workspace rather than their own usage and remaining allowance.

It would be useful if OpenChamber used OpenRouter's per-key usage data when the configured API key has an individual limit.

OpenRouter exposes this through:

`GET https://openrouter.ai/api/v1/key`

which returns per-key fields including:

* `usage`
* `usage_daily`
* `usage_weekly`
* `usage_monthly`
* `limit`
* `limit_reset`
* `limit_remaining`

For example, if a user's OpenRouter key has a $25 monthly limit and they have spent $7.50, OpenChamber could show something like:

`$17.50 left · $7.50 spent this month`

rather than the total credits/spend for the entire OpenRouter workspace.

Ideally the behavior could be:

1. Query `/api/v1/key` using the OpenRouter credential already stored in `auth.json`.
2. If the key has a configured `limit`, display that key's usage/remaining limit using the appropriate reset period.
3. If the key has no individual limit, fall back to the existing `/api/v1/credits` behavior.

This would make the Usage UI much more useful for teams that provision separate OpenRouter keys and budgets to individual OpenChamber users, while preserving the existing behavior for ordinary personal/unlimited OpenRouter keys.

### Extra context (optional)

This is related to #1085 / PR #1127, which changed the OpenRouter display from a misleading lifetime-credit percentage to the current `$X left · $Y spent` display.

The current OpenRouter quota provider still calls:

`GET https://openrouter.ai/api/v1/credits`

and uses `total_credits` and `total_usage`. That works well for a single-user OpenRouter account, but those values represent shared account/workspace spending in an organizational setup.

OpenRouter now specifically supports per-key recurring limits (`daily`, `weekly`, or `monthly`) and exposes the key's own current usage and `limit_remaining` through `/api/v1/key`.

My use case is centrally deploying OpenChamber to multiple users, each with their own OpenRouter API key and spending limit. In that setup, showing each user's own allowance and usage in OpenChamber would be much more actionable than showing the organization's shared credit balance.

Tested with OpenChamber Desktop 1.19.0 on Windows.

[...]

Issue #1085, opened 2026-04-30
[Bug] OpenRouter Rate Limit tracker is functionally useless in its current state


sprakes:
<img width="323" height="128" alt="Image" src="https://github.com/user-attachments/assets/2b4f7c6a-869e-43ba-a885-15630efd4f86" />

Hi, the openrouter rate limit seems to be comparing total $ vs current $ which means eventually hundreds of total dollars later it will be showing extremely small %. Needs to change to total $ value instead.

### Where does it happen?

VS Code extension

PR #3411, opened 2026-09-07
fix(quota): read OpenRouter usage from /api/v1/key

leducmaxime:
## Intent

OpenRouter quota showed `$0.00 left · $0.00 spent` for keys that clearly had money behind them.

The provider called `GET /api/v1/credits`. OpenRouter documents that endpoint as requiring a management key, and it does return `403 Only management keys can perform this operation` in that case. What it does with a normal inference key is worse than a 403: it answers `HTTP 200` with `{"total_credits":0,"total_usage":0}`. Because the response is a success, the existing `!response.ok` guard could never catch it, and the provider happily formatted two zeros into a money label. That behavior is observed rather than documented, so this PR treats it as an OpenRouter quirk to design around rather than a contract to rely on.

Reproduced against a real funded key, both calls seconds apart:

```
GET /api/v1/key      usage 3.17561396   usage_monthly 3.17561396
                     limit 30           limit_remaining 29.9999982
                     limit_reset "daily"  is_management_key false

GET /api/v1/credits  HTTP 200  {"total_credits": 0, "total_usage": 0}
```

A single billed request moved `usage` from `3.17561216` to `3.17561396` and `limit_remaining` from `30` to `29.9999982`, while `/credits` stayed at zero throughout.

The provider now reads `GET /api/v1/key`, which OpenRouter documents as callable with any valid API key. A key with a spending limit reports its own usage against that limit, in the window named by `limit_reset`. A key without a limit reports `usage_monthly`.

Window usage is `limit - limit_remaining`, not `usage`. This matters and is easy to get wrong: `usage` is all-time, while `limit_remaining` tracks the current reset window. On the key above they differ by six orders of magnitude, so pairing `usage` with `limit` would have replaced a wrong number with a different wrong number. `limit_remaining` is also computed server side and already accounts for `include_byok_in_limit`, which is why the `byok_*` fields are ignored rather than added in.

The same commit brings the provider up to the standard `providers/deepseek.js` already sets, since it was missing all of it: a 15 second timeout, 401 and 403 mapped to a session-expired message, parse failures mapped to `Invalid response from provider`, an explicit `No quota data in response` result, and the `aliases` export that `quota/DOCUMENTATION.md` requires. It adds the missing `openrouter.test.js`, and keeps the VS Code copy in sync.

## Non-goals

Surfacing an account's overall OpenRouter dashboard balance. This came up because the affected account shows roughly $3000 there while the key reports far less, and it is worth being direct about why that number is not in this PR. No documented OpenRouter endpoint exposes an organization credit pool, grant, or promotional balance to a non-management key. That is a conclusion drawn from the documented endpoint permissions rather than an explicit statement that no such endpoint exists, so it could change if OpenRouter documents one. On the account in question `is_free_tier` is `true`, which means the balance is not purchased credit anyway. The honest outcome is accurate per-key usage and limits, not that figure.

Changing how the shared UI renders quota. `packages/ui` is untouched. Reworking the other 15 providers that declare `aliases` without exporting it.

## On the `/credits` fallback from the issue thread

The thread landed on preferring `/api/v1/key` when a key has a limit and keeping `/credits` as the fallback when it does not, so that personal and unlimited keys behave as they do today. I started there, and I want to flag what I hit rather than quietly do something else.

I could not get that fallback to produce a useful number for an unlimited key. A key with no individual limit still gets `HTTP 200` and `{"total_credits":0,"total_usage":0}` from `/credits`, the same response that produces the label this issue is about, so the fallback path lands on `$0.00 left · $0.00 spent` for exactly the keys it was meant to cover. Since a limit is opt-in when a key is created, I would expect that to be the common shape rather than an edge case.

What this PR does instead is report `usage_monthly` for those keys, in a monthly window with no percent. They still show a real money label, which I think preserves the intent behind keeping the fallback even though the mechanism is different.

One case the fallback did serve is a management key, which gets a genuine balance from `/credits`. I have left that out on the reasoning that management keys cannot call completion endpoints, so such a key cannot work as an OpenChamber model provider anyway, and keeping the call would add a second request and a second failure path for it. The provider detects `is_management_key` and returns an explanatory error instead of a misleading `$0.00 spent`. If you would rather keep that case working, gating a `/credits` call on `is_management_key` is a small change and I am glad to add it.

If I have misread the intent here, or if there is `/credits` behavior I did not manage to reproduce, I am happy to rework this part.

## Affected surfaces

`packages/web/server/lib/quota/providers/openrouter.js` is the provider used by web, desktop, and Electron through the quota dispatcher. Its `providerId` stays `openrouter` and the result shape is unchanged, so nothing downstream needs updating.

`packages/vscode/src/quotaProviders.ts` carries its own copy of this parsing logic rather than importing the web provider, the same arrangement `DOCUMENTATION.md` already records for Kimi, Copilot, and Z.ai. `fetchOpenRouterQuota` had the identical bug and gets the identical fix, so the two runtimes do not disagree about the same key. The two implementations are deliberately structured the same way to make that easy to audit.

The user-visible change is the quota label and window. OpenRouter previously produced a single `credits` window with no percent. It now produces one window keyed `daily`, `weekly`, or `monthly` for a limited key, or `credits` for a limit with no reset period, or `monthly` for an unlimited key. Those keys are already localized, so no locale files change.

Window contract detail worth flagging: the provider now emits `usedPercent` and `windowSeconds`, which it never did before, so OpenRouter rows can render a progress bar and participate in headline selection. A `valueLabel` is still always emitted, which preserves the money-label behavior from #1085 and #1127. The percent that #1085 objected to was derived from lifetime total against current balance. This one is bounded by a real reset window, which is a different quantity.

`packages/ui` is unaffected because `formatQuotaValueLabel(valueLabel, percent)` returns `valueLabel` whenever the provider supplies one, and this provider always does.

## Repository guidance

|---|---|---|
| `packages/web/server/lib/quota/DOCUMENTATION.md`, response contract | It states provider modules must export `providerId`, `providerName`, `aliases`, `isConfigured`, and `fetchQuota`. The file exported everything except `aliases`. | `aliases` is now exported. Results are still built with `buildResult` and `toUsageWindow`, so the payload shape is unchanged. |
| `DOCUMENTATION.md`, web and VS Code parity | The file records that the VS Code extension duplicates quota parsing for Kimi, Copilot, and Z.ai and must stay in sync. OpenRouter had the same duplication and the same bug. | Both runtimes changed together, with the same check order, window mapping, labels, and error strings. A new `OpenRouter key semantics` section documents the field semantics and the `/credits` behavior so this is not reintroduced. |
| `.agents/skills/openchamber-change-discipline` | Risk classification. This is a module contract change, since `aliases` becomes an export, and a cross-workspace contract change, since two runtimes carry the logic. | Both runtimes and their tests were validated. The diff stays inside the OpenRouter path, and the 15 other providers that keep `aliases` module-private were left alone. |
| Same skill, structural discipline | It rejects blind casts and asks that the normal path read top to bottom. | The VS Code version was restructured into the same explicit limited and unlimited branches as the web version so TypeScript narrows naturally, removing two non-null assertions and a five-level nested ternary. |
| `providers/deepseek.js` as the local pattern | Nearest hardened provider with the same single-endpoint money-label shape. | Timeout, status mapping, parse-error mapping, and the no-quota-data result are copied from it rather than invented. `deepseek.test.js` sets the test conventions this file follows. |
| `.agents/skills/communication-style` | Governs this PR body and the documentation section. | Documentation states behavior as fact rather than instruction, and avoids bold-label list formatting. |
| `.agents/skills/update-changelog` | It gates changelog edits to explicit maintainer requests. | No changelog entry. `changelog/` is untouched. |

## Validation

| Check | Result |
|---|---|
| `bun run type-check` | Pass |
| `bun run lint` | Pass |
| `bun run test` | 407/408 files. The one failure is `packages/ui/src/lib/persistence.test.ts`, which is pre-existing and unrelated. Details below. |
| `bun run build` | Pass, 3m24s |
| `bun run --cwd packages/web test` | Pass, 2013 tests |
| `bun run --cwd packages/vscode test` | Pass, 35/35 files |
| Live check against the configured key | Provider output matched `GET /api/v1/key` read at the same moment on window key, label, percent, and reset |

On the failing file. It also fails on unmodified `origin/main`, checked out into a separate worktree with no part of this branch applied, where it reported 16 failing tests with `warn: offline` and 5 second timeouts against a runtime settings API. Across four runs it failed, passed, failed, and failed, so it is flaky here and sensitive to network availability in this sandbox. `git diff --stat` shows this branch changes no file under `packages/ui`, so it cannot be the cause. `packages/ui/src/lib/relay/tunnel-client.test.ts` behaved the same way on the first baseline run.

Live verification was a read of `GET /api/v1/key`, a read of `GET /api/v1/credits`, and one trivial billed inference request costing $0.0000018 to confirm which counters move. The key was never logged or committed.

Not verified: behavior against a real management key, and against a real key whose `limit_reset` is `weekly` or `monthly`, since the available key is daily-limited. Those paths are covered by unit tests using documented payload shapes but have not been exercised against live OpenRouter responses. Also not verified: the VS Code extension running inside VS Code, as opposed to its test suite.

## Visual evidence

No screenshot. The browser available in this environment runs on a different machine with no network route to the local OpenChamber server, so the usage panel could not be loaded and captured. Rather than mock something up, here is the same evidence at the layer the UI reads from.

Both captured from `GET /api/quota/openrouter` on a locally running OpenChamber server, same key, same machine, minutes apart, with only the provider file swapped between runs.

Before, on upstream `main`:

```json
{
  "providerId": "openrouter", "ok": true, "configured": true,
  "usage": { "windows": { "credits": {
    "usedPercent": null, "remainingPercent": null,
    "windowSeconds": null, "resetAt": null,
    "valueLabel": "$0.00 left · $0.00 spent"
  } } }
}
```

After, on this branch:

```json
{
  "providerId": "openrouter", "ok": true, "configured": true,
  "usage": { "windows": { "daily": {
    "usedPercent": 0.000005999999999062312, "remainingPercent": 99.999994,
    "windowSeconds": 86400, "resetAt": 1788825600000,
    "resetAtFormatted": "Tue, Sep 8, 2:00 AM",
    "valueLabel": "$0.00 / $30.00"
  } } }
}
```

In the panel that is the difference between an OpenRouter row reading `$0.00 left · $0.00 spent` with no bar, and a `Daily` row reading `$0.00 / $30.00` with a bar and a reset time. The label text is `valueLabel` verbatim, because `formatQuotaValueLabel` returns it whenever it is present. This particular key has spent almost nothing inside its daily window, which is why the used figure rounds to `$0.00`. Its all-time `usage` is `$3.18`, and deliberately not what the daily row reports. A screenshot on request from anyone able to run the UI against a funded key.

## Risks and failure behavior

The main behavioral risk is the one described above: a user with a management key configured loses a working balance readout and gets an explanatory error instead. That is intentional, and the population should be effectively zero because such a key cannot run inference.

Every failure path returns `ok: false` with `configured: true` and `usage: null`, so the row degrades to an error rather than a misleading number. That covers 401 and 403, other HTTP errors, timeouts after 15 seconds, malformed JSON, a missing or non-object `data`, a limit with no `limit_remaining`, and an unlimited key with no `usage_monthly`. There is no caching and no retry, so a failed refresh affects only that refresh.

Unlimited keys are the case most likely to be got wrong, since `limit`, `limit_remaining`, and `limit_reset` are all documented as present and null there. Null means unlimited, never missing data, and the tests pin that a fully null trio still produces a valid monthly window rather than an error.

Zero is treated as a real value throughout. A key that has spent nothing shows `$0.00 / $30.00` rather than falling into the no-data path. A `limit` of `0` yields a null percent instead of dividing by zero. A negative `limit_remaining`, which would mean overspend, clamps the percent at 100 while leaving the money label unclamped so the overage stays visible.

`resetAt` is derived rather than returned by the API, since `limit_reset` is a period string. The derivation uses `Date.UTC` only, which normalizes overflow, and was checked across all seven weekdays plus December 31, January 31, and February 29. It is a display value, not a billing figure.

No credentials are logged or persisted. Provider IDs, the result contract, and the shared UI are unchanged, so rollback is reverting this commit with nothing to migrate.

One housekeeping note: exporting `aliases` makes `bun run dead-code` list `aliases openrouter` as an unused namespace export. It already reports the same for claude, exe-dev, google, hyper, and opencode-go, so this adds a line to an existing pattern rather than a new one. Say the word and I will make it module-private instead, though that would put the file back out of step with the documented export contract.

### DIFF
--- a/packages/vscode/src/quotaProviders.test.ts
+++ b/packages/vscode/src/quotaProviders.test.ts
@@ -17,6 +17,7 @@ const AUTH = JSON.stringify({
   crof: { key: 'test-token' },
   neuralwatt: { key: 'test-token' },
   'opencode-go': { key: 'test-token' },
+  openrouter: { key: 'test-token' },
   'zai-coding-plan': { key: 'test-token' },
   deepseek: { key: 'test-token' },
   hyper: { key: 'test-token' },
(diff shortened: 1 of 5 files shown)


btriapitsyn:
Thanks for the live investigation and for keeping the web and VS Code implementations aligned. The per-key calculation and monthly usage fallback make sense. Merged.

Closes #3060.

PR #3418, opened 2026-09-07
fix: forward agent to session creation

Factor-scd:
## Intent

`createSession()` only sends `{ directory, title }` to the OpenCode server, dropping the `agent` field. Sessions are always created with the OpenCode default agent (`build`) regardless of what was requested. The agent is only applied later at prompt time, which is too late — the session already exists on the wrong agent.

This matters for any caller that spawns sessions with a specific agent (e.g. `--agent manager` for long-running orchestration tasks). Without this fix, the session starts on `build` and the agent flip happens silently after the verification window.

## Non-goals

Model and variant forwarding. The OpenCode session-create endpoint (`POST /session`) uses a different wire shape for model than `prompt_async` — `model: { id, providerID, variant? }` vs `model: { providerID, modelID }`. Forwarding those fields with the current shapes would either fail with 400 or silently no-op. That needs separate investigation with proper schema alignment.

## Affected surfaces

- `packages/web` — server route only (`openchamber-sessions/routes.js`)
- No UI, desktop, VS Code, or mobile impact
- No persisted data changes
- Request/response shape unchanged (agent was already accepted in the create payload, just ignored)

## Repository guidance

|---|---|---|
| `openchamber-change-discipline` | Source change in server route module | Smallest complete change: one field added to function signature, body, and call site. Preserves existing behavior when agent is absent (spread is conditional). |
| `ui-api-decoupling` | Server API route modifying OpenCode SDK interaction | Uses existing `opencodeClient` SDK path. No new endpoints, no hardcoded URLs, no credential changes. |
| `communication-style` | PR body and comments | Applied checklist: no puffery, no AI vocabulary, direct voice. |

## Validation

| Check | Result |
|---|---|
| `bun run --cwd packages/web test server/lib/openchamber-sessions/routes.test.js` | 21/21 passed |
| `bun run --cwd packages/web test` (full suite) | 1563/1566 passed (1 pre-existing flaky timeout in `walkthrough/routes.test.js`, unrelated) |

What was not verified: runtime behavior against a live OpenCode server. The test suite mocks the SDK client, so the actual HTTP body sent to OpenCode is not exercised by tests. The fix is a straightforward field pass-through with no logic changes.

## Visual evidence

No visible change. This is a server-only route modification — the session create endpoint now includes `agent` in the POST body where it previously did not. The UI and all runtime surfaces are unaffected.

## Risks and failure behavior

- **Risk: agent string reaches OpenCode unsanitized.** The `agent` value passes through `asNonEmptyString()` (already applied at the call site in `create()`) and is validated by `validateRequestedSelection()` before session creation when a prompt is present. When no prompt is present, validation is skipped — this is existing behavior, not introduced by this change.
- **Rollback:** Revert the commit. Sessions revert to always using the default agent, which is the current broken behavior.
- **Failure mode:** If `agent` is an unknown value, the OpenCode server will reject the session create request (400). This is correct behavior — fail loud instead of silently using the wrong agent.

### DIFF
--- a/packages/web/server/lib/openchamber-sessions/routes.js
+++ b/packages/web/server/lib/openchamber-sessions/routes.js
@@ -193,7 +193,7 @@ const runPromptAsync = async ({ baseUrl, authHeaders, sessionID, directory, payl
   }
 };
 
-const createSession = async ({ baseUrl, authHeaders, directory, title }) => {
+const createSession = async ({ baseUrl, authHeaders, directory, title, agent }) => {
   const sessionUrl = new URL(`${baseUrl}/session`);
   sessionUrl.searchParams.set('directory', directory);
   const response = await fetch(sessionUrl.toString(), {
@@ -204,7 +204,11 @@ const createSession = async ({ baseUrl, authHeaders, directory, title }) => {
       'content-type': 'application/json',
       accept: 'application/json',
     },
-    body: JSON.stringify({ directory, ...(title ? { title } : {}) }),
+    body: JSON.stringify({
+      directory,
+      ...(title ? { title } : {}),
+      ...(agent ? { agent } : {}),
+    }),
   });
 
   if (!response.ok) {
@@ -630,6 +634,7 @@ export const createOpenChamberSessionService = (dependencies) => {
       authHeaders,
       directory: sessionDirectory,
       ...(title ? { title } : {}),
+      ...(agent ? { agent } : {}),
     });
 
     let dispatch = { model, agent, variant, promptDispatched: false, dispatchedAsCommand: false };


btriapitsyn:
Hey @Factor-scd  
Why does it matter during session creation if it always specified with the first prompt?

Factor-scd:
Hey @btriapitsyn — fair question. The agent isn't always tied to the first prompt: `session create` is a standalone call, and when no prompt is sent, the session gets created on the default agent (`build`). The requested agent only kicks in later when a prompt is dispatched.

A few real cases where that matters:
- **External agent orchestration** — an outside agent creates the session first, verifies everything (directory, agent, model, worktree) before sending the prompt. The prompt is what starts token burn, so the session needs to be right before that point.
- **Pre-provisioning** — scripts or CI that spin up sessions ahead of time (e.g., one per agent) before any prompt is sent.
- **Status & monitoring** — `session list`/`status` report the session's agent, so a session requested as `manager` would show up as `build` until a prompt arrives.

This just makes the create call send what the CLI already passes along.

Issue #2039, opened 2026-07-05
[Bug] Cannot enable auto-accept permissions before starting a new conversation


severin-ye:
## Problem

When starting a new conversation, users want to enable "Auto-accept permissions" before sending the first message, so the agent can run without requiring manual confirmation on every file operation.

Currently, the auto-accept toggle is only available after the conversation has already started running (i.e. the agent is actively executing tool calls). This means the first few operations of every new conversation require manual approval, breaking the flow.

## Expected behavior

- The "Auto-accept permissions" toggle should be available in a new conversation window before any message is sent
- Alternatively, a global or project-level "default auto-accept" setting would also solve this

## Steps to reproduce

1. Create a new conversation in OpenChamber
2. Before sending any message, try to click the "Auto-accept permissions" button
3. Observe: the button is not clickable or not present
4. Send a message and wait for the agent to start executing
5. Now the toggle becomes available — but too late

bashrusakh:
related 
https://github.com/openchamber/openchamber/issues/1903
https://github.com/openchamber/openchamber/issues/1466

PR #3421, opened 2026-09-08
fix(desktop): keep Electron UI alive with external runtimes

alvins82:
## Intent

When Electron starts with an external local runtime such as the OMP sidecar as the default host, the renderer could navigate to the runtime's API URL instead of the local OpenChamber UI. On macOS, an `activate` event could race initial bootstrap and create a second window before runtime state was ready. The result was a blank screen or a failed session creation.

## What changed

- Serialize initial Electron window startup so activation cannot race `resolveInitialUrl()`.
- Create the existing startup splash before runtime probes and reuse that window during activation or deep-link races.
- Keep the local HMR or packaged UI as the renderer while routing shared runtime calls to the selected external API base URL.
- Wait for the Electron init script to publish the resolved runtime before the renderer imports the shared app. The wait is Electron-only and has a 10-second fail-open timeout.
- Treat the trusted development HMR origin as a local page for preload and runtime credential exposure.
- Rebind the shared SDK after runtime configuration so it cannot retain the UI origin when the runtime is external.
- Apply the same local-UI and external-runtime split to host switching, additional windows, and mini-chat.
- Wait for the same desktop bootstrap event before Mini Chat creates its runtime APIs or imports its renderer, so it cannot initialize against stale globals during HMR startup.
- Update sync test doubles for the child-store and global-session APIs introduced by the merged upstream changes.

## Non-goals

- No OpenCode or OMP server changes.
- No new dependencies or runtime API contract changes.
- No redesign of host selection or health-check behavior.
- Background or login startup remains windowless.

## Affected surfaces

- Electron main-process startup, window lifecycle, and window URL selection.
- Electron preload trust boundary for the development HMR UI.
- Shared web runtime API bootstrap and SDK initialization.
- User-visible startup behavior: the existing splash appears immediately and remains until runtime selection completes.

## Repository guidance

|---|---|---|
| `AGENTS.md` | Always-on rules govern runtime boundaries, security, dependency scope, and validation. | Kept domain logic in the owning Electron and web modules. Added no dependencies or secrets. |
| `.agents/skills/openchamber-change-discipline/SKILL.md` | The change modifies executable Electron startup and shared runtime behavior. | Kept the fix focused, preserved existing abstractions, and validated affected packages. |
| `.agents/skills/desktop-shell/SKILL.md` | The PR changes Electron startup, window lifecycle, preload trust, and bundled/HMR routing. | Restored the splash before probes, kept background startup windowless, and ran isolated HMR startup plus bundled asset validation. |
| `.agents/skills/ui-api-decoupling/SKILL.md` | External runtime URLs, SDK binding, preload trust, and shared API bootstrap are in scope. | Kept the local UI separate from the selected runtime API and rebound the SDK after configuration. |
| `packages/electron/README.md` | It documents Electron ownership and requires HMR and bundled validation for startup changes. | Ran Electron syntax/architecture checks and the bundled asset build. |
| No applicable `packages/electron/DOCUMENTATION.md` | No nearer Electron module documentation exists. | Used the package README and owning skills. |

## Validation

| Check | Result |
|---|---|
| `bun run --cwd packages/web test -- src/runtimeConfig.test.ts` | Passed at `8b9b284d4`: 8 tests, including the no-event timeout fallback. |
| `bun run --cwd packages/web type-check` | Passed. |
| `bun test src/sync/__tests__/issue-1637-2270.test.ts` and `bun test src/sync/__tests__/issue-2039.test.ts` | Passed: 19 tests, 68 expect calls. These fixtures now match the merged sync-store contracts. |
| `bun test src/lib/relay/tunnel-client.test.ts` | Passed: 21 tests, 0 failures. A full UI-suite run had one timing-sensitive failure in this file; the isolated rerun passed. |
| `bun run --cwd packages/electron type-check` | Passed: Electron main and preload syntax checks. |
| `bun run --cwd packages/electron test:architecture` | Passed: 39 tests, 0 failures. |
| `bun run --cwd packages/electron bundle:main` | Passed at `8b9b284d4`. |
| `bun run --cwd packages/electron build:web-assets` | Passed at the current PR HEAD: Vite transformed 3,065 modules and staged `packages/electron/resources/web-dist`. Existing font, chunk-size, and dynamic-import warnings remain. |
| `./node_modules/.bin/oxlint packages/web/src/mini-chat-main.tsx` | Passed with no diagnostics. |
| Isolated Electron HMR startup with a remote default host | Started at the current PR HEAD with a separate temporary profile, HMR UI/API ports, and `http://127.0.0.1:4096` as the default host. The Electron renderer process launched, but the existing Electron singleton remained the frontmost automation target, so no UI-level `createSession()` result is claimed. |
| `bun run dead-code` | Completed non-blocking. It reports the repository backlog of unused files, exports, and types plus one duplicate export; no new finding was isolated to this change. |

## Visual evidence

No new persistent visual surface is introduced. The change preserves the existing launch splash, which now appears before runtime probes and is replaced by the local HMR or packaged UI in the same window after bootstrap. The current-HEAD isolated HMR run launched a separate Electron renderer, but the existing Electron singleton remained the frontmost automation target, so no screenshot or UI-level session-creation result is claimed.

## Risks and failure behavior

- Background and login startup remain windowless.
- Normal startup creates one splash window before probes. Activation and deep-link races reuse it instead of creating another main window.
- If an external runtime is unavailable, the local OpenChamber UI remains available and the existing host health/unreachable fallback applies.
- If the init script does not publish readiness, the renderer waits at most 10 seconds before proceeding.
- Runtime credentials remain restricted to trusted local, HMR, packaged, or explicitly authorized relay surfaces. They are not added to remote page URLs.
- Manual HMR UI-level session creation could not be independently observed because the existing Electron singleton stayed frontmost; the isolated renderer launch, package checks, and bundled asset staging passed.
- No persisted-data migration, dependency change, or cross-runtime API contract change is introduced.


alvins82:
Pushed follow-up commit bba8cec0e to address this startup regression.

The PR had moved the first main-window creation behind resolveInitialUrl(). When an external default host is unreachable, the 2s probe plus 10s retry ran before any BrowserWindow existed, so the app could appear not to launch. The follow-up creates the existing startup splash before those probes, guards against activation/deep-link races, and keeps background startup windowless.

Validated locally:
- bun run type-check:electron
- bun run --cwd packages/electron test:architecture (39 passing)
- bun run electron:dev:bundled (web build and asset staging)
- node syntax checks for main.mjs and preload.mjs

Only packages/electron/main.mjs is included in the follow-up commit.

alvins82:
Updated the PR body to include the required Repository guidance section and current validation for the follow-up head.

Also pushed 35b994303 to clarify the runtime SDK rebinding comment flagged in the previous review; this is wording-only and the focused runtime-config tests still pass. The bundled validation entry now explicitly records that web build/asset staging passed, while an isolated live bundled launch was not completed because an existing Electron singleton held the app lock.

alvins82:
Addressed in `8b9b284d4` and `0630c22b4`:

- Mini Chat now waits for `waitForDesktopRuntimeBootstrap()` before creating runtime APIs or importing its renderer, so dev HMR uses the resolved external host instead of stale module-scope globals.
- Fixed the existing `timeoutId` lint error in `runtimeConfig.ts`.
- Updated the two UI sync test doubles to match the child-store and global-session APIs introduced by the merged upstream changes; the previously failing issue #1637/#2270 and #2039 files now pass (19 tests).
- Re-ran the merged-head web tests/type-check/lint, Electron type-check and architecture tests, main bundling, packaged asset staging, dead-code scan, and diff check. The isolated relay rerun passed 21 tests after one timing-sensitive failure during the full local UI sweep.

The PR Validation table is updated. The isolated Electron HMR process and renderer launched with the remote default host, but the already-running Electron singleton stayed the frontmost desktop automation target, so I did not claim a UI-level `createSession()` result. Fresh GitHub validation is pending for `0630c22b4`.

Issue #3259, opened 2026-08-31
tracking: productionize OpenCode V2 Beta alongside stable V1


bashrusakh:
> **Policy change (2026-09-03).** The target architecture changes from capability-by-capability V1→V2 migration inside one mixed runtime to two explicit, mutually exclusive modes: Stable/V1 (`opencode`) and opt-in Beta/V2 (`opencode2`). Runtime selection becomes an instance-level setting resolved at bootstrap. V1 remains a first-class production runtime. V2 Beta does not fall back to V1 authority. Prior work stays reusable evidence and is re-evaluated against runtime ownership.

> **Refresh (2026-09-08).** #3007 was re-audited from its current source diff at a fresh HEAD (`8840378c4`). The prior audited SHA `5a424cf7` is **no longer in #3007's history** (the branch was rewritten). #3424 is re-scoped as the **selector unit**, not a second V2-compatibility implementation. See `#3007 audit refresh` and `Current program status` below.

---

## Body layout

```text
CURRENT PLAN        → the selected dual-runtime proposal, selector unit, and next actions
HISTORICAL /        → superseded hybrid strategy plus old audit notes (kept as evidence,
SUPERSEDED STRATEGY   relabelled by SHA and marked not-current)
ENGINEERING EVIDENCE → source-backed audits (authority, permissions, database, state/cache)
```

---

## Goal

OpenChamber supports two explicit, mutually exclusive OpenCode runtime modes:

```text
Stable
→ OpenCode V1 runtime
→ `opencode`

Beta
→ OpenCode V2 runtime
→ separately shipped `opencode2`
```

The previous goal, gradually replacing individual V1 capabilities with V2 calls inside one mixed authoritative runtime, is superseded. The new goal is:

```text
preserve a production-stable V1 mode
+
add an explicit opt-in V2 Beta mode
+
keep runtime ownership coherent inside each running OpenChamber instance
```

OpenCode remains an external stock runtime.

```text
OpenCode modifications: NONE
```

---

## Decision status for this issue

This body records the **#3259 proposal** selected by the issue author on 2026-09-03.

It does **not** claim project-wide maintainer approval.

Use this wording consistently:

```text
PROPOSED
SELECTED FOR #3259
```

Do not read that as:

```text
approved by the maintainer
approved for the project
required for another contributor's PR
```

The maintainer may accept this proposal, change it, or reject it.

---

## Architecture proposal selected for #3259

[...]

PR #3424, opened 2026-09-08
feat(runtime): add Stable/Beta OpenCode runtime selector

bashrusakh:
## Intent

An instance-level Stable/Beta selector for the OpenCode runtime. The `opencodeRuntime: 'stable' | 'beta'` setting persists through the existing settings architecture, the OpenCode CLI settings section offers a two-option control, and the selection applies at the next OpenCode restart. There is no hot switching.

## Architecture

- Stable is the default and runs the existing V1 path unchanged.
- Beta is an opt-in intent. It is persisted now and resolved at the bootstrap boundary.
- Current `main` has no V2 compatibility infrastructure (that work lives in #3007, which we treat as read-only prior art). So a Beta selection on this branch fails explicitly at startup with `OpenCode Beta runtime is not available in this build yet. Switch the runtime back to Stable.` It never silently falls back to V1, and this PR does not carry a second V2 implementation.
- The resolution snapshot keeps configured intent and active protocol separate: Stable reports `resolvedProtocol: 'legacy'`; Beta reports `null` (no V2 runtime resolved, and legacy is deliberately not started).

## Dependency

V2 compatibility infrastructure (opencode2 lifecycle, `service start`, `Service.discover()`, V2 API adapter, V2 SSE/event normalization, shared-service auth) is external to this PR. #3007 is the expected compatibility layer and was not modified. All V2 infrastructure previously carried on this branch was removed.

## Scope

- Settings registry: `opencodeRuntime` as an instance-scoped field, server and VS Code sanitizers with a Stable default for old or invalid documents.
- Resolution snapshot: `opencodeRuntime` intent plus a truthful `resolvedProtocol` (`legacy` for Stable, `null` for Beta).
- Settings UI: Stable/Beta radio options in the existing OpenCode CLI section, Stable selected by default, Beta marked experimental.
- Selection persists immediately through the existing save mechanism and records the existing deferred-restart marker. Changing it marks OpenCode restart required; the current process keeps running unchanged.
- The binary-path save stays independent: it persists only `opencodeBinary` and records the `opencode-binary` restart marker, exactly as on main.
- Startup gate (web server and VS Code): Beta selection with no V2 infrastructure present produces an explicit unavailable state, never a V1 fallback.

## Non-goals

- V2 lifecycle, V2 API adapter, SSE/event normalization, question/permission/agent V2 mapping, V2 execution parity, Queue/Steer. Those belong to #3007 or later #3259 units.
- No new dependencies; the `@opencode-ai/client` beta closure is gone.

## Affected surfaces

- Shared UI: settings registry, settings search, persistence round-trip, `OpenCodeCliSettings` selector UI, 12 locale files.
- Web server: settings sanitizers and response default, resolution snapshot, lifecycle Beta gate wired through `index.js`.
- VS Code: bridge settings read/persist sanitization, extension-host Beta gate reading the shared settings file.
- Persisted contract: settings document gains one additive field; old files read as `stable`.

## Repository guidance

|---|---|---|
| AGENTS.md correctness invariants | No silent fallback, no authority fiction | Beta without V2 infrastructure is an explicit unavailable state; `resolvedProtocol` is `null` for Beta and `legacy` only when the legacy runtime actually resolves |
| openchamber-change-discipline | Cross-workspace settings contract | Setting added through the registry; generated JSON snapshots regenerated and snapshot-checked |
| settings-ui-patterns + locale-ui-patterns | Settings control and copy | Two-option control on the section's existing primitives; full locale translations; no raw protocol strings in the UI |

## Validation

| Check | Result |
|---|---|
| `bun test packages/ui/src/components/sections/openchamber/OpenCodeCliSettings.test.tsx` | 8 pass / 0 fail (default Stable, no write on mount, Beta and Stable persist + restart marker, no marker on redundant click, no persist while loading, binary save persists only the binary with the binary marker, no marker on failed save, no marker on redundant write) |
| `bun test packages/ui/src/lib/settings/registry.test.ts` | 9 pass (snapshot freshness for both generated JSONs) |
| `bun test packages/ui/src/lib/persistence.test.ts` | 52 pass / 0 fail (incl. redundant write resolves written:false) |
| `bun test packages/ui/src/lib/i18n/messages.test.ts` | 5 pass / 0 fail (incl. Beta-copy regression guard) |
| `bunx vitest run packages/web/server/lib/opencode/settings-helpers.test.js` | 57 pass (extended timeout needed for the pre-existing tarball-pack case) |
| `bunx vitest run packages/web/server/lib/opencode/settings-runtime.test.js` | 18 pass / 1 skipped |
| `bunx vitest run packages/web/server/lib/opencode/lifecycle.test.js` | 30 pass / 0 fail, includes Beta gate explicit-failure and Stable legacy-path tests |
| `bunx vitest run packages/web/server/lib/opencode/opencode-resolution-runtime.test.js` | 4 pass (stable/missing/invalid → `legacy`; beta → `null`) |
| `node scripts/run-isolated-tests.mjs packages/vscode/src` | 36/36 test files pass, includes new runtime selector tests |
| `bun run type-check` | 0 errors, all workspaces |
| `bunx oxlint` on changed paths | 0 findings on authored lines; reported pre-existing backlog only |

Not runtime-verified: activating Beta against a live V2 daemon (this branch has no V2 infrastructure by design; that is gated on #3007).

## Visual evidence

The selector renders in the existing OpenCode CLI settings section using the section's existing control primitives. No screenshots were captured in this environment; the interaction contract (default selection, immediate persistence, restart marker, no-op on redundant click, disabled while loading, independent binary save) is covered by the component tests above.

## Risks and failure behavior

- Beta selected before #3007 lands: explicit startup error `OpenCode Beta runtime is not available in this build yet. Switch the runtime back to Stable.`; OpenChamber never starts V1 in its place. Recovery: switch back to Stable and restart.
- Stable path: unchanged vs current main; the gate reads the setting once per start attempt and defaults to false.
- Invalid or foreign stored values never become Beta; old documents read as Stable.
- The fire-and-forget selection save surfaces failures through the shared settings save indicator.
- Rollback: revert the branch; Stable defaults are unchanged.

## Tracking

- #3259 (architecture tracker).
- #3007: external, read-only prior art; this PR no longer duplicates it. Beta activation remains dependent on it.


bashrusakh:
Closing this because the repository's V2 direction has changed underneath this PR.

This PR was intentionally reduced to a Stable/Beta runtime selector, with simultaneous V1/V2 compatibility expected to come from #3007. The official repository now has an actively maintained `opencode-v2-refactoring` branch that explicitly implements a direct V1 → V2 one-cutover, with its own V2 preview tooling and migration work.

That is a materially different architecture and ownership model. Continuing this selector in parallel would duplicate or conflict with maintainer-owned V2 work and would likely require another rewrite once that branch lands.

I'm therefore withdrawing this PR rather than continuing a parallel V2 implementation.

This closure is about project direction and coordination, not a newly discovered correctness problem in the implementation here.

PR #3425, opened 2026-09-08
feat(browser): add the server browser workspace

JosueGalRe:
## Intent

Let an agent use Chrome on the OpenChamber server without keeping a desktop client open, and let a connected user watch and take control of that same session from the web or mobile workspace.

This is a draft for architecture feedback and hands-on testing. It adds a separate **Server browser** workspace alongside the existing local **Browser**. The feature is off by default. The complete implementation is included so reviewers can run it, with commits separating the server, shared UI and owning documentation. I am happy to split the work into smaller mergeable pieces after agreement on the behavior and boundaries.

The server owns isolated session contexts, tabs and control leases. The viewer includes real keyboard and pointer input, clipboard actions, automatic or fixed viewport sizing, a Console/Network inspector and Chrome DevTools. Settings show the active CDP endpoint and allow an automatic or fixed port for attaching an external MCP to the same managed Chrome.

To try it, run the web server from this branch, enable Server browser in Settings > General > OpenChamber Tools, then open Server browser in the workspace. The server requires Chrome or Chromium 109+. If automatic discovery fails, set `OPENCHAMBER_CHROME_PATH` to the executable.

## Non-goals

- Replacing the local Electron browser or changing the default browser backend for all users.
- Installing browser extensions, adding remote annotation, or porting the clear-cookies, clear-cache and open-external buttons.
- Making external CDP/MCP clients participate in OpenChamber control leases. A direct loopback CDP connection remains a privileged connection to the managed Chrome instance.
- Shipping a new dependency, installer or release entry.

## Affected surfaces

- **Web/server:** Chrome process ownership, egress policy, session and tab identity, user takeover, authenticated browser streaming, DevTools asset/protocol transport, settings and shutdown cleanup.
- **Shared UI:** the separate workspace, session/page selection, viewport controls, clipboard and inspectors. Selection is scoped to the server and directory, and is checked against authoritative live state when reopening.
- **Hosted mobile and Capacitor:** the workspace uses the server browser and detaches when hidden, disconnected or covered by an app overlay. It does not launch Chrome on the mobile device.
- **Electron:** the shared remote viewer uses the server backend; the existing local browser remains available.
- **VS Code:** the new browser workspace remains unavailable and the runtime-status route returns an explicit unsupported response. Its generated settings registry stays aligned with the shared contract.
- **Persisted contract:** browser enablement and CDP-port choice belong to the server instance. Saving a different port affects the next Chrome launch and does not relocate or close an already-running browser.

## Repository guidance

|---|---|---|
| Root AGENTS and change discipline | Shared contracts span web, UI and runtime consumers | Domain logic is owned by browser modules; entrypoints wire those modules. No dependency or changelog changes are proposed. |
| UI API decoupling and browser/auth/runtime references | Clients can connect directly or through the runtime transport | Shared clients use the authenticated runtime transport, scoped attachments and explicit unsupported behavior. |
| Sync invariants and store/surface documentation | Session selection, ownership and reconnect can race | Live session and attachment identities are authoritative; stale operations are rejected and hidden viewers detach. |
| Relay transport and performance engineering | Screencast, inspector and DevTools traffic can burst | Queues and bytes are bounded, acknowledgements control progress, and stalled work closes with explicit errors and cleanup. |
| Theme, locale and Settings guidance | New controls and Settings states are user-visible | Shared primitives and tokens are reused; strings are registered in the locale catalogs and settings search. |
| Owning browser, browser-control, agent-tool, stores, surfaces, OpenCode and mobile documentation | Module contracts and runtime behavior changed | The owning documents explain lifecycle, isolation, failure behavior and platform limits. |

## Validation

Current HEAD: `a573af57548b59c2dd8f2be79b70a37c97d44d4b`. The focused checks below were rerun at this commit in an isolated Linux checkout on September 12, 2026. Bun runtime tests used Bun 1.4.0. Vitest runs the server suites under Node.

| Check | Result |
|---|---|
| `bun run --cwd packages/web test server/lib/browser server/lib/ui-auth/ui-auth.test.js` | 27 files passed: 592 tests passed, 1 skipped. Includes session-manager idle sweep, surface-gateway, ui-auth, DevTools policy, lifecycle, backpressure, and the real WebSocket transport test. |
| `bun test ./packages/web/server/lib/browser/devtools-runtime.test.js` | Passed under Bun: 1 test, 3 assertions. Two messages fill the delivery window, an ACK resumes the third in order, and peer close produces the expected closure. This exercises Bun itself, not only the Node branch. |
| `bun test ./packages/ui/src/components/browser/RemoteBrowserCanvas.test.tsx` | 16 passed. Ctrl and Command shortcuts stay inside the remote input boundary; host shortcuts outside it remain active. |
| `bun test ./packages/ui/src/components/browser/RemoteBrowserContextMenu.test.tsx` | 16 passed. Open-menu Ctrl and Command shortcuts do not reach host shortcut handlers; navigation, clipboard and dismissal coverage also passed. |
| `bun test ./packages/ui/src/lib/browser/remoteSurface.test.ts` | 26 passed, including failed URL-token requests opening no tokenless socket and authenticated reconnect behavior. |
| Real Bun server and Chrome DevTools interaction | Before the transport correction, DevTools failed to open even for a blank page. With the final source, DevTools opened, Console evaluation of `document.title` returned the fixture title, Network showed successful document/script reloads, and closing/reopening DevTools succeeded. The user subsequently installed the corrected TGZ on their remote server and confirmed that DevTools works. This is functional confirmation, not a performance measurement. |
| Final-source static checks before commit | Web type-check and lint, Node syntax checks, and oxlint on the changed transport files passed. Dead-code completed with existing repository findings and no changed-path finding. The extra no-excuse helper could not run because it requires TypeScript 7 unstable APIs while this checkout uses TypeScript 5.9.3. |
| Package verification | The web build completed before the server-only transport correction; those UI assets were reused. The final TGZ was audited to match the committed DevTools policy/transport source and compiled UI. Packaged CLI `--version` and `--help` passed under Node and Bun. No dependencies were added. |

The final-source runtime, static and packaging checks were performed before creating the two commits, with no subsequent source changes. The automated checks in the first five rows were then rerun at the exact HEAD above. The installed TGZ SHA-256 is `6821828c4c99bfcd7861131dbfc77af7033dafda1cc5c05f5e45598a133bed8c`.

The earlier full-workspace runs remain historical evidence at `ba1ed11`, not a claim that every workspace suite was rerun at the current HEAD. They reported an upstream UI event-pipeline timing failure and an upstream CLI child-exit assertion failure; those unrelated failures were not changed. Native Electron interaction/installers, the VS Code extension host, physical iOS/Android devices, Windows/Zen, and a deployed private relay were not manually validated by these Linux checks.

## Visual evidence

The attachments retain their original capture revisions. The two open-menu screenshots, final stress page and closed-menu references were captured at `ba1ed11`; the other state screenshots and the 18-second recording were captured at `84a24f3`.

The static captures remain representative of layout, settings and theme appearance. The subsequent fixes change keyboard event propagation, idle cleanup, authentication failure handling and DevTools protocol transport; they do not change the CSS, layout or theme shown in those captures. The old recording is evidence of basic editing and DevTools interaction only, not proof of the later Ctrl+P fix or Bun compatibility. Current keyboard regression results and the fresh Bun interaction check are reported above. No new keyboard recording was captured for this handoff refresh.

Upstream has no equivalent Server browser workspace. The disabled/enabled Settings pair shows the opt-in change; other pairs compare interaction and theme states. Phone layouts use touch-enabled browser emulation, not physical devices. The recording does not measure frame rate.

**Keyboard editing and Chrome DevTools interaction, 18 seconds. Captured at 84a24f3.**

https://github.com/user-attachments/assets/2108d1fc-a5e8-4c55-bcc3-228bc2700cd5

<details>
<summary>Opt-in and session setup</summary>

**Server browser disabled in Settings.**

Server browser disabled in Settings.

**Server browser enabled in Settings.**

Server browser enabled in Settings.

**Authenticated OpenChamber workspace.**

Authenticated OpenChamber workspace.

**Server browser session in the workspace.**

Server browser session in the workspace.

**Local test page opened through localhost.**

Local test page opened through localhost.

**Local test page opened through the loopback address.**

Local test page opened through the loopback address.

</details>

<details>
<summary>Keyboard editing and context menus</summary>

The Light and Dark fallback menus were captured at `ba1ed11` with the opaque background. The later shortcut propagation fix is covered by the current canvas and context-menu regression tests above.

**Keyboard editing in the remote page.**

Keyboard editing in the remote page.

**The page's own context menu.**

The page's own context menu.

**Fallback context menu in Light with an opaque background. Captured at ba1ed11.**

Fallback context menu in Light with an opaque background. Captured at ba1ed11.

**Fallback context menu in Dark with an opaque background. Captured at ba1ed11.**

Fallback context menu in Dark with an opaque background. Captured at ba1ed11.

**Fallback menu dismissed with Escape.**

Fallback menu dismissed with Escape.

</details>

<details>
<summary>DevTools and script loading</summary>

**Chrome DevTools Console reads the page title.**

Chrome DevTools Console reads the page title.

**The native Chrome DevTools context menu.**

The native Chrome DevTools context menu.

**The 1,000-script test page after loading.**

The 1,000-script test page after loading.

**The 1,000-script page with DevTools connected.**

The 1,000-script page with DevTools connected.

**Fresh 1,000-script page from the final run, also checked through external Chrome DevTools MCP. Captured at ba1ed11.**

Fresh 1,000-script page from the final run, also checked through external Chrome DevTools MCP. Captured at ba1ed11.

</details>

<details>
<summary>CDP port settings and browser lifecycle</summary>

**Automatic CDP port with the active endpoint visible.**

Automatic CDP port with the active endpoint visible.

**Server browser stopped in Settings.**

Server browser stopped in Settings.

**Server browser enabled again, before Chrome launches.**

Server browser enabled again, before Chrome launches.

**Server browser active again after launch.**

Server browser active again after launch.

**Validation for an invalid fixed CDP port.**

Validation for an invalid fixed CDP port.

**A saved fixed port waits for the next Chrome launch.**

A saved fixed port waits for the next Chrome launch.

**Chrome running on the configured fixed CDP port.**

Chrome running on the configured fixed CDP port.

**Launch error when the configured fixed CDP port is occupied.**

Launch error when the configured fixed CDP port is occupied.

</details>

<details>
<summary>Light and Dark desktop and phone layouts</summary>

Desktop captures are 1440 × 900. Phone captures are 390 × 844 with browser touch emulation.

**Server browser Settings on desktop in Light.**

Server browser Settings on desktop in Light.

**Server browser Settings on desktop in Dark.**

Server browser Settings on desktop in Dark.

**Server browser workspace on desktop in Light.**

Server browser workspace on desktop in Light.

**Server browser workspace on desktop in Dark.**

Server browser workspace on desktop in Dark.

**Server browser Settings in Light at the emulated phone size.**

Server browser Settings in Light at the emulated phone size.

**Server browser Settings in Dark at the emulated phone size.**

Server browser Settings in Dark at the emulated phone size.

**Server browser workspace in Light at the emulated phone size.**

Server browser workspace in Light at the emulated phone size.

**Server browser workspace in Dark at the emulated phone size.**

Server browser workspace in Dark at the emulated phone size.

</details>

<details>
<summary>Closed-menu references</summary>

Additional closed-menu views from ba1ed11 for comparison with the open-menu screenshots above.

**Light reference with the fallback menu closed. Captured at ba1ed11.**

Light reference with the fallback menu closed. Captured at ba1ed11.

**Dark reference with the fallback menu closed. Captured at ba1ed11.**

Dark reference with the fallback menu closed. Captured at ba1ed11.

</details>

[Download the visual evidence packet and capture manifest.](https://github.com/user-attachments/files/31957458/server-browser-pr-visual-evidence.zip)

## Risks and failure behavior

- Server Chrome is a privileged capability. It is opt-in, uses a loopback CDP listener and applies egress policy in the owning server runtime. Runtime routes remain authenticated; UI visibility is not the authorization boundary.
- Browser contexts isolate session storage, while control leases and attachment generations scope operations to the intended session and tab. Direct external CDP clients bypass these OpenChamber controls by design.
- A fixed port already in use must fail the managed launch instead of adopting another Chrome. Settings distinguish the saved choice from the active port. Disabling the feature stops owned Chrome and closes its sessions.
- Inspector bodies are opt-in, and capture/transport queues have explicit limits. DevTools may disconnect when progress stalls or a limit is exceeded; reconnecting does not turn a failed request into an empty success.
- The Ctrl/Cmd+P host-shortcut conflict is fixed. Current canvas and open-menu regression tests pass. Idle ephemeral cleanup is scheduled, and failed URL-token requests stop before opening a socket; current lifecycle and auth suites cover those paths.
- Chrome 109 remains the enforced minimum. Compatibility of its per-context proxy parameters has not been independently demonstrated here; the review thread disputed the original version concern. No Chrome-109 runtime validation is claimed.
- The deployed private-relay path remains untested end to end. Direct web/Bun validation does not establish relay or physical-device behavior.
- The separate private/loopback session-grant error is not claimed fixed by these DevTools commits. The stream-policy change retains the egress boundary.
- The work is substantial and remains a draft. Native Electron, VS Code, physical iOS/Android devices and the user's Windows 11/Zen installation are not claimed as manually validated by Linux web tests.


mattv8:
I can help review this one. @JosueGalRe I recommend rather than locking us into Chrome weight consider building in the Playwright+Chrome toolset. This feels like it might be re-inventing the wheel (although Openchamber browser might already have this builtin, correct me if I'm wrong).

JosueGalRe:
@mattv8 Playwright would still require us to launch and control a browser process, so it wouldn't really remove the Chrome/Chromium dependency here; it would mostly add another abstraction layer on top of it.

The reason for targeting Chrome/Chromium directly is CDP. It's the common interoperability layer already used by agent tooling such as Chrome DevTools MCP and Agent Browser. By exposing the managed browser's CDP endpoint, OpenChamber, the user, and external CDP-compatible agent tools can all attach to and control the same browser session instead of each tool owning a separate browser instance.

The other important difference from OpenChamber's existing built-in browser is where the browser actually runs. The existing browser is client-side, while this PR runs Chrome on the OpenChamber server and streams/control it from the client. That matters when moving between devices: `localhost` continues to mean the server, and services with strict CORS/origin rules don't suddenly need to be accessed through a different LAN, tunnel, or remote hostname just because I'm controlling OpenChamber from another machine.

So I don't see Playwright as an alternative to the server-browser architecture itself. It could potentially be used as an implementation detail, but we'd still want a server-owned Chromium process and direct CDP access for interoperability with external agent tooling.

mattv8:
I agree with your pushback, the server-owned Chrome browser architecture holds up, and the direct CDP fits the shared server-browser goal. Chrome 109 supports the required proxy parameters.

I found three behavior gaps that need fixing:
1. Fix keyboard input. Ctrl/Cmd+P currently reaches Chrome and opens OpenChamber’s command palette.
2. Fix cleanup. Idle ephemeral browser sessions are never automatically expired.
3. Fix auth errors. A failed token request triggers a tokenless WebSocket connection that the server always rejects.

You will also need to resolve conflicts and stale docs. The PR conflicts with main, and one comment incorrectly says server-chrome is unimplemented.

JosueGalRe:
@mattv8 Thanks for the follow-up review. We'll address the three behavior gaps: keep remote browser keyboard input from also triggering OpenChamber shortcuts, automatically expire idle ephemeral sessions, and stop the WebSocket connection attempt when requesting its auth token fails.

We'll also fix the stale server-chrome documentation and sync this branch with the latest main, resolving the conflicts. We'll check the tokenless upgrade error noted in the bot review as part of the auth fix, and verify the Chrome proxy compatibility claim against the protocol before changing the minimum version.

Work is underway. We'll share the fixes and validation results once they're ready.

JosueGalRe:
@btriapitsyn Following up on your Discord suggestion to convert this into an extension. I reviewed the SDK against main at `ea3ffa863` and wanted to keep the migration discussion here.

The backend looks like a good fit for `contributes.service`: the extension could own Chrome, CDP, browser contexts, tabs, egress policy, control leases, and cleanup. Those responsibilities can stay in the extension rather than becoming browser-specific SDK methods.

To preserve the current experience, I see a few gaps to work through:

- **A bidirectional service stream.** The current `serviceRequest` proxy collects a text response, caps it at 256,000 characters, and uses a 20-second request timeout. The [service contract](https://github.com/openchamber/openchamber/blob/ea3ffa863bebe4d25463ec56c88ba67a1fe78639/packages/sdk/GUEST_SERVICES.md) explicitly defers streaming. The viewer needs continuous frames and input, with ordering, acknowledgments, backpressure, and reconnect. A scoped service channel could cover this without adding a browser-specific API.
- **Activation without an open panel.** Services start on the first `serviceRequest`. Agents need to start and use the browser while no viewer is open. Is there a preferred way for an agent/MCP integration to activate and address an extension service? A separately configured MCP could own startup, but that would change lifecycle ownership.
- **DevTools assets and embedding.** The panel's sandbox is appropriate, but our dynamic DevTools documents/assets and protocol bridge need a supported route through it. This still needs a prototype; I am not proposing relaxing the sandbox.
- **Mobile parity.** Extensions currently target web and desktop. Moving the entire feature today would drop the mobile viewer unless mobile extension hosting is added or we deliberately scope the first version to web/desktop.

This is a source-based assessment, not a completed migration prototype. Would a generic streaming service API plus a supported non-panel activation path fit your SDK direction, or is there an existing route I missed? Once that boundary is clear, I would start with a small extension prototype and reuse the current browser modules and tests.

Separately, I synced the PR with main and resolved the conflicts in `362dbc3f3`. Workspace type-check/lint and web/VS Code builds passed, along with 609 server tests, 81 focused UI tests, and the Bun transport test. One server test was skipped. The branch remains the working reference while we discuss the extension approach.

btriapitsyn:
@JosueGalRe let's figure out what extension shape works for you, and we'll add the missing SDK pieces for it.

One option that needs a lot less from the SDK than your list: a browser provider. A service with no panel owns Chrome, CDP, contexts, egress proxy and cleanup, and answers the existing 10 `browser.*` actions over CDP with the same request and result shapes the desktop client returns today. We'd add a "Browser provider" dropdown in Settings → OpenChamber Tools. With nothing installed it shows "OpenChamber Web" alone, disabled. Once your extension is installed and approved it shows up as a second option, and picking it sends every `browser.*` action to your service whether the desktop is open or not. Disabling or removing it resets the choice to the built-in with a toast.

SDK for that would be a manifest field on `contributes.service` saying it provides the browser, a `POST /browser-control` contract on the loopback, and the broker spawning the service on the first action with an idle timeout.

You'd give up watching the agent's own tab in real time, but the existing browser panel stays, and for a dev server it shows the same app. Which brings me to a question I should have asked earlier. What in the current browser view doesn't work for you? Your viewer looks like a second implementation of a panel we already have, and I'd like to understand what it gives you that ours doesn't beyond console and network on the web build. If it's mostly "see what the agent is doing on the server", there may be a cheaper way to get that than a full remote viewer.

If it turns out you do need the full experience, then the streaming channel and non-panel activation are what we'd design, and DevTools embedding needs the prototype you mentioned. Either way is fine with us, we just want to agree on the shape before writing SDK code. What do you think?

JosueGalRe:
@btriapitsyn Thanks for laying out both options and being open to adding the missing SDK pieces. The provider approach makes sense, and I agree that we should settle the shape before asking for more infrastructure.

The main reason I built the viewer is to let the agent and me work in the **same browser session on the server**, including when I connect from another computer or my phone. Console and Network are useful, but they are not the main distinction.

A concrete example is an agent working on a dev server that is only reachable on the OpenChamber host. I want it to open the app, navigate to a particular screen, and reproduce something. Then I want to inspect that exact state, interact with the page if needed, and let the agent continue. That could include completing a login or stepping through a flow that needs human input.

Opening the same URL in another browser is useful, but it gives me a separate session. It may have different cookies, local storage, authentication, form values, or navigation state. Even with the same account, it will not necessarily reproduce what the agent currently sees. Sharing the actual tab makes the handoff much more direct.

There is also the host boundary. With server-owned Chrome, `localhost` consistently refers to the server. Changing the device I use to access OpenChamber does not change where the app runs or require that app to be exposed through another hostname. That matters for local services and applications with origin-sensitive authentication or CORS rules. Forwarding a port can solve reachability, but it does not share the browser state.

So, to answer your question about the existing panel: I am not saying it is broken or needs replacing. It serves the case where I want to open and use a page myself. What I need beyond that is:

- Observe the page the agent is actually using, including its current state.
- Take control of that same page for a manual step, then let the agent continue.
- Keep the browser session on the server when I switch clients or disconnect.
- Access that shared session from web and mobile, without requiring a desktop client to stay open.

I do not have a strong attachment to a separate panel or to keeping the current implementation in core. If the existing browser panel could display and control a selected provider's server-side tab, I would prefer reusing it. The separate viewer was how I implemented that behavior with the boundaries available at the time.

Your provider proposal sounds like a good foundation. Keeping Chrome, CDP, contexts, egress policy, and cleanup inside the extension, while the host owns provider selection and activation, is a separation I would be happy to work toward. It also solves an important part of the problem on its own: agent browsing without an open desktop client.

The part it would leave out is the shared human/agent session. I would be comfortable treating that as a second stage rather than making the first SDK addition carry everything. Full embedded DevTools could come later as well; preserving the live page and the ability to interact with it matters more to me initially.

Would it make sense to start with the provider contract you described, while leaving room for an optional viewing/control capability that the existing panel could consume? We could then prototype the smallest version of that shared-tab experience before deciding on the streaming API or DevTools embedding. If you have a simpler way to preserve the same-session handoff, I am very open to it.

The current PR can remain a working reference for the behavior and tests while we agree on the extension shape. I would rather reuse the existing UI and move this out of core where practical than maintain a parallel browser panel unnecessarily.

btriapitsyn:
Sounds good, let's go with the provider contract first. I'm starting on the SDK side now, both pieces, and will post here when there's something to look at.

On the second stage, the SDK side of it shouldn't know about browsers. Your extension still does exactly what it does now, but what you describe (see the surface the agent is working in, take over for a manual step, hand it back, keep the session alive on the server) applies just as well to an extension driving a simulator or a desktop app. So the shape I have in mind is a generic "shared surface" capability on a service: frames out, input in, one controller at a time, rendered by the existing panel. Your viewer would be its first real user, and most of what you built for the screencast and control lease maps onto it directly.

btriapitsyn:
@JosueGalRe both SDK pieces are in main now (#3734).

- `contributes.service.provides: ["browser"]`: your service answers the ten `browser.*` actions at `POST /browser-control` on its loopback, the user picks it under Settings → General → OpenChamber Tools → Browser provider, and the host starts it on the first action and stops it after ten idle minutes. No panel needed.
- `contributes.service.surface: true`: the host pulls frames from `GET /surface/frame` (long-poll, one frame per viewer ack), draws them in your extension's rail panel, sends pointer/keyboard/paste/resize/clipboard back, and owns who is in control (nobody, agent, user). The user takes control by clicking or typing, hands it back with one button, and while they hold it your browser actions get a 409 telling the agent to wait.

Contract and parsers: `packages/sdk/GUEST_SERVICES.md` (sections "Browser provider" and "Shared surface"), types exported from `@openchamber/sdk`. `packages/sdk/examples/browser-provider-stub` implements both roles with a fake page painted as PNG, so you can see the host side working before wiring Chrome in.

That's it from our side. Carve the service out of this branch whenever you're ready, and shout here if the contract is missing something your Chrome side needs.

JosueGalRe:
@btriapitsyn Thanks for shipping both pieces in #3734. We now have an extracted extension using the official browser-provider and shared-surface contracts: [openchamber-server-browser](https://github.com/JosueGalRe/openchamber-server-browser). The shared-tab workflow is working. Here is where the migration stands and what we still need to reach the behavior of this PR.

### What we have tested

We built OpenChamber main at `be8eed934df11d6508ee544a723c37ce00d87e80` and installed extension commit `8ab3b843de83c74eedef853fc3d6963873c87560` from its GitHub URL through the official extension installer. This was a Linux web-host test with real Chrome and a real managed OpenCode backend. We exercised the browser actions through the host control API, not a model conversation.

- All ten `browser.*` actions completed, including navigation, inspection, resizing, and saving a screenshot.
- The provider can start and operate without an open viewer.
- The shared panel displays the same Chrome target the provider controls. We clicked, typed/pasted into a form, saved it, handed control back, and confirmed the provider could read the changed form state.
- User takeover blocks provider actions, and handing back control lets them resume.
- An animated test page updates visibly in the panel. This establishes live updates, not a measured frame rate or latency guarantee.

Earlier extraction checks also covered blocked private origins, explicitly allowed local origins, and stopping the service/Chrome when disabling the extension. The initial automated suite passed 15 tests under both Node and Bun. Those are earlier checks, separate from the latest-host run above. Electron, Windows/macOS, relay, mobile, VS Code, and a real model-driven conversation remain unverified.

### Findings from the shared panel

A native HTML `<select>` on the test page receives input, but its options popup is not visible. ArrowDown and Enter change the selected value. We added speed buttons so animation testing can continue, while keeping the select as a reproduction. That workaround does not fix arbitrary websites.

We have not isolated this to the SDK. Both the original implementation and the extension use `--headless=new` and `Page.startScreencast`; the extension sends the resulting JPEGs to the host panel. We remember this interaction working in the original, but have not yet repeated a controlled comparison with the same page and Chrome binary. The next diagnostic is to inspect the frame before it enters the SDK and compare both implementations. That will distinguish capture/input behavior from host transport/rendering before proposing a fix upstream.

One smaller observation: while the user held control, our request through `/api/openchamber/control` was correctly rejected, but the outer response was HTTP 500 with the interaction-blocked message. Your comment describes 409. We still need to trace whether that status is lost in the outer control route; the ownership protection itself worked.

### What is still missing compared with this PR

The extension currently controls one target in one shared browser context per service. It has no project/chat isolation, tab manager or popup-target switching, embedded DevTools, interactive console/network inspector, or original browser toolbar. Local access uses a manual origin allowlist rather than live dev-server discovery and scoped grants. Selection across frames/shadow roots, editing shortcuts, double-click behavior, IME, and viewport coordination need more implementation and testing.

Chrome uses a temporary profile, so state survives a live handoff but not a service restart. Durable profiles would be a separate feature; the original also used temporary profiles.

### Where SDK guidance or changes would help

We want to keep browser-specific logic in the extension and use generic host capabilities, as you suggested. The main contract questions are:

1. **Authoritative scope and viewer routing.** Provider requests carry `requestId`, `action`, and `parameters`, but no originating project/chat identity. The current shared panel also attaches at service level. We need a supported way to bind provider actions, frames, input, control ownership, and cleanup to the same isolated scope. Inferring it from whichever project is currently visible would be unsafe.
2. **Controls beside the shared panel.** `service.surface: true` excludes `panel.entry`. What is the intended composition pattern for a tab strip, address/navigation controls, or an inspector beside the host-owned viewer? We would like to preserve the host's control arbitration.
3. **DevTools integration.** We still need a small prototype for serving the frontend/assets and carrying authenticated, target-scoped bidirectional CDP traffic with cancellation and backpressure. We have not established that this requires a new SDK API; guidance on the supported extension mechanisms would help us test the smallest approach first.
4. **Live dev-server discovery and authorization.** Is there a supported extension API for consuming authoritative host discovery and scoped local-origin grants? Static configuration works today but does not reproduce the original lifecycle or revocation behavior.
5. **Release compatibility.** The extension currently vendors a pinned official SDK snapshot because the published package inspected during extraction lacked these exports. We need a compatible published SDK and a reliable minimum host version before advertising a normal release install.

Chrome target tracking, action implementation, input fidelity, selection/clipboard handling, cleanup, and regression tests remain our responsibility. Agent tab selection also needs an agreed routing approach; additional extension tools may cover part of it, so we are not treating every missing feature as an SDK blocker.

The [README](https://github.com/JosueGalRe/openchamber-server-browser#roadmap-to-parity) records the full parity matrix and separates extension work from contract questions. The next useful design decision seems to be shared scope identity and viewer routing, because tabs, permissions, and DevTools all need to attach to the right session. Does that fit the direction you have in mind, or is there an existing supported pattern we should use?

btriapitsyn:
@JosueGalRe thanks for the detailed run. Three things landed in main (959d179c6) after your notes; they'll ship with the next release.

- The 500 was ours: browser errors kept their status in a field the outer route didn't read. You now get the real 409 for "user holds control" (and 400 for "no such element").
- Every provider request now carries `context: { directory, sessionId }`: the project and chat the action came from, filled by the host from the tool call (a `null` means unknown, e.g. a call from the CLI). Key your targets on it however you like. The surface stays one per service, so you decide which target it shows.
- `surface: true` no longer excludes `panel.entry`. Declare both and the host docks your page beside the picture: `panel.dock` is the edge (`top`, `bottom`, `left`, `right`; default `top`) and `panel.size` its thickness in px (default 40). It's a normal panel page talking to your service through `host.serviceRequest`; frames, input and control arbitration stay with the host. The stub example now has an address bar docked on top.

DevTools inside the panel and dev-server discovery/grants we'd rather keep out of the SDK for now; that's the part of the original we didn't want in core. The docked page gives you room for a console toggle of your own if you need one.

The published `@openchamber/sdk` with these exports goes out with that release too. Watch the releases page: the release version is the minimum host version to put in your `engines.openchamber`.

JosueGalRe:
@btriapitsyn Thanks for the additions in `959d179c6`. We’ve adopted the authoritative project/chat context and docked panel in [Server Browser](https://github.com/JosueGalRe/openchamber-server-browser), now at experimental version 0.3.0.

Each project/chat has an isolated browser runtime. The dock provides session switching, navigation, and an optional native-select compatibility mode, while the host retains control arbitration. The updated README records the remaining parity gaps.

We also completed the select comparison: the original integrated implementation reproduced the invisible native popup too, so we are not reporting this as an SDK regression. Our opt-in workaround uses Chromium’s `appearance: base-select`; options become visible and clickable without replacing the controls. It is reversible, but can change their appearance. Custom Radix menus worked in our tests.

Separately, we fixed an unhandled proxy socket `EPIPE` in our extension that could terminate its service. The suite now passes 41 tests, with real Chrome and live Linux web-host checks. Other platforms and a model-driven conversation remain unverified.

The main remaining contract question concerns viewer identity and stale input. Dock requests do not identify the controlling viewer, and input does not identify the view generation it came from. We currently disable dock mutations while anyone holds control. Could a generic viewer/lease identity and view generation eventually cover both paths, allowing us to reject input from an outdated frame after switching sessions?

A smaller layout question: is there a supported pattern for expanding an extension-owned console/network inspector beyond the fixed dock size, while retaining the host’s control indicators and handback action?

Understood on keeping embedded DevTools and dev-server discovery/grants outside the SDK for now. We’ll continue the extension-owned work and adopt the published SDK and minimum host version when that release lands.

JosueGalRe:
I will be closing this PR as the extension https://github.com/JosueGalRe/openchamber-server-browser has been released and can be used with the 2.0.0 release. Thanks for all the help in enabling us to make this work directly as an extension :)

PR #3437, opened 2026-09-08
fix(small-model): use selected runtime model endpoint

### LINKED ISSUE #3436: [Bug] Small-model runtime provider uses the first model endpoint for every plugin model
### What's wrong?

OpenChamber's server-side small-model path treats a runtime OpenCode provider as if it had one API endpoint for every model. That is not true for many OpenCode plugins.

The runtime provider snapshot in `packages/web/server/lib/small-model/runtime-providers.js` currently derives the provider endpoint from the first model returned by OpenCode:

```js
const firstModel = Object.values(record(record(raw).models))[0];
baseURL: endpoint(options.baseURL) ?? endpoint(record(firstModel.api).url)
```

This discards the selected model's `api.url` and `api.npm`. Later, `packages/web/server/lib/small-model/call.js` uses that provider-level `baseURL` for the request and sends the generic OpenAI-compatible `/chat/completions` request.

A provider plugin can register models backed by different APIs, paths, or SDK adapters. OpenCode already exposes that information per model and uses it during normal chat execution. OpenChamber loses it in the small-model path.

### Our reproduction

We use the `@mcowger/opencode-plexus` plugin, which registers a `plexus` provider and exposes models through a Plexus gateway. The provider has models using different target APIs:

| Model | OpenCode `api.npm` | OpenCode `api.url` |
| --- | --- | --- |
| `gpt-5.6-luna` | `@ai-sdk/openai` | `https://<plexus-host>/v1` |
| `gemini-3.5-flash-lite` | `@ai-sdk/google` | `https://<plexus-host>/v1beta` |
| `deepseek-v4-flash-0731` | `@ai-sdk/openai-compatible` | `https://<plexus-host>/v1` |

OpenCode chat works. Its log shows the selected model running through the OpenCode AI SDK path:

```text
stream providerID=plexus modelID=gpt-5.6-luna ... small=true
llm runtime selected llm.runtime=ai-sdk llm.provider=plexus llm.model=gpt-5.6-luna
```

OpenCode's `/provider` response reports the per-model endpoints above. The provider itself does not have a single `options.baseURL`; the plugin uses its own `plexusBaseURL` setting and publishes the resolved endpoint on each model.

When OpenChamber runs a small-model action with `plexus/gpt-5.6-luna`, the runtime snapshot can choose the first listed Gemini model's `/v1beta` endpoint. OpenChamber then sends the Luna request to the wrong API base instead of the selected model's `/v1` endpoint. The result is a failed small-model action even though the same model works in OpenCode chat.

### Steps to reproduce

1. Configure an OpenCode plugin that registers one provider with models using different `api.url` and/or `api.npm` values.
2. Confirm that a model using the non-first endpoint works in a normal OpenCode chat session.
3. Start OpenChamber with the same managed OpenCode instance.
4. Select or configure that model as the Small Model, then trigger a small-model feature such as a commit summary, title, note summary, goal audit, or Changes Walkthrough.
5. Observe that the request uses the first model's endpoint and generic wire format rather than the selected model's endpoint and adapter.

### Expected behavior

Small-model dispatch should preserve the selected OpenCode model's runtime information:

- resolve `api.url` for the selected `providerID/modelID`, not the first model in the provider;
- honor the selected model's `api.npm` or another explicit runtime adapter;
- preserve plugin-owned request handling where the plugin supplies a custom `fetch` implementation;
- use the same OpenCode execution path as normal chat when direct dispatch cannot preserve the plugin transport.

A provider-level endpoint may remain as a fallback for providers whose models genuinely share one API, but it must not override model-specific runtime data.

### Related issue and regression

This is related to #2666, which was closed after runtime provider discovery was added in `f1b602019`. That fix made plugin providers visible to the small-model path, but the snapshot still stores only one provider-level endpoint by taking the first model's endpoint. #2666's fix therefore does not cover a plugin provider with multiple target APIs.

### Environment

- OpenChamber installed web package: `1.22.2`
- Managed OpenCode with `@mcowger/opencode-plexus`
- Linux
- Affects server-side small-model consumers, including commit summaries, titles, notes, goals, and walkthrough generation

### Logs

No credentials or prompt contents are included. The relevant OpenCode evidence is shown above. OpenChamber's `[small-model:diagnostic]` logging identifies the provider and model but currently cannot show the selected model endpoint, which also makes this failure harder to diagnose.

mcowger:
## Intent

Small-model calls now use the endpoint OpenCode resolved for the selected model. A runtime provider can expose models backed by different APIs, so using the first model's endpoint could send a valid model request to the wrong path and fail even though the same model worked in OpenCode chat.

Fixes #3436.

## Non-goals

- Selecting a different wire adapter from `api.npm`.
- Preserving plugin-owned `options.fetch` implementations that OpenCode cannot serialize through `/provider`.
- Changing the existing OpenAI-compatible request format or provider/config credential precedence.
- Changing normal OpenCode chat execution.

## Affected surfaces

`packages/web/server/lib/small-model`: runtime provider discovery and server-side small-model dispatch used by summaries, titles, notes, goals, and walkthrough generation.

The runtime snapshot now keeps each model's `api.url` and `api.npm`. Dispatch uses the selected model's endpoint, then falls back to the provider-level runtime endpoint and existing catalog resolution. The provider-level endpoint remains available for providers whose models share one endpoint.

Only the web server's small-model path changes. Shared UI, Electron, VS Code, mobile, relay transports, persisted settings, credentials, and public routes are unchanged.

## Repository guidance

|---|---|---|
| `AGENTS.md`, `openchamber-change-discipline` | Server source, tests, and module contract change | Keeps the fix inside the owning small-model module, preserves existing fallback behavior, adds focused regression coverage, and introduces no dependency or persisted-format changes. |
| `packages/web/README.md` | Web package work | Confirms the package-scoped commands and server ownership. |
| `packages/web/server/lib/small-model/DOCUMENTATION.md` | Small-model runtime contract | Updates endpoint precedence to document selected-model resolution before provider fallback. |
| `ui-api-decoupling` and `references/implementation-map.md` | Runtime/OpenCode provider discovery | Keeps provider state at the server runtime boundary and does not bypass the existing OpenCode connection or shared runtime contract. |
| `communication-style` | PR description and documentation | Uses direct, concrete language and records the intentionally limited scope. |

## Validation

| Check | Result |
|---|---|
| `bun run test` from `packages/web` | 186 test files passed; 2,081 tests passed; 2 skipped. |
| `bun run type-check` from `packages/web` | Passed. |
| `node --check` on the two changed modules and two changed test files | Passed. |
| Focused small-model Vitest run | 3 test files passed; 70 tests passed. |
| Manual runtime/plugin verification | Verified. |
| `bunx oxlint` on changed JavaScript files | Reports existing legacy `typeof` narrowing and module-mocking findings in these files; no new lint-specific suppression was added. |

## Visual evidence

No rendered UI change. This is a server-side endpoint selection fix, so screenshots or recordings cannot show the affected behavior. The relevant evidence is the request URL assertion in `packages/web/server/lib/small-model/call.test.js`.

## Risks and failure behavior

- Explicit `provider.<id>.options.baseURL` continues to take precedence, so configured custom gateways are unchanged.
- If the selected model has no endpoint, dispatch falls back to the existing provider-level runtime endpoint, then the catalog endpoint. Existing providers with one shared endpoint retain their behavior.
- `api.npm` is preserved in the snapshot for runtime fidelity but is intentionally not used to introduce a second manual transport in this PR.
- Plugin-specific `options.fetch` behavior remains unsupported when the plugin depends on serialized-away functions. Such providers continue to fail through the existing direct-call path rather than being guessed at.
- No credentials are logged or persisted. Reverting the commit requires no migration.

### DIFF
--- a/packages/web/server/lib/small-model/DOCUMENTATION.md
+++ b/packages/web/server/lib/small-model/DOCUMENTATION.md
@@ -117,8 +117,9 @@ other runtime API.
   - Everything else: OpenAI-compatible `/chat/completions` against the
     provider's base URL, resolved from (1) `provider.<id>.options.baseURL`
     in the OpenCode config, (2) the hardcoded `https://api.openai.com/v1`
-     endpoint, (3) the endpoint OpenCode resolved at runtime, or (4) the
-    provider's `api` field from the models.dev catalog. The credential follows
+     endpoint, (3) the selected model's endpoint OpenCode resolved at runtime,
+     (4) the provider-level runtime endpoint, or (5) the provider's `api` field
+     from the models.dev catalog. The credential follows
     the same shape: config `options.apiKey`, then the runtime credential, then
     the auth.json entry. `provider.<id>.options.headers` is sent with the
     request and overrides the bearer default, so gateways that authenticate on
--- a/packages/web/server/lib/small-model/call.js
+++ b/packages/web/server/lib/small-model/call.js
@@ -606,6 +606,8 @@ const readProviderConfig = (workingDirectory, providerID) => {
   }
 }
 
+const getRuntimeModel = (runtimeProvider, modelID) => runtimeProvider?.models?.get(modelID) ?? null;
+
 // ---------------------------------------------------------------------------
 // Dispatch
 // ---------------------------------------------------------------------------
@@ -650,6 +652,7 @@ export async function callSmallModel({ auth, catalog, workingDirectory, sessionI
   const tokens = Number(maxOutputTokens) > 0 ? Number(maxOutputTokens) : DEFAULT_MAX_OUTPUT_TOKENS;
   const providerConfig = readProviderConfig(workingDirectory, providerID);
   const runtimeProvider = await getRuntimeProvider(providerID);
+  const runtimeModel = getRuntimeModel(runtimeProvider, modelID);
   // Match OpenCode's resolveSDK precedence: config `provider.<id>.options`
   // wins, then what OpenCode itself resolved at runtime (the only place a
   // plugin's credential exists), and the auth.json entry last.
@@ -759,9 +762,10 @@ export async function callSmallModel({ auth, catalog, workingDirectory, sessionI
   // base URL for that provider (openai itself included). When a custom provider
   // is not in the catalog (e.g. a user-configured OpenAI-compatible proxy),
   // fall back to its baseURL from the OpenCode provider config, then to the
-  // endpoint OpenCode resolved at runtime — which for a plugin provider is the
-  // only place it exists, and for several of them is a local proxy the plugin
-  // itself runs. The openai provider also respects
+  // selected model's endpoint OpenCode resolved at runtime, then to the
+  // provider-level runtime endpoint. For a plugin provider, the runtime listing
+  // is the only place those endpoints exist, and several are local proxies the
+  // plugin itself runs. The openai provider also respects
   // provider.openai.options.baseURL — OpenCode itself uses the same config for
   // all providers including openai.
   const provider = getCatalogProvider(catalog, providerID);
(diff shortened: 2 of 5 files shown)


btriapitsyn (approved):

btriapitsyn:
Thanks for tracking this down and sending the fix!

PR #3440, opened 2026-09-09
fix(queue): re-read the server queue after a stream gap

### LINKED ISSUE #3360: [Bug] Queued messages remain in UI dock after being auto-dispatched (empty directory in broadcast snapshot drops cleanup)

**Summary**
When a queued message is auto-dispatched by the server upon a session becoming idle, the message is successfully sent to OpenCode and the agent starts processing it. However, the "Queued messages" card (`QueuedMessageChips`) remains visible above the composer, displaying the already-sent message. Clicking `send` or `edit` on this ghost card fails with 404 ("Couldn't load the queued message. Please try again.").

**Steps to Reproduce**
1. Send a prompt while an agent turn is running so it gets queued (or set follow-up behavior to Queue).
2. Wait for the agent to finish the current turn.
3. The server auto-dispatches the queued message to OpenCode; OpenCode begins executing the new turn.
4. Observe the composer input area: the "Queued messages 1" chip still stays in the UI.
5. Clicking `edit` or `send` on the chip triggers a 404 toast because the server queue has already been emptied.

**Root Cause Analysis**
1. **Server-side (`server/lib/message-queue/runtime.js`)**:
   In `tick(sessionId)`, after successfully dispatching the item via `sendItem`:
   ```javascript
   const after = queues.get(sessionId);
   if (after) setQueueItems(sessionId, after.directory, after.items.filter((entry) => entry.id !== item.id));
   commit(sessionId);
   ```
   When the last queued item is removed, `items.length === 0`. In `setQueueItems`:
   ```javascript
   const setQueueItems = (sessionId, directory, items) => {
     if (items.length === 0) {
       queues.delete(sessionId);
       return;
     }
     queues.set(sessionId, { directory, items });
   };
   ```
   `queues.delete(sessionId)` immediately deletes the entry from the `queues` Map.
   Immediately afterwards, `commit(sessionId)` calls `broadcast(sessionId)`, which invokes `sessionSnapshot(sessionId)`:
   ```javascript
   const sessionSnapshot = (sessionId) => {
     const queue = queues.get(sessionId);
     return {
       sessionId,
       directory: queue?.directory ?? '',
       items: (queue?.items ?? []).map(toPublicItem),
       sendingId: sending.get(sessionId) ?? null,
     };
   };
   ```
   Since `queues.get(sessionId)` was just deleted, `queue` is undefined and `directory` falls back to `''` (empty string).
   The server broadcasts `openchamber:message-queue.updated` with `{ revision, session: { sessionId, directory: '', items: [] } }`.

2. **Client-side (`packages/ui/src/stores/messageQueueStore.ts`)**:
   Upon receiving `openchamber:message-queue.updated`:
   ```javascript
   const applyServerSession = (session, revision, runtimeKey) => {
     if (runtimeKey !== currentRuntimeKey()) return;
     const target = makeSessionTarget(session.sessionId, session.directory, runtimeKey);
     if (!target) return;
     ...
   };
   ```
   `makeSessionTarget` validates that `directory` is non-empty (`normalizeDirectory(directory)`). Since `session.directory` is `''`, `makeSessionTarget` returns `null`.
   `if (!target) return;` triggers, silently dropping the server's queue-clearing update.
   Consequently, the client's local Zustand store never removes the sent message, leaving the chip dock permanently visible until manual deletion or window reload.

**Suggested Fix**
- **Server**: In `runtime.js`, preserve `directory` when snapshotting an empty queue (e.g. retain last known directory or pass `directory` into `commit`/`sessionSnapshot`), so the broadcast snapshot includes the valid `directory`.
- **Client**: In `applyServerSession`, when `session.items` is empty, fallback to matching and clearing target keys by `sessionId` even if `session.directory` is empty.

### Where does it happen?

Desktop (macOS)

### Version (if known)

1.22.1

### Logs (optional)

```shell
[2026-09-05 15:09:06.689] [info] [message-queue] sent queued message to ses_f8fafe8b6ffe7DScsYKIVA7Jno
```

IbrahimKhan12:
Fixes #3360

## Intent

The 1.22.2 fix clears the queue card when the client receives the final `openchamber:message-queue.updated` event. A disconnected client, or one using the `/api/global/event` SSE fallback, can miss that event and keep a message the server already dispatched.

This change re-reads the authoritative queue after stream reconnects and transport switches. A failed `take` also refreshes the session before returning the error. Snapshot revision checks preserve broadcasts newer than an in-flight refresh.

## Non-goals

- Carrying OpenChamber events over the `/api/global/event` SSE proxy.
- Applying the same reconnect policy to the permission store.

## Affected surfaces

- Shared UI queue state for web, desktop, hosted mobile, and Capacitor mobile.
- VS Code is unchanged because it owns its queue locally.
- No server API, persistence format, or route changes.

## Repository guidance

| Guidance | Application |
|---|---|
| `AGENTS.md`, `openchamber-change-discipline` | Reconcile the projection in its owning store and keep the sync provider thin. |
| `sync-state-invariants` | Preserve newer broadcasts by revision and keep prior state when refresh fails. |
| `relay-transport`, queue/store documentation | Document the SSE transport gap without changing transport behavior in this PR. |

## Validation

| Check | Result |
|---|---|
| `bun test packages/ui/src/stores/messageQueueStore.server.test.ts` | 20 pass, 0 fail; four new regressions failed before the fix. |
| Related queue, deletion cleanup, and event-pipeline tests | 34 pass, 0 fail, each file run separately. |
| `bun run lint:ui` | 0 errors; one unrelated existing warning in `MobileChangesSurface.tsx`. |
| `bun run type-check:ui` | No new errors; five existing Codemirror errors reproduce on `main` due to duplicate local package installs. |
| Changed-file oxlint and `bun run dead-code` | No new findings. |
| Headless Chromium, stock 1.22.2 vs this branch | After a missed clear event and stream reconnect, 1.22.2 kept the card for 20 seconds; this branch fetched `/api/message-queue` and cleared it in 3.7 seconds. |

## Visual evidence

Same desktop session after reconnect:

| Stock 1.22.2 | This branch |
|---|---|
| Queued message remains after reconnect on OpenChamber 1.22.2 | Queue card cleared after reconnect with this branch |

No styling changed, so narrow layouts and themes are unaffected.

## Risks and failure behavior

- Each reconnect adds one `GET /api/message-queue` request.
- A failed refresh preserves the current projection for the next event or reconnect.
- `resync()` waits for initial hydration, which prevents legacy local messages from being uploaded twice.


btriapitsyn:
Thanks for tracking down the stream-gap case. Merged with stable SSE updates through the control stream, bootstrap recovery, stale-snapshot protection, and migration recovery finished on our side; no queue poller was added. The final focused run passed 27 queue tests and 3 control-stream tests.

PR #3450, opened 2026-09-09
test(ui): pin buffer reset when a project action reruns

### LINKED ISSUE #3448: [Bug] Re-running a project action superimposes the previous run's output

Re-running a project action in a terminal tab that already holds a previous run paints the new output on top of the old output. The xterm screen ends up with both runs interleaved: lines wrap at wrong columns, fragments of the old run sit between lines of the new one, and cursor-positioning sequences from the new command redraw over stale content. See the screenshot below.

Steps:

1. Run a project action so its terminal tab collects some output.
2. Let it finish (or not, either way).
3. Re-run the same action from the actions dropdown.
4. The tab shows the new run mixed into the previous run's scrollback.

Expected: the tab shows only the new run, replayed cleanly from the fresh attach snapshot.

### Why it happens

A re-run reuses the tab id as the server terminal session id, and the recreated server session restarts its sequence counter at 0. Before #3362, `allocateActionExecution()` kept the tab's old buffer and its higher `lastSequence`, so:

1. The fresh attach snapshot arrived at sequence 0 and `replaceBuffer()` rejected it as stale (`buffer.lastSequence > sequence`).
2. New output was dropped by the same guard until its sequence passed the old `lastSequence`, then appended onto the old chunk list.
3. The viewport saw a plain append, skipped its reset-and-replay path, and wrote the new run over the stale screen.

The sequence guards are correct for one session incarnation; nothing marked the boundary when a new incarnation reused the session id.

### Status

Already fixed on `main` by #3362 (merged minutes after #3287): `allocateActionExecution()` now drops the tab's buffer and nulls `terminalSessionId`, and reconciliation drops the buffer when an action tab's execution changes. Builds cut from branches or forks that predate #3362 still show the bug — that is where the screenshot below comes from. Syncing past 3ffd5d3f4 resolves it.

#3362 shipped without a store-level regression test for the local re-run path, so this issue closes with a test-only PR pinning it: re-run drops prior output and the sequence-zero snapshot is accepted.

### Where does it happen?

Desktop Web (any runtime; the code is in the shared terminal store).

Previous run superimposed on the new run

mattv8:
## Intent

Closes #3448 with a regression test. The superimposed-output bug on action re-run (previous run's screen mixed into the new run) was fixed by #3362, but that PR shipped without a store-level test for the local re-run path. This adds one: starting a new execution drops the tab's old buffer, so the recreated session's sequence-zero attach snapshot is accepted instead of being rejected as stale and appended over the old screen.

The test pins the exact production shape: the tab id doubles as the server session id, so `setTabSessionId`'s change-of-id buffer reset never fires on a re-run and `allocateActionExecution` must do the drop itself.

## Non-goals

No source changes. Reconciliation-path buffer behavior (rebind drops, same-run preservation) is already covered by the tests #3362 added.

## Affected surfaces

`packages/ui` test code only (`useTerminalStore.test.ts`). No runtime, contract, or persisted behavior changes on any surface.

## Repository guidance

|---|---|---|
| `openchamber-change-discipline` | Any source modification | Smallest complete change: one test, no speculative source edits; an earlier draft broadened non-action rebind drops but was cut for lack of a reproducing scenario |
| `sync-state-invariants` | Terminal buffer lifecycle and sequence reconciliation | Test asserts the owner-side transition (`allocateActionExecution`) enforces the incarnation boundary, matching the documented invariant that buffers belong to the tab's current session |
| `packages/ui/src/stores/DOCUMENTATION.md` | Owning module docs | Test exercises `getBuffer`/`replaceBuffer`/`appendToBuffer` through the public store API, consistent with the documented buffer ownership invariants |

## Validation

| Check | Result |
|---|---|
| `bun test packages/ui/src/stores/useTerminalStore.test.ts` | 42 pass, 0 fail |
| `bun run type-check:ui` | pass |
| `bunx oxlint packages/ui/src/stores/useTerminalStore.test.ts` | pass |
| Test fails without the #3362 fix | verified by analysis of the pre-#3362 code (`replaceBuffer` rejects `lastSequence > sequence`); the branch this was investigated on reproduced the user-visible superimposition |

## Visual evidence

No visible change; test-only diff cannot affect rendered behavior. The bug this guards against is pictured in #3448.

## Risks and failure behavior

None identified: the diff adds one deterministic store test with no shared fixtures or timing dependence.

### DIFF
--- a/packages/ui/src/stores/useTerminalStore.test.ts
+++ b/packages/ui/src/stores/useTerminalStore.test.ts
@@ -386,6 +386,25 @@ describe('terminal state reconciliation', () => {
     expect(tab.purpose).toEqual({ type: 'project-action', actionId: 'build', executionId: secondExecution });
   });
 
+  // Regression for openchamber/openchamber#3448: a rerun reuses the tab id as
+  // the server session id, and the recreated server session restarts its
+  // sequence counter at zero. If allocateActionExecution kept the old buffer,
+  // the stale lastSequence would reject the new run's attach snapshot and its
+  // later output would be appended over the previous run's screen.
+  test('starting a rerun drops prior output so its sequence-zero snapshot is accepted', () => {
+    const tabId = setup();
+    useTerminalStore.getState().setTabPurpose('/repo', tabId, { type: 'project-action', actionId: 'build', executionId: 'exec-old' });
+    useTerminalStore.getState().setTabSessionId('/repo', tabId, tabId, { expectedExecutionId: 'exec-old' });
+    useTerminalStore.getState().appendToBuffer('/repo', tabId, 'old output', 5);
+
+    useTerminalStore.getState().allocateActionExecution('/repo', tabId, 'build');
+
+    expect(buffer(tabId).chunks).toEqual([]);
+    useTerminalStore.getState().replaceBuffer('/repo', tabId, 'new snapshot', 0);
+    expect(buffer(tabId).chunks.map((chunk) => chunk.data)).toEqual(['new snapshot']);
+    expect(buffer(tabId).lastSequence).toBe(0);
+  });
+
   test('applies snapshots atomically and deduplicates output by sequence', () => {
     const tabId = setup();
     useTerminalStore.getState().replaceBuffer('/repo', tabId, 'prompt', 4);


mattv8:
Closing this because I realized the underlying bug was already fixed by #3362, and the regression coverage this adds isn't worth carrying as a standalone PR. Analysis lives in #3448 if the re-run path ever regresses.

PR #3451, opened 2026-09-09
fix(chat): keep image export working with external links

ChangeHow:
## Intent

Keep message image export working when rendered Markdown contains an external-link favicon. The export clone now drops that decorative remote image before `html-to-image` processes resources, so the browser does not attempt the CORS-blocked favicon fetch.

## Non-goals

- Do not change how favicons render in chat.
- Do not proxy or rewrite third-party image requests.
- Do not suppress failures from message images that are part of the exported content.

## Affected surfaces

`packages/ui` message image export is affected. The clone preparation runs before the runtime-specific save or share branch, so web, desktop, VS Code, hosted mobile, and Capacitor use the same behavior. The on-screen message DOM, persisted data, API contracts, authentication, and runtime transports are unchanged. Exported images omit the decorative external-link favicon but retain message text and content images.

## Repository guidance

|---|---|---|
|  |  |  |
| Root `AGENTS.md` and `openchamber-change-discipline` | This changes executable shared UI source and adds a focused module. | The change stays in the owning UI code, adds no dependency or public package contract, and uses focused tests plus package-level static checks. |
| `ui-api-decoupling` and `references/browser-assets-and-auth.md` | The failure starts with a third-party browser image that JavaScript cannot read across origins. | The fix does not proxy the favicon, add credentials, rewrite runtime URLs, or weaken the browser security boundary. It removes only the marked decoration from the export clone. |
| `diagnose` | This is a reported export failure with a known external-resource trigger. | A regression test established the failing clone behavior before the fix. Two one-variable ablations confirmed that both the marker and clone cleanup are necessary. |
| `packages/ui/src/components/chat/message/parts/DOCUMENTATION.md` | `MessageBody.tsx` owns message-part rendering and documents how assistant Markdown handles images. | The fix preserves on-screen Markdown and gallery behavior. It changes only the temporary image-export clone. |
| `CONTRIBUTING.md` | The PR changes shared UI behavior. | The PR names affected runtimes, exact validation results, omitted checks, and failure behavior. |

## Validation

| Check | Result |
|---|---|
|  |  |
| `bun test src/components/chat/message/imageExport.test.ts` from `packages/ui` | Passed: 2 tests and 6 expectations. The tests assert that Markdown favicons receive the export-only marker, the export clone omits the remote favicon, content images remain, and the original DOM is unchanged. |
| Ablation with the same focused test | Removing the favicon marker produced 1 failure. Restoring the marker and removing clone cleanup produced 1 failure. Restoring both produced 2 passing tests. |
| `bun run type-check` from `packages/ui` | Passed. |
| `bun run lint` from `packages/ui` | Passed with 0 errors. It reported one unrelated existing `react-hooks/exhaustive-deps` warning in `MobileChangesSurface.tsx:212`. |
| Targeted ESLint for the four changed files | Passed with no output. |
| `bunx oxlint packages/ui/src/components/chat/message/imageExport.ts packages/ui/src/components/chat/message/imageExport.test.ts` | Passed with no output. A broader changed-file run still reports existing anti-slop findings outside the changed lines in `MessageBody.tsx` and `decorate.ts`. |
| `bun run dead-code` | Completed. It reported no finding for the new `imageExport` module; the repository's existing non-blocking report remains. |
| Full workspace test/build and a real Chromium export | Not run. The focused DOM test proves that the CORS-triggering remote image is absent before `html-to-image` runs, but it is not an end-to-end browser export. |

## Visual evidence

Not included. This changes export success or failure and removes a non-critical favicon only from the off-screen export clone. Deterministic DOM assertions prove the relevant contract better than a screenshot: the remote favicon is absent while message content, content images, and the on-screen source DOM remain intact.

## Risks and failure behavior

The exported PNG no longer includes external-link favicons. The selector only removes elements carrying the dedicated export-exclusion marker, so other images keep the existing `html-to-image` failure behavior rather than being silently ignored. On-screen favicons still load normally. The temporary wrapper cleanup remains in `MessageBody`'s existing `finally` block. There are no persisted-data, API, authentication, dependency, or migration changes. Reverting this commit restores the previous export behavior.

### DIFF
--- a/packages/ui/src/components/chat/markdown/decorate.ts
+++ b/packages/ui/src/components/chat/markdown/decorate.ts
@@ -2,6 +2,7 @@ import { copyTextToClipboard } from '@/lib/clipboard';
 import { getExternalFaviconUrl, isExternalHttpUrl, isLoopbackHttpUrl } from '@/lib/url';
 import { dropdownMenuItemClass, dropdownMenuPopupClass } from '@/components/ui/dropdown-menu.styles';
 import type { IconName } from '@/components/icon/icons';
+import { MESSAGE_IMAGE_EXPORT_EXCLUDE_ATTRIBUTE } from '../message/imageExport';
 import { getMermaidViewerController } from './mermaidViewer';
 
 // ---------------------------------------------------------------------------
@@ -581,6 +582,7 @@ const decorateLinks = (root: HTMLElement, ctx: DecorateContext): void => {
     const faviconUrl = getExternalFaviconUrl(href);
     if (faviconUrl) {
       const favWrap = document.createElement('span');
+      favWrap.setAttribute(MESSAGE_IMAGE_EXPORT_EXCLUDE_ATTRIBUTE, 'true');
       favWrap.className =
         'mr-1 inline-flex size-[18px] items-center justify-center rounded border border-[var(--border)] bg-[var(--interactive-hover)] align-middle';
       const img = document.createElement('img');
--- a/packages/ui/src/components/chat/message/MessageBody.tsx
+++ b/packages/ui/src/components/chat/message/MessageBody.tsx
@@ -57,6 +57,7 @@ import { useProviderLogo } from '@/hooks/useProviderLogo';
 import { getAgentColor } from '@/lib/agentColors';
 import { isCapacitorMobileApp } from '@/apps/mobileNativeChrome';
 import { WorktreeRequiresGitRepositoryError } from '@/lib/worktrees/worktreeCreate';
+import { cloneMessageImageExportSource } from './imageExport';
 
 
 const CONTAIN_LAYOUT_STYLE = { contain: 'layout' as const, transform: 'translateZ(0)' };
@@ -1502,7 +1503,7 @@ const AssistantMessageBody = React.memo(({
                     display: inline-block;
                 `;
 
-                const clone = originalElement.cloneNode(true) as HTMLElement;
+                const clone = cloneMessageImageExportSource(originalElement);
                 clone.style.cssText = `
                     ${computedStyle.cssText}
                     transform: none;
--- a/packages/ui/src/components/chat/message/imageExport.test.ts
+++ b/packages/ui/src/components/chat/message/imageExport.test.ts
@@ -1,0 +1,57 @@
+import { describe, expect, test } from 'bun:test';
+import { Window } from 'happy-dom';
+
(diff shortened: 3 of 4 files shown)


ChangeHow:
/oc-review Re-evaluate the NEEDS_EVIDENCE verdict using the deterministic dependency path and pre-fix Chromium reproduction below. Treat export success/failure as nonvisual behavior; do not require a screenshot or recording if the changed contract is proven more directly.

Additional evidence for the current HEAD:

- The implementing session reproduced the failure in real Chromium before the fix. A cross-origin mock returned a valid favicon without `Access-Control-Allow-Origin`. The normal `` displayed, while `htmlToImage.toPng()` rejected with `REJECTED: [object Event]`. DuckDuckGo's favicon response likewise returned `200 image/x-icon` without `Access-Control-Allow-Origin`.
- The exact locked dependency is `html-to-image@1.11.13`. Its pinned upstream source is deterministic: [`resourceToDataURL`](https://github.com/bubkoo/html-to-image/blob/995728147e4459a58188af51b5fcc3f8a7010207/src/dataurl.ts#L82-L110) catches the rejected CORS fetch and returns `options.imagePlaceholder || ''`; [`embedImageNode`](https://github.com/bubkoo/html-to-image/blob/995728147e4459a58188af51b5fcc3f8a7010207/src/embed-images.ts#L53-L83) assigns that empty value to the cloned image and rejects its `error` event when no `onImageErrorHandler` is provided. This resolves the prior review's residual hypothesis that 1.11.13 might tolerate the failed fetch.
- The focused tests prove the post-fix boundary: Markdown-generated favicons receive the export-exclusion marker; the clone passed onward for export contains no DuckDuckGo image; message content and content images remain; and the live source DOM is unchanged.
- The ablation results show both changed steps are necessary. Removing the marker causes the marker test to fail. Keeping the marker but removing clone cleanup leaves the remote favicon in the export tree and causes the clone test to fail.

A recording would show one successful click, but it would not prove the relevant contract more strongly than these assertions and the pinned dependency source. Please evaluate the evidence requirement from this path rather than from the PR body's explicit note that a post-fix Chromium run was not performed.

ChangeHow:
@btriapitsyn Could you please review this PR and merge it when ready? The automated re-review is PASS and the branch is conflict-free. The checks workflow still shows a failure in its Tests step, so that remains to be resolved before merging. Could you also advise on that check? Thanks!

PR #3454, opened 2026-09-09
fix(worktrees): refresh sidebar after external changes

### LINKED ISSUE #1817: [Bug] New worktree created in mini chat dont display in the main window's sidebar

- Open a new mini chat in your project
- Create a new worktree in the mini chat and send messages
- Change to the main window
- The worktree created in the previous operation dont display in the main window's sidebar, until you restart or reload webview in your openchamber

### Where does it happen?

Desktop (macOS)

### Version (if known)

1.13.2

_No response_

### Logs (optional)

```shell

```

AmanTahiliani:
## Intent

Keep worktree topology current when external tooling/terminal (`git worktree add`) or another OpenChamber window creates or removes a linked worktree. The server now watches Git's shared worktree metadata, sends a repository-scoped control event, and refreshes only the affected repository in web, desktop, and hosted-mobile UI state.

Fixes #1817
Addresses #3455

## Non-goals

- No periodic or focus-based polling.
- No VS Code runtime change; that surface does not use the OpenChamber server control stream.
- No changes to worktree creation, deletion, or Git command semantics.

## Affected surfaces

- `packages/web`: watches canonical Git common directories after the first worktree listing, emits `openchamber:worktree-changed`, bounds watcher resources, and disposes them during shutdown.
- `packages/ui`: handles the control event in desktop/web and hosted mobile, bypasses the 30-second listing cache for the affected repository, and reconciles repository topology without dropping unrelated projects.
- Git worktree-list failures now remain failures instead of becoming authoritative empty lists. Non-repositories still return an authoritative empty list.
- VS Code is unaffected because it has a separate runtime boundary and no server control SSE connection.

## Repository guidance

|---|---|---|
| Root `AGENTS.md` and `openchamber-change-discipline` | This changes source, runtime lifecycle, and shared contracts. | Kept ownership in web Git/runtime modules and shared UI modules, added focused tests, ran package and repository checks, and left unrelated worktree changes untouched. |
| `ui-api-decoupling` | Shared UI consumes an OpenChamber-owned server event. | Reused the existing control-event parser and runtime fetch boundary instead of calling server details directly from components. |
| `relay-transport` | The change adds an SSE control event. | Reused `/api/openchamber/events`, kept client cleanup explicit, and added shutdown disposal. |
| `sync-state-invariants` | The event refreshes cached repository topology. | Force-refreshes only the affected repository, preserves prior topology on failure, and partitions authoritative results across registered projects. |
| `performance-engineering` | File watchers and topology refreshes can become hot or unbounded. | Debounces metadata changes, deduplicates by canonical common directory, caps watchers and tracked directories, and does no idle polling. |
| `packages/web/server/lib/git/DOCUMENTATION.md` and `packages/ui/src/components/session/sidebar/DOCUMENTATION.md` | These documents own the changed watcher and sidebar synchronization contracts. | Updated both with watcher lifecycle, failure behavior, and event-driven refresh invariants. |

## Validation

| Check | Result |
|---|---|
| Focused UI tests for control-event parsing, repository replacement, menu refresh, and create/event deduplication | 45 passed. |
| Focused Git route and graceful-shutdown tests | 12 passed. |
| Real temporary-repository watcher integration test | Passed for external worktree creation and removal metadata changes. |
| `bun run lint` | Passed with one pre-existing `react-hooks/exhaustive-deps` warning in `MobileChangesSurface.tsx`. |
| `bun run build:web` | Passed. |
| Server JavaScript syntax checks | Passed. |
| `bun run dead-code` | Completed; reported existing repository findings. |
| `bun run test` | Scripts, UI, VS Code, and Electron passed. Web completed with 2,083 passed, 1 skipped, and 3 unrelated failures: two existing post-checkout-hook tests in `git/service.test.js` and one environment-path test in `opencode/env-runtime.test.js`. |
| `bun run type-check` | Blocked by duplicate installed CodeMirror versions (`@codemirror/language` 6.12.2 and 6.12.4) in `languageByExtension.ts`; mobile and Electron type-checks passed. |

## Visual evidence

No screenshot is attached because this changes when existing worktree rows appear, not their rendering or layout. The behavior is covered by the control-event/UI tests and the real Git watcher integration test. A local server check also confirmed that `/api/git/worktrees` reported a worktree created after server startup; the available browser bundle was older than this branch and was not used as UI evidence.

## Risks and failure behavior

- Filesystem watcher behavior varies by platform. Watcher errors remove the failed entry so a later authoritative list can establish a new watcher.
- Resource growth is bounded to 50 canonical repositories and 100 requested directories per repository. Least-recently-used watcher entries are disposed.
- Events are debounced and repository-scoped. They force a fresh list for that repository without refreshing unrelated projects.
- A transient Git listing failure returns an error and leaves the current UI topology intact. It cannot erase unrelated or previously known worktrees.
- Shutdown explicitly closes all watcher handles.


yulia-ivashko:
Hi @AmanTahiliani, thanks for this one. The client side in particular was well thought out: the repository-scoped event, the forced refresh that leaves unrelated projects alone, and the dedup in `createWorktree` all held up under review, and the failure-contract fix (a listing error is no longer an empty list) was a real catch.

We ended up shipping a different implementation in #3474, which closes both #1817 and #3455, so I'm closing this one.

The part we could not take was the detection source. Watching the common `.git` directory means a watcher per repository, and resolving the watcher key runs two extra `git rev-parse` processes before every worktree listing, even when the watcher already exists. On macOS that is cheap. On Windows a git process is slow to start and gets scanned by Defender each time, so a sidebar with a few dozen projects would feel it on every startup and every refresh. We talked this through with @btriapitsyn, who pointed out exactly that, and the conclusion was that cost here has to scale with what the user is doing, not with how many projects they have registered.

So instead of a watcher, the server now emits the `worktree-changed` event from its own create and remove routes, and it checks the mtime of `.git/worktrees` during git status and worktree listing requests that clients already make. That covers the mini chat case, worktrees created by an agent mid-turn (a completed bash tool already triggers a status refresh on the client), and worktrees added from a terminal the next time the repository is looked at. Nothing runs while the app is idle. Your client-side approach and the event shape carried over almost as is.

Thanks again for the careful work and for the detailed writeup, it made the tradeoff easy to see.

AmanTahiliani:
Sounds good! Thanks @yulia-ivashko. The explanation makes sense as well. I must admit I had not thought of the additional penalty of the system on windows machines. Appreciate you for taking the time to explain that well. 🙌

PR #3456, opened 2026-09-09
fix(chat): prevent model and variant leakage across pinned agents

### LINKED ISSUE #3391: [Bug] Agent settings keep getting messed up, unsure of how

As an example, i'll have my build agent set to 5.6 sol low, and my build - gemini agent set to 3.8 high.
I'll switch from build (main) to build gemini, and it shows 5.6 sol low. This happens all the time, and did NOT happen when using opencode cli directly.

### Where does it happen?

Desktop (macOS)

### Version (if known)

1.22.2

_No response_

### Logs (optional)

```shell

```

maxiedaniels:
## Intent

Fixes an issue where switching between primary agents with pinned models in their configuration files (or sending a message with one pinned agent) would cause that model and its thinking-level variant to leak into other pinned agents and get persisted as unintended session-level overrides.

Fixes #3391.

## Non-goals

- Altering the behavior of issue #2531: switching between agents *without* pinned models (e.g. built-in build and plan modes) continues to preserve user-selected manual models across mode switches.
- Changing how explicit per-agent model or variant overrides are selected by the user via the dropdown controls.
- Introducing a schema migration for existing local storage files: ambiguous previously persisted overrides are avoided moving forward but are not rewritten.

## Affected surfaces

`packages/ui/src/stores/useConfigStore.ts`:
- Agent switches now capture the previous agent before state mutation and do not mark `selectionSource: "manual"` unconditionally. Applying an agent-pinned model or settings default explicitly marks `selectionSource: "auto"`.
- In `setAgent()`, the fallback path that preserves the active model across unpinned agent mode switches (issue #2531) is prevented from running when the target agent configures its own pinned model or when the active model was inherited from the previous agent's pinned model.
- Setting a variant override via `setCurrentVariantOverride()` no longer marks the model `selectionSource` as `"manual"`.

`packages/ui/src/components/chat/ModelControls.tsx`:
- Persisted session model selections now hydrate the session once. Live agent switches are subsequently owned by `setAgent()` and are not overwritten by historical session selections.
- The message-echo preserve path no longer marks restoration complete before the saved agent can be applied through `setAgent(savedAgentName)`.
- When no saved agent model override exists for a saved agent, hydration respects the saved agent's pinned model rather than clobbering it with a session-level fallback.
- Draft variants inherit the selected pinned agent before the global settings default and remain inherited rather than becoming explicit overrides.

`packages/ui/src/sync/session-ui-store.ts`:
- Message sends and draft session materialization save per-agent model overrides only when the user explicitly selected a model (`selectionSource === "manual"`). Inherited agent defaults no longer become sticky overrides merely because a message was sent.

## Repository guidance

|---|---|---|
| OpenChamber Change Discipline | Changes shared UI state and observable agent-switch behavior | Keeps the fix in existing state owners, adds focused regression coverage, and preserves the documented unpinned-agent behavior |
| Sync State Invariants | Session hydration and send configuration must not overwrite newer live choices | Hydration completes once, live agent changes remain authoritative afterward, and sends persist only explicit user selections |
| Store documentation | Model, variant, agent, and selection provenance are owned by the config/session stores | The change preserves those ownership boundaries instead of repairing leaked values in individual controls |

## Visual evidence

- [Before: pinned model and effort leak across agents](https://github.com/user-attachments/assets/5373615d-deba-47f7-a909-a1640c601b12)
- [After: each pinned agent retains its own model and effort before and after sending](https://github.com/user-attachments/assets/d92b7dbb-04a2-4552-8b8d-7797f9983325)

The after recording cycles between pinned agents, sends a message, and cycles again. Each agent retains its configured model and effort.

## Risks and failure behavior

- Existing incorrectly persisted overrides are not migrated because stored data does not identify whether an override was explicit or leaked. This change prevents new leakage without guessing at user intent.
- Explicit user-selected session and per-agent overrides remain supported and continue to win over inherited defaults.
- Unpinned build/plan switching continues to preserve manual choices as required by #2531.
- If a saved model references a provider that has not loaded yet, hydration waits rather than replacing it with an unrelated fallback. The saved agent still receives one restoration attempt when agents become available.

## Validation

| Check | Result |
|---|---|
| `bun test packages/ui/src/components/chat/ModelControls.variant.behavior.test.tsx` | 14 passed |
| `bun test packages/ui/src/stores/useConfigStore.test.ts` | 57 passed |
| `bun test packages/ui/src/sync/__tests__/issue-2039.test.ts` | 11 passed |
| `bun run --cwd packages/ui type-check` | Passed |
| `bunx oxlint` on affected UI files | No new findings; existing findings are unchanged from the previous PR head/upstream baseline |
| Manual Electron verification | Before/after recordings attached above |

Running the three focused files in one Bun process exposes existing cross-file `mock.module` pollution; the same upstream tests fail in that combined invocation. Each owning test file passes through the repository's isolated-test pattern.

### DIFF
--- a/packages/ui/src/components/chat/ModelControls.tsx
+++ b/packages/ui/src/components/chat/ModelControls.tsx
@@ -650,6 +650,7 @@ export const ModelControls: React.FC<ModelControlsProps> = ({
 
     const prevAgentNameRef = React.useRef<string | undefined>(undefined);
     const latestLoadedUserChoiceRestoreRef = React.useRef<string | null>(null);
+    const restoredSessionSelectionRef = React.useRef<string | null>(null);
 
     const currentSessionDirectory = currentSessionId ? getDirectoryForSession(currentSessionId) : undefined;
     const hasRenderableCurrentSessionSnapshot = useSessionRenderable(
@@ -897,6 +898,7 @@ export const ModelControls: React.FC<ModelControlsProps> = ({
                 );
             }
             latestLoadedUserChoiceRestoreRef.current = restoreKey;
+            restoredSessionSelectionRef.current = currentSessionId;
             return;
         }
 
@@ -939,6 +941,7 @@ export const ModelControls: React.FC<ModelControlsProps> = ({
         }
         saveSessionModelSelection(currentSessionId, latestLoadedUserChoice.providerID, latestLoadedUserChoice.modelID);
         latestLoadedUserChoiceRestoreRef.current = restoreKey;
+        restoredSessionSelectionRef.current = currentSessionId;
 
     }, [
         currentSessionId,
@@ -960,6 +963,13 @@ export const ModelControls: React.FC<ModelControlsProps> = ({
     React.useEffect(() => {
         if (!currentSessionId) {
             latestLoadedUserChoiceRestoreRef.current = null;
+            restoredSessionSelectionRef.current = null;
+            return;
+        }
+
+        // Persisted selections hydrate a session once. Live agent changes are
+        // resolved by setAgent and must not be overwritten by session history.
+        if (restoredSessionSelectionRef.current === currentSessionId) {
             return;
         }
 
@@ -985,13 +995,19 @@ export const ModelControls: React.FC<ModelControlsProps> = ({
                     if (result === 'provider-missing') {
                         return 'waiting';
                     }
-                } else if (currentAgentName !== savedAgentName) {
-                    setAgent(savedAgentName);
+                } else {
+                    const savedAgent = agents.find((agent) => agent.name === savedAgentName);
+                    if (currentAgentName !== savedAgentName) {
+                        setAgent(savedAgentName);
+                    }
+                    if (savedAgent?.model?.providerID && savedAgent.model.modelID) {
+                        return 'resolved';
+                    }
                 }
             }
 
             if (savedSessionModel) {
-                const result = tryApplyModelSelection(savedSessionModel.providerId, savedSessionModel.modelId, savedAgentName || currentAgentName || undefined);
+                const result = tryApplyModelSelection(savedSessionModel.providerId, savedSessionModel.modelId, savedAgentName ?? undefined);
                 if (result === 'applied') {
                     if (savedAgentName && currentAgentName !== savedAgentName) {
                         setAgent(savedAgentName);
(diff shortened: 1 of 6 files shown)


maxiedaniels:
Evidence:
(before)
https://github.com/user-attachments/assets/5373615d-deba-47f7-a909-a1640c601b12

(after)
https://github.com/user-attachments/assets/d92b7dbb-04a2-4552-8b8d-7797f9983325

yulia-ivashko:
Thanks for this. I ran the branch with three agents pinned to different models and the #3391 repro is fixed.

I pushed one commit on top. Removing the manual flag from `setAgent` fixed the leak, but that flag also kept a picked agent from being reset when the config reloads. After a reload the composer showed Plan with Build's model. The agent now has its own flag, `agentSelectionSource`, so the pick survives reloads and a pinned model still never gets saved as an override.

The Defaults page also no longer replaces the model of an open chat whose agent pins its own. Two store tests and a docs paragraph are included.

yulia-ivashko:
Live run for the two maintainer commits, since the PR body predates them. Web dev build on macOS, isolated data directory, a project with three primary agents pinned to different models: Build on Big Pickle, Gem on MiMo, Plan on GPT-5.5 Fast.

On `14890c15f`, before the commits, I picked Plan in a session and reloaded the agent list from the app's config store. The composer showed Plan with Big Pickle.

On `e3009a727`:
- Sent a message with Gem, then cycled the agents. Each one kept its own model, and the session stored no model override for Plan or Build.
- Picked Plan, pressed Reload OpenCode in Settings. The chat stayed on Plan with GPT-5.5 Fast.
- Picked Plan, set the default model to Big Pickle in Settings. The chat stayed on Plan with GPT-5.5 Fast.
- Opened a new draft. It started from the defaults, Build with Big Pickle.
- Reloaded the page and reopened the session. Gem and MiMo came back.

`c454e9ce3` closes the effort-only finding from the bot review. I did not run that one live. The new store test fails without the guard, the effort comes back `undefined`, and passes with it.

Checks on `c454e9ce3`: `useConfigStore.test.ts` 67 passed, `ModelControls.variant.behavior.test.tsx` 17, `issue-2039.test.ts` 17, `DefaultsSettings.test.tsx` 1, plus `type-check` and `lint` for `packages/ui`. The recordings in the PR body still show the send and cycle path, which these commits do not change. Electron, VS Code and mobile were not run separately. The change lives in the shared store.

PR #3457, opened 2026-09-09
One identity per repository, and GitLab as a first-class provider

deatheros:
## Intent

A repository used to carry two separate, separately-configured things: a provider account for the API, and a transport credential per remote for Git itself. Neither one settled who the commit was authored by. The three lived in three different places in the UI, and the person had to keep them consistent by hand.

This PR replaces that with **one identity per repository**. An identity is an account, an auth method, and a signature. You pick one when you add a repository, when you clone, or from the chip on the Git panel, and OpenChamber applies it everywhere that repository talks to Git: the commit signature, the transport grant for the bound remote, checkout hydration, the agent's own `git`, and the mobile Changes surface.

What changes for the user:

- The Git panel names the identity a repository is signed as, instead of showing a binding strip and a separate transport editor.
- Add, clone, and the configure dialog offer the same identity menu. The add screen proposes the identity the repository is already signed as.
- Identities from earlier releases keep working. They are labelled "signature only, from an earlier release" and still sign commits; they just do not carry an account or transport.
- The System identity is an explicit choice, meaning "use this machine's own Git configuration". Choosing it asks for confirmation once, because it hands the transfer to credentials OpenChamber did not verify.
- GitLab is a first-class provider: OpenChamber ships its own OAuth application for gitlab.com, self-managed instances are discovered from the credential store, and GitLab has a built-in integration card next to GitHub.
- The agent's shell gets managed push, pull and fetch, and refuses raw Git transfers where a managed action exists.
- Checkout hydration asks for an identity per endpoint rather than a transport.

## Non-goals

- The persisted binding contract keeps its authority model. Endpoint-exact grants, compare-and-swap by revision, and fail-closed readiness are unchanged; only the UI that produces them and the identity that fills them in are new.
- A remote the person did not name is still never granted. Applying an identity extends it to remotes that already have a ready grant and fit the identity's host, and the configure dialog offers an explicit "answer as ..." button for the rest. Nothing is auto-granted.
- No changelog entry. Release notes are the maintainer's release-time work.
- `../opencode` is untouched.
- The public error contract for clone planning is unchanged, so cloning into a path that already exists still reports the generic "Invalid Git network operation request". See Risks.

## Affected surfaces

| Area | What moved |
|---|---|
| `packages/ui` (191 files) | Identity model and pickers, Git panel chip and menu, add/clone screens, configure-repository dialog, hydration editor, `applyIdentity`, `baseBranch` qualification, mobile Changes surface, 12 locale dictionaries |
| `packages/web` (164 files) | Source-control binding service and contract, GitLab provider module and OAuth app, Git identity storage, agent-tool Git boundary, audit records, credential helper reachable behind a UI password |
| `packages/vscode` (13 files) | Webview projects remotes as a ready System-transport binding read; still persists no bindings and exposes no binding mutations |
| `packages/mobile` (1 file) | Capacitor shell picks up the shared mobile Changes surface change |
| Root | One new dependency in `packages/web` |

Runtimes covered: web, Electron, VS Code webview, hosted mobile, Capacitor mobile. Electron and hosted mobile go through the same server routes as web, so they need no separate identity path; VS Code is deliberately read-only here.

Persisted and external contracts touched: `source-control-bindings.json` (version 2, fail-closed on any other version), the Git identities store, the source-control audit schema, and the `/api/source-control/*` and `/api/gitlab/*` routes. GitLab token rejection now carries a stable `INVALID_TOKEN` code so the UI can say the token was refused instead of "Operation failed".

New dependency: `@gitbeaker/rest@43.8.0` in `packages/web`, for the GitLab REST API.

## Repository guidance

|---|---|---|
| `AGENTS.md` always-on constraints | Every commit on the branch | No secrets, tokens or pairing credentials are logged; bindings stay references-only and endpoint metadata stays redacted; `../opencode` untouched; `changelog/` untouched |
| `openchamber-change-discipline` | Module contracts, cross-workspace contracts and persisted state all change | Each risk class validated at its own level: workspace-wide type-check, lint and build for the shared contracts, focused suites for behaviour, `dead-code` after files were added and removed, Docker QA for runtime behaviour |
| `ui-api-decoupling` | Shared UI reaches Git and source control through `RuntimeAPIs` | New intents (`configureTransportBinding`, `configureAuxiliaryBinding`, remote removal, confirmed reset) go through the runtime API surface; no runtime-specific fetch was added to shared UI |
| `sync-state-invariants` | `repository-binding.ts` owns binding reads, their lifecycle and their invalidation | Mutation scopes still pin runtime generation, directory, repository ID and revision; late reads and obsolete results are rejected; a failed read keeps its last complete snapshot as stale rather than publishing empty success |
| `locale-ui-patterns` | New user-facing copy on the panel, pickers, dialogs and settings | Every new key exists in all 12 dictionaries with a real translation, not English placeholders; `t(...)` stays inside render scope |
| `settings-ui-patterns` | Git Settings gained the identity list and lost the transport editors; GitLab gained an integration card | Identity management lives in Git Settings; the active repository's identity lives in the Git panel, not in Settings search |
| `theme-system` | Panel chip, menu, dialog and card styling | Tokens and existing primitives only; light and dark verified in the evidence below |
| `communication-style` | All of the new UI copy, and this description | Plain sentences, no em dashes, mechanism named rather than described in feelings |
| `packages/web/server/lib/source-control/DOCUMENTATION.md` and `packages/ui/src/lib/source-control/DOCUMENTATION.md` | These modules own the binding contract and its UI authority | Both updated where the implementation changed their truth, along with the `git`, `github`, `gitlab`, `agent-tool`, `walkthrough`, `opencode` and `fs` module docs and `packages/web/README.md` |

## Validation

Run on the final HEAD of this branch, after the latest merge of `origin/main`.

| Check | Result |
|---|---|
| `bun run type-check` | Pass, all five workspaces |
| `bun run lint` | Pass, all five workspaces |
| `bun run test` | Pass. scripts 4/4, UI 469/469, VS Code 42/42, Electron 23/23, web 243 files / 3398 tests (1 skipped) |
| `bun run build` | Pass, including the VS Code webview bundle and the staged mobile assets |
| `bun run dead-code` | Inspected. Report is non-blocking; the remaining entries are pre-existing |
| `bunx oxlint` on added and rewritten files | Clean |
| Manual QA, two clean instances | Two Docker containers with empty data directories, one built from this branch and one from `origin/main`, plus a self-managed GitLab container for the local-instance fixtures. Journeys exercised: add a directory, clone, switch identity from the panel, publish, sync, checkout hydration, linked worktree, repository with two remotes, repository with no remote |

Not verified: packaged desktop installers, iOS and Android device builds, and relay or tunnel transport. Nothing in the diff is specific to those paths, but they were not exercised.

## Visual evidence

Captured on the two clean instances described above, in a headless browser at 1440x950 (mobile frames at a phone viewport). Personal email addresses are masked; the GitHub and GitLab account handles are the maintainer's own and are shown deliberately. The frames are committed under [`.github/pr-evidence/`](https://github.com/openchamber/openchamber/tree/4b5bac27d09bbdc0aa9c36b96fe76ecf1783ce02/.github/pr-evidence) beside the evidence from earlier changes.

| Surface | Before (`origin/main`) | After (this branch) |
|---|---|---|
| **Git panel, dark** | Git panel, dark before<br>A branch name and a small remote glyph. Nothing says who the repository is signed as, and the action reads "sync". | Git panel, dark after<br>An identity chip names what the repository is signed as, and the action reads "publish". |
| **Git panel, light** | Git panel, light before<br>The same absence in the light theme. | Git panel, light after<br>The same chip in the light theme. |
| **Add project dialog** | Add project dialog before<br>The dialog asks for a folder and nothing else. | Add project dialog after<br>The identity picker sits above the path, pre-filled with the identity the repository is already signed as. |
| **Git settings, identities** | Git settings, identities before<br>Three identities, each tagged `ssh`, each an email address. No System identity, no accounts. | Git settings, identities after<br>The System identity, the three earlier ones marked "Signature only, from an earlier release", and the account-backed identities under them. Email addresses masked. |
| **Integrations** | Integrations before<br>GitHub and Linear. | Integrations after<br>GitHub, GitLab and Linear, each reporting how many accounts it has connected. |
| **Mobile Changes surface** | Mobile Changes surface before<br>Branch and a plain sync action. | Mobile Changes surface after<br>The same answer as the desktop panel: the identity glyph beside the branch, and publish. |

Three surfaces have no "before" because they do not exist on `main`:

| Surface | This branch |
|---|---|
| **Identity menu** | Identity menu<br>One menu lists the System identity, the three signature-only identities carried over from an earlier release, a self-managed GitLab identity, and the four account-backed identities. Account rows say which host they answer for. |
| **Configure repository** | Configure repository dialog<br>What one identity cannot settle: the agent shell switch, the repository's other addresses with an explicit "answer as ..." button each, and checkout hydration. |
| **Git settings, light theme** | Identity list, light theme<br>The same identity list in the light theme. |

## Risks and failure behavior

- **Identities from earlier releases.** They survive as signature-only and keep signing commits. The binding store itself is new here and accepts version 2 only: a file with any other version, or one that fails the strict parse, is rejected with `INVALID_SOURCE_CONTROL_BINDINGS` and left untouched rather than converted or replaced. They cannot grant transport, which is the intended outcome, but a person who relied on one for a push will now be asked to pick an account.
- **A second remote is not granted by itself.** A repository with `origin` and a fork answers only for the remote its identity was applied to. The configure dialog names the other addresses and offers to answer for them. This is deliberate, and it is the part of the design most worth a second opinion.
- **Adding a remote flips the binding to `needs-attention`.** The repository's config revision no longer matches the binding's. Individual grants stay ready, and choosing the identity again clears it. The tooltip now says so instead of only "Needs attention".
- **A repository with no remote** previously had its identity applied silently on first read, which could write a signing key that did not exist and break the first commit. It now gets the picker, and choosing an identity writes the signature alone: no binding, no signing key, no `gpgsign`.
- **Mobile does not qualify the comparison base.** `qualifyBaseRef` runs in the desktop diff view. The mobile Changes surface still passes the raw base, so a base branch that exists only on a second remote can still produce an empty comparison there.
- **Cloning into an existing directory** still reports "Invalid Git network operation request". The server knows the real reason. Fixing it means changing the public error contract, which is outside this PR.
- **New dependency.** `@gitbeaker/rest` is added to `packages/web` for GitLab. It is server-side only and does not reach any client bundle.
- **Pull request review from main, adapted.** Main added reviewing a published pull request diff in Changes and walkthrough. Here it reads through the bound GitHub context: the picker lists through `SourceControlAPI.changeRequestsList`, and `/api/walkthrough/pr-diff` validates the same read context as the other walkthrough routes. A named source repository is honoured only inside the bound repository's network, the repository and its upstream, which is what the bound list already reads. Anything outside it is refused with `409` before any GitHub request.
- **A worktree with no bootstrap record reads as ready.** This follows main's new proxy gate, which waits for worktree checkout before forwarding to OpenCode. This branch used to treat a missing record as a repair blocker, and main's gate turned that into a `503` for every ordinary repository. The trade-off is deliberate: a worktree whose record was lost, for example to capacity eviction, is now treated as populated. A pending record with no live bootstrap and an unreadable store still fail closed.
- **Locks and recovery are unchanged.** The manual orphan-lock recovery contract in the source-control module still applies, and this PR does not add automatic lock stealing.


deatheros:
Both findings are addressed on `4b5bac27d`.

**1. Visual evidence.** The fifteen frames are now committed under `.github/pr-evidence/`, beside the evidence from the interface-scale change, and embedded in the Visual evidence table above. Six surfaces show a before and an after; three (the identity menu, the configure-repository dialog, the light-theme identity list) have no before because they do not exist on `main`. They were captured on the two clean instances the Validation table describes, not re-used from an earlier HEAD.

**2. The v1 conversion claim.** Correct, and thank you for catching it. `parseBindingStore` accepts `version === 2` and nothing else; a different version is rejected with `INVALID_SOURCE_CONTROL_BINDINGS` and the file is left on disk untouched, which `binding-storage.test.js` pins. Rewrote the three passages: the `binding-storage.js` and `binding-contract.js` entries in `packages/web/server/lib/source-control/DOCUMENTATION.md`, the pointer in `packages/ui/src/lib/source-control/DOCUMENTATION.md`, and the matching line in the Risks section here. They now say the store fails closed and never converts, repairs or replaces an unreadable file.

deatheros:
CI note: the one remaining failure on this branch is not from this branch.

`server/lib/relay/flow-control.test.js > small HTTP response overtakes bulk output without losing bytes or reconnecting` fails with `expected 256546 to be less than 196608`. The same test fails with the **same byte count** on [`feat/combined-git-changes`](https://github.com/openchamber/openchamber/actions/runs/34392998026), a branch that shares none of this work and runs 195 test files to this one's 238. It passes locally on macOS.

Why it is machine-dependent: `peakBytes` measures how much the harness lets pile up before its own drain releases it, and that drain is a `setInterval(..., 5)` handing out 4 KiB of credit per tick. On a contended runner the drain never keeps up, so the peak settles on the scheduler's own window rather than on anything the test controls, and 256546 is a constant rather than jitter.

Two clone-ownership failures from the previous run are fixed and green: `fileIdentity` now folds creation time in, because Linux reuses an inode number when a directory is removed and recreated in the same parent, so `dev:ino` alone could not tell a replaced clone directory from the one the operation created. That was a real gap on Linux, not a test problem.

I have deliberately not touched the relay assertion here. Loosening a backpressure bound to turn a red check green would hide the thing the test exists to prove, and the fix belongs in its own change.

deatheros:
Both non-blockers are addressed on `39ca0c062`.

**1. Birthtime fallback.** You were right, and it was worse than a comment problem: on a mount without a real creation time Node fills the field from the change time, and a directory's change time moves as the clone writes its own checkout, so every clone would have hit the 409. Rather than document that, the operation now establishes the answer for itself. Each directory it creates is changed once on purpose (a subdirectory made and removed), then re-stat'ed. If the reported creation time sat still through that change, it is real and ownership uses it; if it moved, or if the filesystem's clock is too coarse to tell, ownership falls back to device and inode instead of to a false conflict. The same treatment covers the submodule checkout directory, which had the identical shape.

Measured, not reasoned: with a stub that mirrors the no-birthtime behaviour, on Linux the guard reports `birthtime trusted=false, survives our own writes=true, detects replacement=false`, and on a real filesystem `trusted=true, survives=true, detects=true`. So the exotic-mount case returns exactly to the previous `dev:ino` behaviour rather than breaking, and the Linux inode-reuse gap stays closed everywhere else. `network-operations.test.js` 152/152, full web suite 238 files / 3327 tests.

**2. Stale line in Affected surfaces.** Dropped; it now reads "version 2, fail-closed on any other version", matching the two module docs and the Risks section.

The one red check remains `relay/flow-control.test.js`, which fails identically on an unrelated branch. Details in the comment above.

deatheros:
Merged `origin/main` (twelve commits) and resolved the one conflict, in `packages/ui/src/stores/DOCUMENTATION.md`. Both sides had added a paragraph at the same spot: main's note on diff-prefetch concurrency and this branch's note on `useGitIdentitiesStore`. Both are kept. Main also reopened a `### useGitHubPrStatusStore.ts` heading that this branch had already replaced with the provider-neutral `### Source-control stores` section directly below it, so that heading is not carried over and no content is lost.

Two commits worth pointing at, because they touch files outside this branch's scope:

- `0d7733a2f` repairs the stubs in `issue-1637-2270.test.ts` and `issue-2039.test.ts`. `createSession` now writes the created session into its directory's child store, and directory resolution now reads the global store's entity index, but those two older regression tests stub both with objects that lack `getState` and `entityById`. The four files involved are byte-identical to `origin/main` here and this branch never touched them, so the failure arrives with the merge rather than from it.
- The relay flow-control assertion that was the one red check on this branch is gone: main removed it in `5fcae5446` and moved the credit bound to `downstream-scheduler.test.js` under a controlled clock, which matches what the byte-identical failure across unrelated branches pointed at.

Full suite on the merged HEAD: type-check, lint and build pass; scripts 4/4, UI 459/459, VS Code 42/42, Electron 23/23, web 238 files / 3331 tests. The Validation table is updated to these numbers.

mattv8:
I'd separate repository identity from first-class GitLab support so we can review each change on its own. @btriapitsyn, do you want both capabilities in core, and is there a reason to land them together? The author has been addressing feedback; we need to settle the scope.

PR #3459, opened 2026-09-09
feat(ui): add back session running spinners, but opt-in and performance enhanced

mattv8:
## Intent

Restore a visible spinner for running sessions without bringing back the original always-on animation cost. The existing static dot remains the default. Users can enable a stepped spinner from Appearance, and the choice applies to desktop session rows, header session tabs, collapsed project and folder indicators, the session switcher, and mobile session lists.

The preference is local and persisted. The OpenChamber toggle is authoritative across runtimes, including when the OS requests reduced motion.

## Non-goals

- Do not change the one-second elapsed-turn timer or any live-status subscription.
- Do not change global `.animate-spin`, the switcher's existing default pulse, or the worktree-move spinner.
- Do not sync this local display preference through desktop settings.
- Do not change server APIs, runtime transport, native shell code, or changelogs.

## Affected surfaces

- `packages/ui`: shared indicator component, persisted display preference and v5 to v6 migration, Appearance setting and search entry, stepped CSS animation, desktop header/sidebar and mobile integrations, tests, and locale dictionaries.
- Web, desktop, VS Code, hosted mobile, and Capacitor mobile consume the shared UI behavior. No native mobile or Electron boundary changes are required.
- `scripts/perf`: a fixture variant and measured notes for the stepped transform.
- Persisted contract: `session-display-mode` adds `animatedActivityIndicators`, defaulting to `false`. No external contract changes.

## Repository guidance

|---|---|---|
| Root `AGENTS.md` and `openchamber-change-discipline` | This changes shared source, persistence, UI behavior, and a package-level component contract. | The implementation uses one focused leaf component, keeps runtime entrypoints untouched, adds migration and behavior tests, runs package checks and dead-code analysis, and updates the owning sidebar documentation. |
| `theme-system` | This adds an icon and animation. | It reuses `loader-4` and theme tokens, animates only `transform`, throttles with `steps(30)`, and keeps the explicit local preference authoritative. |
| `settings-ui-patterns` | This adds an Appearance preference and settings search result. | It uses the existing section and checkbox primitives, stable `data-settings-item` anchors, local persistence, and a matching search registry ID. Wide and narrow settings states were checked. |
| `locale-ui-patterns` and `communication-style` | The setting adds visible labels, help text, and aria text. | Four keys have real translations in all 12 settings dictionaries. The help text discloses the Reduce Motion override, and existing localized running and unread status labels are reused by the indicator. |
| `performance-engineering` | The original spinner was removed for sustained frame-rate work. | The dot stays the default. The opt-in spinner follows VS Code's stepped transform, and the fixture records zero style recalculations and layouts. |
| `sync-state-invariants` | Running state comes from live session channels. | The change is display-only. `useGlobalSessionStatus`, collapsed activity selectors, reconciliation, and the timer remain unchanged. |
| `packages/ui/src/components/session/sidebar/DOCUMENTATION.md` and `scripts/perf/DOCUMENTATION.md` | These documents own sidebar activity behavior and animation measurement. | They describe the default dot, opt-in spinner, local-setting precedence, unchanged timer, and measured transform cost. |

## Validation

| Check | Result |
|---|---|
| `bun run --cwd packages/ui test` | Passed, 537 of 537 test files. |
| `bun test packages/ui/src/components/session/SessionActivityIndicator.test.tsx` | Passed, 4 tests. Includes the header session-tab integration guard. |
| `bun run --cwd packages/ui type-check` | Passed. |
| `bun run --cwd packages/ui lint` | Passed. |
| `bun run --cwd packages/electron build:web-assets` | Passed; bundled Electron web assets completed successfully. |
| Electron 43.3 animation regression probe on macOS | The installed stylesheet reproduced advancing animation time with a frozen transform inside `button` and `[role="button"]` ancestors. The rebuilt stylesheet advances the real transform in plain, button, and role-button contexts, including with Reduce Motion enabled. |
| `bunx oxlint` on the review-follow-up component, test, and 12 locale files | Passed. |
| Branch scope check | Merge-base is upstream `56f33fd59`; the spinner-only delta is 8 commits, 31 files, +414/-53. PR #3008 tip `a37d31a7e` is not an ancestor. |
| `bun run dead-code` | Completed. The report contains the existing repository backlog and does not flag `SessionActivityIndicator` or a new export. |
| Animation fixture, 8 elements for 10 seconds | Smooth transform: 0 style recalculations, 0 layouts, 0.07% main-thread busy. Stepped transform: 0 style recalculations, 0 layouts, 0.02% main-thread busy. |
| Repeated 32-element animation fixture | Median main-thread busy was 0.04% smooth and 0.02% stepped. Both remained at 0 style recalculations and 0 layouts. These tiny values are environment-sensitive; the stable result is the zero-recalc, zero-layout transform contract. |
| Browser runtime checks against reviewed spinner-only branch assets | Verified default static dot, enabled `activity-spin 1.5s steps(30)`, desktop row, collapsed aggregate, session switcher, mobile session sheet, visible/persisted Appearance setting, and light/dark narrow/wide layouts. The follow-up reuses the tested indicator in header tabs and extends help text only. |

Not validated: a native iOS or Android build. The mobile check used the hosted mobile app with `?surface=mobile`; this change does not touch Capacitor or native code.

## Visual evidence

These captures use production assets built from the reviewed spinner-only feature head before the review follow-up. The follow-up reuses the same indicator in header session tabs and extends the setting help text; it does not change the captured spinner, setting control, or layouts. The evidence files live on the dedicated `pr-3459-visual-evidence` branch and the links are pinned to commit `e618feeaf5644f94f3f2bb360535356fe65cc35a`.

### Default and opt-in states

Default static dot while the setting is off:

Default static running dot

Wide Appearance pane before enabling the option:

Wide Appearance pane with animated indicators off

Wide Appearance pane after enabling the option:

Wide Appearance pane with animated indicators on

Desktop running row with the opt-in spinner:

Desktop running row with spinner

Stepped spinner motion, captured over one 1.5 second cycle:

Stepped running spinner animation

### Other surfaces

Session switcher:

Session switcher running spinner

Hosted mobile session sheet:

Mobile session sheet running spinner

Narrow dark Appearance pane:

Narrow dark Appearance pane

## Why this compromise

The static dot was never a bug. PR #2480 (commit `4c647ed3`) introduced the original spinner; commit `faa9c243` ("perf(ui): swap the session activity spinner for a dot and a turn timer", Aug 5 2026) removed it on purpose because each spinner repainted a composited layer about 60 times per second for every active session. The replacement was a static `bg-primary` dot plus an elapsed timer updated once per second, applied across desktop, mobile, folders, projects, and collapsed groups. That rationale stays in `SessionNodeItem.tsx` and `SessionActivityDuration.tsx` on this branch.

This PR does not reverse that decision. It changes the animation technique so opting in costs a fraction of what `faa9c243` removed:

- The dot stays the default for every existing and new user. Motion is opt-in per device.
- The spinner copies VS Code's approach from `src/vs/base/browser/ui/codicons/codicon/codicon-modifiers.css`, where all codicon spinners, including Copilot Chat's running indicators and the `codicon-tree-item-loading` tree rows that mirror our sidebar rows, share `animation: codicon-spin 1.5s steps(30) infinite` with the comment "Use steps to throttle FPS to reduce CPU usage".
- The old spinner was `1s linear infinite`: the rotation value changed every vsync, so the compositor produced roughly 60 frames per second per spinner. `steps(30)` over 1.5s changes the value 20 times per second, and the browser skips frame production between steps. The fixture above confirms both variants stay transform-only (0 style recalculations, 0 layouts) with the stepped variant at roughly a third of the smooth variant's main-thread cost.
- The one-second timer and its live-status subscription are untouched. The explicit local toggle controls whether running sessions show the dot or spinner.

If frames drop under load, a stepped spinner reads as deliberately ticky rather than janky, so degraded conditions do not look broken.

## Risks and failure behavior

- Enabling the option adds sustained compositor work for each running indicator. The setting is off by default, the animation is stepped to 20 updates per second, and users can disable it immediately.
- Persisted v5 state migrates to v6 with the option set to `false`; existing display keys are preserved. An already-enabled v6 value remains enabled.
- The setting is off by default. Users who enable it get the stepped spinner across runtimes, including when the OS requests reduced motion; turning it off restores the dot.
- All runtimes share the same leaf component and store. Runtime-specific status subscriptions and native boundaries are unchanged.
- No security, network, server, payment, or data-loss behavior changes.
- Rollback is a normal code revert. Users can also restore prior behavior by turning the option off.


mattv8:
Addressed in `8ba064050`.

- Restored the spinner-only branch directly on upstream `fdef13cbd`. PR #3008 tip `00e72f918` is not an ancestor; the PR delta is now 7 commits and 31 files.
- Header session tabs now render running and unread state through `SessionActivityIndicator`, while AI rename keeps precedence.
- The setting help text in all 12 locales now says the opt-in animation still plays when Reduce Motion is enabled.

Validation: focused indicator tests pass 4/4, UI type-check passes, changed-file oxlint passes, and Electron web assets build successfully. The full UI suite passes 443/444 files; the remaining `event-pipeline.test.js` timing failures reproduce unchanged on pristine upstream `fdef13cbd`.

mattv8:
Fixed the static macOS desktop spinners in `197c5555c`.

The animation clock was running, but a global button-icon rule applied a fixed `translateZ(0)` transform to every SVG except `.animate-spin`. Sidebar session rows receive `role="button"` from dnd-kit, so that fixed transform overrode `.activity-spinner` rotation. The rule now excludes both animated spinner classes.

Verified with Electron 43.3 on macOS using the installed app stylesheet as the failing baseline and rebuilt desktop assets as the passing case. The rebuilt spinner's real transform advances inside plain, native-button, and role-button contexts, including with Reduce Motion enabled. `bun run --cwd packages/electron build:web-assets` also passes.

mattv8:
Addressed at `0dafb719e`.

- Rebuilt the eight spinner commits directly on upstream `db183e2f0`; PR #3008 tip `a37d31a7e` is not an ancestor. The PR delta is 31 files, +415/-55.
- Preserved the new upstream streaming CPU fixes while resolving the CSS and performance-documentation overlaps. The WebKit-only icon-layer rule now excludes `.activity-spinner` without restoring the old Chromium-wide layer hint.
- Verified `bun run --cwd packages/ui test` (516/516 files), UI type-check, UI lint, and `bun run --cwd packages/electron build:web-assets`.
- Updated the PR validation table to match the current HEAD.

mattv8:
Rebased onto upstream `56f33fd59` at `316e62572`.

- Preserved main's new provider-logo animation alongside `activity-spin`.
- Preserved collapsed aggregate permission/question badges and their precedence; only active/unread states use `SessionActivityIndicator`.
- Verified the spinner-only delta remains 8 commits and 31 files, and PR #3008 tip `a37d31a7e` is not an ancestor.
- Passed 537/537 UI test files, UI type-check, UI lint, and Electron web-assets build.
- Updated the PR validation table for this HEAD.

btriapitsyn (approved):
Thanks, Matt! I appreciate keeping this opt-in. I added coverage for the new Timeline views and kept the current status colors.

PR #3462, opened 2026-09-09
fix(chat): preserve triple-click text selection

### LINKED ISSUE #3416: [Bug] Can't triple-click to select whole sentence

On macOS, you can triple click to select a whole sentence.
If I try to do it, the double click menu pops up and I unintendedly add words to notes instead

### Where does it happen?

Desktop (macOS)

### Version (if known)

1.22.2

Image

### Logs (optional)

```shell

```

IbrahimKhan12:
## Intent

Fixes [#3416](https://github.com/openchamber/openchamber/issues/3416). On macOS, the second mouseup of a triple-click can reveal the selection toolbar before the third click. When the selection starts on the first line of a message, the placement clamp can also put the toolbar over that text, so the third click can trigger a toolbar action instead of extending the selection.

This change delays the toolbar during multi-click sequences and places it below the selection when it cannot fit above. A fresh drag also cancels any pending reveal.

## Non-goals

The setting to disable the selection toolbar and the separate `FilePreviewCommentMenu` behavior remain separate work.

## Affected surfaces

This changes the shared desktop selection toolbar in `packages/ui`, used by web, Electron, and the VS Code webview. Mobile keeps its existing bottom-bar placement. No persisted data, routes, package contracts, or external APIs change.

## Repository guidance

|---|---|---|
| Root `AGENTS.md` and `openchamber-change-discipline` | This is executable shared UI code | The change stays local to the owning component and existing positioning module, preserves runtime contracts, and adds focused regression tests for placement |
| `theme-system` | The change affects a UI component | The diff changes only event timing and geometry; existing tokens, controls, icons, and animations remain intact |
| `packages/ui/src/components/chat/message/parts/DOCUMENTATION.md` | It is the nearest chat-message documentation | It covers message part renderers rather than the selection toolbar, so its documented ownership remains accurate |

## Validation

| Check | Result |
|---|---|
| `bun test packages/ui/src/components/chat/message/__tests__/selectionMenuPosition.test.ts` | 12 pass, 0 fail |
| `node scripts/run-isolated-tests.mjs packages/ui/src/components/chat/message` | 15/15 test files passed |
| `bun test packages/ui/src/lib/ime.commentInputs.test.ts packages/ui/src/lib/addSelectionToChat.test.ts` | 15 pass, 0 fail |
| `bunx eslint` on the three changed files | Passed |
| `bunx oxlint` on the three changed files | No new findings; the same five existing `TextSelectionMenu.tsx` findings remain on `origin/main` |
| `bunx tsc --noEmit -p packages/ui/tsconfig.json` | No new errors; the same five existing duplicate-CodeMirror errors remain in `languageByExtension.ts` on `origin/main` |
| `bun run dead-code` | Passed; no changed-file or new-export findings |

## Visual evidence

The animation compares `origin/main` with the amended PR implementation in the same Chromium runtime. The orange ring marks the fixed pointer position. Before the fix, click 2 opens the toolbar under the pointer and click 3 activates it, clearing the selection. After the fix, click 2 keeps the toolbar hidden, click 3 selects the full sentence, and the settled toolbar appears below the text.

Before and after triple-click behavior

## Risks and failure behavior

The 300 ms multi-click delay is a heuristic because browsers do not expose the macOS click interval. It adds latency after a double-click but leaves drag selection on the existing 10 ms path. The placement change keeps slower sequences safe by moving the toolbar away from the selected line. Tall selections can move the toolbar below the selection, with the existing viewport clamp keeping it visible.

## References

- [Issue #3416](https://github.com/openchamber/openchamber/issues/3416)

### DIFF
--- a/packages/ui/src/components/chat/message/TextSelectionMenu.tsx
+++ b/packages/ui/src/components/chat/message/TextSelectionMenu.tsx
@@ -24,8 +24,11 @@ import { collectSelectionOverlayRects } from '@/lib/selectionOverlayRects';
 import {
   DESKTOP_MENU_FALLBACK_HEIGHT_PX,
   DESKTOP_MENU_FALLBACK_WIDTH_PX,
+  DESKTOP_MENU_SELECTION_GAP_PX,
+  DESKTOP_MENU_SIDE_MARGIN_PX,
   getDesktopClampedX,
   getDesktopClampedY,
+  getDesktopSelectionAnchorY,
 } from './selectionMenuPosition';
 
 interface TextSelectionMenuProps {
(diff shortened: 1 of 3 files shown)


yulia-ivashko:
Thank you for digging into this, and for the before/after recording. It made the mechanism obvious. The toolbar landed under the pointer and swallowed the third click. While this PR was open, #3626 landed on main and changed the placement, so the toolbar now goes above the selection or flips below it, never over it. That removes the cause here, because the third click now reaches the text. After a rebase, what would remain of this PR is the 300 ms multi-click delay, which would slow the toolbar on every double-click without fixing anything that still breaks. I'm closing it for that reason. Thanks again for the careful work.

Issue #3277, opened 2026-09-01
[Bug] Text before question tool is still hidden until an option is selected (was claimed fixed in #503)


ouyangjian28:
### What's wrong?

The assistant's text output immediately before a `question` tool call is still hidden. The question options render, but the reasoning/context text the agent printed right before asking is not visible — it only appears **after** I select an option and submit.

This defeats the purpose of the question: I'm asked to make a decision without seeing the context that explains the options, which directly impacts decision quality.

### Steps to reproduce

1. Ask the agent something that makes it call the `question` tool (with a text preamble explaining the choice).
2. While the question form is pending, the preamble text is not rendered in the chat.
3. Select an option → the previously hidden text suddenly appears.

### Environment

- Desktop Web
- Version: 1.22.0

### Context

This is the same behavior reported in #503, which was closed as fixed on 2026-03-13. However it still reproduces on 1.22.0. Notably, in #503 the last reports (2026-03-22/23) already indicated it persisted in the sorted chat view after the fix landed for live view.

### Expected behavior

Text parts emitted before a `question` tool call should be rendered normally (or at least collapsible-but-visible) while the question is still pending, so the user can read the context before choosing an option.

EmberCraze:
I am also seeing this issue. if there is text before a question, it is hidden until i reject och answer the question

ouyangjian28:
Reproduced on current `main` and put up a fix: #3495.

The residual path is live mode's block-commit reveal (`commitStreamedText` holds the tail after the last complete line, and a turn blocked on a pending question never finalizes to release it). The fix releases any text part sealed with `time.end` — the SDK stamps it as soon as the model moves on to the question part — so the preamble renders in full, including its last line, while the question waits. Reproduction measurements (DOM text length 208 vs the API part's 233, the missing 25 chars being exactly the held last line) are in the PR.

ouyangjian28:
Fix resubmitted as #3498 (supersedes #3495, same change with cleaner evidence: before/after screenshots of the pending-question state).

PR #3485, opened 2026-09-10
fix(chat): show a finished text part while its message still runs tools

### LINKED ISSUE #3484: [Bug] OpenChamber is behind on messages, they appear after the next message

I've run OpenChamber for a while, and I found that over time I experience the same bug regularly. For example, if a question is asked, immediately before the question, there will be a text (by openai gpt) describing the last "thought" or reasoning for the question. But the message always seemed.. partial. Then, when I answered the question, the message was completed.

This is odd. But I hooked up OpenCode CLI to the running OpenChamber OpenCode-instance, to see the behaviour side by side, and sure enough, OpenChamber is consistently one message behind. I've captured the video in this issue, where you can see OpenCode on the right and OpenChamber on the left. OpenChamber is consistently one though, or message, or a line of a message, behind.

Its visible in all "Thought:"'s, but around 00:10 there is a text where OpenChamber shows the half message for a long time while OpenCode already has the full message rendered. Only on the next response, does OpenChamber show the full message.

### Where does it happen?

Desktop Web

### Version (if known)

1.23.0

https://github.com/user-attachments/assets/7c693549-b4f4-48ac-bd80-abcdce059637

### Logs (optional)

```shell

```

IbrahimKhan12:
## Intent

Closes #3484.

OpenChamber showed each assistant text one step late. The paragraph before a question or a long tool call stayed hidden, or showed all but its last paragraph, until the next step's message appeared.

Nothing arrived late. Since 1.21.0 the block-level reveal (`commitStreamedText`) holds a paragraph back until its line break arrives, and `AssistantTextPart` ran that reveal whenever the message phase was `streaming` or `cooldown`. OpenCode closes a text part with `time.end` as soon as the model stops writing it, then keeps the message busy for its tool calls. A finished part with no trailing newline therefore had its last paragraph held until `message.time.completed` or `finish === 'stop'`, which land exactly when the next step message shows up.

`AssistantTextPart` now treats a part with `time.end` as not streaming, the rule `ReasoningPart` already applies (#2020). A closed part renders in full at once. A part still receiving deltas keeps the reveal.

## Non-goals

- `ReasoningPart`, `commitStreamedText`, the 100 ms text throttle, and `sorted` mode's deferral of inline text until `finish === 'stop'` are unchanged.
- No sync-layer change. The bot comment on the issue suspected event coalescing and the materialization cooldown; neither was involved.

## Affected surfaces

`packages/ui` only: `AssistantTextPart.tsx`, its new test, and the parts `DOCUMENTATION.md`. Every runtime that renders the shared chat in `live` mode (web, desktop, VS Code, hosted and Capacitor mobile) now shows a finished text part's last paragraph while the message still runs tools. `sorted` mode never runs the reveal for an in-progress message, so it is unaffected. No persisted data, routes, exports, or entrypoints change.

## Repository guidance

|---|---|---|
| `AGENTS.md`: prefer authoritative state over heuristics | The message phase, a heuristic for the part, overrode the part's own `time.end` | `isStreaming` reads the part's `time.end`, which the SDK defines on `TextPart`, before the phase |
| `openchamber-change-discipline` | Source change in one package | One derivation moved and extended, no new abstraction. Focused tests, package type-check and lint, `bun run dead-code` for the added file |
| `performance-engineering` | `AssistantTextPart` is on the streaming render path | No new subscriptions, effects, or allocations. For a closed part the throttle flushes and `MarkdownRenderer` leaves live mode sooner, so less work |
| `sync-state-invariants` | The symptom looked like sync lag | Traced the live part overlay (`useSessionPartsForMessages`); parts arrive on time, no sync code touched |
| `parts/DOCUMENTATION.md` | Held the reasoning-only form of this rule | Rule now covers text parts |

## Validation

| Check | Result |
|---|---|
| `bun test .../AssistantTextPart.test.tsx` before the fix | 2 fail, 2 pass: a finished part rendered nothing under `streaming` and `cooldown` |
| Same, after the fix | 4 pass |
| Guard removed and restored | 2 fail, then 4 pass |
| `node scripts/run-isolated-tests.mjs packages/ui/src/components/chat/message/parts` | 11/11 files pass |
| `bun test` on the new test, `ReasoningPart.test.tsx`, and `streamTextCommit.test.ts` in one process, both orders | 24 pass |
| `bun run type-check:ui` | 0 errors in touched files. 5 pre-existing errors in `src/lib/codemirror/languageByExtension.ts`, identical at `origin/main` |
| `bunx eslint` on both files | Clean |
| `bunx oxlint` on both files | Test file clean. Component: 4 pre-existing findings on lines 31-34, down from 5 at `origin/main` |
| `bun run dead-code` | Exit 0, nothing for the touched files |
| Not verified | No recording against a running model. Electron, VS Code, and mobile shells not run; the change is shared UI with no runtime branch |

## Visual evidence

Before: the reporter's recording on #3484 (clearest at 00:10; OpenCode CLI on the right has the full paragraph, OpenChamber on the left holds it).

After: no like-for-like recording; it needs a model that writes a paragraph and then calls a tool. The component test pins the behavior: with `time.end` set and the phase still `streaming` or `cooldown`, `AssistantTextPart` renders where it rendered nothing before, an in-progress part holds its incomplete paragraph, and a completed message renders a part without `time.end`. I can attach a recording if you want one before merge.

## Risks and failure behavior

- The guard only turns `isStreaming` off in more cases, so it cannot bring back the #2020 false-streaming state for cached parts without `time.end`.
- If OpenCode appended text after stamping `time.end`, that text would render unthrottled. The protocol sets `time.end` on `text-end`, and `ReasoningPart` has relied on that since #2020.
- For a closed part, the throttle flush, `MarkdownRenderer` cache eligibility, and `parseGeneratedJsonResult` now run at part end. All of them already ran at message end.
- No state, storage, or network is touched.


mattv8:
I recommend consolidating this with #3498. Both changes release finalized text while the message continues running; #3498 has passing checks and DOM evidence from this review pass. We should carry over your renderer regression tests and contribution credit, and link #3484 alongside #3277 so we track both reports.

IbrahimKhan12:
Agreed. Same root cause and same gate; #3498 has the evidence and the older report, so it should be the one that lands. Closing this in favor of #3498.

Carry-over is done: the component test from this branch is up as ouyangjian28/openchamber#1, a one-commit PR into #3498's branch. On #3498's head (`90a764c19`) it passes 4/4 unchanged alongside `assistantTextVisibility.test.ts`. Its footer closes #3484, so both reports track once #3498 merges; suggest #3498's body read `Fixes #3277, closes #3484` too in case the footer doesn't survive the squash.

PR #3494, opened 2026-09-11
fix(github): resolve issue/PR pickers by number or URL directly

### LINKED ISSUE #3412: [Bug] GitHub issue/PR pickers: searching by number/#n/URL never finds the item (search times out on 8s octokit cap)
### What's wrong?

Searching a GitHub **issue or PR by number** when creating/linking it in a session never finds it in the New Worktree dialog, and is awkward in the chat link pickers:

- New Worktree dialog -> "GitHub issue/PR" -> search by PR or issue number (`123`, `#123`, or a pasted `…/pull/123` / `…/issues/123` URL): result list stays empty or keeps loading; the item never appears.
- Mixed input (e.g. `123 bug`) essentially never matches.
- Chat "link issue/PR" pickers: typing a number shows a clickable "Use issue/PR #N" row, but the full-open-issues list still loads anyway and the lookup only happens after an extra click.
- After a number search on the PRs tab, switching to the Issues tab can show a stuck/blank list until the dialog is reopened.

Expected: a bare number, `#number`, or pasted GitHub URL shows that specific issue/PR directly; text search continues to work.

### Root cause (code-traced; live API checks)

There is **no direct by-number search path**: depending on the form of the input, the number either requires an extra manual step or is dropped into GitHub's free-text Search API, which does index numbers (verified live: `repo:openchamber/openchamber 3407 type:issue` found #3407) but is extremely slow — measured 30–120s+ per query in testing — while `packages/web/server/lib/github/octokit.js` applies a hard **8s per-request timeout**. The search dies on the timeout and the route's catch returns an empty list, so the UI shows "nothing found" (or spins).

Affected paths (all confirmed in current `main`):

1. `packages/ui/src/components/session/GitHubPrPickerDialog.tsx` + `GitHubIssuePickerDialog.tsx` (chat "link issue/PR" flow): `parsePrNumber` / `parseIssueNumber` detect `#123` / bare number / URLs (`directNumber`), and a clickable "Use pull request #N" / "Use issue #N" row **is** rendered (`GitHubPrPickerDialog.tsx:352‑370`, `GitHubIssuePickerDialog.tsx:530‑548`), which does go to the targeted `prContext` / `issueGet`+`issueComments`. The number is not entirely ignored — the problem is UX: typing a number still triggers a full-list `refresh()` (`if (!debouncedQuery.trim() || directNumber) { void refresh(); }`, i.e. the open list loads unnecessarily) and the targeted lookup only runs after an **extra manual click** on that row. The search never presents the number's result directly.
2. `packages/ui/src/components/session/GitHubIntegrationDialog.tsx` (New Worktree dialog): no number/URL detection at all. Numbers, `#123`, and pasted URLs are forwarded verbatim as free-text `query` to `issuesList`/`prsList`; URLs are never recognized, so they can never match. This is the "search by number finds nothing" path from the bug report.
3. Server `packages/web/server/lib/github/routes.js`:
   - `GET /api/github/pulls/list` (query non-empty) -> `octokit.rest.search.issuesAndPullRequests` with `repo:o/r <query> type:pr state:open`, then one `pulls.get` per hit (up to 50).
   - `GET /api/github/issues/list` (query non-empty) -> same search with `type:issue state:open`.
   - Under slow search responses (common: 30–120s+), the 8s octokit per-request timeout aborts the call, the catch block returns `{ issues: [] }` / `{ prs: [] }` (no error surfaced), and both timeouts compound for PR search (search + per-hit `pulls.get`).
4. Stuck spinner: `GitHubIntegrationDialog.tsx` only clears `loading` in `.finally()` when `!controller.signal.aborted`. Its `AbortController` is created in the effect but **never passed into `runtimeFetch`/the request**, so aborting does not cancel the in-flight fetch — a cancelled search can leave `loading=true` indefinitely. That same stale state also explains the "issues tab broken afterwards" symptom (stale loading/list state after an aborted PR search); no shared server state is involved.

### Suggested fix direction

- In `GitHubIntegrationDialog` (New Worktree dialog), route purely numeric / `#number` / issue-or-PR URL input to the **targeted** endpoints (`issueGet` + `issueComments`, or `prContext`) and present the single result — or surface the same clickable "Use #N" row the chat pickers already have. This is deterministic, one request, and independent of search indexing and `state:open` filtering.
- In the chat pickers, consider showing the targeted result directly for number/URL input instead of loading the full open list plus requiring the extra click.
- Keep text and mixed text+number (`123 bug`) as free-text search — that is a legitimate search query, not a number lookup; for it, make the timeout behavior explicit (longer/timeout-aware handling for search calls, or surface a visible "search timed out" message instead of an empty list).
- Clear `loading` on effect cleanup regardless of abort state, and either wire the `AbortSignal` through `runtimeFetch` for these calls or drop the ineffective controller.

### Where does it happen?

All runtimes share this code path (`packages/ui/src/components/session/*`, `packages/web/src/api/github.ts`; VS Code proxies the same server routes), so Desktop / Desktop Web / VS Code are all affected.

### Version

Current `main` (latest local: `3d0673a54`, v1.11.x series).

### Notes

- Confirmed by code tracing + live GitHub Search API checks; the exact user-visible spinner sequence was not reproduced in a running app, so the UI-stuck mechanics are code-traced rather than captured live.

bashrusakh:
Fixes #3412.

## Intent

Searching a GitHub issue or PR by a bare number (`123`), `#123`, or a pasted GitHub issue/PR URL in the New Worktree dialog or the chat link pickers never surfaced the item: the number fell into GitHub's Search API, which routinely exceeds the server's 8s octokit bound and was swallowed into an empty list, and the pickers also loaded the full open list before requiring an extra click on a "Use #N" row.

The query resolution now lives at the owning boundary. `issuesList`/`prsList` accept any user query and the **server** list route decides:

- bare `N`, `#N`, or a `github.com` issue/PR URL → that single item through `issues.get`/`pulls.get` on page 1 (one GitHub call, no Search);
- free text and mixed queries (`123 bug`) → GitHub Search, unchanged.

Failures reject with machine-readable codes and localized copy: search timeout → `504 search_timeout`, exact reference miss or kind mismatch → `404 not_found`, a URL outside the project's repo/fork network → `422 repo_unavailable`. A successful empty list now only ever means "no results". Resolved exact items carry `sourceRepo`, and each picker has one load owner: page-1 loads abort their predecessors, superseded page-2 loads are dropped, and only the current request generation may settle its spinner. A failed load renders only its localized error - the empty state is gated on `!error`, so a failure is never shown as "no results".

## Non-goals

- Network-wide bare-number resolution: a bare `#N` resolves against the project's own repo; a URL can address the fork network. Same-number collisions across the network are not guessed.
- Cross-kind auto-switch: an issue URL in the PR tab (or vice versa) is a normal `not_found`, not a tab switch.
- Per-item `pulls.get` failures inside a free-text PR search stay isolated (a failed item is dropped) - pre-existing partial-failure behavior.
- VS Code's GitHub backend stays disabled by design; the webview client only forwards `signal` at the bridge level.

## Affected surfaces

- `packages/ui`: `GitHubIntegrationDialog`, `GitHubIssuePickerDialog`, `GitHubPrPickerDialog`, `usePullRequestComparison`, new `lib/api/github-errors.ts`, `lib/api/types.ts`, and 12 i18n dicts (add `searchTimedOut` and `repoNotResolvable`; remove two dead quick-row keys).
- `packages/web` server: exact-resolution branches in `issues/list` and `pulls/list`, new `lib/github/reference.js`, repo-identity preservation in `resolveRepoForRequest`; `src/api/github.ts` throws errors carrying `code`.
- `packages/vscode/webview/api/github.ts`: `signal` forwarded via `sendBridgeMessageWithOptions` (bridge-level only).
- Runtimes: web/desktop/hosted mobile share the UI + server path; VS Code proxies the same server routes.
- User-visible states: New Worktree issue/PR picker and chat link pickers (loading, empty, not-found, timeout, foreign-repo).

## Repository guidance

|---|---|---|
| `openchamber-change-discipline` | Source/contract changes across shared UI and server | Server-side resolution replaces the client resolver and parser; no new deps/store; tests at the owning layer |
| `ui-api-decoupling` | Shared UI data access | Signals flow through the existing `runtimeFetch`/bridge wrappers; one shared error-code parser |
| `sync-state-invariants` | Async loading/abort state in pickers | Single load owner per picker, generation guards, aborts, and failure is never rendered as authoritative empty |
| `locale-ui-patterns` | New user-facing strings | Keys added to all 12 dicts; `repoNotResolvable` values match the existing translations; parity tests green |
| `packages/web/server/lib/github/DOCUMENTATION.md` | Route contract changed | Exact-reference and failure contract documented next to the routes |
| `CONTRIBUTING.md` PR contract | PR handoff requirements | Body refilled for the current HEAD; validation and unverified items stated |

## Validation

| Check | Result |
|---|---|
| `bun run --cwd packages/web test server/lib/github src/api/github.test.ts` | 5 files, 49 tests passed |
| `bun test src/lib/api/github-errors.test.ts src/lib/i18n/messages.test.ts src/lib/i18n/store.test.ts` (`packages/ui`) | 11 passed, 439 assertions |
| `bun run --cwd packages/web test` (full web suite) | 195 files passed, 2133 tests passed; 3 suites fail only on missing modules (`happy-dom` x2, `@pierre/diffs`), zero assertion failures |
| Independent `@tester` pass + delta pass | Contract checks PASS; dialog runtime tests blocked by missing `happy-dom` (no component harness) |
| Independent whole-PR `@reviewer` (`Coverage: full PR`) | PASS, no blocking findings at the reviewed HEAD |
| `bunx eslint` on changed UI files | Clean |
| `bunx oxlint <changed paths>` | No findings on delta-added lines except documented boundary-parser/test-mock baseline |
| `bun run dead-code` | No new unused exports/files |
| Scoped type-check (`packages/ui`, `packages/web`) | No errors in changed files; package checks blocked by pre-existing missing-declaration errors |
| Reconciled base (`a9944a28e`) + merge integration | PR contract intact (27 files; the merge changed no PR file except the required `assetWithUrlToken` mock line); vscode `settings.test.ts` green; final whole-PR review PASS at `f1ef1ee71` |

Not verified: live GitHub API/runtime behavior, dialog DOM tests (missing `happy-dom`), and VS Code runtime (GitHub backend disabled by design).

## Visual evidence

User-visible change, but this environment has no running app with a GitHub-connected project to drive the pickers, so no meaningful before/after capture was possible. Manual check recommended: open each picker, type `#N` and a pasted issue/PR URL, and confirm the single item appears (closed/merged states included), while free text still searches.

## Risks and failure behavior

- Failure contract: empty list = no results; timeouts, rate limits, and server failures reject. The UI maps codes to localized copy, and `usePullRequestComparison` / DiffView / WalkthroughView / MobileChangesSurface render an error instead of an empty list on a PR search timeout.
- URL safety: only `github.com`/`www.github.com` hosts parse; owner/repo is matched case-insensitively against the project repo or its fork network, and anything else is rejected with `422` - no silent fallback to another repo.
- Cancellation: page-1 abort, page-2 abort plus generation guard; only the owning generation settles its loading flag, and cleanup/close clears it. Known latent gap (unreachable in the shipped UI): the Issue picker's manual Refresh only renders in non-`select` mode, and its unsignaled response is only generation-guarded.
- Compatibility: search/list pagination is unchanged (`page > 1` never exact-resolves); the removed list `error` fields were added by this same unmerged PR; the VS Code GitHub backend is disabled.
- Rollback: revert the branch; no migrations or persisted state.

## Non-blocking follow-ups

- Boundary-parser `unknown` params keep 5 anti-slop findings (same pattern as the existing `lib/api/files-errors.ts`).
- `usePullRequestComparison` can leave `loadingMore` set if `refresh` identity changes mid-page-2 (pre-existing).
- Per-item `pulls.get` failure isolation inside a free-text PR search (pre-existing; tracked in #3546).
- Upstream `main` was merged into the branch (`a9944a28e`), so the reviewed base is current; the previously flaky `packages/vscode/webview/api/settings.test.ts` was fixed upstream (PR #3542) and passes on the merged tree.


bashrusakh:
Leaving both non-blocking findings as-is. Neither is worth another candidate + full re-review + CI cycle, and one has zero practical gain.

**Duplicate list request on empty query** (`GitHubIssuePickerDialog.tsx:136`, `GitHubPrPickerDialog.tsx:124`, `GitHubIntegrationDialog.tsx:134`)

The extra request is a conditional GET: the server sends `If-None-Match` and GitHub answers `304` from the ETag cache (`packages/web/server/lib/github/octokit.js:20-25,44-57`), so the revalidation does not count against the REST rate limit and returns the cached body. The only cost is one extra round trip; there is no user-visible difference and no rate-limit impact. Net practical gain from fixing: zero.

For transparency: the query-clear variant is new in this branch (the race fix added `debouncedSearchQuery` to the open effect's deps); the open-time duplicate is pre-existing. A clean fix is to let one effect own the empty-query load, but that would restructure effect ownership across three dialogs for no observable benefit.

**No automated coverage for the dialog abort/selection state machine** (`GitHubIntegrationDialog.tsx:154`)

The value here would be regression protection, not runtime behavior. Closing it requires `happy-dom` (a dependency addition) plus a new dialog test harness — disproportionate to this fix and a test-infra concern rather than a defect in this PR. Tracking as a follow-up.

The fix and its verified behavior are unaffected. Happy to split either item into a separate follow-up PR if a maintainer wants it.

PR #3498, opened 2026-09-11
fix(ui): render sealed text parts in full while a question pends

### LINKED ISSUE #3277: [Bug] Text before question tool is still hidden until an option is selected (was claimed fixed in #503)

ouyangjian28:
Supersedes #3495 — same fix, resubmitted as a single clean commit with rendered evidence attached.

Fixes #3277

## Problem

While a `question` tool call pends, live mode's block-commit reveal keeps holding the text after the last committed line, and the turn never finalizes until the user answers. So the tail of the preamble — typically its final line — stays invisible for as long as the question card waits, even though the model already wrote it. #2655 fixed this class of hold for the `sorted` render mode; the live path kept the same gap.

## Fix

Gate the hold on **part** finalization instead of **message** streaming. The SDK stamps the text part's `time.end` as soon as the model moves on to the question part, so a sealed part is provably complete even though the turn is still open: it now renders in full immediately, while a part without `time.end` keeps the existing hold unchanged. This mirrors the reasoning-part gate from #2020. One behavior-note line added to `parts/DOCUMENTATION.md` beside that invariant.

## Evidence

Before (`main`, question pending — the list ends at `Line 9`; `Line 10` is absent from the DOM entirely, confirmed by tree walk — not a CSS/overlay effect):

before: pending question on main hides the list's final line

After (this branch, **same session, same pending question** — the full list renders above the question card):

after: same pending question with the fix shows the full list

Both frames are the same session viewed at 1280×900, question card centered; the tenth line carries a fixed marker phrase so the difference is unambiguous. DOM list items on the same stored message: 9 → 10. Screenshots live on the fork's `assets/3495-repro` branch so this PR's diff stays code-only.

## Testing

- New `assistantTextVisibility.test.ts` (7 cases): sealed vs unsealed parts across live/sorted modes, pending-question turns, and the empty-tail edge.
- `packages/ui` type-check and eslint pass; the `chat/message` suite passes (135 tests).

### DIFF
--- a/packages/ui/src/components/chat/message/parts/AssistantTextPart.tsx
+++ b/packages/ui/src/components/chat/message/parts/AssistantTextPart.tsx
@@ -3,7 +3,7 @@ import type { Part } from '@opencode-ai/sdk/v2';
 import { MarkdownRenderer } from '../../MarkdownRenderer';
 import type { StreamPhase, ToolPopupContent } from '../types';
 import { useStreamingTextThrottle } from '../../hooks/useStreamingTextThrottle';
-import { resolveAssistantDisplayText, shouldRenderAssistantText } from './assistantTextVisibility';
+import { resolveAssistantDisplayText, resolveAssistantTextStreaming, shouldRenderAssistantText } from './assistantTextVisibility';
 import { streamPerfCount, streamPerfObserve } from '@/stores/utils/streamDebug';
 import { GeneratedJsonResultCard } from './GeneratedJsonResultCard';
 import { parseGeneratedJsonResult } from './generatedJsonResult';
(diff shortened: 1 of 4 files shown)


IbrahimKhan12:
Following @mattv8's suggestion on #3485, the component-level regression test from that PR is available as ouyangjian28/openchamber#1, a single commit against this branch. It renders `AssistantTextPart` and asserts on the DOM marker (fails 2/4 with the `isFinalized` gate removed, 4/4 on `90a764c19`). Merging it there adds the commit here; its footer closes #3484 alongside #3277. #3485 is closed.

PR #3515, opened 2026-09-11
feat(usage): support command-backed multi-account quota sources

maxiedaniels:
## Intent

External tools, authentication plugins, and local credential brokers may manage multiple accounts or proxies for an upstream model provider. OpenChamber's built-in quota fetchers cannot see secondary accounts or external availability states.

This change allows a trusted local command to replace the quota source for an existing registered provider ID such as `codex`. It adds:

- A versioned command protocol configured in local-only `usage-providers.json`
- Account-scoped usage windows and status metadata
- Account groups in expanded usage cards and the Usage settings page
- A collapsed headline that selects the shortest window from the current available account
- A runnable reference provider for testing

## Non-goals

- Dynamic provider registration or custom provider IDs
- Reading third-party plugin credential stores directly
- Managing account rotation or routing decisions inside OpenChamber
- Storing executable commands in remotely writable or synchronized settings

## Affected surfaces

- `packages/web`: command runner, local configuration, provider registry, and directory-aware quota route
- `packages/vscode`: extension-host command runner and runtime bridge parity
- `packages/ui`: response validation, account rendering, headline selection, and directory-change refresh
- `packages/docs`: custom provider documentation, sidebar entry, and all supported locales
- `scripts`: runnable reference provider

## Repository guidance

|---|---|---|
| OpenChamber Change Discipline | Extends a cross-workspace response contract and runtime behavior | Updates every real consumer, validates at the shared payload boundary, and preserves runtime parity |
| UI API Decoupling | Shared UI receives quota data through runtime-owned APIs | Command execution stays in web/VS Code hosts; shared UI only consumes the versioned result contract |
| Sync State Invariants | Directory-scoped data must follow the authoritative session/worktree | Quota requests resolve the active session directory at call time and refresh immediately when it changes |
| Quota provider documentation | Defines registration and response contracts | Extends provider usage with an optional accounts array and documents the command protocol |

## Configuration

Create `~/.config/openchamber/usage-providers.json` (or `$OPENCHAMBER_DATA_DIR/usage-providers.json`) with absolute executable and script paths:

```json
{
  "version": 1,
  "commands": {
    "codex": ["/absolute/path/to/node", "/absolute/path/to/openchamber/scripts/example-usage-provider.mjs"]
  }
}
```

Commands execute as argv without a shell, with a 30-second timeout and 1 MiB output cap.

## Visual evidence

- [Multi-account Codex usage in the work-status sidebar](https://github.com/openchamber/openchamber/pull/3515#issuecomment-5692399530)

The screenshot shows multiple accounts with labels, masked details, availability/status metadata, plan labels, and separate usage windows.

## Risks and failure behavior

- **Command failure or timeout:** the provider returns `ok: false` with a normalized error; command stderr is never returned to the client.
- **Malformed configuration:** provider discovery remains available instead of becoming a server-wide 500. A quota request receives a provider-scoped error naming `usage-providers.json` so the local user knows what to fix.
- **Partial account failure:** one account may report `error` without erasing unrelated account usage.
- **Directory changes:** switching sessions or worktrees triggers a refresh using the authoritative session directory rather than retaining the previous directory's sample until the polling interval.
- **Local trust boundary:** an authenticated local OpenChamber client supplies the active session/worktree directory used as command `cwd`. Clients cannot configure, replace, or append executable argv; argv remains fixed in the host's local configuration file. Invalid working directories fail the provider request rather than falling back silently.
- **Command configuration:** commands remain outside settings read/write routes, so browser and remote clients cannot register arbitrary executables.

## Validation

| Check | Result |
|---|---|
| Real `fetchQuota` multi-account parser regression | Passes; verified to fail when `accounts` is removed from the Zod schema |
| Web quota tests | 19 files, 188 passed |
| VS Code isolated test suite | 42/42 test files passed |
| UI isolated test suite | 479/481 files passed; the same two quota files fail on the upstream baseline |
| Web test suite | 211/212 files passed; the same environment-dependent `network-defaults` test fails with changes stashed |
| Workspace `type-check` | Passed across all packages |
| Workspace `lint` | Passed |
| Workspace `build` | Passed |
| `bun run docs:validate` | Passed: 583 pages, 53 sidebar links |
| `bun run dead-code` | No findings in changed quota/docs files; report contains only existing unrelated entries |

The two UI failures are existing quota test isolation problems: `useQuotaStore.refresh.test.ts` and `useQuotaStore.same-origin.test.ts` fail identically with the new worktree changes stashed. The web failure is likewise reproducible without this PR's uncommitted follow-up changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,6 +130,7 @@ Go deeper with the OpenChamber guides:
 - [GitHub workflows](packages/docs/content/docs/github.mdx)
 - [Mobile](packages/docs/content/docs/mobile.mdx)
 - [Security](packages/docs/content/docs/security.mdx)
+- [Custom usage providers](packages/docs/content/docs/custom-usage-providers.mdx)
 - [Troubleshooting](packages/docs/content/docs/troubleshooting.mdx)
 
 For self-hosting details, see the [reverse proxy guide](docs/REVERSE_PROXY.md). For custom theme authoring, see the [custom themes guide](docs/CUSTOM_THEMES.md).
--- a/packages/docs/content/docs/custom-usage-providers.mdx
+++ b/packages/docs/content/docs/custom-usage-providers.mdx
@@ -1,0 +1,83 @@
+---
+title: Custom usage providers
+description: Supply account and quota data from a trusted local command.
+---
+
+# Custom usage providers
+
+OpenChamber can use a local command instead of its built-in quota fetcher for
+an existing provider. This is useful when an authentication plugin, credential
+broker, or account pool owns usage data that OpenChamber cannot read directly.
+
+Create `usage-providers.json` in the OpenChamber data directory:
+
+- macOS and Linux: `~/.config/openchamber/usage-providers.json`
+- Custom server data directory: `$OPENCHAMBER_DATA_DIR/usage-providers.json`
+
+Configure an argv array under the provider ID:
+
+```json
+{
+  "version": 1,
+  "commands": {
+    "codex": ["/absolute/path/to/usage-helper", "limits", "--json"]
+  }
+}
+```
+
+A runnable reference script is provided in the repository at
+`scripts/example-usage-provider.mjs`. You can point your configuration directly
+to it to inspect multi-account rendering:
+
+```json
+{
+  "version": 1,
+  "commands": {
+    "codex": ["node", "/path/to/openchamber/scripts/example-usage-provider.mjs"]
+  }
+}
+```
+
+Use an absolute executable path when launching OpenChamber outside a terminal,
+because desktop applications may not inherit your shell `PATH`.
+
+The command runs without a shell and receives the active session directory as
+its working directory. It must print one JSON object to stdout:
+
+```json
+{
+  "version": 1,
+  "providerName": "Codex",
+  "accounts": [
+    {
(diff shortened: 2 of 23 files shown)


maxiedaniels:
right sidebar usage

Screenshot of sidebar usage with multiple Codex accounts configured.

tamatyan99:
Nice work getting the automation to PASS. Two things are keeping this parked, and neither is code:

1. `merge-conflict:true` — the branch is unmergeable again after `main` moved.
2. `needs-discussion` — a PR that introduces a new setting/command surface carries a product decision, and CONTRIBUTING wants a linked Ideas discussion before review. A short, vendor-neutral discussion (single-host command overrides; argv only from a host-local file, never from settings or the network) would give the maintainer something concrete to say yes to.

On the review nit: I can add the missing pin between `SUPPORTED_QUOTA_PROVIDERS` and the `fetchQuotaForProvider` switch cases, so the two cannot drift when the next provider lands. I checked your branch — the set has 22 ids and the switch has the same 22 cases today, only in a different order — so the test is small and should stay green.

I am happy to do either of these: open a small PR against your branch with the pin test (and the conflict resolution if you want a hand), or paste a patch you can apply yourself. Just say which.

Issue #3532, opened 2026-09-12
[Bug] iOS: long dictation transcript overflows the composer — caret, scroll and Insert/Send controls unreachable (dictated text lost)


Tobias-Conrad:
### What's wrong? Maybe some steps to reproduce.

On the OpenChamber **iOS** app, a **long dictation transcript overflows the composer overlay**. The action/recovery row (`Cancel` · `Insert` · `Insert & Send`) is pushed below the visible viewport, the transcript can no longer be scrolled, and the **caret cannot be placed into the text** — so the dictated text can neither be edited nor sent and is effectively lost. The only way out is to force-quit and reopen the app.

Steps to reproduce:

1. Open a chat on iOS and start dictation (microphone).
2. Dictate a **short** message → the transcript appears, scrolls, and can be inserted/sent. This works.
3. Start dictation **again** and speak for several minutes (a long, uninterrupted thought).
4. Try to scroll the transcript, place the caret, or reach the action row.

Actual behavior:

- The transcript grows beyond the visible viewport.
- There is **no scroll**, and the recovery row (`Cancel` / `Insert` / `Insert & Send`) is off-screen.
- The **caret cannot be placed** — only word/range selection works.
- The dictated text cannot be submitted and is **lost**.

Expected behavior:

- The overlay/transcript is bounded to the visual viewport above the keyboard; long content scrolls inside it.
- The action row stays **pinned and reachable** regardless of transcript length.
- The caret stays placeable, and the dictated text can always be inserted/sent.

Note: this is **not deterministic on the first attempt** — it depends on transcript length. Long dictation triggers it reliably, which is why it appears "irregularly".

### Where does it happen?

Native **iOS** application (Capacitor shell, shared `packages/ui` composer). The closest option in the template is *Mobile (Web/PWA)* — the composer is shared across web, desktop, hosted mobile and Capacitor mobile, so the fix belongs in the shared composer.

### Version (if known)

1.23.1 (latest release, 2026-09-11). iOS version: not captured yet.

### Screenshots / recordings (optional)

**Original screenshot from the reporter** (native iOS app, stuck state):

![OpenChamber iOS — long dictation stuck in the composer overlay](https://files.catbox.moe/qurti7.jpg)

Only the **top** of the dictated transcript is visible. Everything below the viewport is unreachable — see the data-loss section below. Schematic before/after for context: https://files.catbox.moe/zw567w.png

### Logs (optional)

[...]

PR #3533, opened 2026-09-12
fix(ui): keep failed-dictation salvage controls on screen

### LINKED ISSUE #2533: [Bug] Android: long dictation transcript pushes insert/send controls off-screen

Tobias-Conrad:
## Intent

Fixes #2533. Refs #3532.

A long dictation that fails partway (e.g. the server answers `Dictation stream not started` after several minutes) renders the partial transcript as **salvage text** in the composer overlay. The host grew the composer to the full transcript height through an uncapped `minHeight`, so on mobile the overlay outgrew the viewport:

- the transcript tail could not be scrolled to,
- the caret could not be placed (read-only text; iOS offers only **Copy**, never **Select All**), and
- the `retry` / `discard` / `insert` row was pushed off-screen.

Minutes of dictation were trapped with no recovery path; force-quitting discards them. This is the failure the reporter hit on a physical iPhone (iOS/Capacitor).

This change applies the editor's measured **line and screen-height caps** to the salvage height before using it as a floor, so the composer stops at the viewport and the salvage area scrolls internally. The limit math moves into `heightLimit.ts` (shared with the editor) and is measured through `useComposerHeightLimit`; a height is reported only while salvage text is shown, and re-reported after rotation or any width change that rewraps the text.

**This is a rebase of #3026 (by @karimodm) onto current `main`** — that branch is `mergeable: false / dirty`. The intent is unchanged; the surrounding composer moved (floating composer, etc.), so the patch was ported and re-verified. Full credit to @karimodm.

## Non-goals

- No change to speech capture, transcription providers, or the dictation WebSocket protocol.
- No live transcript while recording/transcribing (unchanged; only the failed-salvage path grows the composer).
- No change to recovery button behaviour, keyboard handling, or the fullscreen composer.
- No change to desktop or VS Code layouts beyond routing the editor cap through the shared helper.

## Affected surfaces

- `packages/ui`: the shared composer used by web, desktop, hosted mobile and Capacitor mobile.
- The failed-dictation salvage overlay on mobile (the reported iOS path); the editor's own cap is refactored onto the shared helper.
- No persisted data, API, route, setting or external contract changes.

## Repository guidance

|---|---|---|
| `AGENTS.md` + change discipline | shared UI and mobile runtime behaviour | diff stays in composer sizing, the shared helper, focused tests, and the owning documentation |
| `theme-system` | affects shared layout | no colours, tokens, controls, icons or animations changed |
| `packages/ui/src/components/chat/composer/DOCUMENTATION.md` | owns composer sizing and mobile behaviour | records the shared typed/salvage limit and the transient non-positive viewport fallback |
| `changelog/` is read-only until the maintainer asks | release notes are release-time work | no changelog entry added |

## Validation

| Check | Result |
|---|---|
| `bun test packages/ui/src/components/chat/composer/editor/__tests__/composerHeightLimit.test.ts` | **7 pass, 0 fail** (9 assertions) |
| `bun test packages/ui/src/components/chat/composer packages/ui/src/components/dictation` | **380 pass, 0 fail** |
| `bun run --cwd packages/ui lint` | pass |

**Hardware note:** the report is from a physical iPhone (iOS). The original screenshot and reproduction are in #3532. I could not reproduce on a physical iOS device here, so a device check of the salvage overlay before merge is welcome.

## Visual evidence

This is a user-visible mobile fix, so here is the before state and the reasoning for the after:

- **Before** (reporter's physical iPhone, failed dictation with long salvage text): the transcript fills the viewport, the `retry` / `discard` / `insert` row is below the fold, and the tail cannot be reached — https://files.catbox.moe/qurti7.jpg (also in #3532). The same screenshot on the reporter's macOS app shows the underlying cause the server reports: `Dictation worker request timed out: session.commit`.
- **After:** the composer stops at the measured bound, so the salvage area scrolls internally and the action row stays pinned above the keyboard. I could not capture this on a physical iOS device from here — the height math is pinned by `composerHeightLimit.test.ts` (7 tests) and the composer+dictation suite (380 tests), and a device check is requested below.

## Risks and failure behavior

- **Composer folding to zero on a transient measurement:** `getComposerHeightLimit` keeps the line cap whenever the measured screen budget is non-positive, so a resize race cannot collapse the editor. Pinned by the "no positive screen budget" test.
- **Feedback loop (the limit growing with the host it limits):** the growing host is excluded from its own budget (`getComposerHostHeightLimit` subtracts `hostHeight` from `branchHeight`). Pinned by the "does not feed the applied salvage height back into its limit" test.
- **Scroll position:** the salvage area only auto-follows while the reader is already at the end, so a re-report after rotation no longer yanks someone who scrolled up to copy text (second commit on this branch).
- **Cross-runtime:** desktop inherits the 8-line cap for salvage text; VS Code is unaffected because dictation is disabled there. No persisted data, API, route, or setting changes.
- **Rollback:** revert the two commits; nothing is persisted, so no migration or cleanup is needed.

### DIFF
--- a/packages/ui/src/components/chat/ChatInput.tsx
+++ b/packages/ui/src/components/chat/ChatInput.tsx
@@ -121,6 +121,7 @@ import {
     type ComposerChange,
     type ComposerEditorHandle,
 } from './composer/editor/ComposerEditor';
+import { useComposerHeightLimit } from './composer/editor/useComposerHeightLimit';
 import { createComposerEditorViewStore } from './composer/editor/viewStore';
 import { composerAutoCorrect } from './composer/editor/autocorrect';
 import {
@@ -2197,14 +2198,23 @@ const ChatInputComponent: React.FC<ChatInputProps> = ({
         }
     }, [agents, currentAgentName, currentSessionId, setAgent, saveSessionAgentSelection]);
 
-    // Height the dictation transcript needs (null when idle). Its overlay sits
+    // Height the failed-dictation salvage text needs. Its overlay sits
     // absolutely over the composer, so the composer must be able to grow for
-    // it. The editor sizes itself to its own content; this is the one external
-    // constraint, applied as a floor on the editor's container.
+    // it. Apply the editor's line and screen bounds before using that height as
+    // a floor, otherwise long salvage text can push the action row off-screen.
+    const dictationHeightHostRef = React.useRef<HTMLDivElement | null>(null);
     const [dictationContentHeight, setDictationContentHeight] = React.useState<number | null>(null);
     const handleDictationContentHeightChange = React.useCallback((height: number | null) => {
         setDictationContentHeight((prev) => (prev === height ? prev : height));
     }, []);
+    const dictationHeightLimit = useComposerHeightLimit({
+        active: dictationContentHeight !== null,
+        disabled: isComposerExpanded,
+        hostRef: dictationHeightHostRef,
+        maxLines: isMobile ? MAX_MOBILE_COMPOSER_LINES : MAX_VISIBLE_COMPOSER_LINES,
+        boundSelector: isMobile ? '[data-composer-bound]' : undefined,
+        boundGapPx: isMobile ? MOBILE_COMPOSER_BOUND_GAP_PX : 0,
+    });
 
     const updateAutocompleteState = React.useCallback((
         value: string,
@@ -3500,14 +3510,20 @@ const ChatInputComponent: React.FC<ChatInputProps> = ({
                             {!isBtwActive ? <ActiveEditorFileSuggestion /> : null}
                         </div>
                         <div
+                            ref={dictationHeightHostRef}
                             className={cn("relative overflow-hidden", isComposerExpanded && 'flex flex-1 min-h-0 flex-col')}
                             onDragEnter={handleDragEnter}
                             onDragOver={handleDragOver}
                             onDropCapture={handleDropCapture}
                             onDrop={handleDrop}
                             onDragEnd={handleDragEnd}
-                            style={dictationContentHeight !== null
-                                ? { minHeight: `${dictationContentHeight}px` }
+                            style={dictationContentHeight !== null && !isComposerExpanded
+                                ? {
+                                    minHeight: `${Math.min(
+                                        dictationContentHeight,
+                                        dictationHeightLimit ?? dictationContentHeight,
+                                    )}px`,
+                                }
                                 : undefined}
                         >
                             <ComposerEditor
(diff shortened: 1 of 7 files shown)


Tobias-Conrad:
Thanks — both points addressed.

**Handoff (body sections):** the PR body now includes the template's **Visual evidence** and **Risks and failure behavior** sections.

**Non-blocker (salvage scroll jumps back on rotation/rewrap):** fixed in `e1b7b84`. The salvage area now follows the newest words only while the reader is already at the end; a re-report after a width change keeps a user who scrolled up exactly where they were. The first report still jumps to the bottom.

Re-validated: `bun run --cwd packages/ui type-check` pass; `bun test packages/ui/src/components/chat/composer packages/ui/src/components/dictation` → 380 pass, 0 fail.

The remaining evidence gap (no physical-iOS capture) is stated in the body; the reporter's device screenshot of the trapped state is linked there.

Tobias-Conrad:
Companion to #3535. This ports #3026 (by @karimodm) onto current `main`: the failed-salvage overlay now stops at the viewport and scrolls internally, so `retry` / `discard` / `insert` stay reachable on mobile instead of trapping minutes of dictation. Credit for the original fix to @karimodm.

Green, `review:ready` since Sep 12, and it still merges cleanly although the composer moved on Sep 13 and Sep 15. Happy to re-verify against current `main` if that helps.

btriapitsyn:
Thanks, Tobias, for updating this fix and catching the scroll-position problem. We landed the combined implementation in #3026, keeping the original contribution and crediting your follow-up. Closing this port now that the fix is in main.

PR #3535, opened 2026-09-12
fix(dictation): keep segments short enough to commit without timing out

### LINKED ISSUE #3537: [Bug] Long dictation fails with `Dictation worker request timed out: session.commit` (server-side: macOS app, iOS, web)

A **long dictation fails** and the transcript is dropped, with the composer overlay showing:

> `Dictation worker request timed out: session.commit`

Reporter observations (Apple M1, 16 GB):

- ~15 s dictation returns instantly.
- ~30 s gets progressively slower.
- ~2 minutes fails; the recognition process climbs to several GB of RAM in Activity Monitor.

**This is not an iOS-only bug.** It is a **server-side** dictation failure and reproduces in the **macOS app** and (per the architecture below) any client, including iOS and web.

Steps to reproduce:

1. Start dictation.
2. Speak continuously for more than roughly a minute or two.
3. Stop.
4. The commit fails with `Dictation worker request timed out: session.commit`; the transcript is not delivered.

### Where does it happen?

OpenChamber **server / dictation worker**, reached by every client. The native macOS app reproduces it directly; the iOS app hits the same server.

**Architecture (why it is one bug, not one per client):** the client only captures audio and streams 16 kHz mono PCM16 over `/api/dictation/ws`; the **server transcribes**. The local engine is `sherpa-onnx` Parakeet running in a forked worker process owned by the OpenChamber web server on the host machine. iOS does **not** transcribe on-device — it sends audio to the same worker. So a fix in the worker/manager covers macOS, iOS and web at once.

### Version (if known)

1.23.1 (latest release, 2026-09-11). Reproduced on macOS (M1, 16 GB) and iOS.

Failed-dictation overlay on the macOS app showing the error: reported by the same user as #3532 (the overlay in #3532 is the iOS view of the same failure). The overlay keeps the salvage text reachable only because of the fix in #3533.

### Logs (optional)

Error surfaced by the client as `Dictation worker request timed out: session.commit`.

---

### Root cause

- The worker decodes a committed segment **synchronously**: `packages/web/server/lib/dictation/local/worker-process.js` handles `session.commit` by calling `SherpaSegmentTranscriptionSession.commit()` inline, and `sherpa-recognizer.js` runs `decodePcm16` on the worker's thread. `sendOk` only runs after the decode returns.
- The client's request watchdog is **30 s** (`worker-client.js`, `DEFAULT_REQUEST_TIMEOUT_MS`). A decode that exceeds it rejects with exactly this message.
- Segments were allowed up to **90 s** (`stream-manager.js`, `DEFAULT_SEGMENT_MAX_SECONDS`). Parakeet decode cost and peak memory grow **quadratically** with segment length (documented: 60 s → 2.1 s / +90 MB; 300 s → 21.3 s / +1.5 GB on reference hardware). On a loaded or low-core machine a long segment blows past 30 s — and the transcript is lost.

So the failure is length-dependent, which is why short dictations always work and long ones fail "irregularly".

### Fixes proposed

- **#3535** — cap segments at **30 s** (pause split from 20 s) and let a slow commit run **120 s** before it counts as a failure. This is the direct fix for the timeout, on the server, for all clients.
- **#3533** — keep the failed-dictation overlay bounded to the viewport so the salvage text and the `retry` / `discard` / `insert` row stay reachable instead of trapping the transcript (the iOS view of this failure).

### Impact

Dictation is meant for long, uninterrupted input — exactly the case that fails. Several minutes of speech are lost. Because the transcript lives only in the client's memory while the overlay is up, a failed commit plus an unreachable overlay (iOS) means the text cannot be recovered.

Related: #2533 · #3532

Tobias-Conrad:
## Intent

Fixes #3537 (server-side: covers the macOS app, iOS and web, since the client only streams audio and the host transcribes).

A long dictation can fail with **`Dictation worker request timed out: session.commit`** — reported on a physical iPhone (#3532) and on the macOS app, reproducible with a ~2 minute dictation on an M1 / 16 GB. The worker decodes each committed segment **synchronously** (`worker-process.js` calls `SherpaSegmentTranscriptionSession.commit()` inline, `sherpa-recognizer.js` runs `decodePcm16` on the worker's thread), so a long segment blocks the worker past the client's 30 s watchdog and the transcript is dropped.

This caps segments at **30 s** (pause split from **20 s**) so every decode stays short, and lets a slow commit run **120 s** before it counts as a failure. On the documented curve (60 s → 2.1 s; 300 s → 21.3 s on reference hardware) a 30 s segment decodes in well under a second, while a 90 s segment on a loaded machine can take tens of seconds.

## Non-goals

- No change to the WebSocket protocol, the composer UI, or the recovery controls (that is #3533).
- No change to the local model, thread count, or provider selection.
- No change to the finalization timeout budget.

## Affected surfaces

- `packages/web` dictation server: `DictationStreamManager` segment defaults and `DictationWorkerClient.commitSession`.
- Every local (`sherpa-onnx`) dictation longer than 20 s. Shorter ones still decode as a single segment.
- The `openai-compatible` provider shares the manager, so its segments also become ≤30 s.
- No persisted data or external contract changes.

## Repository guidance

|---|---|---|
| `AGENTS.md` + change discipline | server runtime behaviour on a data-loss path | the change stays in the dictation module and its owning documentation |
| `packages/web/server/lib/dictation/DOCUMENTATION.md` | owns segmentation and provider behaviour | records the new caps and the synchronous-commit / watchdog reason |
| `changelog/` is read-only until the maintainer asks | release notes are release-time work | no changelog entry added |

## Validation

| Check | Result |
|---|---|
| `bun test packages/web/server/lib/dictation` | **15 pass, 0 fail** |
| `bun run --cwd packages/web type-check` | pass |

**Not verified:** end-to-end decode timing on the reporter's device. The numbers are the documented measurement; a device check with a 5-minute dictation is requested.

## Visual evidence

Server-side change with no rendered difference. The user-visible symptom is in #3532 (screenshot): the failed-dictation overlay showing `Dictation worker request timed out: session.commit`. The overlay containment that keeps that state recoverable is #3533.

## Risks and failure behavior

- **Transcription quality:** shorter segments can cut mid-sentence at the 30 s hard cap for pauseless speech. The 20 s pause threshold prefers a natural pause first, and the reporter explicitly preferred short parts over losing the dictation. The caps can be tuned independently of the timeout change if the maintainers want different numbers.
- **More commits:** a 5-minute dictation now commits ~10 segments instead of ~3–4. Decodes run while the user is still speaking, so this stays off the critical path, and each decode is strictly cheaper.
- **Slow hardware:** the 120 s commit watchdog is a backstop for a slow machine, not a hang detector. A genuinely stuck worker still fails the stream — and the salvage text is now reachable thanks to #3533.
- **Rollback:** revert the commit; no migration or cleanup.

### DIFF
--- a/packages/web/server/lib/dictation/DOCUMENTATION.md
+++ b/packages/web/server/lib/dictation/DOCUMENTATION.md
@@ -75,8 +75,8 @@ openaiCompatible?: { baseUrl, model, apiKey } }`.
 ## Segmentation
 
 A dictation is one segment unless it runs long. Past `segmentMinSeconds`
-(60 s) the manager commits on the first silent chunk, so cuts land at a pause
-rather than mid-word; `segmentMaxSeconds` (90 s) is a hard cap for speech with
+(20 s) the manager commits on the first silent chunk, so cuts land at a pause
+rather than mid-word; `segmentMaxSeconds` (30 s) is a hard cap for speech with
 no pause in it. Client chunks are ~1 s, so "silent chunk" is roughly a second
 of silence.
 
@@ -85,6 +85,15 @@ and peak memory grow quadratically with segment length. Measured on Parakeet
 v3 int8 with 2 threads: 60 s took 2.1 s and +90 MB, 180 s took 9.3 s and
 +490 MB, 300 s took 21.3 s and +1.5 GB. Committed segments decode while the
 user is still speaking, so only the tail is left to transcribe on stop.
+
+The commit decode is **synchronous inside the worker process**, so a segment
+that is too long for the machine blocks the worker and can trip the client's
+request watchdog (`Dictation worker request timed out: session.commit`). The
+caps keep every decode short enough to stay well inside it: on the same
+hardware a 30 s segment decodes in well under a second, while a 90 s segment
+on a loaded machine can take tens of seconds and drop minutes of dictation.
+`commitSession` therefore also allows a longer watchdog (120 s) than the
+default 30 s, so a merely slow commit finishes instead of failing the stream.
 
 ## Invariants
 
--- a/packages/web/server/lib/dictation/local/worker-client.js
+++ b/packages/web/server/lib/dictation/local/worker-client.js
@@ -94,7 +94,10 @@ export class DictationWorkerClient {
   }
 
   commitSession(sessionId) {
-    void this.sendRequest({ type: 'session.commit', sessionId }).catch((err) => {
+    // Decoding a segment is synchronous inside the worker, so a long segment on
+    // a loaded or low-core machine can exceed the default watchdog. A commit
+    // that is merely slow must not fail the dictation and drop the transcript.
+    void this.sendRequest({ type: 'session.commit', sessionId }, { timeoutMs: 120000 }).catch((err) => {
       this.emitSessionError(sessionId, err);
     });
   }
--- a/packages/web/server/lib/dictation/stream-manager.js
+++ b/packages/web/server/lib/dictation/stream-manager.js
@@ -26,10 +26,12 @@ const DEFAULT_FINAL_TIMEOUT_MS = 10000;
 // quadratically with segment length (measured: 60s -> 2.1s/+90MB,
 // 300s -> 21.3s/+1.5GB). Segmenting keeps a long dictation off that curve and
 // lets committed segments decode while the user is still speaking, so only the
-// tail is left to transcribe on stop. Typical dictations are shorter than the
-// minimum and are decoded as a single segment.
-const DEFAULT_SEGMENT_MIN_SECONDS = 60;
-const DEFAULT_SEGMENT_MAX_SECONDS = 90;
+// tail is left to transcribe on stop. The decode is synchronous inside the
+// worker, so a long segment on a loaded machine blocks it and can trip the
+// client's request watchdog; the caps keep every decode short. Typical
+// dictations are shorter than the minimum and are decoded as a single segment.
+const DEFAULT_SEGMENT_MIN_SECONDS = 20;
+const DEFAULT_SEGMENT_MAX_SECONDS = 30;
 const FINAL_TIMEOUT_MAX_MS = 5 * 60 * 1000;
 const FINAL_TIMEOUT_PER_PENDING_SEGMENT_MS = 15 * 1000;
 const FINAL_TIMEOUT_PER_PENDING_AUDIO_SECOND_MS = 1500;
--- a/packages/web/server/lib/dictation/stream-manager.test.js
+++ b/packages/web/server/lib/dictation/stream-manager.test.js
@@ -250,3 +250,19 @@ describe('DictationStreamManager', () => {
     expect(session.clears).toBe(1);
   });
 });
+
+describe('default segment caps', () => {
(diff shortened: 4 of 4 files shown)


Tobias-Conrad:
**Evidence for the causal claim** (the review's evidence gap), measured on the reporter's affected machine — Apple **M1, 16 GB** — through the production path: model `parakeet-tdt-0.6b-v3-int8`, `numThreads: 2`, `SherpaOfflineRecognizerEngine.decodePcm16`, on 16 kHz mono PCM16 German speech generated with macOS `say` and converted with `afconvert`.

| segment | decode | RSS after |
|---|---|---|
| 15 s | 3.0 s | 1048 MB |
| 30 s | **7.9 s** | 987 MB |
| 60 s | 17.6 s | 1159 MB |
| **90 s** | **29.4 s** | 806 MB |
| 120 s | 39.6 s | 1050 MB |
| 180 s | 51.7 s | 1232 MB |
| 300 s | 146.9 s | 2521 MB |

(model load: 9.4 s, one-time)

A **90 s segment takes 29.4 s — right at the 30 s watchdog.** Any load on the machine pushes it over, which is exactly the reported length-dependent, intermittent failure (15 s always fine, ~2 min fails). The documented reference curve (300 s → 21.3 s) does **not** hold on this hardware (146.9 s here), so the premise is measured, not assumed.

With the fix (max 30 s, pause split 20 s) the worst case is ~8 s, comfortably inside the watchdog; a 5–10 minute dictation commits ~10–20 short segments instead of 3–4 long ones, each strictly cheaper.

Repro script (standalone, no app needed): generate speech with `say -v Anna -f speech.txt -o base.aiff`, `afconvert -f WAVE -d LEI16@16000 -c 1 base.aiff base16.wav`, then decode slices of 15/30/60/90/120/180/300 s and time them.

Tobias-Conrad:
@openchamber-bot review — evidence for the timing claim is now in the comment above (measured on the reporter M1/16 GB: 90 s segment = 29.4 s, the 30 s watchdog).

Tobias-Conrad:
Non-blocker closed in `897489e`: `stream-manager.test.js` now pins the 20/30 defaults with the measured reason, so reverting the constants fails CI. Re-run: `bun test packages/web/server/lib/dictation` → 16 pass, 0 fail.

Tobias-Conrad:
Review nudge: this has been at `review:ready` since Sep 12 and it is the fix for the dictation failures still reported on the released build (v1.23.2).

- **Symptom / path:** a long dictation fails with `Dictation worker request timed out: session.commit`, a short one passes — the "sometimes works, sometimes not" pattern. The worker decodes a committed segment synchronously while the client watchdog is 30 s; a 90 s segment decodes in ~29 s on the reported M1/16 GB and races it.
- **Change:** segment caps 60 s/90 s → 20 s/30 s (`stream-manager.js`) plus a 120 s `commitSession` watchdog (`worker-client.js`). 4 files, 43 lines; the caps are the only behaviour change.
- **State:** `bun test packages/web/server/lib/dictation` green; merges cleanly onto current `main` (the dictation module is unchanged since Sep 12). #3533 is the companion UI half.

If the 20 s/30 s caps are the sticking point, say so and I will retune them.

PR #3536, opened 2026-09-12
perf(sidebar): isolate search input and bound search result rendering

### LINKED ISSUE #3523: perf(sidebar): session search causes full sidebar rerenders and large DOM commits
## Summary

Typing in the Session Sidebar search freezes the sidebar. On the current default branch the raw search query re-renders the entire sidebar on every keystroke, and the debounced query then mounts every matched session row because search bypasses the list limits and the archived-bucket virtualization.

- Commit: `961f1611cb50cb7dd8d24c82e688ca6d8eef37bb` (`main`, v1.23.1)
- Surfaces: web, desktop, VS Code (shared `packages/ui` sidebar)

## Observed symptoms

- Typing in search visibly lags/janks the whole sidebar, not just the input.
- A broad query on a large project mounts hundreds/thousands of session rows at once.
- Large archived buckets are especially affected.

### 1. Raw keystroke re-renders the list subtree (before the 120 ms debounce)

`SessionSidebar` owns raw `sessionSearchQuery` / `isSessionSearchOpen` (`SessionSidebar.tsx:83-84`, debounce at `:139`). Every keystroke re-renders it, and the raw values also ride down through `actions.rowActions`:

`SessionSidebar` -> `SessionProjectCollection` -> `SessionProjectScroller` -> `SessionGroupSection` -> `SessionTreeItem` -> `useSessionActions`

Observed memo breaks on the raw-keystroke path:

- `SessionProjectCollection` is not `React.memo` (`SessionProjectCollection.tsx:648`) and receives fresh inline `topology={{...}} view={{...}} actions={{...}}` literals every render (`SessionSidebar.tsx:643-704`).
- `SessionProjectScroller` is memoized (`SessionProjectScroller.tsx:460`) but its `model`/`actions` props are rebuilt because `groupProps`/`recentSection` depend on `rowActions`, and `openWorktreesPage` is an inline arrow.
- `SessionGroupSection`'s comparator explicitly compares the raw `sessionSearchQuery` (`SessionGroupSection.tsx:251`), so every group fails the memo check.
- `useSessionActions.handleSessionSelect` closes over `sessionSearchQuery` / `isSessionSearchOpen` (`useSessionActions.ts:89-120`), so its identity changes per character; `SessionNodeItem`'s comparator compares it (`SessionNodeItem.tsx:1841`). `SessionTreeItem` is also unmemoized and passes a fresh `children` array, so `SessionNodeItem` fails `:1852` regardless.

Two corrections to the earlier analysis: filtering itself already waits for the debounce (`useSessionSidebarSections` consumes only the debounced, normalized query) - the pre-debounce cost is render propagation; and the scroller's `view` prop is not affected by the raw query.

### 2. Effective query: per-group search work

After the debounce, `useSessionSidebarSections` runs per group (`useSessionSidebarSections.ts:192-235`):

- `filterSessionNodesForSearch` recursively filters the tree (`useSessionGrouping.ts:47-73`), then
- a second recursive `countNodes` walk counts the filtered tree (`:200-206`), plus group-text and folder-name matching.

### 3. Large result DOM

- Active groups show **all** matched roots during search, bypassing the batch limit (`SessionGroupSection.tsx:556-560`).
- Archived virtualization is explicitly disabled while search is active: `shouldVirtualize = isArchivedBucket && !hasSessionSearchQuery && count >= 50` (`SessionGroupSection.tsx:569-571`).

### 4. Shift range selection depends on DOM order

`SessionNodeItem.handleRowSelect` builds range order from `document.querySelectorAll('[data-session-row]')` (`SessionNodeItem.tsx:898-905`). That works only while every row is mounted, so it blocks virtualizing large result sets.

## Measurements before the fix

Harness to be included in the fix PR (`packages/ui/src/components/session/sidebar/sessionSearch.performance.test.tsx`): real 120 ms debounce, real matcher, real sections hook, real `useSessionActions`. Component-level mount is not reachable from `bun test` because the sidebar import graph reaches the Vite-only `import.meta.glob` in `hooks/useProviderLogo.ts`; existing component tests avoid it with module mocks, which the repo lint policy rejects adding.

| Scenario | sessions | filter calls | filter ms | effective-query flush ms | matched rows |
|---|---|---|---|---|---|
| active | 100 | 1 | 0.90 | 2.84 | 100 |
| active | 1000 | 1 | 4.86 | 11.61 | 1000 |
| archived | 100 | 2 | 0.45 | 1.96 | 100 |
| archived | 1000 | 2 | 4.38 | 10.01 | 1000 |
| mixed | 500+500 | 2 | 4.70 | 11.29 | 1000 |
| live debounce flush (act) | 100 | 2 | 0.56 | 2.13 | 110 |
| live debounce flush (act) | 1000 | 2 | 4.38 | 5.77 | 1010 |

Pre-debounce measurements from the same harness: the search data path is not entered on a raw keystroke (filter calls 0, section references stable), but `handleSessionSelect` identity changes on the keystroke - the referential-contract test fails on `main` and is expected to pass after the fix. Matched-row counts are the exact number of nodes handed to `SessionGroupSection`; because virtualization is disabled under search, that is also the number of rows the component mounts.

## Expected behavior

- A raw keystroke before the debounce re-renders only the search input/header; the list subtree receives no new search prop or callback identity.
- One effective (debounced) search update produces the result set.
- Large active and archived search results mount a bounded number of rows (threshold-based virtualization); small results keep normal DOM.
- Shift range selection works across virtualized/offscreen rows using logical visible order.
- Search matching semantics stay identical.

## Proposed fix

1. Keep raw query ownership in `SessionSidebar`; pass only the debounced normalized query down; expose a single stable `resetSessionSearch()` intent to row actions instead of raw query/setters/open-state.
2. Add a real memo boundary: memoize `SessionProjectCollection` and stabilize `topology`/`view`/`actions` with `useMemo`/`useCallback` (including `openWorktreesPage`). No custom comparator unless proven necessary.
3. Derive range-selection order from the logical render model instead of `document.querySelectorAll`, and feed it to the existing `useSessionMultiSelectStore.setRange`.
4. Generalize the existing archived virtualizer so large search results (archived and active) mount a bounded row window; keep current threshold behavior for small lists.
5. Only if profiling still shows the matcher as hot: collapse the double traversal into one pass returning `{ nodes, matchedCount }`, preserving current matching semantics exactly.

Deliberately not changing: search matcher semantics (multi-token AND, punctuation-insensitive compact matching, exact `ses_` ID mode, parent/child tree behavior, group/folder search, standalone managed chats), `buildGroupedSessions`, the project-section cache, DnD/folders, and unrelated Zustand subscriptions. The `ses_` exact-ID behavior shipped in #3428 and managed-chat search from #3212 are treated as canonical.

## Regression risks

- Search semantics drift (parity tests planned in the PR).
- Selection behavior across groups/scopes/descendants and with offscreen rows.
- Virtualization: nested/expanded sessions, search-forced expansion, folders, DnD, editing, context menus, scroll position, dynamic row heights.
- Hidden coupling where a component relies on raw query for its own UI.

## Acceptance criteria

- Raw keystroke: no `SessionProjectScroller` / `SessionGroupSection` / `SessionNodeItem` render caused by search-prop or callback identity change.
- Effective query: one search pass, bounded mounted rows for large result sets (active and archived), small results unchanged.
- Shift range selection uses logical row order and works with virtualized rows; anchors/scope/descendants preserved.
- Existing search parity/lifecycle behavior preserved (tests in the PR).
- Typecheck, lint, focused tests, and the repository build pass.

bashrusakh:
## Intent

Issue #3523 reported sidebar freezes while typing and searching large session workspaces. The final Candidate keeps raw search input local to the header, updates the session list only from the committed query, virtualizes the global search result model, and derives selection and DnD identity from the same render model that produces the rows.

The follow-up fixes also cover archived-folder authority, complete active owner-scope cleanup for search DnD, archived fallback ordering, sticky-header lifecycle and bounded state, live worktree metadata, goal/project-directory row semantics, shared header presentation, one archived-scope identity contract, and honest search-match counting.

Fixes #3523

## Non-goals

- No server, IPC, public API, dependency, worker, search index, global result cap, or product redesign changes. The persisted UI-state representation of folder-collapse state DOES change; see the persistence section.
- No change to matcher semantics or ordinary non-search folder behavior.
- No change to archived-folder ownership. Archived folders remain read-only derived projections.

## Persistence and migration

This Candidate changes one persisted UI-state representation, so the earlier "no persistence changes" claim was wrong and has been removed.

- **What changed:** `useSessionFoldersStore` now persists `collapsedFolderIds` (browser localStorage key `oc.sessions.folderCollapse`, and the `collapsedFolderIds` array inside the server disk snapshot posted to `/api/session-folders`) as **scoped identity keys** `scopeKey\0folderId` instead of bare folder ids. Folder ids are only unique inside their scope, and merged/flat groups plus archived scopes can legitimately carry the same folder id in several scopes, so collapse state needs the scope to stay unambiguous.
- **Migration:** reads go through `normalizePersistedCollapsed`. Scoped (`\0`-containing) values are kept verbatim. Legacy bare ids are migrated only when the id resolves to exactly one scope; an ambiguous legacy id that exists in several scopes **cannot be restored unambiguously** and is dropped (those folders reopen expanded once).
- **Multi-client compatibility:** an older client reading new-format data compares bare ids and matches nothing, so it shows folders expanded; its own writes (bare ids) are tolerated and re-migrated by a new client on the next read. Mixed old/new clients therefore do not corrupt data, but collapse state written by the old client cannot be displayed by it for keys it never wrote. This is the real cross-version risk if the same browser profile or server disk snapshot is shared between versions.
- There are no other persisted-format changes. Reverting the Candidate is a normal code revert.

## Affected surfaces

- `packages/ui` shared session sidebar, search, selection, folders, DnD, sticky headers, worktree row actions, and shared group/activity header presentation. Every effective changed file is under `packages/ui`.
- Web, Electron, VS Code, hosted mobile, and Capacitor mobile consumers of the shared UI.
- User-visible behavior includes committed-query search, virtualized offscreen selection, stable duplicate-row DnD, archived-folder protection, complete active-scope folder moves, sticky headers that remain correct while virtual rows mount and unmount, and search headers that now match normal-mode presentation (action tooltips, collapsed activity indicator, and the alwaysShowActions toggle affordance on tablet/mobile).
- The header result count keeps base's occurrence-based match semantics. See the behavior note below.
- The latest upstream extensions/SDK guest behavior is preserved through the refreshed Base; this PR does not add or change the SDK contract.

## Repository guidance

|---|---|---|
| `AGENTS.md` | Shared UI and multi-runtime behavior | Kept domain logic in owning sidebar modules and preserved runtime-neutral contracts. |
| `.agents/skills/openchamber-change-discipline/SKILL.md` | Source, exports, tests, persisted UI state, and module ownership changed | Kept the diff focused, added no new dependency, documented the persisted-state change and its migration, and ran package checks. |
| `.agents/skills/sync-state-invariants/SKILL.md` | Session selection, topology, archived authority, and live state | Used authoritative topology for folder cleanup, explicit incomplete-owner no-op, and bounded sticky lifecycle. |
| `.agents/skills/performance-engineering/SKILL.md` | Virtualization, memo boundaries, render hot paths, and measurements | Removed per-render full-model scans, preserved model identity contracts, and measured production builds before and after. |
| `.agents/skills/drag-to-reorder/SKILL.md` | Folder and session DnD occurrence identity | Unified row keys and occurrence drag keys; archived and incomplete scopes are not droppable. |
| `.agents/skills/theme-system/SKILL.md` | Shared header presentation extraction | Group/activity header geometry, icons, hover, PR presentation, and action placement now exist once in `groupHeaderPresentation.tsx`. |
| `packages/ui/src/components/session/sidebar/DOCUMENTATION.md` | Owning sidebar behavior and invariants | Followed the documented render-model selection, virtualization, loading, and archived-folder rules. |
| `packages/ui/src/stores/DOCUMENTATION.md` | Folder and selection store contracts | Preserved scoped folder identity, collapse ownership, and the documented migration behavior. |

## Validation

| Check | Result |
|---|---|
| Focused search counting suites (`sessionSearchRowModel`, `useSessionSidebarSections`, `sessionSearch.performance`) | `45/45` passed, 199 expect calls. |
| Header presentation batches (`SessionSearchRows.behavior`, `SessionGroupSection.behavior`, `recent/`, `useStickyProjectHeaders.behavior`) | `42/42` passed. |
| Selection/registry/multiselect suites (`sessionRowOrder`, `sessionRowOrderUtils`, `useSessionMultiSelectStore`) | Passed (see suite notes). |
| Folder/DnD/bulk suites (`folders/`, `useSessionFoldersStore`, `useSessionMultiSelectStore`, deletion cleanup) | Passed per file; see suite notes. |
| Full UI suite | `492/494` files. The two failures are not from this change: `MobileChangesSurface` times out identically on the pre-fix Candidate, and `useFilePreviewScrollPosition` passes standalone (load-order timing flake). |
| UI type-check (`tsc --noEmit`) | Pass. |
| UI lint (eslint) | Pass. |
| UI build / production web build | Pass (both used for the measurements below). |
| Targeted oxlint (anti-slop) | No findings on authored lines; pre-existing backlog untouched. |
| `bun run dead-code` | One new non-blocking hint: the exported `ARCHIVED_FOLDER_SCOPE_PREFIX` constant (part of the canonical archived-scope contract; consumed internally). |
| Suite notes | `bun test` with several suites in one process re-mocks `@/lib/desktop`/safeStorage, so mixed batches show pre-existing cross-file pollution (folders, keyboard, searchLifecycle, SessionNodeItem batches). Every affected file passes standalone, and the same mixed batches fail identically on the pre-fix Candidate. |
| `git diff --check upstream/main...HEAD` | Pass against Base `2a1a4c553`. |

## Performance evidence

Re-measured after the final changes (Candidate `8eb5194bd`) with the same methodology as the earlier evidence: real headless Chromium against production builds of both commits, isolated copy of a real workspace (3469 sessions, 270 roots, 197 directories), same server, viewport `1440x900`, query `e`, three interleaved runs per build. `packages/web` is byte-identical between the two measured commits; the only variable is `packages/ui`.

| Metric | Base `2a1a4c553` | Candidate `8eb5194bd` | Change |
|---|---|---|---|
| Mounted `[data-session-row]` after search | 2630 | 23 | 114x fewer |
| Unique sessions mounted | 2598 | 23 | 113x fewer |
| Longest long task | 13.9–14.6 s | 344–378 ms | ~38x lower |
| Elapsed to first DOM change | 13.9–14.6 s | 341–374 ms | ~38x faster |
| Mounted rows after scrolling to the end | 2630 | 24 | 110x fewer |
| Mounted rows after typing without Enter | 66 (unchanged) | 66 (unchanged) | draft-only input invalidates nothing |

Run-by-run:

| Run | Elapsed ms | Longest task ms | Mounted | Unique | Post-scroll |
|---|---|---|---|---|---|
| base-1 | 14006 | 14009 | 2630 | 2598 | 2630 |
| base-2 | 13868 | 13872 | 2630 | 2598 | 2630 |
| base-3 | 14600 | 14603 | 2630 | 2598 | 2630 |
| cand-1 | 358 | 364 | 23 | 23 | 24 |
| cand-2 | 374 | 378 | 23 | 23 | 24 |
| cand-3 | 341 | 344 | 23 | 23 | 24 |

- Base runs report `2751 matches`; candidate runs report `2717 matches` for the same query (see behavior note).
- `hasSearchVirtualContent` is `true` on the candidate and `false` on base; the candidate scrolls to `87758` and still mounts 24 rows, so the tail is reachable and bounded.
- Both builds pass through the same pre-search steady state (66 mounted rows) and apply the query only on Enter, so no run reports a draft-only result.

## Behavior note on the result count

The header count means "how many search matches the list contains". The final Candidate keeps base's per-group triple semantics — session matches + folder-name matches + one per matching group, occurrences not unique ids, exact-ID matches only for `ses_` queries — counted over the groups it renders, then adds every rendered Recent session row on top (base did **not** count Recent rows; base summed grouped-section triples plus the standalone managed-Chats group only), and folds base's per-group `+1` term into one per merged flat group, because base summed the grouped projection's triples while rendering the flat projection. Measured as `2717 matches` on the Candidate versus base's `2751` for the same dataset and query; the delta is exactly that composition difference. An earlier intermediate Candidate briefly reported unique session ids (2598); that semantics was reverted. The count is a search-match count again, not a unique-session count.

## Visual evidence

No theme or layout redesign is included. The change affects interaction, mounting, selection, DnD, sticky-header behavior, and header affordance parity (search group headers now carry the same tooltips, collapsed activity indicator, and alwaysShowActions icon swap as normal mode) rather than visual styling. A behavior test pins the alwaysShowActions parity via `data-gh-icon-*` markers.

## Risks and failure behavior

- Virtualized rows may be unmounted, but the render-model registry keeps offscreen selection and range operations available.
- Search folder DnD is disabled when active owner topology is incomplete and never mutates archived scopes; archived-scope detection now comes from one canonical identity module.
- Sticky fallback state is bounded to one selected header and clears on replacement, root loss, disable, and unmount.
- The persisted collapse-state representation changes with a read-time migration; ambiguous legacy ids are dropped, and old clients cannot see new-format collapse state. See the persistence section.
- Inherited from base, not introduced here: during active search, group and folder collapse toggles render force-expanded rows, so clicking one flips the persisted collapse state (`oc.sessions.groupCollapse` / `oc.sessions.folderCollapse`) invisibly until search closes. Both base and the Candidate behave this way. Deciding the search-collapse contract — make the toggles actually work in search, or disable them there — is a tracked follow-up outside this PR's frozen row-model scope.
- Sticky-header switching was additionally verified in a real Chromium (CDP, production Candidate build, desktop-runtime gate emulated, 3469-session dataset, broad query): with sticky zone headers enabled, both the normal sidebar and search mode switch the shared identity overlay at the same Chats→Recent boundary (normal: 120px, search: 120px in the paired run), the Recent header survives virtualizer eviction through the keep-until-replacement fallback, and projects take over through their own sentinels. A paired Base-vs-Candidate run of the same scenario (scroll down to a stuck project identity, then an instant jump back to scrollTop 0) shows both builds behave identically: the overlay briefly keeps the last stuck project identity for roughly one frame to 600 ms (the IntersectionObserver reports asynchronously), then returns to Chats — in 2x2 interleaved runs all four measurements settled at `chats`. No stale-overlay defect exists in either build.
- The remaining unmeasured surface is the mobile viewport, paint and compositor cost.


bashrusakh:
@yulia-ivashko

yulia-ivashko:
@deatheros please check this, since you're working on potentially related task.

deatheros (commented):
Thank you, Leonid. Your work on [#3536](https://github.com/openchamber/openchamber/pull/3536) provided an important foundation for our implementation in [sidebar-performance-tuning](https://github.com/openchamber/openchamber/tree/sidebar-performance-tuning), particularly the flat row model, global virtualization, and decoupling selection and DnD from the mounted DOM. We extended this direction into a unified renderer for the entire sidebar, but your investigation, benchmarks, and tests helped validate the architecture and significantly reduced the exploration required. We sincerely appreciate the quality and depth of your work.

PR #3549, opened 2026-09-13
fix(composer): isolate Android IME workaround

### LINKED ISSUE #3514: [Bug] Android Gboard correction causes caret jumps and sentence-fragment reinsertion

In the Android app with **Gboard**, tapping a word correction sometimes leaves the chat composer in a broken editing state:

1. Type a message containing a word Gboard offers to correct.
2. Tap the correction.
3. The caret moves unpredictably. When placing it elsewhere, a truncated part of the sentence inserts itself at that position.
4. Moving the caret to the end of the sentence no longer works.

Expected: only the corrected word changes, the rest of the draft stays intact, and the caret can move anywhere in the draft.

The occurrence is intermittent. These are the reported device symptoms, not a reliable automated reproduction.

### Where does it happen?

Native Android app, Capacitor/Android WebView, using Gboard. This is not the Electron runtime. Whether the same device also reproduces it in Chrome or the hosted mobile UI is not yet known.

### Version (if known)

Reported on the latest Android build; the exact installed app version, phone model, Android version, Gboard version and Android System WebView version still need recording.

Code investigated: `084eeabc3`, package version `1.23.0`, pinned `@codemirror/view@6.43.9`.

### Related issues and patches checked

- #2527 / merged #2691 address a related IME caret jump. Their `view.compositionStarted` guard is already present in `ComposerEditor.tsx`, before controlled-value writeback. Reapplying that patch will not address this report.
- #3013 / merged #3207 address CRLF selection coordinates. The composer already uses `replaceWithCaret`; this is not a CRLF report.
- [codemirror/dev#1456](https://github.com/codemirror/dev/issues/1456), "Android Webview: Text changes when cursor is moved", includes Gboard reports with very similar symptoms. It closed without an isolated CodeMirror reproduction or a fix.
- [codemirror/dev#1504](https://github.com/codemirror/dev/issues/1504) describes predictive text leaving an EditContext composition stuck. That report uses Samsung Keyboard, so it is related evidence rather than confirmation of this Gboard failure.
- [codemirror/dev#1518](https://github.com/codemirror/dev/issues/1518) describes Gboard failing to move to the document end. It does not establish that the two failures share a cause.
- [w3c/edit-context#111](https://github.com/w3c/edit-context/issues/111) discusses text/selection updates during active composition producing stale offsets.

No existing OpenChamber issue specifically covering this Gboard correction/reinsertion case, or an unapplied patch proven to fix it, was found.

### Investigation and evidence

Current CodeMirror uses native `EditContext` on supported Android Chromium versions. It handles composition there rather than through its fallback DOM composition handlers. A theory based only on the DOM `compositionend` delay therefore does not establish this report's cause.

An isolated browser experiment mounted both the **real controlled `ComposerEditor`** and **plain CodeMirror 6.43.9**. It used desktop Edge `152.0.4191.66`, an Android user-agent injected before CodeMirror loaded, native Chromium EditContext, and CDP IME commands. This is **not physical Android/Gboard validation**.

Sample draft:

```text
Android autocorrect should replace teh midword without truncating the remaining sentence and preserve the reachable end.
```

| Experiment | Observed result |
|---|---|
| Compose `the` over `teh` at offsets 35..38, then commit `the` | Both editors preserve the full 120-character sentence; caret 38. |
| `beforeinput` / `insertReplacementText` with correction text and a DOM target range | Both EditContext editors apply the correction correctly. |
| Correct `teh` to `their`, then compose elsewhere | Both preserve the corrected 122-character sentence. |
| Correct another word outside an active composition | Both preserve the changes. |
| Native mouse placement followed by composition | The composer and EditContext selection agree; no stuck caret observed. |
| Deliberately commit stale `teh` after replacing that active composition with `their` | Both produce `tehir`. This demonstrates sensitivity to stale input, but does **not** prove Gboard emits that event sequence. |
| Disable EditContext in an isolated diagnostic view and repeat the synthetic overlap using native DOM editing | Produces `theirteh` and leaves composition active. This does not support disabling EditContext as a fix. |

The real composer did not diverge from plain CodeMirror in these experiments. A React self-echo failure was not observed. The exact device symptom, especially repeated sentence fragments and the unreachable end, remains unreproduced.

### Screenshots / recordings

I will capture a recording from the affected Android device is still needed. A desktop simulation recording would not demonstrate the reported Gboard behavior.

frogconn:
## Intent

Fix #3514. Thai Gboard input could move the composer caret into the middle of the draft, reinsert a sentence fragment, and make the end of the draft unreachable in Chrome PWA.

The composer now opts out of CodeMirror EditContext. The workaround is scoped to the composer so other CodeMirror editors retain their current behavior.

## Non-goals

This PR does not update CodeMirror, change controlled writeback, alter selection rendering, or change the generic Files, Plan, or Skills editors. It also does not claim to fix every Android keyboard or IME issue.

## Affected surfaces

- `packages/ui` composer in Web, Chrome PWA, Capacitor, Electron, and VS Code webview.
- On Android Chromium, composer construction uses `ComposerEditorView`, whose `EDIT_CONTEXT` static is `false`.
- Generic `CodeMirrorEditor` instances used by Files, Plan, and Skills still construct the base `EditorView`, so they retain EditContext.
- No persisted or network contract changes.

## Repository guidance

|---|---|---|
| `AGENTS.md` and `openchamber-change-discipline` | Shared UI runtime behavior changed. | The diff stays in the composer owner, maps all `EditorView` constructors, and leaves unrelated editor consumers unchanged. |
| Composer `DOCUMENTATION.md` | It owns the CodeMirror composer lifecycle and mobile IME caveats. | Documents the Android workaround, diagnostic mode, and generic-editor boundary. |
| `performance-engineering` | Composer construction runs on an interaction path. | Adds only a static subclass property. It does not add input-path work, subscriptions, or allocations. |

## Validation

| Check | Result |
|---|---|
| `bun test ./packages/ui/src/components/chat/composer/editor/__tests__/ComposerEditorView.test.ts` | Passed |
| `bun test ./packages/ui/src/lib/imeTrace.test.ts` | Passed |
| `bun test ./packages/ui/src/components/chat/composer/editor/__tests__/writebackCompositionGuard.test.ts` | Passed |
| `bunx oxlint` on new IME source and test files | Passed |
| `bun run dead-code` | Completed. It reports existing unused exports and duplicates, with no new blocker from this subclass. |
| POCO F8 Pro, Android 16, Chrome 152.0.7977.83 PWA, Gboard 18.1.3 | Before: default composer reproduced #3514. Control: `?imeEditContext=off` did not reproduce. After this patch at the default URL: the issue no longer reproduced. |

The shared Capacitor, Electron, and VS Code runtimes were not run on hardware for this PR.

## Visual evidence

The before-state recording is attached to #3514. The current PR head was manually verified on the affected POCO device after deployment; no separate after-recording is attached.

## Risks and failure behavior

CodeMirror 6.43.9 checks `view.constructor.EDIT_CONTEXT !== false` when constructing a view. Setting the static property on the base `EditorView` would affect every later CodeMirror view, so this PR instead uses a composer-only subclass. Existing kept-alive composer views are constructed from that subclass before they enter the view store. The diagnostic query remains available for investigation and can still disable EditContext globally for that diagnostic page.


frogconn:
Before and after recordings from the affected POCO F8 Pro, running Android 16, Chrome PWA, and Gboard with Thai input.

### Before the fix

The original build reproduces the caret/text jump during Thai input.


https://github.com/user-attachments/assets/36944187-4057-4bed-b81e-f03262b19812



### After the fix

Tested the deployed build containing commit `5632a567`, without the `imeEditContext=off` query parameter. The caret/text jump no longer occurs in my testing.


https://github.com/user-attachments/assets/0edb9c49-929f-48b0-8b42-9ffc823d7c11



The fix disables EditContext only for the chat composer through a dedicated `EditorView` subclass. It does not change the default behavior of the Files, Plan, or Skills editors.

mattv8:
Yes! This is exactly the issue I was having!! Thank you for reproducing it and posting the evidence and for the PR!

PR #3550, opened 2026-09-13
fix(electron): reuse managed SSH remote servers across reconnects

theshasha:
## Intent

Every connect to a managed-mode SSH remote instance started a new OpenChamber server on the remote host and never recorded or stopped it. Reuse only happened for instances with a pinned preferred port; otherwise the port was a fresh `Date.now()` hash each time. A day of normal use left 5 orphan servers on my host, each with its own `opencode serve` child, and orphans can hold the private-relay host lock.

This PR persists the port of app-started servers and reuses the running one on reconnect:

- Ports are persisted per instance under a new settings root key `desktopSshManagedServers`. It sits outside `desktopSshInstances` because renderer saves rewrite that array from known fields and would drop the port, which is what broke #2954's `lastUsedPort`. `setInstances` prunes entries of deleted instances; malformed entries are dropped at the read boundary.
- Connect probes the preferred port, then the persisted port, and only starts (and persists) a new server when neither answers.
- A server found on the persisted port is adopted as ours. Disconnect stops it when keep-running is off. If its reported version differs from the app, or its persisted config tag (truncated SHA-256 of UI password and bind host, never plaintext) no longer matches, it is restarted on a fresh port via `openchamber stop --port`, which also works on password-protected instances. Without the tag check, a password change deadlocks every connect on a 401 from the client-token request.
- A server on the user-pinned preferred port is reused untouched, since it may be externally managed.

Supersedes the managed-server reuse part of #2954 (thanks @yulong-ge for the prior art). Differences: the port lives outside user-editable instance config, the stale server is stopped through the CLI instead of the HTTP shutdown route, and no changelog edits.

## Non-goals

- The version-manager PATH preamble from #2954 (unrelated failure mode).
- Cleaning up existing orphans or surfacing them in the Remote Instances UI. Both are follow-ups.
- Distinguishing auth-rejected probes from connection-refused; both still mean "not running" (shared follow-up with #2954).
- Retrying when a fresh random port is occupied (pre-existing).
- Serializing settings writes across Electron main, web server, and SSH manager (pre-existing last-writer-wins; see Risks).
- External mode, the keepRunning default, and Windows transport are unchanged.

## Affected surfaces

| Surface | Effect |
|---|---|
| `packages/electron/ssh-manager.mjs` | Managed-mode probe/adopt/restart/persist logic and the new settings key. |
| Desktop settings JSON | New optional root key, written only by the SSH manager. Old app versions preserve it on rewrite and ignore it; new versions tolerate its absence. Web-server settings saves carry unknown root keys through untouched. |
| Shared UI / web / VS Code / mobile | Unaffected: no renderer-facing contract, route, or SDK change. `startedByUs` has no UI consumer; its only behavioral reader is the disconnect stop-gate, where adoption is the intended semantics. |
| User-visible states | Reconnects skip the server-start phase and finish faster. A restart over a version or config mismatch goes through the normal server-start phase once, with the reason in the instance's SSH log. Keep-running off now stops the remote server on disconnect. |

## Repository guidance

|---|---|---|
| AGENTS.md runtime boundaries, `desktop-shell` skill | SSH server lifecycle is inherently native Electron behavior | All logic stays in `packages/electron/ssh-manager.mjs`; no shared-UI or web changes; stop uses the owning CLI lifecycle command. |
| `openchamber-change-discipline` skill | Persisted contract plus platform behavior | Additive settings key with both-direction compatibility, boundary parsing of malformed entries, and explicit failure behavior; smallest change that closes the leak. |
| `packages/electron/README.md` | Documents SSH invariants | Added the managed-server port persistence/reuse invariant, including the config-tag restart rule. |

## Validation

| Check | Result |
|---|---|
| `bun test packages/electron/ssh-manager.test.mjs` | 33/33 pass, incl. 19 new tests: probe order, adoption vs preferred-pin, stale/newer-version restart, `unknown`-version no-restart, stop-failure fallback, auth-rejection fresh start, malformed entries, prune on delete, and 5 config-tag tests. |
| `node scripts/run-isolated-tests.mjs packages/electron` | 23/23 test files pass. |
| `bunx oxlint` on both changed files, `node --check` | 36 findings, all pre-existing anti-slop backlog; new lines clean. Syntax pass. |
| Real SSH integration, before/after in one harness | A plain Node driver runs the real `ElectronSshManager` over real SSH to a dedicated test account, with a fresh manager per cycle (equivalent to an app restart). Unfixed code from `git show HEAD`: remote server count grew 0 -> 1 -> 2 -> 3 over 3 cycles. Fixed code: count stayed at 1 over 5 cycles, reconnects ready in 0.9s vs 2.1s, and keep-running-off disconnect stopped the server (back to 0). Adding a password mid-sequence and changing it both triggered a `configuration changed` restart on a fresh port with the count at 1; the new server then answered 401 for a wrong password, 401 for the previous password, and 200 for the current one. A live server in another account on the same host was untouched. |
| Not verified | Windows as the SSH client (transport code untouched; probe/persist ride the platform-agnostic `runRemoteCommand`). Packaged-build click-through (validation drove the manager class directly, not the packaged UI). |

## Visual evidence

No rendered-UI change; the diff alters Electron main-process bookkeeping and tests. The user-visible effect is fewer orphan remote processes. Process counts from the control/experiment pair in the same harness:

    unfixed code, 3 connect/disconnect cycles:
      after disconnect: 1, then 2, then 3 remote servers, each on a fresh random port
    fixed code, 5 cycles (fresh manager per cycle):
      constant 1 on the reused port; keep-running off stops it on disconnect, back to 0

Config edits now take effect on the next connect instead of silently not applying:

    password added at cycle 3:  "configuration changed" restart, new port 21313, count stays 1
                                auth on 21313: wrong password = 401, current = 200
    password changed:           restart on the next connect, new port 48090, count stays 1
                                auth on 48090: wrong = 401, previous = 401, current = 200

## Risks and failure behavior

- An auth-rejected probe counts as "not running": a fresh server starts and overwrites the persisted entry, leaving the old one running and untracked. This only affects setups that gate `/api/system/info`, which current servers answer publicly. Locked in by a test; follow-up planned.
- The config-tag check only runs on adopted servers. With a pinned preferred port the server is reused untouched, so a password change on such an instance still deadlocks on the 401 (pre-existing; the pinned port may name an externally managed server). Follow-up: adopt when the persisted tag matches, and report a targeted error when it is stale.
- A config-edit restart stops and starts only the remote OpenChamber server process; the remote host is never rebooted or otherwise affected. That one connect takes a few seconds longer.
- Settings writes are whole-root read-modify-write with no cross-writer serialization (established pattern). A lost update costs one port record and degrades to a single leaked server; the atomic write prevents file corruption.
- With keep-running off, an auto-reconnect now stops and restarts the remote server instead of leaking it. Semantics match the setting; reconnects are capped at 5 attempts.
- Adoption trusts the persisted port as ownership proof; a foreign OpenChamber landing on exactly that port would be adopted. The port is a time-hash over a 30000-wide range, so this needs a deliberate or very unlucky collision.
- A fresh start fails the connect if its random port is occupied (pre-existing; observed once in a self-SSH harness where both ends share one loopback namespace). The next attempt rolls a new port.
- A daemon that refuses to stop is left behind and its replacement starts on a fresh port, so it cannot block the connect.


btriapitsyn:
Thanks for working on this. #3711 now handles managed SSH server reuse through the remote CLI registry, including version changes and stopping adopted daemons when keep-running is disabled. Closing this PR as already addressed.

PR #3556, opened 2026-09-13
fix(ui): show failed subagents instead of Done

alvins82:
## Intent

The work-status panel treated every child session that was no longer busy as `Done`. When the OMP sidecar reported a child failure, that failure was collapsed into idle state, so the interface hid the reason and made a failed review look complete. This change keeps terminal failures in a directory-scoped live store and shows `Failed` with the provider message on the affected subagent row and section summary.

The paired sidecar change is [omp-openchamber-server#35](https://github.com/alvins82/omp-openchamber-server/pull/35), now merged. After review feedback, this PR also removes the stale `session.next.step.failed` documentation claim and makes the Spanish and Brazilian Portuguese failure summaries count-safe.

## Non-goals

This does not hydrate failure state from historical transcripts, change parent-session completion semantics, modify upstream OpenCode, or retroactively change an already-ended session. The sidecar must emit the failure event during the live run.

## Affected surfaces

- Shared `packages/ui` session synchronization and work-status components.
- Web, desktop, VS Code, hosted mobile, and Capacitor mobile surfaces through the shared UI package.
- All shipped locale dictionaries receive the new failure labels.
- The failure store is in-memory and scoped by runtime directory plus session ID; it adds no persisted or external data contract.

## Repository guidance

|---|---|---|
| `AGENTS.md` and `openchamber-change-discipline` | This is a shared UI source and contract change. | Keeps the fix focused, preserves unrelated worktree edits, updates owning docs, and runs focused validation plus dead-code analysis. |
| `sync-state-invariants` | Failure state is derived from live session events and must reconcile with busy and deleted states. | Records `session.error` live, clears on a new busy attempt, scopes by directory, and clears on deletion; idle alone never means success. |
| `theme-system` | The work-status row needs a distinct failure presentation. | Reuses the existing `WorkStatusValue` error tone and title behavior. |
| `locale-ui-patterns` | `Failed` and the failure summary are user-facing strings. | Adds both keys to English and every shipped locale, with count-safe summaries in Spanish and Brazilian Portuguese. |
| `packages/ui/src/sync/DOCUMENTATION.md` and `packages/ui/src/components/chat/work-status/DOCUMENTATION.md` | These files own the synchronization and work-status data contracts. | Documents the live failure store and its work-status consumer. |

## Validation

| Check | Result |
|---|---|
| `bun test packages/ui/src/sync/session-failure-store.test.ts packages/ui/src/sync/session-error-log.test.ts packages/ui/src/lib/i18n/messages.test.ts` | 9 passed, 0 failed. |
| `bun run type-check` | Passed for mobile, UI, web, Electron, VS Code, and the root package. |
| `bun run dead-code` | Exit 0; reports the repository's existing unused files/exports and does not report the new failure store as unused. |
| `bunx oxlint packages/ui/src/sync/session-failure-store.ts packages/ui/src/sync/session-failure-store.test.ts packages/ui/src/components/chat/work-status/WorkStatusSubagentsSection.tsx` | Passed. |
| Broader oxlint over all touched TypeScript files | Reports pre-existing anti-slop findings in the large existing `sync-context.tsx` and `WorkStatusPrimitives.tsx` files; the new store and subagent-row changes have no findings. The newly updated locale files retain their pre-existing open-dictionary findings. |
| `git diff --cached --check` | Passed. |

## Visual evidence

Codex Image 14 Sept 2026, 11_40_19

## Risks and failure behavior

Failure state is intentionally live and in-memory, so old sessions do not change retroactively and a reconnect does not reconstruct a failure from persisted history. A later busy event clears the old failure for a new attempt, and session deletion removes it. If the sidecar does not emit the failure event, the UI retains its existing status behavior; the paired PR supplies that event.


alvins82:
Added screenshot evidence.

bashrusakh:
@alvins82 Took another look at this, including the paired sidecar change. Overall I think the layering is right: the sidecar now preserves the actual terminal outcome and emits `session.error` before settling the child to idle, while the UI consumes that authoritative signal instead of trying to infer failure from “not busy”.

A separate failure store also makes sense to me here. I wouldn't put `failed` into `global-session-status`: that store seems specifically built around live/activity state (`busy | retry`, with absence meaning idle), while failure is a terminal outcome. Mixing those would give one store two different meanings.

For the same reason I don't think `session-error-log` should be reused for this either. That's diagnostic history with a bounded buffer, whereas this state is reactive and affects what the UI currently shows.

There is one lifecycle case that still looks missing to me though: `session-failure-store` is module-global and keyed by `directory + sessionId`, but it isn't cleared by `resetAppForRuntimeEndpointChange()`.

So a failure from runtime A can theoretically survive an endpoint switch and be picked up in runtime B if the same directory/session key exists there. The other runtime-scoped sync state is explicitly reset on a runtime change, so I think this store should follow the same rule — probably a small `resetSessionFailureStore()` / `clearAllSessionFailures()` called from `resetAppForRuntimeEndpointChange()`.

I'd also add coverage one level above the store itself. The current unit test proves record/scope/clear, but the important behavior is really the event ordering:

```text
session.error -> idle
```

should still render as failed, while:

```text
session.error -> idle -> busy
```

should clear the old failure for the new attempt.

A runtime-endpoint-switch regression test would cover the leak above as well.

The in-memory/live-only behavior seems fine to me as long as that's intentionally the scope of this fix, which the PR already says it is. If failures are ever expected to survive reload/reconnect and still render as Failed from history, I'd rather see that reconstructed from durable/server-side state than patched with more UI heuristics.

One small point on the suggestion to cap `session-failure-store` like `session-error-log`: I'm not sure I'd do that blindly. The error log is history, so evicting old entries is harmless. This store represents current UI outcome state; evicting an older but still-visible failed child could make it fall back to Done again. Lifecycle cleanup (deletion/runtime reset/etc.) seems safer than an arbitrary history-style cap.

Other than the runtime reset gap, this looks like the right level for the fix to me.

alvins82:
@bashrusakh Fixed in `5135c97c9`.

- Added `resetSessionFailureStore()` and called it from `resetAppForRuntimeEndpointChange()`, so live failure records cannot cross runtime switches.
- Added `handleEvent` coverage for `session.error -> idle` retaining the failure and `session.error -> idle -> busy` clearing it for the next attempt.
- Added an endpoint-reset regression test and documented the runtime-switch lifecycle.

I left the store uncapped because it represents current visible outcomes, not bounded error history. Focused tests, UI type-check/lint, docs validation, and oxlint for the new files pass.

bashrusakh:
@alvins82 Thanks — this looks good now. The runtime-switch reset is at the right lifecycle boundary, the new event-order coverage closes the case I was concerned about, and I agree with keeping the failure store uncapped since it represents current UI outcome state rather than history.

One small nit I noticed while tracing the switch path: the old event pipeline flushes any queued events during cleanup, which happens after the runtime reset has already cleared the failure store. In theory, a queued `session.error` from the previous runtime could therefore repopulate the module-global failure store after the reset.

I don't think this should block the PR. For it to become user-visible in the new runtime, the new runtime would also need a matching directory + session ID, so this looks more like a defensive runtime-isolation edge case than a practical bug.

If you happen to touch this area again, a stale-runtime guard around failure-state mutation would make the invariant fully airtight. Otherwise, I'd leave it as a nit.

Nice follow-up overall — the original lifecycle gap is addressed.

alvins82:
@bashrusakh Fixed and pushed in 38e6f9f27.

Failure-store writes now require the event pipeline's runtime key to match the active runtime, so queued events flushed by an old pipeline after a runtime switch cannot repopulate or clear the new runtime's failure state. I also added regression coverage for stale record and clear events.

Validation: focused sync tests pass (18 tests), UI type-check passes, and UI ESLint passes.

alvins82:
Updated the PR body with the requested evidence: current results for `runtimeEndpointReset.test.ts` (1 passed) and `session-switch-resync.test.ts` (16 passed, including the stale-runtime guards), plus a live-run note for Web HMR on macOS 27.0 arm64. The local managed runtime does not include the merged OMP sidecar, so I could not stage the child-failure event live; the automated suite covers that transition. No source changes were needed for this evidence gap.

PR #3561, opened 2026-09-14
fix(npm): honor user registry for plugin metadata

Ex0ry:
## Intent

Use existing user npm configuration for plugin metadata checks instead of assuming the public registry.

## Non-goals

- Plugin installation and update behavior.
- The separate VS Code registry lookup path.

## Affected surfaces

- `packages/web` server plugin metadata checks used by Desktop.
- Existing npm configuration is read only.

## Repository guidance

|---|---|---|
| `AGENTS.md` and `CONTRIBUTING.md` | Define source boundaries and validation workflow. | Kept the change in the server integration layer and ran focused and web validation. |
| `openchamber-change-discipline` and `ui-api-decoupling` | The plugin state is driven by a server metadata request. | Changed only the server resolver; UI contracts are unchanged. |
| `packages/web/server/lib/opencode/DOCUMENTATION.md` | Documents module ownership. | Added the resolver entry. |

## Validation

| Check | Result |
|---|---|
| `bun test "packages/web/server/lib/opencode/npm-registry-config.test.js" "packages/web/server/lib/opencode/npm-registry.test.js"` | Passed: 26 tests |
| `bun run type-check:web` | Passed |
| `bun run lint:web` | Passed |
| Local source-server plugin check | Passed |
| `bun run dead-code` | Not run: the environment registry returned HTTP 401 before analysis started |

## Visual evidence

No screenshot is attached because a real verification view would expose private configuration or session data. UI layout and styling are unchanged.

## Risks and failure behavior

- Reads existing npm configuration only when metadata is requested; it never writes configuration.
- Matching bearer/basic credentials are sent only in request headers, never in package URLs.
- Invalid configuration fails the lookup rather than falling back to the public registry.

## Related work

Related to #2296. Complements #2348.


Thijmen:
@btriapitsyn What can we do to get this prioritized? It's quite annoying that we can't load private registries, this worked fine with OpenChamber 1 by the way.

PR #3564, opened 2026-09-14
fix(git): continue rebases and merges without GIT_EDITOR rejection

anandghegde:
## Intent

Continue in the Git view's "Rebase in progress" and "Merge in progress" banner never works right now. After resolving conflicts, clicking it shows `Use of "GIT_EDITOR" is not permitted without enabling allowUnsafeEditor` and the operation stays in progress. The only way out is a terminal.

The cause is in `packages/web/server/lib/git/service.js`. `continueRebase` and `continueMerge` call `git.env('GIT_EDITOR', 'true')` before `rebase --continue`, `rebase --skip` and `commit --no-edit`. simple-git 3.36.0 (the version in `bun.lock`) refuses that variable unless the instance enables `allowUnsafeEditor`, and `createGit` never does.

Enabling the flag alone is not enough. I tried it: `createGit` does not apply its `env` today (the constructor option is ignored), so `.env(name, value)` leaves Git with an environment holding only `GIT_EDITOR`. No `HOME`, so no global identity, and the continue fails with `Committer identity unknown`. Giving simple-git a full env copy instead brings in its environment scan, which rejects inherited `PAGER` or `GIT_ASKPASS` (the blocker the review on #3364 found).

So this change runs `rebase --continue` and `rebase --skip` through the existing direct `runGitCommand` path, with the normal `buildGitEnv()` environment plus `GIT_EDITOR=true`. `runGitCommand` takes an optional env override for that. `commit --no-edit` never opens an editor, so the merge path just drops the override and stays on simple-git. Conflict and "nothing to commit" detection is unchanged. The thrown message now carries Git's stderr and stdout, which the existing `includes` checks already expect.

After the fix, Continue finishes the rebase or merge, shows the success toast, and the panel refreshes to a clean tree.

## Non-goals

- `createGit`, `buildGitEnv` and how simple-git receives its environment. #3364 (Windows MSYS2 paths) rewrites that part and applies the env through `.env()` with a filter. This PR stays clear of those lines: `service.js` merges cleanly with #3364's branch, and the only conflict in that trial merge is one #3364 already has with `main` in `service.test.js`.
- Abort, start rebase and start merge, which never set `GIT_EDITOR`.
- The VS Code runtime. `packages/vscode/src/gitService.ts` already runs `rebase --continue` with its own `execGit`.

## Affected surfaces

- `packages/web` Git service (`continueRebase`, `continueMerge`), behind `POST /api/git/rebase/continue` and `POST /api/git/merge/continue`.
- Those routes serve the web UI, Electron (in-process server) and hosted/Capacitor mobile clients, so the fix reaches all of them. VS Code uses its own Git bridge and is not affected.
- No persisted data, route shape or response contract changes. `runGitCommand` is an internal helper; its new third argument is optional and existing callers are untouched.

## Repository guidance

|---|---|---|
| Root `AGENTS.md` | Source change in the web server | No dependencies, no changelog edits, change kept inside the owning Git module |
| `openchamber-change-discipline` | Executable source change | Smallest complete fix in `service.js`, reuses the existing `runGitCommand` helper instead of a new layer, focused regression tests plus package type-check, lint and tests |
| `ui-api-decoupling` | The UI reaches this through `/api/git/*` | Fix stays server-side. No UI workaround, route or contract change |
| `packages/web/server/lib/git/DOCUMENTATION.md` | Nearest module docs | Public behavior matches what it already documents for `continueRebase`/`continueMerge`, so no doc change was needed. Tests use real temporary repositories like the existing `cherryPick` suite |
| `communication-style` | PR text | Plain description of cause, fix and evidence |
| `CONTRIBUTING.md` and PR template | Review handoff | Before/after UI evidence and exact validation results below |

## Validation

| Check | Result |
|---|---|
| New tests `continuing a conflicted rebase or merge` in `service.test.js` (finish rebase, report unresolved rebase conflict, finish merge). They set `GIT_EDITOR=false`, `PAGER=less` and `GIT_ASKPASS=false` in the process env | Against unmodified `service.js`: 3 failed with `Use of "GIT_EDITOR" is not permitted without enabling allowUnsafeEditor`. With the fix: 3 passed |
| `bun run --cwd packages/web test server/lib/git` | 5 files, 145 tests passed |
| `bun run --cwd packages/web test` (full web suite) | 194 files, 2223 passed, 1 skipped. One earlier run had a single failure in `server/lib/linear/routes.test.js`. That file passed 4 of 4 times on its own and the full rerun passed, so it looks flaky and unrelated |
| `bun run --cwd packages/web type-check` | Passed |
| `bun run --cwd packages/web lint` | Passed |
| `bunx oxlint` on `service.js` and `service.test.js` | 46 findings before and after, same set. All in existing `service.js` code, none in changed lines |
| `node --check` on both files, `git diff --check` | Passed |
| Node script calling the exported service functions on temp repos, with identity only in a temp global `~/.gitconfig` | Before: `continueRebase` and `continueMerge` threw the `GIT_EDITOR` error and `getStatus` still reported the rebase/merge in progress. After: both returned `{ success: true, conflict: false }`, status was clean, and the log showed the rebased commit and `Merge branch 'feature'`. Same result with `GIT_EDITOR=false`, `PAGER=less` and `GIT_ASKPASS` set |
| Built web UI (`bun run build:web`, `openchamber serve --foreground`, isolated `HOME`), Git panel, Continue on a resolved rebase and a resolved merge, headless Chrome | Before: error toast shown below. After: "Rebase step completed" / "Merge completed" toast and a clean working tree |
| Trial merge of this branch with #3364's head | `service.js` auto-merged. The only conflict was in `service.test.js`, and merging #3364 into `main` without this change conflicts there too |

Not verified: Windows, the Electron shell itself, mobile clients, and a rebase with several conflicting commits in a row through the UI.

## Visual evidence

Desktop web, light theme. The banner and the toasts are unchanged UI; the difference is the result of pressing Continue.

Before, rebase:
openchamber-rebase-before

After, rebase (banner before clicking, then result):
openchamber-rebase-banner
openchamber-rebase-after

Before, merge:
openchamber-merge-before

After, merge:
openchamber-merge-after

No layout or styling changes, so narrow and dark states are not included.

## Risks and failure behavior

- A failed continue behaves as before. Git leaves the rebase or merge in progress, conflicts come back as `{ conflict: true, conflictFiles }`, and other errors throw to the route, which returns 500 with Git's message.
- `rebase --continue`/`--skip` now inherit the same environment as other direct Git commands here (`buildGitEnv()`, including `SSH_AUTH_SOCK` resolution and, once #3364 lands, its Windows `MSYS=noglob`). `GIT_EDITOR=true` overrides any editor the server inherited, so no editor can open or hang.
- These two commands skip simple-git's argument checks. The arguments are fixed literals, so no user input reaches them.
- Rollback is a plain revert. No stored data is involved.


anandghegde:
Pushed 4ff63f58c for the skip fallback the review flagged.

That branch could never run. When `rebase --continue` stops with "No changes", git's hint text also says "Resolve all conflicts manually", so the conflict check matched first and the UI got a conflict with an empty file list. The empty-commit check now runs first. If the `rebase --skip` that follows stops on a conflict in the next commit, it returns `conflict: true` with the conflicted files instead of success.

The new test uses `rebase.backend=apply`, which is the backend that stops on an emptied commit. The default merge backend drops it on its own. The test fails on the previous commit and passes now, and the git service suite passes (121 tests).

The before/after screenshots are now in the PR description under "Visual evidence". These are server-side changes only, so they still apply.

anandghegde:
Merged main to resolve the conflict with the new `runGitCommand` timeout option (dabfe7ab9): it now takes `{ timeoutMs, env }`. The git service tests pass.

Issue #2318, opened 2026-07-19
[Feature Request] LaTeX rendering


happyfish586:
### What should we add/change?

OpenChamber is used by developers and researchers who frequently discuss mathematical formulas, technical notation, or scientific content with AI agents. Currently, the chat interface displays LaTeX code as raw plain text (e.g., $E=mc^2$ or $$\int_{0}^{\infty} e^{-x^2} dx$$), which makes mathematical expressions difficult to read and interpret. This significantly reduces the readability of AI responses that contain technical content.

### Describe the solution you'd like

I would like OpenChamber to render LaTeX expressions in chat messages, similar to how many other chat and documentation platforms handle mathematical notation. Specifically:

Inline LaTeX enclosed in $...$ should be rendered inline within the text flow.

Block LaTeX enclosed in $$...$$ or \[...\] should be rendered as centered, standalone display math.

The rendering should happen automatically when messages are displayed, without requiring user interaction.

### Expected behavior

When a chat message contains LaTeX notation, it is rendered as properly typeset mathematics.

Rendering should not interfere with other Markdown features (code blocks, lists, etc.) already supported by OpenChamber.

For performance, rendering should happen off the main thread where possible, similar to how code-block highlighting was optimized.

makeittech:
Closing as superseded by existing KaTeX support. Chat already renders display math with $$...$$ and \[...\], and inline math with \(...\). Single-dollar $...$ is intentionally not supported because it collides with currency text. If you need $...$ specifically, please open a focused follow-up; otherwise try the supported delimiters.

Issue #1880, opened 2026-06-28
[Bug] Math formulas rendered in upright (roman) font instead of italic — KaTeX font files not loaded from CSS @import


xuexi-xiaobai:
## What's wrong?

Variables in LaTeX math formulas (e.g. \(e\), \(x\), \(\pi\)) are rendered in upright (roman) font instead of italic. Standard mathematical typography requires variables in italic.

Affected formulas include:

- Block math (`$$...$$`) — e.g. `$$ e^{i\pi} + 1 = 0 $$`
- Inline math (`\(...\)`) — e.g. `\(E = mc^2\)`
- Display math (`\[...\]`)

## Where does it happen?

**Versions**: v1.13.2 through v1.13.6 (regression — v1.12.4 was correct)

**Why v1.12.4 was correct**: The KaTeX CSS was loaded via JavaScript `import 'katex/dist/katex.min.css'` in `MarkdownRendererImpl.tsx`. Vite's JS-import pipeline correctly resolves the CSS `url()` font file references — it copies the `.woff2` font files to the build output and rewrites the URLs.

## Root cause

In v1.13.2 the KaTeX CSS import was moved from a JS `import` to a CSS `@import` inside `packages/ui/src/index.css:11`:

```css
@import "katex/dist/katex.min.css";
```

The CSS `@import` is processed by Lightning CSS (Tailwind v4's built-in CSS processor). Lightning CSS inlines the CSS content into the main bundle, **but does not resolve the relative `url()` references** to KaTeX font files. The KaTeX CSS contains `@font-face` blocks such as:

```css
@font-face {
  font-family: KaTeX_Math;
  font-style: italic;
  src: url(fonts/KaTeX_Math-Italic.woff2) format("woff2");
}
```

After inlining into the main CSS bundle, the relative path `url(fonts/KaTeX_Math-Italic.woff2)` no longer points to the correct location. The font files are not copied to the build output, and the URLs resolve to 404 at runtime.

**Consequence**: The KaTeX `@font-face` declarations are never satisfied → browser falls back to the parent container's font (IBM Plex Mono via `--font-sans`) → all math text appears upright.

## Additional context

[...]

PR #3592, opened 2026-09-15
fix(chat): render single-dollar inline math and unescape entity-corrupted display math

Dawnfz-Lenfeng:
## Intent

Dollar math in chat was broken in two ways:

1. **Single-dollar inline math never rendered.** Messages like `$y$：$n\times 1$ 观测向量` reached the chat as literal text — single-dollar was intentionally unsupported after #2318 because of currency collisions, so `$y$`, `$X$`, `$\beta$` all displayed verbatim.
2. **Display math containing `'` or `&` rendered as red soup.** `$$S(\beta)=|y-X\beta|^2=(y-X\beta)'(y-X\beta)$$` passes through marked first, which escapes the rendered text (`'` → `&#39;`, `&` → `&amp;`, `<` → `&lt;`). The `$$` pass then fed those entities to KaTeX as LaTeX, the parse failed, and KaTeX emitted the raw source as a red `.katex-error`: `S(\beta) = … (y-X\beta)&#39;(y-X\beta)`.

This PR renders both dollar forms from the parsed HTML, in one pass over text runs:

- `$...$` inline math, guarded against currency prose:
  - the opening `$` must be followed by a non-space, non-`$` char;
  - the closing `$` must be preceded by a non-space and not followed by a digit;
  - content never contains `$`/`<`/`>`/`"`, so a pair cannot reach across markup or into an attribute (an href may legitimately hold `$`);
  - purely numeric content (`$100$`) stays text.

  → `US$ 680`, `$5 and $10`, `$50M to $72M`, `价格 $100 到 $200` survive as literal text (test-covered), while `$y$：$n\times 1$` renders typeset inline math.
- `$$...$$` display math: the captured LaTeX is HTML-unescaped before KaTeX, so a transpose apostrophe or an `aligned` ampersand parses correctly.
- Both forms are matched by a single alternation so KaTeX output is never rescanned by the same pass.
- Math now runs **per text run** (markup boundaries excluded), mirroring how KaTeX auto-render walks DOM text nodes. This also ends the old behavior where a `$$` pair could match across `</p><p>` with the tags themselves fed to KaTeX as LaTeX — an orphaned opener now stays literal.

`\(...\)` and `\[...\]` keep their existing marked-extension path; behavior there is unchanged.

## Non-goals

- No new delimiters (`\begin{align}` outside `$$`, AMS environments, etc.).
- KaTeX error styling is unchanged: genuinely invalid LaTeX still renders as a red `katex-error`; a follow-up could decide on a different fallback.
- `\$` escapes cannot be honored post-parse (marked has already consumed the backslash by then); noted in code. Currency prose is protected by the guards above instead.
- Not touching the KaTeX font-loading regression tracked in #1880 (upright math fonts on web).

## Affected surfaces

- `packages/ui` — chat markdown rendering only (`markdownCore.ts` math pass).
- All runtimes that render chat messages (web, desktop, VS Code, hosted and Capacitor mobile) consume `packages/ui` and pick this up uniformly; no runtime-specific branch exists or is needed.
- No persisted data, routes, IDs, public exports, or package contracts change — pure render-path fix. `markdownCore.ts` exports are untouched.

## Repository guidance

|---|---|---|
| `AGENTS.md` always-on constraints | Source change in packages/ui | No dependencies added; no changelog edits (maintainer's release-time work); change scoped to two files |
| `openchamber-change-discipline` skill | Source modification | Classified as local implementation risk: private helper behavior in one package; smallest complete change; validated at the owning package level; the existing public behavior of `\(...\)`/`\[...\]` is preserved (tests unmodified and passing) |
| `performance-engineering` skill | The math pass sits on the streaming render hot path | The `indexOf('$') === -1` early return is untouched, so non-math streaming blocks still skip the pass entirely; math-bearing blocks add one split per text run. Adversarial regex benchmarks below establish no backtracking risk — relevant given the catastrophic-backtracking history kept alive by the Linear-comment regression test |
| Nearest `DOCUMENTATION.md` / README | Module discovery rule | No `DOCUMENTATION.md` or README documents the markdown math pass; the in-file section comments are the owning documentation and were updated to match the new behavior |

## Validation

| Check | Result |
|---|---|
| `bun test packages/ui/src/components/chat/markdown/markdownCore.test.ts` | 33 passed, 0 failed (27 pre-existing incl. the Linear backtracking regression, 6 new) |
| `node scripts/run-isolated-tests.mjs packages/ui/src/components/chat/markdown` | 7/7 test files passed |
| `bun run --cwd packages/ui type-check` | Clean (`tsc --noEmit`, no errors) |
| `bun run --cwd packages/ui lint` | 0 errors; 1 warning in `MobileChangesSurface.tsx` — pre-existing, untouched file |
| `bunx oxlint` on both changed files | No findings in changed regions; remaining findings pre-exist in untouched code (extensions cast, image renderer spread, DOMPurify cast, test helpers) |
| Adversarial regex benchmark (new `MATH_DOLLAR_RE`) | Linear, no catastrophic backtracking: 5000 unpaired dollars 10KB → 0.19ms; 200KB single unpaired run → 0.32ms; 50k repeated `$x` 100KB → 1.34ms; 100KB unclosed display → 0.22ms; ~180KB realistic mixed → 0.27ms |
| Before/after render of a full compound message (worktree at HEAD vs this branch) | Before: inline math literal, 2 red `katex-error` blocks with raw `&#39;`. After: 8 KaTeX spans (3 display + 5 inline), 0 errors, no residual `&#39;`; currency, code span, and href `$` untouched |

Not verified: no runtime check on Electron/VS Code/mobile shells (the change is the shared render pass itself, exercised through `renderMarkdownSync`/`renderMarkdownBlocks`, which every shell consumes); no long-session memory measurement (block caches untouched).

## Visual evidence

`renderMarkdownSync` output for an English version of the reporter's message (classical linear regression: model, loss, normal equations), generated from shared repo checkouts — HEAD `6e58589` for before, this branch for after. The harness bundles KaTeX's own stylesheet and fonts, so the capture shows true production typography (no CDN dependency). Font-shape caveat for web production builds remains tracked in #1880.

**Before (upstream/main 6e58589):**

- Inline math arrives as literal text: `$y$: $n\times 1$ vector of observations` — every variable shows with visible dollar signs.
- Display math with a transpose renders as a red raw-source block: `S(\beta) = |y - X\beta|^2 = (y-X\beta)&#39;(y-X\beta)` and `\frac{\partial S}{\partial\beta} = -2X&#39;(y-X\beta) = 0 … \boxed{X&#39;X\beta = X&#39;y}`.

**After (this PR):**

- The same list items render typeset: `y: n×1 vector of observations`, `X: n×(p+1) design matrix` — no visible delimiters.
- The same display formulas render centered KaTeX with real transpose primes: `S(β)=∣y−Xβ∣²=(y−Xβ)′(y−Xβ)` and `∂S/∂β = −2X′(y−Xβ) = 0 ⟹ X′Xβ = X′y` (boxed).
- Currency prose (`about $5 thousand and $10 thousand per run`), the code span `` `echo $PATH` ``, and `$` inside a link href all remain literal.

Side-by-side capture (left: before, right: after):

before/after dollar math

The interactive self-contained demo (`demo.html` + `katex.min.css` + `fonts/`, no external requests) is on the same asset branch: https://raw.githubusercontent.com/Dawnfz-Lenfeng/openchamber/pr-assets/3592/demo.html

## Risks and failure behavior

- **Currency false positives:** a `$x$` pair in prose that the author meant literally now renders as math. The four guard rules keep known money shapes as text (test-covered); the remaining exposure is deliberate math-looking prose, which is what the feature request in #2318 asked for.
- **Hot path:** unchanged for the no-dollar case (early return). For math-bearing blocks the added cost is one `split(TAG_RE)` + join per text run; measured in the adversarial benchmark above.
- **Failure mode:** KaTeX parse failure still degrades to the existing red `katex-error` (explicitly unchanged); `renderKatex` already falls back to the raw text if `renderToString` ever throws.
- **Security:** the tag-run split makes the pass strictly tighter than before — math can no longer be matched into attributes or across markup; DOMPurify sanitization and the code/pre/kbd exclusion are unchanged.
- **Rollback:** single-commit revert; no persisted state, no migration, no contract change.


yulia-ivashko (approved):

Dawnfz-Lenfeng:
Complementary note: the app itself was still rendering math in a fallback typeface because the KaTeX font files were never emitted into the build (openchamber/openchamber#1880). That is fixed separately in #3669. The before/after images here are rendered with KaTeX's stylesheet bundled, so they already show the intended typefaces.

PR #3612, opened 2026-09-16
fix(web): merge session-folder snapshots per scope instead of whole-file replace

### LINKED ISSUE #3611: Session folders are lost across devices: stale device replaces the whole stored snapshot
### Summary

On an instance accessed from multiple devices (desktop browsers, phones, PWA), user-created session folders are silently lost. Two visible symptoms:

1. Folders created on one device never reach the others, so users end up recreating them per device.
2. Folders disappear after upgrading, even though the server-side data file survives upgrades untouched.

### Root cause (traced on v1.23.2)

- The client persists its **whole foldersMap** on any folder mutation: `packages/ui/src/stores/useSessionFoldersStore.ts` (`schedulePersistToDisk` POSTs the entire snapshot, stamped with wall-clock `Date.now()`).
- Hydration from disk runs **once at bootstrap**; there is no re-sync before a mutation, and a device whose bootstrap GET failed (e.g. pre-unlock 401) or whose tab predates a change never sees the newer state.
- The server (`packages/web/server/lib/session-folders/routes.js`, POST handler) accepts the incoming snapshot **only as a whole-file replacement** gated on `updatedAt` — no merge, no tombstones.

So any stale writer — a long-open tab, a device that failed hydration, or a clock-skewed client — replaces the entire stored map with its older view on its next mutation. Upgrade-forced reloads reshuffle which device "wins" next, which is why it also looks like upgrades lose folders.

### Evidence (captured live on a self-hosted v1.23.2 instance)

- `11:08` — folder `coco` created on device A (scope `/home/yyj/ai`, 2 sessions); present in `sessions-directories.json` at `11:09`.
- `11:32:26` — another device POSTed a snapshot that did not contain that scope; the whole file was replaced and the folder was gone. Device A's next reload dropped it locally as well (server state is adopted at bootstrap).

### Reproduction

1. Device A creates a folder and adds a session.
2. Device B (tab opened before the folder existed, or otherwise stale) moves any session into any folder.
3. A's folder is erased server-side; every device that rehydrates loses it too.

### Suggested fix

Treat absence as "not seen", not "deleted": the server should **merge snapshots per scope** — keep scopes/folders the writer never saw, let the incoming version win per folder id — instead of whole-file replacement. That eliminates the data loss with a server-only change.

The complete fix is per-folder revisions plus tombstones (or explicit delete events), which makes cross-device deletion unambiguous; that requires a client+server protocol change.

I have a server-side per-scope merge implemented with regression tests and can open a PR.

ouyangjian28:
## Intent

Fixes #3611.

A `POST /api/session-folders` body is one device's **whole foldersMap**. Clients hydrate from disk only at bootstrap and stamp updates with wall-clock `Date.now()`, so a stale writer (long-open tab, failed bootstrap GET, clock skew) replaces the stored snapshot with its older view on its next mutation and erases every folder it never saw — losing user data across devices.

This PR changes the POST handler from whole-file replace to a **per-scope merge**:

- scopes/folders the incoming snapshot does not mention are **kept** (absence is not deletion);
- folders in shared scopes are unioned by id, with the incoming version winning per folder (renames/session-list edits propagate);
- same-name folders with different ids (independently auto-created archive folders on two devices) have their session lists unioned;
- `updatedAt` stays monotonic (`max` of both sides) so the client hydration gate is unaffected.

The previous `ignored: true` short-circuit for older-stamped writes is removed: an offline device's delayed write still carries folders the server lacks, so it merges instead of being discarded.

## Non-goals

- **Cross-device folder deletion.** Without per-folder revisions or tombstones, "deleted on one device" is indistinguishable from "never seen by this writer" in a whole-map POST, so deletions do not propagate (a folder removed locally reappears after the next hydration). Making deletes unambiguous needs an explicit protocol change (client + server) and is deliberately out of scope; this PR stops the data-loss direction, which is the one destroying user folders today.
- Client-side hydration improvements (re-hydrate on focus, etc.) — the server-side defense protects against every stale writer regardless of client vintage.
- The `ignored: true` response field is gone. The client persists fire-and-forget and does not branch on it.

## Affected surfaces

- `packages/web` — `POST /api/session-folders` only. `GET` is untouched. Persisted file format is unchanged (still a version-1 snapshot; only the write semantics and merged content differ). Empty folder lists are not persisted; the read side already treats them as absent.
- All runtimes share this server route (web, desktop, VS Code, hosted mobile; Capacitor mobile when connected to a server), so all benefit. No runtime-specific behavior is introduced.
- Client stores and localStorage contracts are untouched.

## Repository guidance

|---|---|---|
| AGENTS correctness invariants ("One failed entity must not erase or block unrelated complete entities"; "Prefer authoritative state over heuristics") | A stale device's partial map erasing unrelated complete scopes is exactly the forbidden direction | Merge isolates per scope; the stored snapshot's scopes/folders survive unknown writers |
| `sync-state-invariants` ("Never infer a disappearance event from … partially loaded scopes"; persisted-snapshot ordering: "Do not rely on uncontrolled wall-clock timestamps") | The whole-map POST from a stale client is a partially-loaded view; wall-clock `updatedAt` LWW is the failure mode | Absence is not treated as deletion; merge does not grant authority by timestamp |
| `openchamber-change-discipline` (persisted/external contract risk class) | Stored data + route behavior changed | Failure/ordering cases covered by tests: stale-vs-newer writes, duplicate revision, missing-vs-empty, malformed-state repair, concurrent-save serialization unchanged |
| `ui-api-decoupling` ("Parse at the boundary … treat persisted payloads as unknown until validated") | Merged input is persisted external data | Both merge sides are only used after `hasValidFolderSnapshotShape` passes; malformed stored state still falls back to repair-by-valid-snapshot |
| Nearest docs: `packages/ui/src/stores/DOCUMENTATION.md` (client persist/hydrate contract); no `DOCUMENTATION.md` exists for `session-folders` | Established the bootstrap-only hydration + debounced whole-map persist that motivates server-side defense | Behavior kept compatible: response shape, validation, temp-file+rename atomicity unchanged |

## Validation

| Check | Result |
|---|---|
| `bunx vitest run server/lib/session-folders/routes.test.js` | 12/12 passed — 6 pre-existing cases kept (concurrent temp files, rename cleanup, GET guards, malformed repair) with the two LWW cases rewritten for merge semantics, plus 6 new merge regressions (scope preservation, shared-scope union incl. dropped folder kept, per-id incoming-wins, same-name session-list union, older-stamp merge with monotonic `updatedAt`, duplicate-revision union) |
| `bun run --cwd packages/web test` (full package) | 2365 passed / 1 failed — the failure (`network-defaults.test.js`, fresh-Node import of `@openchamber/sdk`) reproduces on a clean tree without this diff (local workspace-link issue), pre-existing |
| `bunx oxlint` on both changed files | 17 findings, all pre-existing patterns on untouched lines (clean tree: 18 — this diff removes one) |
| Live validation | The identical merge has been running as a local patch on a production v1.23.2 instance since the incident in #3611: the captured loss (folder created 11:08, erased by a stale-device POST at 11:32) did not recur, and multi-device state (3 scopes incl. auto-archive folders) stays preserved across writes |
| Not verified | No browser-level E2E in this PR; server behavior is covered by route tests |

## Visual evidence

No rendered behavior changes: this is a server-side merge only; response shapes are unchanged apart from the removed `ignored` flag, and no UI copy or layout is touched.

## Risks and failure behavior

- **Data loss**: strictly reduced — the union never drops a scope/folder the stored snapshot had. The residual direction is over-retention: a folder deleted on one device persists on the server and reappears there after the deleting device rehydrates, until a tombstone protocol exists (see Non-goals).
- **Same-name merge**: two same-name folders with different ids become one folder with the union of their session lists. This matches multi-device auto-archive folder creation; a user who intentionally creates two same-name folders on two devices gets one merged folder.
- **Concurrency/atomicity**: `saveQueue` serialization and tmp-file+rename atomic writes are unchanged; a write failure still unlinks the temp file and reports 500; malformed stored state is still repaired by the next valid snapshot (test kept).
- **Rollback/compatibility**: persisted format unchanged, no migration; reverting the commit restores previous behavior. Written file size can grow to the union of both snapshots; the 4 MB request-body cap still bounds each incoming side.


yulia-ivashko:
Thanks for chasing this down — the loss you're fixing is real, and your diagnosis matches the code: the route replaced the stored file wholesale, so a client that hydrated once could erase scopes it had never seen.

The merge as written can't land, though, because it turns every write into a union, and a union cannot express a deletion. I ran this branch's route handler against an in-memory fs:

- delete one of two folders in a scope → both folders are still stored
- delete the only folder in a scope → an empty incoming array is read as "never seen", and the folder comes back
- expand a collapsed folder → `collapsedFolderIds` is a union, so an id can never be removed
- rename "Beta" to "Alpha" while "Alpha" exists → the same-name branch drops the original "Alpha" and moves its sessions into the renamed folder; two folders become one

The first two are not cross-device cases. They happen on a single device: delete a folder, reload, it is back.

What done looks like:

1. Merge per scope, not per folder. A scope missing from the incoming map is kept from disk; a scope present in the incoming map is authoritative, including when its array is empty. That closes #3611 and leaves deletion working.
2. Apply the same rule to `collapsedFolderIds`: keep ids that belong to scopes the writer does not know, and take the incoming list for the scopes it does.
3. Drop the same-name twin merge. Folder names are not unique — `createFolder` defaults to "New folder" and checks nothing — so matching by name conflates unrelated folders and destroys one of them. The id is the identity.
4. Cover each of these with a test: deleting one of two folders, deleting the last folder in a scope, expanding a collapsed folder, and two folders sharing a name in one scope. The current tests only assert unions, which is why none of this showed up.
5. Check the size cap against the merged document, not only the incoming body. Today the merged result is written unchecked, so the file can grow past 4 MB, after which every client that reloads and re-posts gets a 413 and can never save again.

PR #3626, opened 2026-09-16
fix(chat): keep the selection menu off the selected text

### LINKED ISSUE #3596: [Bug] Since 1.23.2, can't copy text because of overlay
### Language

Most of the time, when highlighting text since 1.23.2, the "Comment, By the way, Add to Notes" box now covers the highlighted text - so we can't right click to copy text.

Happens about 80% of the time, especially when under a horizonal rule (hr), OR when selecting text from the first line

Tested on Linux desktop app.

### Where does it happen?

Desktop (macOS)

### Version

1.23.2

Image

https://github.com/user-attachments/assets/6bb1cfb5-dfd9-44e8-a4d7-906d49a7885c

### Logs (optional)

```shell

```

yulia-ivashko:
Fixes #3596

## Intent

On desktop, selecting text in the first lines of a chat message opened the "Comment / By the way... / Add to notes" menu on top of the selection. The menu sits in a portal above everything, so right-click copy hit the menu instead of the text.

The cause is the guard added in `4ece3d608` (v1.21.0) to keep the comment box off the desktop drag header. It pushed the menu down until its top edge was inside `containerRef`. That ref is the message body (`MessageBody.tsx` passes `messageContentRef`), not the chat viewport. Any selection within about 50px of a message's top edge pushed the menu onto the text. The first line of every message was affected, wherever it was on screen, which fits the reporter's "first line" and "under an hr" cases.

Now the menu opens above the selection. When there is no room above inside the chat viewport, it flips below the selection. The upper boundary is the scroller that holds the message: the chat viewport (`[data-scrollbar="chat"]`), or the "By the way" panel body, which renders messages in its own scroller and now carries `data-selection-menu-boundary`. The header guard from `4ece3d608` still holds in both. The menu can only cover text when the selection fills the whole visible area and neither side has room. In that case it is clamped on screen as before (#2257).

## Non-goals

- Horizontal placement is unchanged. The menu still centers on the selection and clamps to the window edges.
- The menu does not follow the selection while the chat scrolls. It didn't before either.
- The mobile bottom bar and mobile comment overlay are unchanged. They ignore these coordinates.

## Affected surfaces

- `packages/ui` shared chat UI: `TextSelectionMenu.tsx`, `selectionMenuPosition.ts`, and one boundary attribute on the `BtwPanel` body scroller.
- Desktop layout on web, Electron, and VS Code. All three render the same component, so the fix applies to each.
- Mobile web and Capacitor are unaffected because `isMobile` renders a bottom bar that doesn't use `position.x/y`.
- No persisted data, routes, or exported package contracts change. `getDesktopClampedY` is replaced by `getDesktopMenuY`, and nothing outside this component used it.

## Repository guidance

|---|---|---|
| `openchamber-change-discipline` | Source change in shared UI | The fix stays in the owning component and its existing position helper. Placement logic is a pure function with focused tests. The separate "push down" layout effect is removed rather than patched, so one function owns vertical placement. |
| `AGENTS.md` correctness invariants ("prefer authoritative state over heuristics") | The old guard measured the wrong element | The boundary is now the real chat viewport element, with a viewport-top fallback when it isn't found. |
| `AGENTS.md` validation (`oxlint`, `dead-code`) | Rewritten TS in `packages/ui` | Both run. See Validation. |
| `communication-style` | PR text | Applied to this description. |

`theme-system` and `locale-ui-patterns` don't apply: no colors, tokens, or strings changed. There is no `DOCUMENTATION.md` for `components/chat/message/`.

## Validation

| Check | Result |
|---|---|
| `bun test packages/ui/src/components/chat/message/__tests__/selectionMenuPosition.test.ts` | 13 pass, 0 fail. New cases cover above placement, the flip below near the chat top, "never overlaps when either side has room", a tall comment box, #2257 (selection starting above the viewport), a selection scrolled below the viewport, and a viewport shorter than the menu. |
| `bun run type-check` in `packages/ui` | Pass |
| `eslint` on the three changed files | Pass |
| `bunx oxlint` on the three changed files | No findings in changed code. Three pre-existing findings remain in `TextSelectionMenu.tsx` (lines around 428, 508, 536, untouched). Two earlier findings went away with the removed `typeof window` guards. |
| `bun run dead-code` | No findings for the changed files after un-exporting the input/result interfaces. |
| Manual, web dev server (`bun run dev`), Chrome | Selected the first line of an assistant message at several distances from the chat top and measured the menu and selection rects. On `main` the menu overlapped the selection (menu 197-239px, selection 215-232px). On this branch it didn't overlap in any case: above at 70px and 150px from the top, below at 4px and 30px. |
| Manual, "By the way" panel | Not exercised in the browser. The boundary attribute reaches the DOM through `ScrollShadow`'s prop spread; placement uses the same function. |
| Manual, comment mode | Opened Comment near the chat top and typed four lines. The box grew below the selection without covering it or reaching the header. |

Not verified: packaged Electron, the VS Code webview, and a real right-click copy. Those runtimes render the same component with the same coordinates, but I did not launch them.

## Visual evidence

Desktop web, dark theme, 900x560, first line of an assistant message selected ("Не вмерла: статус").

Before (`main`): the menu covers the selected line.

before

After: the menu sits above the selection and the text stays reachable.

after

The below-selection flip is covered by unit tests and the manual rect measurements above. I didn't screenshot it because the sticky user message header covers the top of the chat at this window size.

Light theme and mobile are not shown. No colors or styles changed, and mobile doesn't use this placement.

## Risks and failure behavior

- A future surface that mounts `ChatMessage` in a scroller with neither boundary marker falls back to the window top. The menu stays on screen but could overlap an app header there. Today both mounts (chat and the "By the way" panel) are marked.
- While typing in the comment box, the box re-measures its height. If it outgrows the space above the selection, it flips below mid-typing. That jump is deliberate. Before, the box slid down over the quoted text.
- Rollback is a plain revert. No state or data is involved.


yulia-ivashko:
I ran the review bot locally before opening this, with the same instructions as the CI job (`.opencode/agent/pr-review-bot.md`), so we don't wait on a round-trip for things that are quick to fix. The bot on this PR will still run as usual.

Local verdict: **PASS**, with one non-blocking finding.

The "By the way" panel renders messages in its own scroller, outside `[data-scrollbar="chat"]`. There the menu fell back to the window top as its boundary, so a tall comment box could still reach the desktop drag header. I fixed it before pushing: the panel body now carries `data-selection-menu-boundary`, and the menu uses the nearest of the two markers. The PR description is updated to match.

PR #3627, opened 2026-09-16
perf(sidebar): virtualize the unified session list

deatheros:
## Intent

Large session histories made the unified sidebar mount and maintain every visible session row. At the 25,000-session reproduction scale, startup took 112.80 seconds, the renderer held about 2.3 GiB, and the sidebar created more than 126,000 DOM nodes.

This PR gives the sidebar one flat row model and one virtualizer. It mounts only the viewport-sized row window while preserving project grouping, nested sessions, selection, bulk actions, search, incremental reveal, and drag-and-drop identities. It also keeps expanded off-screen directories alive while they have foreground bootstrap demand, which prevents eviction and refetch loops once the number of open directories exceeds the child-store soft limit.

The implementation adapts the flat-row and global-virtualizer direction explored in #3536 to the current unified sidebar architecture.

## Non-goals

- No visual redesign or new sidebar controls.
- No changes to persisted session data, server routes, SDK contracts, or runtime authentication.
- No increase to the child-store capacity limit. Foreground demand protects in-use stores; idle eviction still bounds the cache.

## Affected surfaces

- `packages/ui`: shared session sidebar rendering, logical selection identity, bulk actions, search results, drag-and-drop bookkeeping, stream performance diagnostics, and directory child-store eviction.
- Web, Electron, VS Code, hosted mobile, and Capacitor consume this shared UI implementation. There is no runtime-specific transport or native-shell change.
- Existing persisted project and session records remain unchanged. New row and occurrence keys are in-memory presentation identities only.
- Runtime switching clears the new in-memory selection state through the existing endpoint-reset flow.

## Repository guidance

|---|---|---|
| Root `AGENTS.md` and `CONTRIBUTING.md` | This is a shared UI, sync, and performance change with cross-runtime consumers. | Kept ownership in `packages/ui`, updated owning docs, preserved authoritative session state, and validated shared consumers rather than treating a type-check as runtime proof. |
| `openchamber-change-discipline` | The diff adds and removes source files, changes exports, and replaces a rendering path. | Kept the row model, renderer, and identity helpers focused on concrete invariants; added behavior tests; ran workspace checks, build, and dead-code analysis. |
| `performance-engineering` and `scripts/perf/DOCUMENTATION.md` | The reported failure is startup, DOM, CPU, and memory growth at large session counts. | Measured the production build at current HEAD with the identical 25,000-session fixture, including the merged sticky-header observers and a full-list scroll pass. The row window stayed bounded. |
| `sync-state-invariants` and `packages/ui/src/sync/DOCUMENTATION.md` | Virtualized rows can be off-screen while their directory state is still in use. | Foreground bootstrap demand now protects a directory before eviction; the cache may exceed its soft target until entries become idle instead of disposing mounting stores. |
| `packages/ui/src/components/session/sidebar/DOCUMENTATION.md` | This file owns sidebar ordering, grouping, mounting, and interaction invariants. | Updated it to document the unified row model, occurrence identity, bounded mounting, pinned rows, and logical selection rules. |
| `drag-to-reorder` | Sidebar DnD now addresses logical rows that may not all be mounted. | Uses stable occurrence IDs and commits subtree moves through the existing drag-end path; no index-based identity or live reorder loop was introduced. |
| `theme-system` | Shared row components and controls changed, and the merge integrated main's sticky-header crossfade. | Reused existing components, semantic classes, buttons, and icons. The crossfade moves one live header host between the virtual row and sticky layer; it does not duplicate interactive controls. |
| `ui-api-decoupling` runtime-switch guidance | New in-memory selection state must not survive an endpoint switch. | Added cleanup to the established runtime endpoint reset. No transport, URL, auth, or runtime API contract changed. |

## Validation

| Check | Result |
|---|---|
| `bun install --frozen-lockfile` | Passed. |
| `bun run type-check` | Passed after merging current `origin/main`. |
| `bun run lint` | Passed after merging current `origin/main`. |
| `bun run type-check:ui` and `bun run lint:ui` | Passed on PR HEAD. |
| `bun run build:ui && bun run build:web` | Passed on `55e632a8f`; the measurements below use those production assets. |
| `bun run test` | Root scripts 6/6 files, SDK 13/13, UI 505/505, VS Code 44/44, Electron 25/25, and web 215/216 files passed. The sole failure is the unchanged `env-runtime.test.js` login-shell fallback test because this Mac has a real `/opt/homebrew/bin/opencode`; it expects no fallback binary. |
| `sandbox-exec -p '(version 1)(allow default)(deny file-read* (literal "/opt/homebrew/bin/opencode"))' bunx vitest run server/lib/opencode/env-runtime.test.js` | Passed 29/29 when the host binary is hidden, matching CI. A full sandboxed root run is not used as evidence because macOS sandboxing blocks process inspection and causes three unrelated CLI lifecycle tests to fail. |
| `bun test src/stores/utils/streamDebug.test.ts` | Passed 3/3 on PR HEAD. |
| Focused `bunx oxlint` over new and substantially rewritten sidebar, DnD, and selection files | Passed. A broad changed-file run also reports pre-existing anti-slop findings in `App.tsx`, `streamDebug.ts`, and `child-store.test.ts`; no new finding remains in the added code. |
| `bun run dead-code` | Completed. It reports the repository's existing unused-file/export backlog; source changes were inspected. |
| Bundled production UI at 10 projects x 500 sessions | Opened on PR HEAD at desktop and mobile viewport sizes with no page errors. The sidebar mounted the visible bounded window rather than all 5,000 session rows. Long, Unicode, path-like, recent, grouped, and incrementally revealed labels rendered. |
| CDP touch drag replay on the bundled 10 x 500 fixture | Passed. A quick swipe before 200 ms reported `quickDragging=false`; a 300 ms hold reported `longPressDragging=true`; dropping `Sidebar Touch 02` over `koloia` changed their order once. |
| Post-review regression checks | UI 511/511 files, `bun run type-check:ui`, `bun run lint:ui`, focused `oxlint`, `bun run dead-code`, `git diff --check`, and 7/7 focused crossfade/action-context tests passed after the merge and mobile action fix. |
| Recreated 25,000-session fixture | `setup`, `extend-50x500`, `prepare-alteration`, `apply-alteration`, `verify-alteration`, `seed-client-state`, and `verify-client-state` passed. Verification found 50 projects, 25,000 sessions, structural hash `33c6d9b0fc096050339bede3d333ec05a9caa26e44c668252c339d7cf6658dce`, 500 folders, 10,000 folder memberships, 250 collapsed folders, and 1,000 pins. |
| Current-HEAD startup and idle capture | Three startup runs completed with all 50 root and 50 full scopes loaded, no HTTP errors, and a 33.41 s median (33.23–37.13 s). A separate 61.16 s idle window after a 90 s settle measured 1.67% main-thread busy, zero DOM growth, and 337.06 MiB peak heap. Frame liveness was 62 fps with the page visible. |
| Current-HEAD sticky-header scroll pass | A CDP probe crossed the full 190,617 px virtual list at seven positions. The one live crossfade host handed off from `chats` through projects 05, 12, 25, 38, 45, and 50. The renderer kept 34–47 virtual rows and 24–40 session rows mounted during the pass; a separate settled interaction check mounted 22 session rows. |

Physical mobile hardware was not available. The recording uses Chrome touch emulation against the production bundle, not synthetic React events.

## Visual evidence

This PR intentionally preserves the sidebar's visual design. Static before/after screenshots look the same and do not show the reported failure, so the relevant evidence is the production-scale measurement.

Touch activation on the current bundle:

Quick swipe and long-press drag behavior

The base implementation used one distance-activated `PointerSensor`; it had no touch-specific hold period. The recording shows the replacement behavior: movement before 200 ms stays out of drag, while a hold activates drag before the project moves. The fixture order changes once on drop. The later sticky-header merge, bootstrap invalidation fix, and mobile action fix do not change the sensors or DnD path shown here.

Identical 25,000-session fixture:

| Metric | Before | After | Change |
|---|---:|---:|---:|
| Startup to ready | 112.80 s | 33.41 s | -70.4% |
| Idle main-thread share | 35.24% | 1.67% | -95.3% |
| Peak renderer heap during idle | 2,330 MiB | 337 MiB | -85.5% |
| DOM nodes during idle | 126,309 | 21,100 | -83.3% |
| Mounted session rows | about 25,000 | 22 at rest; 24–40 during the scroll pass | bounded to viewport plus pinned headers |

These measurements were taken on current PR HEAD `55e632a8f` from a fresh production build on September 17, 2026. The fixture was recreated from the same scripts and verified before capture. The table replaces the older `941c40849` results because the merged sticky-header crossfade adds pinned adjacent headers, a scroll listener, resize observers, and a mutation observer to the measured path. The current startup and heap values are less favorable than the old 19.33 s and 201 MiB values, while idle CPU is lower; the bounded-mount result still holds across the full-list scroll pass. Local measurement artifacts are gitignored because they contain machine paths.

## Risks and failure behavior

- Virtualization changes mount, measurement, focus, and scroll behavior. The global model keeps stable row keys, pins selected, DnD, current-header, and adjacent-header rows, and retains logical selection independently of mounted DOM. Transition and behavior tests cover grouping, reveal, search, selection, row ordering, and virtual-start crossfade handoff. The current-HEAD scroll pass also exercised every quarter of the 190,617 px list.
- One session can appear in more than one presentation scope. Selection and DnD now use occurrence identity while destructive actions resolve back to logical session IDs, preventing duplicate actions.
- Foreground-demand protection can temporarily keep more than `MAX_DIR_STORES` child stores. This is deliberate: capacity is a soft target, and idle-time eviction resumes when foreground demand clears.
- A stale runtime selection could target an entity from another endpoint. The existing runtime-reset path now clears the selection store.
- Rollback is a code revert. There is no data migration, persisted schema change, or cleanup step for users.


deatheros:
/oc-review Re-check the current-HEAD 25,000-session measurements and sticky-header scroll evidence added to the PR body.

PR #3517, opened 2026-09-11
fix(work-status): correct turn stats across compactions

alvins82:
## Intent

Correct the work-status turn statistics for OMP sessions that cross automatic compaction. Compaction summaries are checkpoints, not model steps, and a sidecar assistant record can represent several raw model requests. This change makes OpenChamber consume the sidecar's optional aggregate usage and timing metadata while preserving the final assistant token snapshot for context-window display.

The companion sidecar implementation is committed separately as `78264de` in `omp-openchamber-server`.

## Non-goals

- No change to compaction behavior, retry behavior, or model execution.
- No change to the separate final-response delivery-rate metric.
- No layout, label, theme, responsive, or interaction redesign.

## Affected surfaces

- `packages/ui/src/components/chat/work-status/telemetry.ts`: excludes compaction summaries, consumes optional OMP aggregate telemetry, and averages TTFT over provider-request samples.
- `packages/ui/src/components/chat/work-status/telemetry.test.ts`: covers compaction boundaries, sidecar aggregate usage, and sidecar timing.
- `packages/ui/src/components/chat/work-status/DOCUMENTATION.md`: documents compaction and provider metadata semantics.
- Optional persisted/live message metadata contract: reads `info.metadata.omp` when supplied by the companion sidecar and falls back to existing message fields for older sessions and other backends.

Only shared UI telemetry calculation changed. No Electron, mobile, web runtime, or native integration files were changed because those surfaces consume this calculation and do not own the provider message mapping.

## Repository guidance

| Guidance | Why it applies | How the change complies |
|---|---|---|
| `AGENTS.md` | Repository and package boundaries govern this UI-only change. | Kept the diff focused, added no dependencies, and did not touch `../opencode` or unrelated packages. |
| `CONTRIBUTING.md` and `.github/PULL_REQUEST_TEMPLATE.md` | Required PR structure and validation reporting. | This description fills every template section and records exact commands and results. |
| `openchamber-change-discipline` skill | Turn-stat fixes must preserve the semantic model instead of hiding incomplete data. | Compaction summaries are filtered explicitly; a trailing summary leaves the newest turn incomplete rather than falling back to an older turn. |
| `ui-api-decoupling` and `sync-state-invariants` skills | The UI consumes server records and must remain stable across live/history synchronization. | Optional metadata is namespaced and backward-compatible; existing fallback calculations remain in place. |
[...]

PR #3504, opened 2026-09-11
fix(chat): hide compaction summaries from live activity

alvins82:
## Intent

OMP records context compactions as assistant messages with `info.summary === true` in the same parent turn as normal assistant activity. The live Activity renderer and the ordinary assistant block treated those records as live model output, so internal Goal and Constraints snapshots could appear in chat.

This filters compaction summaries before selecting the live final message, calculating the live activity summary, rendering live Activity rows, and rendering the ordinary assistant block used by sorted mode and live turns without Activity. Normal progress and final output remain visible.

## Non-goals

- No changes to the OMP sidecar or its session records.
- No changes to persisted data, transport behavior, or activity classification rules.

## Affected surfaces

- `packages/ui` shared chat rendering used by the web, desktop, VS Code, and mobile surfaces.
- Live Activity no longer replays internal compaction text in its rows or header summary.
- Sorted mode and live fallback rendering no longer show compaction snapshots as full assistant messages.
- No API, persistence, runtime bridge, or cross-process contract changes.

## Repository guidance

| Guidance | Why it applies | How the change complies |
|---|---|---|
| `AGENTS.md` | This is a shared UI source and owning-documentation change. | The diff stays in the chat renderers, their regression test, and the owning rendering documentation. Unrelated worktree changes remain outside the branch. |
| `openchamber-change-discipline` | The fix changes local rendering behavior and its documented invariant. | It uses the existing `summary` flag, adds no dependency or speculative abstraction, and validates the owning UI package. |
| `performance-engineering` | The affected renderers run during chat updates. | The live filtered list remains memoized against the projected assistant-message array. The fallback path performs one bounded filter before its existing map and adds no global state or cache. |
| `writing-for-agents` | The owning `DOCUMENTATION.md` is reached through the repository's documentation-discovery pointer. | The documentation records the new visibility rule in the existing live disclosure section without duplicating unrelated guidance. |
| `communication-style` | This pull request is a human-facing review artifact. | The description uses concrete behavior and reports only verified checks. |
| `CONTRIBUTING.md` and `packages/ui/package.json` | They define the pull request contract and package validation commands. | The required intent, non-goals, affected surfaces, validation, evidence, and risk sections are included. |

## Validation

| Check | Result |
|---|---|
[...]

PR #3628, opened 2026-09-16
fix(context): show unknown usage after compaction instead of the summary request

### LINKED ISSUE #3572: [Bug] After compaction, Context meter shows the summarization call's input (pre-compaction size) until the next message

**Expected:** once a session compaction completes, the Context sidebar shows the post-compaction context.

**Actual:** the "Context" meter keeps reporting approximately the *pre-compaction* size (here: `865,110 / 1,000,000`, "86.5% used") until the next assistant message completes, at which point it drops to the true post-compaction value (`108,656 / 1,000,000`, "10.9% used").

**Root cause (verified in v1.23.1 source):** the Context tile equals `input + cache read + output + reasoning (+ cache write)` of the **last assistant message**. After a compaction, that message is the **summarization call** itself, whose input is the entire pre-compaction history, so the meter reports the size of the read operation rather than what survived it. Compaction records (`info.summary === true`) are not excluded from the calculation.

**Token arithmetic, exact in both states:**

State A, immediately after compacting (`Context: 865,110 / 1,000,000`, 86.5% used):
- `Input 860,410` + `Cache Read 0` + `Output 2,511` + `Reasoning 2,189` = **865,110**
- `Cache Read 0` / `Cache Hit 0.0%` is the fingerprint of the one-shot summarization read; the 2,511-token output is the summary text.

State B, after the next normal assistant turn (`Context: 108,656 / 1,000,000`, 10.9% used):
- `Input 40,483` + `Cache Read 53,248` + `Output 536` + `Reasoning 14,389` = **108,656**

**Source pointer:** in `resources/web-dist/assets/useAppFontEffects-DQXEYu1N.js` (v1.23.1), the context sidebar aggregates the last assistant turn from `message.tokens.input / output / reasoning / cache.read / cache.write` and the Context tile sums those fields.

**Steps to reproduce:**
1. Grow a session to a large fraction of the model window (1M in this report).
2. Compact it (`/compact`; auto-compaction produces the same message shape).
3. Check the Context meter immediately: it still shows approximately the pre-compaction size.
4. Send any message: the meter collapses to the true post-compaction context.

**Impact:** every compaction appears to free almost nothing (86.5% shown where the real context is ~9.4%), which invites repeated manual compaction and distrust of the meter and auto-compaction.

**Related:** #880 reports the same class of bug ("/compact does not update used context count") and was auto-closed as stale without a fix. #2914 (merged) previously corrected the meter for internal round-trips. #3517 and #3504 (both open) fix adjacent surfaces (work-status turn stats and live-activity rendering across compactions); #3517 explicitly preserves the final assistant token snapshot for the context-window display, so this meter is unaffected. #2166 (open) shows a related stale-recalc symptom on model switch.

**Possible fixes (any of):**
1. Exclude compaction/summarization messages when selecting the last assistant message used for the context estimate.
2. Base the meter on the effective read of the last normal request (fresh input + cache read) instead of also counting output/reasoning (which overstates it even in steady state: 14,925 extra tokens in State B).
3. If the steady-state semantics are intended, special-case compaction records, or relabel the tile (for example "Last request footprint") so it is not read as occupied context.

### Where does it happen?

Desktop (Windows 11, Electron app; the form options do not list Windows desktop).

### Version (if known)

1.23.1 (latest stable; not checked against the v2 preview)

Two Context-panel captures, transcribed exactly (images attachable on request):
- State A: `Context 865,110 / 1,000,000 - 86.5% used`; last message `Input 860,410 - Output 2,511 - Reasoning 2,189 - Cache Read 0 - Cache Hit 0.0%`
- State B: `Context 108,656 / 1,000,000 - 10.9% used`; last message `Input 40,483 - Cache Read 53,248 - Output 536 - Reasoning 14,389 - Cache Hit 56.8%`

### Logs (optional)

Not applicable (UI metric; no logs involved).

yulia-ivashko:
Fixes #3572

## Intent

Right after a compaction, the Context readout showed the wrong number until the next message arrived. Every counter took the newest assistant message that reported tokens. After a compaction that message is the compaction's own record (`summary: true`). Its tokens describe the summarizing request, not what the window holds afterwards.

The error goes both ways. In a long session the request's input is the whole pre-compaction history, so the counter stays high (the reporter's case). In a short one it goes low, because the compaction request carries no system prompt or tools. My repro on a 200K model: 5.9% before compaction, 1.2% right after it, and 5.9% again (11,789 tokens) on the next real response.

No number from the messages on hand is right at that moment:

| Candidate | Last 12 compactions in my local OpenCode database | Real fill on the next response |
|---|---|---|
| compaction record `total` | 12K to 144K | 18K to 28K |
| compaction record `output` | 1.2K to 2.3K | 18K to 28K |
| last response before the compaction | up to 384K | 18K to 28K |

`output` misses the system prompt, the tools and the recent turns OpenCode keeps after compacting (`preserve_recent_tokens`). OpenCode's own overflow check skips these records too (`lastFinished.summary !== true` in `session/prompt.ts`).

So the readout now says it does not know. After a finished compaction every surface shows a dash with "Context compacted. Usage updates after the next response." The first response that reports tokens replaces it. Two cases keep the previous number on purpose:

- a compaction that is still running has not changed the window yet;
- a compaction that failed or was aborted never changed it. "Finished" uses OpenCode's own rule, `summary && finish && !error`.

All six readouts now go through one helper, `findLatestContextFill` in `stores/utils/tokenUtils.ts`, instead of six copies of the same loop.

## Non-goals

- Whether `output` and `reasoning` should count toward the fill. The counter still uses the server's `tokens.total`, which includes them. That is a separate product question raised in the issue.
- Estimating the post-compaction size. Any estimate would be a guess the next response overrides seconds later.
- Turn stats across compactions (#3517) and hiding summaries from live activity (#3504). Neither touches these files.
- `contextStore.ts` still has its own `getContextUsage` / `updateSessionContextUsage`, but nothing calls them, so they cannot show a number.

## Affected surfaces

`packages/ui` only. It covers every runtime that renders these components:

| Surface | Runtime | Change |
|---|---|---|
| Header counter (`Header.tsx` via `session-ui-store.getContextUsage`) | web, desktop | empty ring, dash, explanation in the tooltip |
| Work-status panel (`contextUsage.ts`, `WorkStatusPrimaryGroup.tsx`) | web, desktop | dash, empty meter, caption |
| Context sidebar (`ContextSidebarTab.tsx`) | web, desktop (context panel in `MainLayout`) | "— / limit" with the caption; last-turn breakdown still shows the compaction turn's own tokens |
| VS Code header (`VSCodeLayout.tsx`) | VS Code | same as the header counter |
| Mini chat header (`MiniChatLayout.tsx`) | desktop mini chat | same as the header counter |
| Mobile metadata popover (`MobileSessionMetadata.tsx`) | hosted and Capacitor mobile | empty ring, "—/limit", caption |

Contracts: `SessionContextUsage` becomes a union with a `state` of `measured` or `compacted`, and `ContextUsageDisplay` takes a `reading` prop instead of `totalTokens`/`percentage`/`colorPercentage`. Every consumer is inside `packages/ui` and is updated here. Nothing is persisted, and no server route changes.

One small alignment: the VS Code header now shows the unrounded percentage like the web header and the work-status panel (6.0% became 5.9%). It uses the same `toContextUsageReading` as the other two headers.

New i18n key `contextUsage.compacted.description`, translated in all 12 dictionaries.

## Repository guidance

|---|---|---|
| `openchamber-change-discipline` | Shared UI contract (`SessionContextUsage`, `ContextUsageDisplay` props) with several consumers | Traced every consumer by grep and updated all six. Removed the duplicated loops and equality checks instead of patching each. Ran package and workspace checks plus `dead-code`, which led me to unexport two types |
| `sync-state-invariants` | The readout is derived from live session messages, and a compaction is a lifecycle transition | The unknown state is an explicit discriminated union, not a nullable pair. It clears as soon as a later response reports tokens. Running and failed compactions keep the previous reading and never fall back to a pre-compaction number |
| `theme-system` | New visual state in the counters | Uses the existing tokens only (`--interactive-border` track, `text-muted-foreground`). The empty state drops the progress arc instead of drawing a zero-length round cap. No new animation |
| `locale-ui-patterns` | New tooltip and caption text | One semantic key, real translations in every dictionary, resolved inside render with `useI18n` |
| `packages/ui/src/components/chat/work-status/DOCUMENTATION.md` | Documents how the panel computes context usage | Added the compaction rule and pointed to the shared helper |
| AGENTS.md "Prefer authoritative state over heuristics" | Choosing what to show without a real reading | Shows unknown until the server reports tokens, rather than estimating |

## Validation

| Check | Result |
|---|---|
| `bun test` on `tokenUtils.test.ts`, `contextUsage.test.ts`, `contextUsageReading.test.ts` | 49 pass, 0 fail. New cases: finished compaction gives `compacted`, a streaming response after it keeps `compacted`, running, failed, and completed-but-not-yet-failed (overflow) compactions keep the previous reading, the next measured response replaces it, `isSameContextUsage` never equates the two states |
| `bun run type-check` (workspace) | pass in all six workspaces |
| `bun run lint` (workspace) | pass in all six workspaces |
| `bun run test` (workspace) | pass: scripts 6/6, sdk 13/13, ui 501/501, vscode 44/44, electron 25/25 files, web 2,396 tests (1 skipped) |
| `bun run build` | pass |
| `bun run dead-code` | no new findings for the changed files (two exported-but-unused types were unexported) |
| `bunx oxlint --no-ignore <file>` for each of the 15 changed TypeScript files | no findings on added or changed lines; the remaining findings are the existing backlog on untouched lines |
| Live repro on `origin/main` (dev build, OpenCode 1.18.25, `opencode/big-pickle`, 200K context) | 5.9% before `/compact`, 1.2% after it, which is the compaction record's own `total` of 2,392 |
| Live check on this branch, same session | header, tooltip, work-status panel, context sidebar and mobile popover show the dash and caption after a finished compaction; the next response shows 5.9% (11,789 tokens) |
| Not verified live | VS Code header and desktop mini chat. They render the same `ContextUsageDisplay` from the same `buildSessionContextUsage`, so type-check and the unit tests cover the wiring but not the runtime. Also not verified: a compaction that fails or is aborted live (covered by a unit test only) |

## Visual evidence

One real session in a live dev build, captured with headless Chrome. "Before" is the same session on `origin/main` code. The session was compacted twice. Every before/after pair shows the second compaction, whose record reports 1,862 tokens (0.9%). The "next response" shot was taken between the two compactions. The last commit only narrows when a compaction counts as finished; I re-ran the header capture on the final HEAD and it still shows the dash for this session.

**Header counter and tooltip, dark (desktop, 900 px wide)**

| Before | After |
|---|---|
|  |  |

**Header counter, light**

| Before | After |
|---|---|
|  |  |

**Work-status panel (desktop, 1400 px wide)**

| Before (dark) | After (dark) | After (light) |
|---|---|---|
|  |  |  |

**Context sidebar**

| Before | After |
|---|---|
|  |  |

**Mobile metadata popover (390 px)**

| Before | After |
|---|---|
|  |  |

**Next response restores the real number:** 11.8K, 5.9%

## Risks and failure behavior

- **A server that stops sending `summary: true`.** The helper would read the compaction record as a normal response, which is exactly today's behavior. Nothing gets worse.
- **A compaction record without `finish`, or with `error`.** It counts as running or failed, so the previous number stands. This is OpenCode's own rule for a completed compaction (`summary && finish && !error` in `message-v2.ts`). A timestamp check would not work: an overflowing compaction request gets `time.completed` before its error is written.
- **A response after the compaction that errors before it reports tokens.** The dash stays until one does. That is honest, because nothing measured the window.
- **Old sessions whose last assistant message is a compaction.** They open with a dash instead of a wrong number.
- **Performance.** Each readout still walks messages from the end and stops at the first match, as before. The only new allocation is in `ContextSidebarTab`, which maps entries to `info` inside a memo that already filters the list twice.


yulia-ivashko:
Before pushing, I ran the review locally with the bot's own instructions (`.opencode/agent/pr-review-bot.md`) on the diff, this description and the screenshots.

Verdict: PASS, with one non-blocking finding and one nit. I fixed both before pushing.

- The helper counted a compaction as finished once it had `time.completed` and no `error`. OpenCode's own rule is `summary && finish && !error`. When a compaction request overflows, OpenCode writes `time.completed` before it writes the error, so for a moment the counter would have shown the dash for a failed compaction. The check now uses `finish`, the same as OpenCode. A new unit test covers a record that has `time.completed` but no `finish` yet. That change is commit a2bb6bad9.
- The description said the new string was translated in 13 dictionaries. There are 12.

PR #3629, opened 2026-09-16
fix(vscode): keep a session read while its chat is on screen

### LINKED ISSUE #3505: [Bug] Stale unread dot revealed on background session after switching (VS Code extension)

Stale unread (blue dot) on a session I already read. It is invisible while the session is active and only appears after I switch away, which makes it look like switching *created* the unread.

Steps to reproduce (VS Code extension):

1. Open session B and let its turn finish while you are viewing it (chat visible; focus may be in the editor, not in the webview).
2. While B is the active session, open the session switcher dropdown — no blue dot on B.
3. Switch to session A, open the switcher dropdown again — B now shows an unread blue dot, even though its reply was already on screen.
4. Switch back to B — the dot is gone (and stays gone afterwards).

Expected: a turn that completes in the session I'm currently viewing should not produce an unseen notification. Switching sessions should not "reveal" stale unread state.

Suspected cause, from reading the source (not fully verified at runtime):

- `isViewedInCurrentSession` (`packages/ui/src/sync/sync-context.tsx`) requires `document.hasFocus()`. In VS Code the webview document often isn't focused while the user is looking at the chat (focus sits in the editor), so the `session.idle` event for the *viewed* session is appended with `viewed: false`.
- The switcher row hides the dot for the active session (`showUnreadDot = ... && !isActive` in `SessionSwitcherDropdown.tsx`), so that stale unseen stays invisible until you switch away — then it pops up on the background session.
- There is no dedup of repeated `session.idle` for the same turn, so a late duplicate arriving after you've switched away can also recreate unseen for an already-read session.

Possible directions: treat the active VS Code webview session as viewed without requiring `document.hasFocus()`, and/or dedupe `session.idle`/`session.error` notifications per turn.

## Where does it happen?

VS Code extension

## Version (if known)

1.23.0 (Marketplace, latest as of 2026-09-11); also present on `main` by code inspection.

yulia-ivashko:
Fixes #3505.

## Intent

In the VS Code extension, a session you had just read got a blue unread dot the moment I switched to another session.

The cause was the check for "is the user looking at this session". It used `document.hasFocus()` inside the webview. When the cursor sits in the code editor, the webview document has no focus even though the chat is on screen. So a turn that finished in the open session was stored as unseen. The session list hides the dot on the active row, which is why it only showed up after switching away.

Now the extension host tells each webview two things it knows for sure: whether the VS Code window is focused, and whether that webview is currently shown. A finished turn in the selected session counts as seen when both are true. When the chat becomes visible again (VS Code regains focus, or the collapsed sidebar opens), the selected session is marked seen. That is the same thing `App.tsx` already does on window focus for web and desktop.

## Non-goals

- Web, desktop and mobile keep their current behavior. They still use document focus, and their "mark seen on return" logic is untouched.
- No dedup of repeated `session.idle` events per turn (third point in the issue). The duplicates never showed up in my runs, and the dot comes from the focus check alone.
- Native notification rules are unchanged. VS Code still suppresses them while its window is focused, reading the same window focus value as before.

## Affected surfaces

- `packages/vscode`: all three webview hosts (sidebar chat, session editor tab, Agent Manager) now send `viewerStateChanged { windowFocused, surfaceVisible }` on resolve, on `webview:ready`, on window focus change, and when the view or panel is shown or hidden. It replaces the old `windowFocusChanged { focused }`. Host and webview ship together, so no version skew exists. `webview/main.tsx` parses the payload with a zod schema and still sets `__OPENCHAMBER_VSCODE_WINDOW_FOCUSED__` for notifications.
- `packages/ui`: new `lib/surfaceAttention.ts` answers "can the user see this surface". Without a host report it returns `document.hasFocus()`, exactly what `sync-context.tsx` used before. `VSCodeApp.tsx` marks the selected session seen when a host report says the webview is seen again (`onHostSurfaceSeen`).
- No persisted data, routes, or settings change.

## Repository guidance

|---|---|---|
| `sync-state-invariants` | The unread marker is session state derived in `sync-context.tsx` | The live source (host-reported focus and visibility) replaces a signal that is wrong in VS Code. Before the first report arrives, the old behavior applies. Once a report arrives, the host report alone decides for that document. |
| `ui-api-decoupling` (runtime parity reference) | Adds a VS Code host to webview bridge message consumed by shared UI | The payload is parsed at the bridge with a schema before shared code sees it. Each runtime's behavior is explicit: VS Code reports, the others keep document focus. |
| `openchamber-change-discipline` | Source change across `ui` and `vscode` | Scoped to VS Code. Web and mobile code paths are unchanged. Focused unit tests plus package checks and a live VS Code run. |
| `communication-style` | PR text | Applied to this description. |
| `packages/ui/src/sync/DOCUMENTATION.md` | Owns the viewed/unread rules | Added "Viewed sessions and surface attention". |
| `packages/vscode/src/DOCUMENTATION.md` | Owns host to webview messaging | Documented `viewerStateChanged` and when each host sends it. |

## Validation

| Check | Result |
|---|---|
| `bun test src/lib/surfaceAttention.test.ts` (ui) | 4 pass: document focus fallback, host report overrides document focus, "seen again" fires on return from a collapsed view or a background window but not on repeated or still-unseen reports, malformed payloads rejected |
| `bun run --cwd packages/ui test` | 501/501 test files passed |
| `bun run --cwd packages/vscode test` | 44/44 test files passed |
| `bun run --cwd packages/ui type-check`, `bun run --cwd packages/vscode type-check` | pass |
| `bun run --cwd packages/ui lint`, `bun run --cwd packages/vscode lint` | pass |
| `bunx oxlint` on every changed TS file | 0 findings on changed lines (existing findings on untouched lines only) |
| `bun run dead-code` | no findings for changed files |
| Live VS Code, `main` (e885faf49) | Reproduced: turn finished in session b with the cursor in `notes.txt` and chat visible, `hasFocus()` false, notification stored `viewed: false`. After switching to a, b showed the blue dot. |
| Live VS Code, this branch, VS Code focused, cursor in editor, sidebar chat visible | Turn stored `viewed: true`. No dot on b after switching to a. No native notification attempted (spied `Notification` constructor stayed empty). |
| Same, session open as an editor tab, cursor in another editor group | Turn stored `viewed: true` |
| Same, sidebar collapsed while the turn finished | Unread (`viewed: false`). Opening the sidebar marked it seen. |
| Same, VS Code in the background (Finder focused) | Unread plus one native notification. Focusing VS Code marked it seen. Rechecked on the final HEAD. |
| Agent Manager panel | Not exercised live. It sends the same message from `onDidChangeViewState`, and the code path matches the editor tab. |

Live runs used an Extension Development Host driven through the Chrome DevTools protocol, reading the notification store inside the webview. Windows and Linux were not run.

## Visual evidence

Before (`main`): the turn finishes in session b while the cursor is in `notes.txt`.

before, turn done

Before (`main`): after switching to a, b shows the unread dot.

before, switched

After: same steps, turn finishes with the cursor in `notes.txt`.

after, turn done

After: switching to a leaves b without a dot.

after, switched

The background and collapsed-sidebar cases have no single-frame picture, so they are covered by the store readings in the validation table.

## Risks and failure behavior

- If a host report is dropped, the webview keeps its last reported state. Every host resends it on `webview:ready` and on the cached-state retries, so a fresh document gets one early.
- Before the first report, the old `document.hasFocus()` check applies. At worst that is today's behavior for a few hundred milliseconds after load.
- A malformed payload is ignored rather than trusted, and the previous state stays.
- A tab that is visible but not active in its editor group counts as seen. VS Code reports it as `visible`, and it is on screen.
- "Seen" is decided for the whole webview, not for the chat screen inside it. In the narrow sidebar, the session list screen keeps the selected session, so a turn that finishes there while VS Code is focused counts as seen, and its permission or question toast is skipped. `main` already behaves this way whenever the webview has focus, and web and desktop treat other screens the same way. This PR makes that case more common because editor focus no longer counts as "not looking".


yulia-ivashko:
Before pushing, I ran a review locally using the review bot's own instructions (`.opencode/agent/pr-review-bot.md`) on the branch diff, this description, and the evidence images.

Verdict: PASS, with two non-blocking findings.

1. No test covered "coming back marks the selected session seen". Fixed: that logic now lives in `onHostSurfaceSeen` in `lib/surfaceAttention.ts`, and a test covers a collapsed view reopening, VS Code going to the background and coming back, and repeated reports that must not count as a return. Checked again in a live VS Code after the change.
2. "Seen" is decided for the whole webview, so the sidebar's session list screen still counts as looking at the selected session, and permission or question toasts for that session are skipped. `main` does the same whenever the webview has focus. I left the behavior as is and added it to the risks section.

PR #3630, opened 2026-09-16
fix(git): explain submodules and nested repositories instead of failing diffs

yulia-ivashko:
Fixes #3586

## Intent

The Changes panel showed "Failed to get git diff: Internal Server Error" for rows that `git status` lists but that have no file to diff. The reporter's log repeats the same failure every second and again 14 minutes later, so a row stayed in the list and kept failing. A one-off race would not look like that.

`resolveGitFileContext` in `packages/web/server/lib/git/service.js` threw `Invalid file path` when a path was missing from the working tree, the index, and HEAD. Neither diff route caught it, so both answered 500. I reproduced three cases on `main` in a scratch repository:

- A nested repository that is not a submodule, such as a clone inside the project. `git status -uall` lists it as `tools/icons-playground/`. It is a directory with its own `.git` and has no index entry, so every check failed and both routes returned 500 on every request. This matches the repeating log best.
- A file removed after the status listing and before the diff request. Also 500, and the row kept showing the error because nothing refreshed the list.
- A submodule. It never returned 500, but only by accident. `cat-file -e :sub` exits 1 with no stderr because the submodule commit lives in the submodule's own object store, and simple-git reports that as success. The results were still wrong. A submodule that only gained untracked files is `M` in status while `git diff` prints nothing, so the diff answered `{"diff":""}` and the panel showed an empty diff that looked like "no changes". `file-diff` returned empty contents for every submodule.

Now:

- The server reads each entry's mode from `ls-files --stage` and `ls-tree` instead of `cat-file -e`. That costs the same two git calls and tells a submodule (`160000`) apart from a file.
- A path that no longer exists answers `404 { code: "path_not_found" }`. A nested repository answers `422 { code: "nested_repository" }`. The message now names the path, which the issue's log was missing.
- Diff responses carry `submodule: { headCommit, indexCommit, worktreeCommit, hasTrackedChanges, hasUntrackedFiles, hasConflict }` for submodules and `null` for everything else. State comes from `git status --porcelain=v2`, including unmerged (`u`) and renamed (`2`) records, so a submodule in a merge conflict says so instead of looking unchanged. `file-diff` returns git's own `Subproject commit <sha>` text on both sides, so an older client still shows the commits instead of blanks.
[...]

PR #3631, opened 2026-09-16
fix(mobile): clear the nested-repository note on every diff read

yulia-ivashko:
Follow-up to #3630, from the bot review there.

## Intent

The mobile Changes detail remembers a `nested_repository` answer in `nestedRepositoryKey` and never clears it. The fetch effect resets only `diffLoadError`. If a later read of the same path succeeds, the "Separate Git repository" note still wins over the diff in `MobileDiffDetail` until the surface remounts. The same applies when the diff is already cached.

The effect now clears the key at the start of every diff read, before the cache check. The note appears only when the current read answered 422. The desktop entry already works this way: `MultiFileDiffEntry` clears its failure state when each request starts.

## Non-goals

- No change to how the server classifies paths or to the desktop Changes panel.
- In normal git use a path rarely goes from nested repository to diffable. When a folder loses its `.git`, status lists the files inside it instead of `folder/`. This fixes state that should never outlive its read, not a common user flow.

## Affected surfaces

- `packages/ui/src/apps/MobileChangesSurface.tsx`: mobile Changes on hosted mobile and Capacitor, and the mobile layout of web and Electron.
- VS Code does not mount this surface. No routes, persisted data, strings, or exports change.

## Repository guidance

|---|---|---|
| `AGENTS.md` correctness invariants ("scope temporary fallbacks narrowly and clear them when authoritative state arrives") | The note is state derived from one read | It is cleared when the next read starts, so only the latest read decides what the detail shows. |
| `openchamber-change-discipline` | Source change in shared UI | One added reset in the owning effect, matching the desktop entry. A regression test drives the real surface. |

`theme-system` and `locale-ui-patterns` don't apply: no styling or text changed.

## Validation

| Check | Result |
|---|---|
| `bun test ./src/apps/MobileChangesSurface.test.tsx` in `packages/ui` | Pass. The existing scenario now opens `nested/` while the file-diff route answers 422 and sees the note, goes back, lets the route answer with content, and reopens. The note must be gone and the content shown. With `MobileChangesSurface.tsx` from `main`, this fails: `Expected to not contain: "Separate Git repository"`. |
| `bunx tsc --noEmit -p .` in `packages/ui` | Pass |
| `eslint` on both changed files | Pass |
| `bunx oxlint --no-ignore` on both changed files | No findings on changed lines |

Not verified: a live run. A real repository does not naturally produce this sequence, as explained in Non-goals, so the component test covers it.

## Visual evidence

None. For a nested repository, the first open looks the same as before, and #3630 shows it. The fix only affects a later read of the same path, which the test above covers.

## Risks and failure behavior

- When a read of a nested repository starts, the note clears and the loading state shows until the 422 arrives. That is the same flow as any other diff read.
- None identified beyond that: a single state reset in one effect, with no new requests.


yulia-ivashko:
I ran the review locally before opening this, following `.opencode/agent/pr-review-bot.md` as written. The verdict was PASS with no findings. It confirmed that the finding from the #3630 review is fixed. It also checked the effect's dependencies, so routine status polling does not re-run the read and make the note flicker. Nothing needed changing.

Issue #3633, opened 2026-09-17
@openchamber/web@1.24.0 uninstallable: pins unpublished @openchamber/sdk@1.23.1


CHUNLEUNG910:
## @openchamber/web@1.24.0 uninstallable: pins unpublished `@openchamber/sdk@1.23.1`

**Version:** @openchamber/web 1.24.0 (npm `latest`)
**Platform:** macOS arm64, bun 1.3.14 (also fails with npm)

### What happened
`bun add -g @openchamber/web@latest` (and plain `npm i -g`) fails to resolve dependencies:

```
error: No version matching "1.23.1" found for specifier "@openchamber/sdk" (but package exists)
error: @openchamber/sdk@1.23.1 failed to resolve
```

### Root cause
`@openchamber/web@1.24.0` declares an **exact** pin:

```json
"dependencies": {
  "@openchamber/sdk": "1.23.1"
}
```

But `1.23.1` was never published to npm. The only versions that exist are:

```
[ '1.23.2-preview.1', '1.24.0' ]
```

### Expected
Either publish the missing `@openchamber/sdk@1.23.1`, or release `@openchamber/web@1.24.1` with the pin corrected (e.g. `1.24.0`).

### Notes
- The desktop app is unaffected (it ships its own bundled `web-dist`).
- Only the standalone/global `@openchamber/web` install path is broken, leaving users stuck on 1.23.2.

GuoChen-thlg:
+1

CHUNLEUNG910:
Confirmed reproduced in a clean isolated directory (no global install touched):

```
$ bun add @openchamber/web@1.24.0
bun add v1.3.14 (0d9b296a)
Resolving dependencies
Resolved, downloaded and extracted [283]
error: No version matching "1.23.1" found for specifier "@openchamber/sdk" (but package exists)
error: @openchamber/sdk@1.23.1 failed to resolve
```

So this is 100% an upstream packaging issue, not environment-specific.

Agree with the suggested fix path: a `@openchamber/web@1.24.1` patch release with the dependency corrected to the existing `@openchamber/sdk@1.24.0` (rather than publishing the orphaned 1.23.1 after the fact). Happy to verify the fix here once 1.24.1 lands.

jakar:
Adding my +1. v1.24.0 is totally uninstallable right now.

presswizards:
[...]

PR #3635, opened 2026-09-17
fix(web): declare zod runtime dependency

presswizards:
## Intent

@openchamber/web crashes on startup on clean npm installs. Eight files under packages/web/server import zod, but package.json never declares it. Bun workspaces hide this because the root node_modules provides it. Plain npm installs do not, so the daemon exits before reporting ready with ERR_MODULE_NOT_FOUND for zod from server/lib/opencode/theme-runtime.js. I hit this on two remote Linux hosts while connecting the 1.24.0 desktop app, right after working around #3633.

This adds "zod": "^4.3.6" to packages/web dependencies. That matches packages/sdk (same range) and the version bun.lock already resolves. v4 is required, not v3: server/lib/guests/storage.js uses z.json(), which only exists in v4. It is the only v4-only API in the package, I checked.

## Non-goals

The unpublished @openchamber/sdk@1.23.1 pin from #3633. That lives in the release pipeline, not in source (web already says workspace:*), and needs a re-release. Also untouched: the openai@4 peerOptional on zod@^3, which conflicts with zod v4 under stock npm. openai v7 allows ^3.25 || ^4.0, so moving the openai line is the follow-up. That call belongs to a maintainer since it is a major bump, and I did not want to smuggle it into this fix.

## Affected surfaces

packages/web server runtime only. No UI, no routes, no persisted formats, no CLI output changes. Desktop, VS Code, and mobile pick this up through the next published web tarball. bun.lock is refreshed so frozen installs stay green. The refresh also syncs workspace versions the lock still had stale at 1.23.x.

## Repository guidance

|---|---|---|
| AGENTS.md PR handoff | Required reading before opening a PR | Read CONTRIBUTING.md and the template, filled below with current evidence |
| openchamber-change-discipline, dependency change | Manifest and lockfile edited | Smallest complete change, one manifest line plus lock sync, validation below |
| AGENTS.md changelog rule | changelog/ is read-only for fixes | No changelog file touched |
| communication-style | Human-facing PR text | Plain wording, no filler |
| Dockerfile workspace rule | Only applies to added/renamed workspaces | No workspace added, Dockerfile untouched |

On the "do not add dependencies" rule: the code already imports zod in eight files, so this declares an existing coupling instead of introducing a new one. There is no way to fix the startup crash without it.

## Validation

| Check | Result |
|---|---|
| JSON parse of edited manifest | Pass |
| npm view zod@^4.3.6 | Resolves, 4.3.6 through 4.6.5 |
| bun install, then bun install --frozen-lockfile | Pass, no changes |
| Runtime, two Linux hosts (node 22/npm 10 and node 24/npm 11) | web@1.24.0 with top-level zod v4 reaches "listening" and launches managed OpenCode. Before: ERR_MODULE_NOT_FOUND from theme-runtime.js |
| Package test suite and type-check | Not run. Manifest-only diff, no source touched, and the only test file present is a shim. Live runtime evidence above covers the risk. |

## Visual evidence

No rendered change. A manifest line cannot affect rendering. The user-visible effect is the daemon starting instead of exiting code 1.

## Risks and failure behavior

Stock npm installs of a future fixed tarball still need the openai peer sorted (see Non-goals). Until then, plain npm install -g keeps failing on ERESOLVE even with this fix, and --legacy-peer-deps stays the workaround. Rollback is one reverted line plus lock sync. No security, data-loss, or performance concerns: same zod major the workspace already runs.


bernhardberger:
Hit the same crash through plain npm, so this isn't bun-specific.

I patched the published web 1.24.0 tarball to point at the existing sdk 1.24.0 (the shipped 1.23.1 pin was never published), installed it globally with npm, and started it under systemd. It crash-looped with `Error: z.json is not a function`, same v4-only call in guests/storage.js this PR names. Rolled back to web 1.23.2, which runs fine.

Environment was Debian 13, node v22.22.2, npm 10.9.7. Happy to retest once a 1.24.1 candidate with this fix lands.

PR #3661, opened 2026-09-17
test(electron): stop the shell-probe timeout test flaking on macOS

yulia-ivashko:
## Intent

`packages/electron/shell-environment.test.mjs` subtest 8, "real probe timeout kills the attempt and falls back", failed intermittently on macOS: it expects the login-only fallback environment `{ FALLBACK: 'yes' }` and got `null`. On current main it reproduced 15 failures in 20 consecutive runs of the file on an otherwise idle machine. It was the only failing test in an otherwise green `bun run test`.

The cause is not a race between the timeout and the fallback assignment. `createShellEnvironmentLoader` applies one `timeoutMs` to **both** probe attempts, so the 100ms deadline the test uses to kill the `-il` probe also governs the `-l` fallback probe, which has to actually complete. On macOS the **first** exec of a freshly written executable costs about 300ms while the system validates it — three times the entire budget — so the fallback was frequently killed at its own deadline and the loader returned `null`.

Measured on this host (Darwin 25.6), spawning the same temp script repeatedly:

```
first exec  305.8 ms
then          3.4 ms
              3.9 ms
              4.1 ms
```

An instrumented run of the failing path shows the fallback probe getting SIGTERM'd with no output:

```
[0]   spawn -il
[107] -il close code=null sig=SIGTERM killed=true bytes=0
[107] spawn -l
[209] -l  close code=null sig=SIGTERM killed=true bytes=0
RESULT null
```

The fix pays that one-time validation before anything is timed, rather than raising the deadline: the test runs the script's cheap `-l` branch once and awaits its exit, then runs the probe with an unchanged `timeoutMs: 100`.

The cold exec had a second consequence: the `-il` probe was usually killed *during* process startup and never reached `exec /bin/sleep 10`, so the test was not proving the kill its name claims — only that something failed. It now records the probe's pid before the exec and asserts the process is gone after `load()` resolves.

## Non-goals

- No production behavior change. `shell-environment.mjs` is untouched; the single shared `timeoutMs` across both probe attempts is the intended contract and stays as is.
- No change to the other real-shell tests in the file. They use the default 5000ms deadline, where a 300ms first exec has no effect on the outcome.
- Not a general warm-up in the `withShell` helper. The scripts differ per test — one always sleeps 10s, others write megabytes — so a blind extra exec would be wrong for them. The warm-up is opt-in at the one call site that needs it.

## Affected surfaces

- `packages/electron` test suite only. One file changed: `packages/electron/shell-environment.test.mjs`.
- No runtime, package export, entrypoint, persisted format, route, or user-facing surface is affected. The Electron main/preload privilege boundary is untouched; no IPC, window, updater, deep-link, or child-process production path changed.
- Web, VS Code, hosted mobile, and Capacitor mobile are unaffected — the file is an Electron-only test and is imported by nothing.
- The subtest is already skipped on Windows (`skip: process.platform === 'win32'`), so Windows CI behavior is unchanged. The defect and the fix are macOS-specific in cause but the change is platform-neutral in effect: on Linux the extra exec costs a few milliseconds.

## Repository guidance

|---|---|---|
| `AGENTS.md` — "Use `package.json` scripts as the command source of truth"; "Report exactly what was and was not validated" | Any change needs the right validation, reported honestly | Validated with `node --test` on the file (the same command `scripts/run-isolated-tests.mjs` uses for it) and `bun run test` in `packages/electron`. The Validation table below states the repeat counts, the load conditions, and what was deliberately not run |
| `AGENTS.md` — "Do not run git or GitHub commands unless the user explicitly asks" | Session-level rule | Nothing was committed or pushed until the maintainer asked for a commit and a PR |
| `desktop-shell` skill — required context: read `packages/electron/README.md` and nearby `packages/electron` code | The changed file lives in `packages/electron` | Read `packages/electron/README.md` and `shell-environment.mjs` before editing, and assigned the behavior: this is test-harness code, not native behavior, so nothing moved into or out of the main process |
| `desktop-shell` skill — keep domain logic out of Electron; Electron owns only inherently native behavior | A tempting "fix" would have been to give the loader a separate fallback timeout for testability | Rejected. No production seam was added and no export widened; the flake is a property of the test's environment, not of the loader |
| `desktop-shell` skill — Validation: run focused Electron tests | Electron-scoped change | `bun run test` in `packages/electron` (26/26 files) plus 40 repeat runs of the changed file |
| `openchamber-change-discipline` skill — "Make the smallest complete change and validate at the narrowest level that covers the real risk" | Governs the shape of the fix | One file, +14/−1, no production code. Validation is package-scoped because the risk is package-scoped |
| `openchamber-change-discipline` skill — Risk Classification: this is "Local implementation" | No exported API, no cross-workspace or persisted contract is involved | Preserved the observable behavior of the suite and validated the owning package |
| `openchamber-change-discipline` skill — Test Design: "Prefer observable contracts, state transitions, failure handling"; "Do not weaken what a test proves" | The fix must not make the test easier to pass | The assertion set grew rather than shrank: the fallback environment is still asserted exactly, and the kill the test is named for is now asserted directly instead of inferred |
| `openchamber-change-discipline` skill — Completion Standard: run the review the automated reviewer will run and address its findings | The PR is a review handoff | The repository's PR reviewer was run locally against this diff and body before pushing. Verdict `PASS` with two non-blocking findings, both fixed in the same commit: the pid read now tolerates a missing file so the deliberate assertion message survives, and `warmExec` resolves on `error` so a failed spawn cannot crash the file |
| `openchamber-change-discipline` skill — Validation Matrix: `bun run dead-code` when files are added/deleted/renamed or export/import shape changes | The change adds an `import { spawn } from 'node:child_process'` to a test file | Not run, and stated as not run below. No source file was added, deleted, or renamed and no export or entrypoint changed; an added import inside a test cannot create unreferenced code |
| `openchamber-change-discipline` skill — run `bunx oxlint` on files you substantially rewrote | A test file was edited | Ran clean. The worktree sits under an ignored `.claude/**` path, so a temporary config without that pattern was used and then deleted; a deliberate-violation probe confirmed the linter really inspected the file rather than silently skipping it |
| `CONTRIBUTING.md` — "Claims about performance, memory, CPU, rendering, startup, or similar empirical behavior require relevant before and after measurements" | The diagnosis and the fix both rest on timing claims | Before/after pass counts and the spawn-cost measurements that explain them are included above and below, with the conditions they were taken under |

No `DOCUMENTATION.md` exists under `packages/electron`; `packages/electron/README.md` was read and describes no invariant that this change affects, so no documentation update is required.

## Validation

All on macOS (Darwin 25.6), Node v22.23.1, in this worktree.

| Check | Result |
|---|---|
| `node --test packages/electron/shell-environment.test.mjs` ×20, **before** the fix | 5 passed, 15 failed — every failure was subtest 8, duration ~207ms (two consecutive 100ms deadlines) |
| `node --test packages/electron/shell-environment.test.mjs` ×25, after the fix | **25 passed, 0 failed** (re-run ×25 after the review fixes: 25 passed, 0 failed) |
| Same ×15 with all 8 cores saturated by busy-loop processes | **15 passed, 0 failed** |
| Warm spawn latency of the temp script, 50 samples, idle | p50 3.0ms, p95 4.8ms, max 5.7ms |
| Warm spawn latency, 4 concurrent samplers + 8 busy cores | p50 8.4ms, p95 10.6ms, max 11.1ms — against the unchanged 100ms deadline |
| `bun run test` in `packages/electron` | 26/26 test files passed |
| `bun run type-check` in `packages/electron` (`node --check` on main.mjs and preload.mjs) | Passed |
| `node --check packages/electron/shell-environment.test.mjs` | Passed |
| `bunx oxlint` on the changed file | No findings. Verified non-vacuous with a deliberate `anti-slop/no-unknown-parameters` violation, which the same invocation reported |
| Repository PR reviewer, run locally against this diff and body | `PASS`. Two non-blocking findings, both fixed and re-validated before pushing |
| Leftover processes after the runs | None — checked for stray `/bin/sleep 10` probes and for the load generators |

Not verified, deliberately:

- `bun run dead-code` — reasoning in the guidance table above.
- Workspace-wide type-check and lint — nothing outside `packages/electron` is touched and the file has no importers.
- Linux and Windows behavior. The subtest is skipped on Windows. On Linux the fix is a few milliseconds of extra exec and the original flake is not expected to occur there, but that was not measured on a Linux host.
- Packaged desktop startup — no production code runs differently, so there is nothing for a packaged build to exercise.

## Visual evidence

No user-visible change. The diff touches only `packages/electron/shell-environment.test.mjs`, a `node:test` file that ships in no build: `packages/electron/package.json` lists `dist-bundle/main.mjs` and `preload.mjs` under `build.files`, so test files are not packaged, and no source file imports this one. There is no rendered surface, no string, and no styling for a screenshot to show.

The empirical claim this PR does make — that the test is now stable — is evidenced by the before/after pass counts and spawn-cost measurements in the Validation table rather than by images.

## Risks and failure behavior

- **Production risk: none.** No production file changed. `git diff --stat` for the branch is one test file, +14/−1.
- **New dependency on the warm-up holding.** If the warm exec were removed or reordered, the test would go back to flaking. That failure mode is now legible rather than silent: when the deadline expires during process startup the script never runs, so the missing pid file is swallowed and the assertion fails with "the timed-out probe must have started", which points at the cause instead of at a mysterious `null`.
- **Warm-up failure is contained.** `warmExec` resolves on `error` as well as `close`. A `spawn` that fails outright — `EAGAIN` under fork pressure, a `noexec` tmpdir — would otherwise throw an unhandled `error` event and take every test in the file down at once, or hang it, since `node --test` applies no default per-test timeout. Resolving is the right behavior because the warm-up is best effort and the real probe below still reports any failure.
- **Pid reuse in the new assertion.** `process.kill(pid, 0)` could in principle observe a recycled pid. The window is the microseconds between the probe's `close` and the assertion, and the existing "quit cancels a real in-flight shell" subtest already relies on the same pattern, so this adds no new class of risk.
- **Reaping, not a second race.** `load()` only resolves after the probe's `close` event, which Node emits after the child is reaped, so `ESRCH` is guaranteed at the point of the assertion rather than merely likely.
- **Residual timing dependence.** The test still spawns real processes, so it is robust rather than mathematically deterministic. The measured margin is the argument: an 11ms worst-case warm spawn under full CPU saturation against a 100ms deadline, with the slow branch sleeping 10s — about 9x headroom where it previously had none.
- **Rollback.** Reverting the commit restores the previous, flaky test and nothing else. No data, no persisted format, and no released artifact is involved.


yulia-ivashko:
A green CI run proves less than usual here. The suite runs this file once, and the bug was a flake that passed roughly one run in four. A single pass is exactly what the broken version did too, so I ran the checks locally and repeatedly instead.

Counts, all on macOS 25.6 / Node 22.23.1:

- before the fix: 5 of 20 runs passed
- after the fix: 25 of 25, and another 25 of 25 once the review findings below were folded in
- 15 of 15 with all 8 cores pinned by busy loops, which is the condition the flake was originally blamed on
- `bun run test` in `packages/electron`: 26 of 26 files

Spawn cost is the whole story. First exec of the freshly written temp script costs 305ms; every exec after it costs 3 to 4ms. The 100ms deadline applies to both probe attempts, so the fallback probe was dying at its own deadline while macOS validated the binary.

I also ran the repository's PR reviewer locally against this diff before pushing, rather than spending a ten minute round trip on a verdict. It came back PASS with two non-blocking findings, both real, both fixed in the same commit:

1. the pid read would have thrown `ENOENT` before the assertion message it was meant to print, so a future regression would surface as a bare "file not found" instead of the explanation
2. `warmExec` listened only for `close`, so a spawn that failed outright would have thrown an unhandled `error` event and taken every test in the file down at once, or hung it, since `node --test` applies no default per-test timeout

Everything above is from the current HEAD, 359cb9c.

PR #3667, opened 2026-09-18
fix(ui): do not hide uninstalled TTS models in downloader

### LINKED ISSUE #3665: [Bug] Local TTS model downloader hides uninstalled alternative models (cannot download non-default voices)
### Description
In the Voice settings page, the UI component responsible for listing and downloading local TTS models filters out models that are not installed, not downloading, and not the default model (`kokoro-en-v0_19`). 

This makes it impossible for users to discover or download alternative TTS models (such as `piper-pt_BR-faber-medium` or other language models) through the UI, because they never appear in the list unless they are already installed or currently downloading.

### Steps to Reproduce
1. Go to **Settings > Voice**.
2. Set **Provider** to `Local`.
3. Look at the available models to download/select.
4. Notice that only the default `kokoro-en-v0_19` (and any already installed models) are listed. Other catalog models like `piper-pt_BR-faber-medium` are completely hidden.

### Root Cause
In the frontend code, the filter function for the models effectively does this:
```javascript
const defaultTts = "kokoro-en-v0_19";
function filterModels(model) {
  return model.id === defaultTts || model.installed || model.downloading;
}
const visibleModels = catalogModels.filter(filterModels);
```
By filtering the catalog array, any uninstalled non-default model is removed from the UI.

### Expected Behavior
The UI should list all available models from the TTS model catalog so the user can click the "Download" button for the language/model they want to use.

Caio-HD:
## Intent

Resolves #3665. The `Local` provider in `Settings` -> `Voice` -> `Playback` hides every uninstalled TTS model, so the picker only ever shows one row and there is no path to install another language from the UI.

The original fix in this PR dropped the client-side filter and dumped the full 14-model catalog as a flat list. That solved the discoverability bug but crowded the panel and buried the active model. This revision keeps installed, downloading and errored rows inline where users expect them, and moves the rest into a collapsed `Available to download (N)` disclosure with a substring filter over id and description.

## Non-goals

- No change to the STT catalog, engine, download mechanism, or `/api/dictation/status` payload.
- No change to the default selected model, the voice picker below the list, or the auto-language behavior.
- No new UI dependency. Reuses `@/components/ui/collapsible` already used by `McpPage` and `RemoteInstancesPage`.

## Affected surfaces

`Settings` -> `Voice` -> `Playback`, `Local` provider only. Shared React UI only; no server, transport, sync, or Electron/mobile shell change. The 12 locale files gain three keys under `settings.voice.page.localTts` for the disclosure label, filter placeholder and empty-filter message.

|---|---|
| Web | Installed and in-flight models render inline. Uninstalled models sit in a collapsed disclosure with a filter. |
| Desktop (Electron) | Same shared UI; no preload/native change. |
| VS Code | Same shared UI inside the webview. |
| Hosted mobile / Capacitor mobile | Same shared UI; layout is vertical and already handled by the existing container. |

## Repository guidance

|---|---|---|
| `AGENTS.md`, `openchamber-change-discipline` | Shared UI change in `packages/ui` with locale contract touchpoints. | Change is scoped to `LocalTtsModelStatus` and its 12 locale keys. Row rendering is extracted to `LocalTtsModelRow` and reused inline and inside the disclosure, so the per-row markup is identical to the current PR HEAD. |
| `settings-ui-patterns` | The change lives in the Voice settings panel. | Reuses the same `Collapsible` primitive other settings sections use, keeps the panel single-column, and preserves the existing spacing tokens around the model list. |
| `ui-api-decoupling` | The list is driven by `/api/dictation/status?provider=local`. | The fetch, the `useLocalTtsModels` hook and the `DictationModelState` shape are unchanged. Partitioning happens client-side on the already-fetched array. |
| `locale-ui-patterns` | Three new user-facing strings. | New keys added to all 12 shipping locales alongside the existing Voice keys. No hard-coded English fallbacks in the component. |
| `theme-system` | New disclosure and filter input inside a themed panel. | Uses existing `Icon` primitive, existing muted-foreground tokens, and the pre-existing `VOICE_TEXT_INPUT_CLASS` used by the OpenAI-compatible fields. |

## Validation

| Check | Result |
|---|---|
| `bun run --cwd packages/ui type-check` | Pass. |
| `bunx oxlint packages/ui/src/components/sections/openchamber/VoiceSettings.tsx` | 3 anti-slop findings, all pre-existing on unchanged lines (two on `LocalModelPicker`, one on the `downloadProgress` typeof narrow that already lived in the row markup and moved verbatim into `LocalTtsModelRow`). No new finding introduced. |
| Playwright run against the local dev server (`http://127.0.0.1:5180`) | 7 of 7 assertions pass: Voice section opens, disclosure trigger visible with the `(N)` count (12 on this environment), expanding it shows the filter input and additional rows, filtering by `spanish` reduces the list to 11 rows, and the no-match message renders for a gibberish query. Screenshots below were captured by that run. |
| Manual smoke on the same dev server | Provider `local` renders `Kokoro TTS (English, natural voices)` and `Piper TTS (Portuguese (Brazil))` inline with their existing install/delete affordances; expanding the disclosure exposes download buttons for the remaining 12 catalogued models. |

Not exercised: packaged Electron, VS Code webview, hosted mobile, and Capacitor. The diff is UI-composition only over primitives already used in those runtimes and does not touch shell code.

## Visual evidence

Same dev server, same `Local` provider, three states of the new disclosure.

| Collapsed (default) | Expanded | Filtered by "spanish" |
|---|---|---|
| Voice settings with the local provider selected. The two installed models render inline and the disclosure Available to download (12) is collapsed. | The disclosure is expanded, showing a filter input and the twelve uninstalled Piper models with a download button on each row. | The filter input contains 'spanish' and the disclosure shows the single Spanish Piper row. |

## Risks and failure behavior

- The disclosure starts collapsed on every mount. The filter input state is local to `LocalTtsModelStatus` and does not persist across settings closes. That is intentional for this scope; adding persistence would require a config store key that is not in this PR.
- Partitioning is `installed || downloading || downloadError` inline, everything else in the disclosure. An errored model therefore stays visible in the main list so the user can see the failure without expanding anything.
- The filter is a case-insensitive substring match over `description` and `id`. Locale-specific collation is not applied; the descriptions in `LOCAL_TTS_MODEL_CATALOG` are English and the id is a stable ASCII slug.
- If the API returns an empty `ttsModels`, the component returns `null` as it does on the current PR HEAD.
- Rollback is a straight revert of this commit; no data migration, no config key, no schema change.


Caio-HD:
The flat list this PR originally shipped was bad. It solved the discoverability bug from #3665 but crowded the panel and pushed the active model below the fold on a normal viewport, which is worse than the original filter it replaced.

Latest push (e6f88db82) keeps the fix but partitions the list. Installed, downloading and errored rows still render inline where the user expects them. The rest sit in a collapsed "Available to download (N)" disclosure with a substring filter over id and description. Reuses the existing `Collapsible` primitive already used by `McpPage` and `RemoteInstancesPage`, no new dependency.

Screenshots and validation are in the updated PR body. Playwright asserts collapsed default, expanded count, filter reduction and no-match state against the local dev server.

/oc-review

PR #1900, opened 2026-06-28
fix: load KaTeX CSS via JS import to resolve font file paths correctly

moisesvalero:
Closes #1880

## Root cause

KaTeX CSS was loaded via CSS `@import` in `packages/ui/src/index.css`. Lightning CSS inlines the CSS content but does not resolve relative `url(fonts/...)` references to KaTeX font files, causing 404s at runtime and fallback to upright (roman) font.

## Fix

1. Remove `@import "katex/dist/katex.min.css"` from `packages/ui/src/index.css`
2. Create `packages/ui/src/styles/katex-css.ts` with a JS import of the KaTeX CSS
3. Import it from all three app entry points (`main.tsx`, `mobile-main.tsx`, `mini-chat-main.tsx`)

This ensures Vite's JS-import pipeline correctly resolves the font file URLs (copies `.woff2` to `dist/assets/` and rewrites URLs), while keeping the CSS loaded eagerly with the main bundle.

PR #3669, opened 2026-09-18
fix(chat): load KaTeX CSS via JS imports so math uses KaTeX typefaces

### LINKED ISSUE #1880: [Bug] Math formulas rendered in upright (roman) font instead of italic — KaTeX font files not loaded from CSS @import

Dawnfz-Lenfeng:
## Intent

Chat math rendered with the wrong typeface: variables came out in a fallback (container/Times) font instead of KaTeX's math fonts, so formulas looked off even though the LaTeX was correct.

Root cause (openchamber/openchamber#1880): KaTeX's stylesheet was pulled in through a CSS `@import "katex/dist/katex.min.css"` in `packages/ui/src/index.css`. Tailwind v4's Lightning CSS inlines that import but does not resolve its relative `url(fonts/KaTeX_*.woff2)` references, so the font files were never emitted into the build. At runtime the CSS asked for `/assets/fonts/KaTeX_*.woff2`, those 404'd, and every formula fell back to the container typeface.

The stylesheet now loads through a JS import (`packages/ui/src/styles/katex-css.ts`). Vite's JS-import pipeline resolves the font URLs: it copies the files into the build and rewrites the references. The import is added at every entry point that already imports `styles/fonts`, before `index.css`, so it stays eager (no late stylesheet injection) and the theme overrides in `index.css` still win.

Closes #1880.

## Non-goals

- No change to math delimiters or parsing. Single-dollar `$...$` support is #3592, a separate change.
- No change to the KaTeX version, the font families, or the theme overrides in `index.css`.
- Not touching native (Capacitor) font bundling or the Markdown renderer itself.

## Affected surfaces

Every runtime that renders chat markdown, since this lives in `packages/ui`:

- Web (`packages/web/src/main.tsx`, `mobile-main.tsx`, `mini-chat-main.tsx`)
- Desktop / Electron mini chat and mobile web surface (they bundle the same shared app modules)
- VS Code webview
- The shared UI entry (`packages/ui/src/main.tsx`)

The VS Code webview is built by `packages/vscode` from `renderVSCodeApp`, not by the web entries. Covering only the three web entries — as the open #1900 does — would remove the KaTeX stylesheet from that build entirely, so all entry points that load `styles/fonts` were updated.

New source file: `packages/ui/src/styles/katex-css.ts`. No exports, persisted data, routes, or package contracts change.

## Repository guidance

|---|---|---|
| `AGENTS.md` always-on constraints | Source change across `packages/ui` and `packages/web` | No dependencies added, no changelog edits, diff limited to the import sites plus the new CSS module |
| `openchamber-change-discipline` skill | Source modification + a new file | Classified as local implementation/dependency-boundary risk; every consumer of `index.css` was traced and updated; the new module is a side-effect import, and its consumers were verified at build-output level |
| `performance-engineering` skill | Stylesheet placement affects first render | The KaTeX stylesheet stays eager with the main bundle (the reason it was moved out of the lazy markdown chunk originally), so no late style recalc / layout flash is reintroduced |
| `bun run dead-code` requirement for added files | `packages/ui/src/styles/katex-css.ts` is new | Knip reports no findings for the new module (see Validation) |
| Nearest `DOCUMENTATION.md` / README | Module discovery | No doc or README documents stylesheet loading; the comment in `index.css` and the new module header are the owning documentation and now describe the JS-import requirement |

## Validation

| Check | Result |
|---|---|
| `bun run --cwd packages/web build` at `6e58589` (before) | `0` KaTeX font files in `dist`; emitted CSS contains `url(fonts/KaTeX_*.woff2)` — the broken relative refs |
| `bun run --cwd packages/web build` on this branch | `19` KaTeX `.woff2` emitted to `dist/assets`; CSS references `/assets/KaTeX_Main-Regular-*.woff2`, and `curl` of an emitted font returns 200 |
| `bun run --cwd packages/vscode build:webview` on this branch | `59` KaTeX font files under `dist/webview/assets`, CSS refs `./KaTeX_*.woff2` colocated with them |
| `bun run --cwd packages/ui type-check`, `packages/web type-check`, `packages/vscode type-check` | Pass, no errors |
| `bun run --cwd packages/ui lint`, `packages/web lint` | `ui`: 0 errors, 1 pre-existing warning in `MobileChangesSurface.tsx` (untouched); `web`: clean |
| `bunx oxlint` on the changed files | Only pre-existing findings in untouched code |
| `bun run dead-code` | No findings for `katex-css.ts`; remaining report is pre-existing backlog |
| All `index.css` importers now import `katex-css` | Verified by diffing the importer lists (8/8) |

Not verified: a real Electron/VS Code/Capacitor runtime session, and PWA/service-worker cache behavior on update (an already-installed app may serve the old `index.html`/CSS until reload — worth a hard refresh after updating).

## Visual evidence

Same KaTeX-rendered message, rendered from the **two real builds** served from their own `dist` directories. The only difference is the stylesheet/font pipeline, so the difference is purely the typeface:

- **Before** (upstream/main build): font files 404 → variables and operators fall back to the container font.
- **After** (this PR build): KaTeX fonts are emitted and loaded → proper Computer Modern math typeface, italic variables, correct fraction and large operators.

KaTeX font fallback before/after

## Risks and failure behavior

- **A missed entry point loses more than fonts.** Removing the CSS `@import` means any entry that does not import `katex-css` loses the whole KaTeX stylesheet for math. Mitigated by mirroring every existing `styles/fonts` import site (8 importers of `index.css`, all covered) and by a comment in `index.css` recording the requirement.
- **Cascade order.** `katex-css` is imported in the same position as `styles/fonts`, before `index.css`, so the light/dark overrides there still apply.
- **Bundle weight.** The woff2 set is ~292 KB and is split into its own CSS chunk; browsers only fetch the faces actually used, so chat without math pays nothing at runtime.
- **Stale caches.** Installed PWAs/service workers can serve the previous CSS/HTML; users may need a reload for the fix to appear. Rollback is a single-commit revert, which restores the previous (fallback-font) state.


yulia-ivashko:
Thanks, this is a careful fix. I checked our shipped builds and they do request `fonts/KaTeX_*.woff2` with no font files in the bundle, and every entry that imports `index.css` picks up the new stylesheet import here, VS Code and mobile included. Merging.

Issue #3389, opened 2026-09-06
[Bug] Hitting 'restart to update' doesn't gracefully shutdown terminal sessions


maxiedaniels:
### What's wrong? Maybe some steps to reproduce.

If i hit 'restart to update' it kills everything within probably a second or two. I have some servers running that need probably 10 to 15 seconds to shutdown. I would think since this app has terminals and is built to run servers, it should NOT forcefully kill things unless it's taking too long, in which case maybe a confirmation to force kill servers would make sense.

### Where does it happen?

Desktop (macOS)

### Version (if known)

1.22.1

### Logs (optional)

```shell

```

maxiedaniels:
Just to be clear, this occurs also when you do "restart" in the File menu on Mac, so that would need to be covered too.

PR #3671, opened 2026-09-18
fix(desktop): ignore duplicate restart-to-update clicks while the install is in flight

### LINKED ISSUE #3670: [Bug] Second click on "Restart to Update" during the install window cancels the quit and the update never installs

A second "Restart to Update" click while the first install is still in flight doesn't just fail — it cancels the app's pending quit, so Squirrel's installer then aborts because the app is still running, and the update is left uninstalled until OpenChamber is fully restarted manually.

Repro (macOS, 1.22.2 → 1.23.2):

1. An update finishes downloading; the dialog shows the "Restart to Update" button.
2. Click it **once**. Nothing visibly happens: the button stays active and the dialog doesn't change — the app needs a while to tear down (window state persist, OpenCode sidecar, terminals).
3. After 3+ seconds, click it again (very natural — the button looks dead).
4. The error block shows "The updater stopped after a failed install. Quit OpenChamber, open it again, and retry the update." The app keeps running the old version.

What actually happens, main-process side:

- The **first** click invokes `autoUpdater.quitAndInstall()` (updater in ready state) — the install flow starts, ShipIt picks up the request, the app begins quitting.
- The **second** click hits Squirrel's one-shot state machine, which rejects with `SQRLUpdaterErrorInvalidState` ("The command is disabled and cannot be executed").
- `installDownloadedUpdate()` treats that as a failure: `fail()` → `rollbackQuitState()` resets `quitRequested`/`installingUpdate` and rejects the IPC — while the first install attempt is still under way. The app therefore never exits.
- ShipIt, which requires the target app to be fully quit, waits and then aborts: `SQRLInstallerErrorDomain Code=-9 "App Still Running Error"`.

Notes:

- The `desktop_restart` handler has no reentrancy guard — `state.installingUpdate` is set but never checked on entry, so every extra click reaches `quitAndInstall()` again.
- This is the missing piece after #3204, which deliberately kept the restart button clickable for retries and documented that later clicks hit the now-disabled Squirrel session. Closest cousins: #3027 (update downloaded but never installed; reported on macOS too), #3389 (ungraceful shutdown on this same path, which widens the window where the button looks dead).
- Not a signature/quarantine issue: old and new bundles verify with `codesign --verify --deep --strict`, both signed by the same Team ID, and the downloaded zip's SHA-512 matches `update-info.json`. The same flow installs cleanly when clicked exactly once — the previous update on my machine went through on a single click, and both attempts that failed today had two clicks 2.6 s / 6.2 s apart.

Suggested fix, two parts:

1. **Make `desktop_restart` idempotent while an install is in flight**: when `state.installingUpdate` is already true (or by tracking the pending `installDownloadedUpdate()` promise), reuse/resolve the pending call instead of invoking `quitAndInstall()` a second time. That removes the invalid-state error path and the quit rollback entirely.
2. **Show the in-progress state in the dialog**: while an install is in flight, **disable the Restart button so it cannot be clicked again**, and make the state visually explicit — a spinner in the button plus a "Restarting…" label — so it's obvious the first click registered and something is happening. Right now the dialog gives zero feedback for the entire teardown (well over 3 s in practice), which is exactly what invites the second click.

### Where does it happen?

Desktop (macOS 26, arm64, app in /Applications)

### Version (if known)

1.22.2 updating to 1.23.2

<details>

<summary>Screenshot 1:</summary>

Image

</details>

### Logs (optional)

`~/Library/Logs/OpenChamber/main.log` — both attempts; the two `desktop_restart` calls are the clicks:

```
[2026-09-16 00:17:19.337] [info]  [electron] update-downloaded version=1.23.2
[2026-09-16 00:17:20.712] [info]  [electron] desktop_restart applyUpdate=true packaged=true   <- 1st click
[2026-09-16 00:17:23.296] [info]  [electron] desktop_restart applyUpdate=true packaged=true   <- 2nd click
[2026-09-16 00:17:23.297] [warn]  Error: The command is disabled and cannot be executed
    at MacUpdater.quitAndInstall (.../app.asar/dist-bundle/main.mjs:25907:30)
    at Immediate._onImmediate (.../main.mjs:32802:19)
[2026-09-16 00:17:23.300] [error] [electron] update install failed Error: The command is disabled and cannot be executed
...
[2026-09-16 00:18:08.548] [info]  [electron] desktop_restart applyUpdate=true packaged=true   <- 1st click (new instance)
[2026-09-16 00:18:14.713] [info]  [electron] desktop_restart applyUpdate=true packaged=true   <- 2nd click
[2026-09-16 00:18:14.716] [warn]  Error: The command is disabled and cannot be executed
```

`~/Library/Caches/dev.openchamber.desktop.ShipIt/ShipIt_stderr.log` — both install attempts aborted because the app never quit:

```
2026-09-16 00:17:38.877 ShipIt[52761] Beginning installation
2026-09-16 00:18:09.480 ShipIt[52761] Aborting update attempt because there are 1 running instances of the target app
2026-09-16 00:18:09.486 ShipIt[52761] Installation cancelled: Error Domain=SQRLInstallerErrorDomain Code=-9 "App Still Running Error" UserInfo={NSLocalizedDescription=App Still Running Error, NSLocalizedRecoverySuggestion=All instances of the target application should be quit during the update process}

2026-09-16 00:20:09.495 ShipIt[55568] Beginning installation
2026-09-16 00:20:16.328 ShipIt[55568] Aborting update attempt because there are 1 running instances of the target app
```

For contrast, the same flow with a single click installed the previous update cleanly:

```
2026-09-08 10:19:36.776 ShipIt[66699] Beginning installation
2026-09-08 10:19:39.499 ShipIt[66699] Moved bundle contents ...
2026-09-08 10:19:41.033 ShipIt[66699] Installation completed successfully
```

slvvko:
## Intent

Fixes #3670.

A second click on "Restart to Update" while the first install is still shutting the app down used to hit Squirrel's one-shot `quitAndInstall()`, throw `SQRLUpdaterErrorInvalidState`, and roll the pending quit back — so ShipIt aborted with "App Still Running Error" and the update never installed. The button also looked dead the whole teardown window, which is exactly what invites the second click.

Two layers:

- Main process: `desktop_restart` with `applyUpdate` short-circuits when an install is already in flight, so a duplicate click joins the in-flight install instead of invoking `quitAndInstall()` again and triggering the `rollbackQuitState()` path mid-install. The guard keys on `state.updateInstallPending` — set synchronously when the install starts — plus `state.installingUpdate` for the tail after the installer takes over the exit.
- Dialog: `useUpdateStore` gains an `installing` flag, set when the restart invoke starts and cleared when it fails (or the dialog is dismissed). While it is set, the button is disabled and shows a spinner with "Restarting..." — the same disabled-spinner pattern the Downloading button in the same footer already uses. A duplicate click now never reaches the invoke at all.

**Update (post-merge of main, `a5b61dfc`):** main landed #3687 while this PR was open, which moves the quit/install flags into `installDownloadedUpdate()` — they are set only after background shutdown, so checking `installingUpdate` at handler entry no longer covers the shutdown window where the double-click lands. The merge re-targets the guard to the new synchronous `updateInstallPending` latch and keeps `installingUpdate` for the post-handover tail; the renderer layer is unchanged by the merge (only `main.mjs` conflicted, everything else auto-merged).

## Non-goals

- #3389, the slow ungraceful shutdown that widens the dead-looking window, is untouched; this fixes the double-click fallout only.
- The retry-after-failure semantics from #3204 are unchanged: a genuinely failed install still surfaces its error and the button becomes clickable again.
- No changelog entry (maintainer's release-time work).

## Affected surfaces

- `packages/electron` — the `desktop_restart` IPC handler only. No preload or contract change: same command, same return shape; a duplicate now resolves `null` instead of rejecting.
- `packages/ui` — `useUpdateStore` (new transient `installing` flag, nothing persisted), `UpdateDialog` (new optional `installing` prop, default `false`), and the two `UpdateDialog` call sites that show the desktop restart button (`AboutSettings`, `SessionSidebar`).
- All 12 i18n dictionaries — new `updateDialog.status.restarting` key, fully translated everywhere; key parity is enforced by `messages.test.ts`.
- Web, mobile, and VS Code runtimes are unaffected: the store guard is desktop-only (`runtimeType === 'desktop'`), the prop defaults to `false`, and the other `UpdateDialog` callers (Header, MobileApp) are unchanged.

## Repository guidance

|---|---|---|
| root `AGENTS.md` (change discipline, validation, PR handoff) | Executable source change across two workspaces | Smallest complete change; focused tests + package-scoped type-check/lint per the validation matrix |
| `desktop-shell` skill | Updater behavior + privileged IPC | Guard lives in `main.mjs`; no preload shape change; renderer checks were never the security boundary, so nothing moves |
| `ui-api-decoupling` skill | Checked whether the renderer needs a new capability | Nothing new is exposed: the existing `desktop_restart` invoke and its resolve/reject contract are reused; the `installing` state is store-local |
| `openchamber-change-discipline` skill | Scope and validation risk | Local implementation risk only: 4-line main-process short-circuit ahead of untouched logic + a transient store flag; no new abstraction or contract |
| `theme-system` skill | Button visual states | The restarting state reuses the disabled-spinner button pattern verbatim from the Downloading button in the same footer (status token at half opacity, shared `Icon name="loader" animate-spin`); no hardcoded colors |
| `locale-ui-patterns` skill | New user-facing string | `updateDialog.status.restarting` added to en plus all 11 non-English dictionaries with real translations, not English stand-ins |

## Validation

| Check | Result |
|---|---|
| `bun test packages/ui/src/stores/useUpdateStore.test.ts packages/ui/src/lib/i18n/messages.test.ts` | 8 pass (4 new store tests + i18n key parity across all locales) |
| `bun run --cwd packages/electron type-check` (includes `node --check` on `main.mjs` and `preload.mjs`) | pass |
| `bun run --cwd packages/electron test` | 26/26 test files pass |
| `bun run --cwd packages/electron test:updater` (new on main) | 18/18 pass |
| `bunx oxlint` on the five changed TS/TSX files | Only pre-existing findings on untouched lines; new code is clean |
| `eslint` (packages/ui config) on the five changed files | pass |

All checks above re-run on the merged HEAD `a5b61dfc`. Not verified: the real install-in-flight flow. It requires a packaged build plus a downloaded signed update — in dev `app.isPackaged` is false, so `desktop_restart` takes the plain-restart branch and Squirrel's install path never engages. The new `updater-testing.md` reference on main prescribes the full loopback-feed e2e run for updater changes; that fixture needs separately packaged builds and has not been run for this PR. What stands behind the guard instead: the store tests, and it being a short-circuit ahead of the (now #3687-restructured) install sequence rather than a reordering of it.

## Visual evidence

**Update (2026-09-18):** after-state screenshot added following the review — no code changed for this.

After state, produced exactly as the review suggested — seeded via `useUpdateStore.setState({ downloaded: true, installing: true, runtimeType: 'desktop' })` in a dev build, the same flags the store test sets, then opened the dialog through the sidebar update button:

Update dialog showing the disabled Restarting… spinner button

The bottom-right button is disabled (`cursor-not-allowed`) with the spinner and "Restarting..." label, in the muted success tone — the Downloading button's pattern with `status-success` at half opacity. The genuine in-flight state only exists inside a packaged install, which dev and CI cannot reach (see Validation). The "before" behavior — a dead-looking button for the whole teardown, then the failure banner after the second click — is recorded in #3670 with the reporter's screenshot and logs.

## Risks and failure behavior

- A duplicate click now resolves `null` (success) while the original invoke keeps governing the outcome — if that install later fails, its rejection still surfaces the error once, as before. One error, not two, and the quit state never rolls back mid-install.
- Genuine post-failure retries are unchanged: after `fail()` rolls the state back, `installingUpdate` is false and the next click reaches the updater again, preserving #3204's documented retry behavior (Squirrel won't accept it in the same session; the existing `updater-disabled` message explains that).
- The store flag is cleared on failure and on `dismiss()`, so a dismissed dialog can never leave a stuck spinner; on success it deliberately stays set because the app is quitting.
- No persisted data, cross-runtime contract, or privilege change: the guard only short-circuits a duplicate of an already-authorized install request, and the capability assertions still run on the first invocation.


slvvko:
After-state screenshot for the review's evidence request — the fix state isn't reproducible through the real install flow outside a packaged build, so this was captured the suggested way: a dev build with the store forced to `{ downloaded: true, installing: true, runtimeType: 'desktop' }` (the same flags the store test sets), dialog opened through the sidebar update button.

Update dialog showing the disabled Restarting… spinner button

The button is `disabled` with the spinner + "Restarting..." label, in the success tone at half opacity — same pattern as the Downloading button. No code changed since `1f0ecc7c`; this is evidence only, also added to the PR description.

slvvko:
/oc-review the evidence screenshot requested by the previous review has been added to the PR description and a separate PR comment — please re-check that finding

PR #3675, opened 2026-09-18
fix(mcp): stop directory bootstrap from initializing MCP fleets

### LINKED ISSUE #3674: 1.23.x launches every known project's full MCP fleet at startup
### What happened

Since 1.23.0 (verified still present on 1.24.1), launching the app initializes the complete MCP server fleet for every known project directory, with zero interaction. On my config (27 MCP servers, ~20 known projects) the managed OpenCode process grew from 459 to 850 child processes within 24 seconds of an untouched launch, and memory kept climbing. I quit around what looked like the path to 20GiB; on 1.22.2 the same config launches quietly and only initializes MCP for directories I actually open.

Repro method: ps snapshots of the managed serve's process tree during a no-click launch. Released 1.23.2: 459 → 568 → 850 descendants in 24 seconds (~15/sec). Each herd is one directory's full config: 27 processes, roughly 1.3-2.2GB on my setup.

### Why

Three links, all verified in source:

1. In OpenCode, any first touch of an instance's MCP state eagerly connects every enabled server in the config (`mcp/index.ts`, the per-instance state initializer). Two endpoints touch that state: `mcp.status`, and `command.list`, because building an instance's command list enumerates MCP prompts (`command/index.ts`).
2. 9000cf7fd "fix(ui): decouple sidebar session loading from workspace initialization" (landed the day before 1.23.2) scoped the bootstrap's per-directory reads explicitly. Good perf intent, but it moved both MCP-touching reads into per-directory enrichment: `sdk.mcp.status({ directory })` and `sdk.command.list({ directory })` in `packages/ui/src/sync/bootstrap.ts`.
3. The sidebar session list declares bootstrap demand for every known project directory (`useSessionListSync`), so every project gets the fleet-init side effect at startup.

Before that commit, both reads were unscoped and funneled into a single instance, which is why 1.22.2 was safe: startup initialized exactly one fleet (the serve process's own cwd), and more fleets only appeared for directories actually used.

### Fix

PR to follow in a comment: stop reading MCP status and command list during directory initialization. Both surfaces already fetch on demand through their own stores (`useMcpStore`, `useCommandsStore`), and slash dispatch does a single live lookup when needed. With the fix, an untouched launch initializes zero fleets, and idle memory lands below the 1.22.2 baseline.

knorby:
## Intent

Fixes #3674.

Since 1.23.0 (still present on 1.24.1), launching the app initializes every known project directory's complete MCP fleet with no interaction. On the reporter's config (27 MCP servers, ~20 known projects) the managed OpenCode process grew from 459 to 850 child processes in 24 seconds of idle, tens of GB total.

Cause: 9000cf7fd correctly scoped the bootstrap's per-directory reads, but two of those reads touch OpenCode's MCP state, and any first touch eagerly connects that instance's entire config. `mcp.status` reads it directly; `command.list` enumerates MCP prompts while building the instance's command list. The sidebar declares bootstrap demand for every known project directory, so each project launched a fleet at startup. Before that commit both reads were unscoped and funneled into a single instance, which is why 1.22.2 was safe.

This PR removes both reads from directory initialization. MCP panels and command surfaces already fetch on demand through their own stores; slash dispatch now does one live lookup when a command matches neither store. An untouched launch now initializes zero fleets, and idle memory lands below the reporter's 1.22.2 baseline.

## Non-goals

- The sidebar's all-projects bootstrap demand is unchanged; only the MCP-touching reads moved out of bootstrap.
- OpenCode's eager MCP-state initialization and prompt enumeration are upstream OpenCode behavior, untouched here. Longer-term hardening (lazy per-server init, non-initializing peek endpoints) belongs there.
- Directories where a session actually uses an MCP tool still initialize that directory's fleet; that is OpenCode's design, not this bug.
- The dead `State.mcp` and `State.command` fields, written only by the removed reads, are removed; nothing else was pruned.

## Affected surfaces

- `packages/ui/src/sync/bootstrap.ts`, `types.ts`, `bootstrap.test.ts`, `DOCUMENTATION.md`, and `session-ui-store.ts` (slash dispatch and goal-mode command lookup).
- Shared UI, so web, desktop, VS Code, and mobile all get the change.
- Behavior: MCP surfaces fetch on first open instead of every project being pre-warmed; slash command dispatch for a name in neither store does one live lookup before the plain-prompt fallthrough.
- No persisted contracts, no server-side changes.

## Repository guidance

|---|---|---|
| `openchamber-change-discipline` skill | Sync source change | Smallest complete change: remove the two reads plus the fields only they wrote; validation scoped to the sync package, plus full-suite and build |
| `sync-state-invariants` skill | Bootstrap behavior change | Failure-is-not-empty preserved (reads removed, none added); dead state pruned with its only writer |
| `ui-api-decoupling` skill | SDK data access in shared UI | Removed calls; the one new live path uses the documented `opencodeClient.listCommandsWithDetails` wrapper |
| `performance-engineering` skill | Reported memory/CPU runaway on a spawn hot path | Structural removal of work, measured at production scale (27-server config, ~20 project dirs) with before/after process counts |

## Validation

| Check | Result |
|---|---|
| `bun test packages/ui/src/sync/bootstrap.test.ts` | 13/13 pass, including the regression case asserting zero `/mcp` and `/command` requests during directory initialization |
| Sync suite under the repo's isolated runner | Pass. Note: running `packages/ui/src/sync/` as one shared process shows 61 load-order-dependent failures, all of which also fail on pristine v1.24.1 and disappear under `scripts/run-isolated-tests.mjs` |
| `bun run test` (full) | 2486+ pass; 2-3 pre-existing failures in `packages/web/server` (`env-runtime.test.js`, which expects no system opencode while this machine has a brew install; `git/service.test.js`, a thousand-file fixture at the timeout boundary). None in the changed area; proven pre-existing by swapping upstream's versions of the five changed files and re-running |
| `bun run --cwd packages/ui type-check` | Clean |
| `bunx oxlint` on changed files | Clean except two pre-existing anti-slop errors in `types.ts`, present identically on pristine, only line-shifted |
| `bun run build` | Pass |
| Manual, macOS arm64, no-click launches | Released 1.23.2: 459→850 processes in 24s of idle. This branch on 1.24.1: flat. Idle memory below the reporter's 1.22.2 baseline |

## Visual evidence

No rendered UI change; the diff removes network reads and dead state. The behavior is empirical, so the before/after process counts in the validation table serve as the measurements, captured by the reporter as ps snapshots of the managed serve on both the released and fixed builds.

## Risks and failure behavior

- Slash dispatch for a name matching neither the commands store nor the skills store does one live `listCommandsWithDetails` lookup; on failure the input falls through to a plain prompt, the pre-existing fallthrough.
- First open of MCP surfaces fetches (brief loading state) instead of being pre-warmed.
- No persisted state, no migrations, no server changes. Rollback is a plain revert.


btriapitsyn:
Thanks, this one was painful for a lot of people and the write-up made it easy to verify. Merged. We'll follow up on the sidebar bootstrap demand itself so unopened projects stop creating instances at all.

PR #3701, opened 2026-09-18
fix(sidebar): close session menus when Rename is chosen

### LINKED ISSUE #3700: [Bug] Sidebar session Rename leaves the menu stuck open and never starts renaming
### Language

Renaming a session from the sidebar menu no longer works. The menu stays open and the inline rename field never appears.

Steps:
1. Right-click a session in the sidebar (or open its `⋯` menu).
2. Click **Rename**.

Expected: the menu closes and the inline rename input opens.
Actual: the menu stays stuck open and nothing else happens.

Regressed in #3627 (sidebar virtualization). The menus became controlled by the shared `openSidebarMenuKey`, and the close handlers in `SessionNodeItem` skip clearing that key while a rename is pending, so the row stays pinned in the virtualizer. The problem is that the same key drives the controlled `open` prop. Refusing to clear it on close keeps the menu open, so `onOpenChangeComplete(false)` never fires, and that callback is the one that starts the deferred rename. The two wait on each other.

### Where does it happen?

Web (all runtimes that use the shared sidebar)

### Version

main @ ebe2e107e

karimodm:
Fixes #3700

## Intent

Choosing **Rename** from a sidebar session's right-click or `⋯` menu should close the menu and open the inline rename input. Since #3627 the menu stays stuck open and the rename never starts.

The cause: the menus became controlled by the shared `openSidebarMenuKey`, which also pins the row in the virtualizer. To keep the row mounted through close, the close handlers skipped clearing the key while a rename was pending. But that same key drove the controlled `open` prop, so the menu never closed, `onOpenChangeComplete(false)` never fired, and the deferred rename never ran.

This PR moves the row's menu open state into `useSessionRowMenuState`. A local close request hides the menu right away. The shared key stays pinned until close completion, then the rename starts and the key is released. The row still stays mounted through the close transition, as #3627 intended.

## Non-goals

No change to rename editing, saving, AI rename, or the virtualizer's pinning rules. Project/folder menus in `sortableItems.tsx` are untouched; they don't defer work to close completion.

## Affected surfaces

`packages/ui` sidebar session rows (`SessionNodeItem`). No persisted or external contract changes.

|---|---|
| Web | Rename from session context/overflow menu closes the menu and opens inline rename |
| Desktop (Electron) | Same, shared UI |
| VS Code | Same, shared UI |
| Hosted mobile | Same for the overflow menu; long-press/context menu path unchanged otherwise |
| Capacitor mobile | Same as hosted mobile |

## Repository guidance

|---|---|---|
| `openchamber-change-discipline` | Source fix in shared UI | Smallest change that keeps #3627's pinning invariant; new hook owns the open/pin split and is covered by focused tests |
| `sidebar/DOCUMENTATION.md` | Documents that the owning occurrence stays mounted through menu close completion | Invariant kept; doc updated to say how the controlled `open` prop and pinned key are separated |
| AGENTS.md changelog rule | Fix PR | No changelog edits |

## Validation

| Check | Result |
|---|---|
| `bun test src/components/session/sidebar/sessions/useSessionRowMenuState.test.tsx` | 4 pass |
| Same test against the old close logic (guard reinstated, no close request) | both rename tests fail: menu stays open, rename never runs |
| `bun test src/components/session/sidebar/sessions/` | 49 pass, 1 fail. The failure (`useSessionActions` descendant snapshot) is pre-existing on `main` when the folder runs in one process; the file passes alone |
| `bun run type-check` (packages/ui) | pass |
| eslint + `bunx oxlint` on changed files | clean |

**Live run:** Not run. I didn't exercise it in a running app; the Base UI menu transition is covered only through the hook's controlled-state contract.

## Visual evidence

Behavior fix, no styling change. The menu now closes on Rename instead of staying open; no new visuals.

## Risks and failure behavior

If Base UI ever skipped `onOpenChangeComplete(false)`, the key would stay set and the close request would keep the menu hidden. Reopening either menu clears the request and resets the key, so nothing gets stuck. A plain close (no pending rename) releases the key immediately as before.


yulia-ivashko:
Thanks for the clear write-up and the fix. I ran this branch in the web app on macOS: Rename from both the context menu and the `⋯` menu closes the menu and opens the inline input, the duplicate row stays untouched, and the menus reopen and close normally afterwards. Merging.

yulia-ivashko (approved):

PR #3704, opened 2026-09-18
fix(chat): match managed chat directories case-insensitively

### LINKED ISSUE #3642: [Bug] Project-less chats vanish from sidebar when home path casing differs (case-sensitive isWithinRoot)
## Summary

On macOS, chats started without a project never appear in the sidebar's **Chats** section — in fact they don't appear *anywhere* — when the letter casing of the user's home directory as reported by `os.homedir()` differs from the canonical on-disk casing. This happens in the wild when the macOS user record's `NFSHomeDirectory` has different casing than the actual home folder (e.g. record says `/Users/wayne`, folder on disk is `/Users/Wayne`).

## Root cause

Two different case-sensitivity policies for path comparison:

1. The chats root is resolved from `os.homedir()` (`/api/fs/home` in `packages/web/server/lib/fs/routes.js`), which returns `$HOME` verbatim — lowercase `/Users/wayne` in the affected setup. The chat classifier then compares with a **case-sensitive** prefix check in `packages/ui/src/lib/chatDirectories.ts`:

   ```ts
   function isWithinRoot(directory: string, root: string): boolean {
     return !directory.split('/').some((part) => part === '..' || part === '.')
       && (directory === root || directory.startsWith(`${root}/`));
   }
   ```

2. OpenCode, however, records the session's `directory` in the canonical on-disk case: `/Users/Wayne/.config/openchamber/chats/2026-09-17/session-<uuid>`.

So `isChatDirectoryPath('/Users/Wayne/.config/openchamber/chats/...')` is checked against root `/Users/wayne/.config/openchamber/chats` → `startsWith` fails → the chat session is misclassified as a **project** session in `partitionSidebarSessions` (`packages/ui/src/components/session/sidebar/list/sessionCollection.ts`).

The misclassified session is then dropped entirely by `isKnownActiveSessionDirectory`, because its directory is not a registered project directory. Ironically, that check *does* compare case-insensitively (`normalizePath(...)?.toLowerCase()`), which is why ordinary project sessions are unaffected by the same casing mismatch — only project-less chats vanish.

Net effect: the session exists in OpenCode's DB and keeps working, but is invisible in the UI. Users report "chats not showing up at all".

## Expected behavior

Path classification on macOS (and Windows) should be case-insensitive, consistent with `isKnownActiveSessionDirectory` — e.g. lowercase both sides inside `isWithinRoot`, or normalize via `realpath` before comparing.

## Reproduction

1. On macOS, have a user record whose `NFSHomeDirectory` casing differs from the actual folder (check with `dscl . -read /Users/<name> NFSHomeDirectory` vs `ls -ld /Users/<name>`). This is more common than it sounds — it survives macOS migrations and some MDM/user-creation tooling.
2. Start OpenChamber, create a chat without selecting a project, send a message.
3. The chat never appears in the sidebar. Verified in the DB: the session exists with a capital-case `directory`, while `/api/fs/home` returns a lowercase `chatsRoot`.

## Workaround (confirmed working)

Fix the user record casing via System Settings → Users & Groups → right-click user → Advanced Options → set **Home directory** to the canonical case (e.g. `/Users/Wayne`), then restart. Chats immediately reappear.

## Environment

- OpenChamber desktop 1.24.0, macOS (APFS, case-insensitive — the default)
- Reproducible on current `main` (the case-sensitive `isWithinRoot` is unchanged)

## Suggested fix

In `packages/ui/src/lib/chatDirectories.ts`, make `isWithinRoot` case-insensitive on case-insensitive filesystems (darwin/win32), matching the treatment already used for project-directory membership. A simple `.toLowerCase()` on both operands would do; a `realpath`-based normalization of the cached roots would be more robust.

bashrusakh:
## Intent

Fixes #3642.

On macOS, `os.homedir()` can report a different letter casing than the canonical on-disk home directory (user record `/Users/wayne`, folder `/Users/Wayne`). The server-resolved chats root inherits that casing, while OpenCode persists session directories in canonical case. The exact-case prefix check in `isWithinRoot` then misclassified managed chats as project sessions, and `isKnownActiveSessionDirectory` dropped them because they are not registered project directories — project-less chats were invisible in the sidebar even though the sessions existed and kept working.

After this change, chat-root membership ignores letter case for classification and client-side delete containment, so those sessions classify as managed chats again. Helpers still return the cached root string unchanged, so scope keys and identities stay stable.

**Scope note (maintainer decision):** this PR deliberately fixes the classification boundary only. Whether to go deeper — canonicalizing server-resolved home-derived roots, or making server-side managed-root containment case-tolerant — is the maintainer's call. I did not make that decision here because it entails significantly larger changes to server-owned path semantics than this bug's boundary.

## Non-goals

- Server-side path canonicalization is untouched: `/api/fs/home` still returns `os.homedir()` casing verbatim, and server resolve/containment (`packages/web/server/lib/fs/routes.js`) still compares case-sensitively on POSIX.
- Server-side delete containment is untouched. On the affected macOS setup, deleting a chat can still have its directory-cleanup request rejected with 400 by `/api/fs/delete`; cleanup is best-effort and only logs a warning, so there is no user-visible failure and no data loss.
- Platform-conditional behavior was intentionally not added. The shared UI has no host-platform signal, and the active runtime can be remote, so a client-side `darwin`/`win32` gate would be wrong by construction.
- No changelog entry: per repository rules that is release-time maintainer work.

## Affected surfaces

- Shared UI path classification: `packages/ui/src/lib/chatDirectories.ts` (sidebar/mobile/header/persist-cache classification and the client-side delete guard), with its tests.
- Owning sync documentation: `packages/ui/src/sync/DOCUMENTATION.md`.
- No persisted or external contract changes: no API shape, storage, config, or dependency change; returned root identities are unchanged.

|---|---|
| Web | Managed chats classify case-insensitively; on unaffected hosts nothing renders differently. |
| Desktop (Electron) | Same shared classification; this is the runtime where the macOS home-casing mismatch was reported. |
| VS Code | Managed chats stay excluded from VS Code surfaces; the case-insensitive check does not re-enable them (regression test preserved). |
| Hosted mobile | Same shared classification, no surface-specific behavior change. |
| Capacitor mobile | Same shared classification, no surface-specific behavior change. |

## Repository guidance

|---|---|---|
| Root `AGENTS.md` | Source change in shared UI; minimal diff, correctness invariants, validation rules | Change limited to one UI module, its tests, and the owning doc; no deps, no server edits, no changelog; focused validation run. |
| `openchamber-change-discipline` skill | Implementing/fixing OpenChamber source | Right-level fix at the shared membership helper; no adjacent refactors; cached roots remain the returned identity. |
| `sync-state-invariants` skill | Directory membership feeds sidebar/session projection state | No store or sync semantics changed; runtime-keyed root caching, root identity, and projection behavior preserved and covered by regression tests. |
| `packages/ui/src/sync/DOCUMENTATION.md` (owning doc) | Documents the managed chat directory invariant | Invariant text updated to state case-insensitive membership and the unchanged protections. |
| `CONTRIBUTING.md` PR contract | Review handoff requirements | This body covers intent, non-goals, surfaces, guidance, validation, and risks; evidence limitations are stated. |

## Validation

All results were produced on commit `d5791ce84`, merge-base with `upstream/main` at `5df72db27`; upstream did not touch any of the four changed files since that base (`git diff --name-status 5df72db27 upstream/main -- <files>` is empty).

| Check | Result |
|---|---|
| `bun test packages/ui/src/lib/chatDirectories.test.ts` | 7 pass / 0 fail (25 expect calls) |
| `bun test packages/ui/src/components/session/sidebar/list/sessionCollection.test.ts` | 20 pass / 0 fail (43 expect calls) |
| `bun run --cwd packages/ui type-check` | Clean (`tsc --noEmit`, no diagnostics) |
| `bunx oxlint packages/ui/src/lib/chatDirectories.ts packages/ui/src/lib/chatDirectories.test.ts` | No findings |
| Failing-before proof | Temporarily restoring the exact-case comparison makes the new classification/root-identity assertions and the sidebar projection assertion fail (`chatSessions: []`); with the fix all pass. |
| Consumer regression | 49 pass / 0 fail across global sessions store, persist-cache, sessionNodeItemUtils, project context owner; 70 pass / 0 fail across session-ui-store and sidebar sections. |

**Live run:** Not performed. The changed behavior only manifests on a macOS host whose user-record home casing differs from the on-disk folder, which this Linux environment cannot reproduce. Evidence is the failing-before/passing-after unit check plus the sidebar-projection regression test.

## Visual evidence

No new rendered state. The change only alters path comparison: on affected hosts the existing Chats section shows sessions it already knows how to render, and on unaffected hosts nothing renders differently. A before/after capture would require the macOS casing mismatch described above, which is unavailable here.

## Risks and failure behavior

- Membership is now case-insensitive on case-sensitive filesystems as well. This matches the sidebar's existing case-insensitive project-directory membership (`sessionCollection.ts:82,175`), but a directory differing from a chats root only by case could be classified as a chat. The chats root is app-owned, so this is practically implausible, and `/api/fs/delete` independently re-checks containment server-side — the active workspace, its worktrees, or the managed roots — so client-side folding does not widen what the server accepts.
- The shared roots remain protected in any casing; dot segments, lookalike paths, and runtime switches are still refused (covered by tests).
- Rollback is a single-commit revert; no migration, persistence, or compatibility impact.
- Performance: two extra `toLowerCase()` calls on short strings per membership check; negligible.


bashrusakh:
Nit addressed: risk wording about server containment updated in the PR description (body only; code unchanged at d5791ce84).

yulia-ivashko:
Thanks for working on this and adding regression tests. The change fixes the reported macOS case, but we found a problem with applying case-insensitive matching to every server.

On Linux, `chats` and `Chats` can be separate directories. Because the same comparison also controls chat-directory cleanup, the change can treat a normal project under `Chats/project` as a managed chat directory. In a controlled test using this helper and the real server delete handler, cleanup deleted the project’s test file. The server permits operations inside the active workspace, so it does not block this case.

We’re going with [#3707](https://github.com/openchamber/openchamber/pull/3707), which asks the server’s filesystem to resolve the actual root paths and keeps exact matching in the client. That fixes the macOS mismatch while preserving distinct Linux directories and existing sidebar folder identities.

Thank you for helping address [#3642](https://github.com/openchamber/openchamber/issues/3642).

PR #3705, opened 2026-09-18
fix(browser): report drivability in browser.open results

### LINKED ISSUE #3641: [Bug] browser.open reports ok/opened:true when no controlling client is connected

**Expected:** when OpenChamber runs as the CLI/web server with no desktop client connected, `browser.open` should tell the agent that the opened page cannot be driven (or fail like every other browser action does).

**Got:** `browser.open` returns a success result (`{ url, opened: true }`), while every subsequent action (`browser.snapshot`, `browser.click`, …) fails immediately with:

```
No OpenChamber client connected here can control a page. Reading and
interacting with a page works when OpenChamber runs as its desktop
application; a web browser tab can display a page but cannot be driven.
```

The agent therefore builds on a false success: it believes it opened a page it can work with, then discovers on the next call that the environment is display-only.

**Steps to reproduce:**

1. Run the CLI server (`bun /usr/lib/node_modules/@openchamber/web/server/index.js --port 3000 --host 127.0.0.1`), version 1.23.2, and open the workspace in a plain browser tab. No desktop app connected.
2. Have the agent call `browser.open` with any URL → returns ok, `opened: true`.
3. Have the agent call `browser.snapshot` → 503 "No OpenChamber client connected…".

**Root cause (code-level):**

- `packages/web/server/index.js` — `emitRequest` counts **any** connected client for `browser.open` (`needsBrowserView = request.action !== 'browser.open'`), so the request reaches the web tab.
- `packages/ui/src/lib/browser/controlClient.ts` — handles `browser.open` via the registered `BrowserOpener` even with no capable view ("it is handled even with no view attached, since opening a tab is precisely what creates one") and posts `{ ok: true, data: { url, opened: true } }` — the response carries no indication that no `browser=1`-capable client is connected.
- Every other action then fails fast in `packages/web/server/lib/browser-control/broker.js` with the 503 above.

## What did you expect to happen?

The `browser.open` result should surface client connection state, so agents can branch immediately (e.g. fall back to local browser tooling) instead of discovering on the next action that the panel is display-only.

Per PR #2883's design intent ("the tool reports that honestly instead of blocking"), opening the panel even without a capable client is deliberate — so the suggested fix is to make the response honest, not to block opening:

- **(a) Preferred:** include a `drivable: false` / capability note in the `browser.open` result when no `openchamberBrowserCapable` client is connected. Preserves the useful behavior of opening the panel for the user to see, while keeping the agent informed.
- (b) Alternatively, treat `browser.open` like other actions in `emitRequest` when no capable client exists, returning the same 503.

## Where does it happen?

Desktop Web

## Version

1.23.2 (`@openchamber/web`)

bashrusakh:
## Intent

`browser.open` answered `{ url, opened: true }` even when the connected client could only display the page (a plain browser tab), so an agent believed it could drive what it had just opened and only found out on the next call, which fails with the display-only error. This makes the open result honest: every successful open reports whether the client that opened the page can also drive it.

Fixes #3641. This follows the issue's preferred option (a): the panel still opens for the user to see, and the response now states the capability so the agent can branch immediately.

## Non-goals

- The deliberate display-only open (PR #2883 design intent) is unchanged.
- Server routing, the zero-listener 503 fail-fast, the claim/result protocol, and all other browser actions are untouched.
- The mixed-client claim race is unchanged; `drivable` reflects whoever actually opened the page, not whether some other connected client could drive a page.

## Affected surfaces

|---|---|
| Web | Open still displays the page; every open result carries `drivable: false`. The viewport fallback note no longer points at `browser.resize`, which cannot run on this host. |
| Desktop (Electron) | Declares `browser=1` and reports `drivable: true`; existing fields and note text unchanged apart from the added field. |
| VS Code | Does not subscribe the control event stream (`isVSCodeRuntime` short-circuit), so `browser.open` is not handled here; unaffected. |
| Hosted mobile | Same shared-UI web path: display-only, `drivable: false`. |
| Capacitor mobile | Same hosted-mobile path: display-only, `drivable: false`. |

## Repository guidance

|---|---|---|
| `AGENTS.md` + `openchamber-change-discipline` | Source change in shared UI plus a server-owned contract | Focused diff, no unrelated refactors; the new helper pays for the shared invariant; focused tests and workspace checks were run |
| `ui-api-decoupling` | Shared UI capability reporting across runtimes; the contract crosses renderer and server | Declaration and report share one capability function (`hostCapability.ts`) so they cannot drift; no new routes or transport changes; runtime parity above |
| `packages/web/server/lib/browser-control/DOCUMENTATION.md` | Owning module documents broker/client invariants | Added the per-claimer `drivable` invariant where `browser.open` handling is specified |
| `packages/web/server/lib/openchamber-control/actions.js` | Agent-facing tool description | One sentence documenting that `drivable: false` means display-only |

## Validation

| Check | Result |
|---|---|
| `bun run type-check` | All workspaces pass |
| `bun run lint` | All workspaces pass |
| `bun run build` | Exits 0 (all workspace builds; working tree stays clean) |
| `bun run --cwd packages/ui test` | 527/527 files pass |
| `bun run --cwd packages/web test` | 225/225 files, 2517 tests pass |
| Focused `controlClient` + `openchamberEvents` tests | 15 pass |
| `bunx oxlint <changed paths>`, `bun run dead-code` | Clean for the new module; only the pre-existing anti-slop backlog elsewhere |
| `bun run test` (full chain) | Blocked locally by environment, not the diff: local bun 1.3.14 vs the pinned `bun@1.4.2`; `scripts/bump-version.test.mjs` and `packages/sdk/scripts/examples.test.ts` fail on packaging/bundling output under the older bun. The ui and web suites were then run separately and pass; the vscode and electron suites were not run locally. |
| CI `checks` job (pinned bun, Linux) | Pass (8m30s): install, build, type-check, lint, changelog check, full `bun run test`, Electron packaging unit tests |

**Live run:** Performed on this candidate (web runtime, headless Linux). The worktree server was started from this branch's built `packages/web/dist` (`bun packages/web/server/index.js --port 5272 --host 127.0.0.1`, isolated data dir), and the UI was loaded in a plain browser tab (no `__OPENCHAMBER_ELECTRON__`, so no `browser=1` capability declaration). Invoking the same route the `openchamber_web` tool uses (`POST /api/openchamber/control`) from that environment returned:

- `{"url":"https://example.com/","opened":true,"drivable":false}` — open succeeds, and the result now reports that this client cannot drive the page.
- `{"url":"https://example.com/","opened":true,"viewportApplied":false,"drivable":false,"note":"The panel had no browser view yet, so the viewport was not applied. This client can display a page but cannot drive one, so browser.resize is not available here."}` — the display-only note replaces the resize follow-up.
- The following `browser.snapshot` still fails with "No OpenChamber client connected here can control a page... a web browser tab can display a page but cannot be driven", which the agent can now anticipate from `drivable: false`.
- The page visibly opened in that client's browser panel.

## Visual evidence

No rendered UI changes: the diff adds a JSON field to the agent tool result and changes one agent-facing note string; no component reads or displays `drivable`. Before/after screenshots of the app would be identical; the live-run output above is the behavior being changed.

## Risks and failure behavior

- Additive field only; existing result fields and notes are unchanged, and other actions get no new field.
- No persisted state, schema, migration, or route changes; reverting the commit restores the previous behavior.
- Known non-blocking review notes: the controller-path merge uses `Object.assign({}, data, { drivable })`, which would coerce a non-object controller result on an open — no production registrar returns non-objects (`BrowserPane` always returns objects), so it is latent; and display-only viewport opens still wait the attach deadline before answering (pre-existing latency, unchanged).


bashrusakh:
/oc-review Live run evidence is now recorded in the PR body (Validation → Live run). Summary: with the built candidate served to a plain non-Electron tab, the tool route returned {"url":"https://example.com/","opened":true,"drivable":false}; the viewport variant returned drivable:false with the new display-only note ("...browser.resize is not available here."); the following browser.snapshot still failed with the display-only message. Also: the order-dependence nit on openchamberEvents.test.ts looks like a false positive — its afterEach deletes globalThis.window, so __OPENCHAMBER_ELECTRON__ cannot leak between tests; no code change was made for it.

PR #3711, opened 2026-09-18
fix(ssh): reuse the running managed server instead of starting one per connect

### LINKED ISSUE #3710: Remote managed servers are never reaped: orphan web + opencode processes accumulate until OOM
## What happens

Each remote-host connect starts a web server on the remote (`server/index.js --port <random>`), and each session spawns managed `opencode serve` children on loopback. When sessions close or the client disconnects, nothing stops them. They accumulate across days. I found 69 opencode servers (about 12 GB) plus 5 web servers on a 15 GB RAM homeserver, and 19 opencode plus 18 web servers on a 62 GB dedicated box. The small box OOM-killed two production VMs overnight.

## Details

- Desktop app 1.24.0, remote web 1.24.0, stock managed-server flow, no special config.
- Orphan web servers keep supervising: killing an opencode child gets it relaunched within minutes (I watched 18 respawns). Cleanup requires killing the parent web servers first.
- Survivors dated back days (oldest Sep 16, still present Sep 18). New ones appear with every connect, including failed connects.
- Live sessions are identifiable: their web port holds ESTABLISHED loopback connections (4 in my case). Orphans hold zero. That distinction is reliable enough to build a reaper on, but the app itself should own lifecycle.

## Expected

Stop remote web and managed servers on session close or disconnect, or add an idle timeout. At minimum, do not spawn a new web server per connect when an existing one answers.

## Environment (redacted on purpose)

server1: PVE homeserver, 15 GB RAM, Debian 13. server2: dedicated PVE host, 62 GB RAM, Debian 13.

yulia-ivashko:
Fixes #3710

## Intent

A managed SSH instance started a new `openchamber serve` on the remote host on every connect, including every automatic reconnect after a dropped tunnel or a sleeping laptop. The old server stays up on purpose, because `keepRunning` defaults to on so agents keep working without the laptop. Nobody looked for it on the next connect, and the start port is random, so servers piled up. Each one supervises its own `opencode`. The reporter counted 69 opencode processes and lost two VMs to the OOM killer.

Now a connect asks the remote CLI what is already running (`openchamber status --json`) and attaches to a server that fits the instance's password. A new server starts only when none fits.

"Fits" is checked, not assumed. `/api/system/info` is public, so a server answers it whatever password you hold. The probe now reports whether `/auth/session` accepted the instance's password, and an instance without a password skips servers the registry marks as protected. Without this, changing the UI password would attach the app to the old server and the connect would fail on every retry.

Among the servers that fit, a running daemon of another app version is stopped and replaced, the same thing `openchamber update` does to running instances. The maintainer picked this over leaving it running next to the new one. The same replacement happens when the server's bind address no longer matches the instance. That covers a loopback server for an instance set to publish to the network, and a server still on `0.0.0.0` after the user turned publishing off.

A registered server that is passed over gets a line in the connect log with the reason, auth status included.

With `keepRunning` off, disconnecting now also stops a daemon the connect adopted. Before, a dropped tunnel left the server up, the reconnect adopted it as not ours, and the setting stopped applying.

`status --json` gains one field, `bindHost`, so the desktop can tell those two cases apart. A server bound to `0.0.0.0` also answers on loopback, so a probe cannot tell.

## Non-goals

- `keepRunning` keeps its default. The server outliving the session is the feature, the missing lookup was the bug.
- Same-version servers that already piled up on a host are not reaped. The app attaches to one of them and leaves the rest, since they may hold live sessions. `openchamber stop --all` on the remote clears them. After an app update the connect does stop every fitting daemon of the old version, because each of them is stale.
- Foreground servers (systemd and the like) and servers with a different password are neither reused nor stopped.
- No idle timeout for remote servers.

## Affected surfaces

`packages/electron` (SSH manager) and `packages/web` CLI (`status --json` gains `bindHost`, additive).

|---|---|
| Web | Not applicable. `openchamber status --json` reports `bindHost` per instance. |
| Desktop (Electron) | Connecting to a managed SSH instance reuses the running remote server instead of starting another. |
| VS Code | Not applicable, no SSH instances. |
| Hosted mobile | Not applicable. |
| Capacitor mobile | Not applicable. |

## Repository guidance

|---|---|---|
| AGENTS.md, "Prefer authoritative state over heuristics" and "Never let fetch failure masquerade as authoritative empty success" | The fix needs to know which servers run on the remote | It reads the remote CLI registry instead of grepping processes or guessing ports. When the registry cannot be read, the connect log says so at WARN before a server is started. |
| `desktop-shell` | SSH lifecycle lives in Electron main | Change stays in `ssh-manager.mjs`. No IPC, preload, or renderer contract changes. |
| `clack-cli-patterns` | `status --json` output changes | JSON-only output kept, the field is additive, human and `--quiet` output unchanged. |
| `openchamber-change-discipline` | External contract (CLI JSON) and platform behavior | Focused tests for both packages plus a live run against the real CLI. `packages/electron/README.md` now documents the one-server-per-host rule. |

## Validation

| Check | Result |
|---|---|
| `bun test ./ssh-manager.test.mjs` in `packages/electron` | 32 pass, 15 new. The fake remote host checks passwords the way the server does, public info route included |
| `bun test ./bin/cli.test.js` in `packages/web` | 108 pass, 1 new |
| `bun run type-check` in `packages/electron`, `node --check ssh-manager.mjs` | pass |
| `bunx oxlint` on the four touched files | no findings on authored lines, the rest is the known backlog |

**Live run:** desktop SSH manager on macOS against a real ssh session. The remote is a Debian bookworm container running sshd, with opencode 1.18.31 installed. The script drives the full `ElectronSshManager.connect()` and `disconnect()` cycle three times: ControlMaster, remote probe, the manager's own npm install of the published `@openchamber/web@1.24.1`, server start, port forward, and a `/health` request through the tunnel. After each disconnect it counts processes on the remote with `pgrep`. `keepRunning` is on, the default. I ran `main` first, restarted the container, then ran this branch.

`main`:

```
connect 1: phase=ready remotePort=35662 startedByUs=true tunnel /health=200
  after disconnect -> web servers: 1, opencode servers: 1, RSS: 531 MB
connect 2: phase=ready remotePort=31584 startedByUs=true tunnel /health=200
  after disconnect -> web servers: 2, opencode servers: 2, RSS: 1130 MB
connect 3: phase=ready remotePort=33427 startedByUs=true tunnel /health=200
  after disconnect -> web servers: 3, opencode servers: 3, RSS: 1577 MB
```

This branch:

```
connect 1: phase=ready remotePort=45979 startedByUs=true tunnel /health=200
  after disconnect -> web servers: 1, opencode servers: 1, RSS: 486 MB
connect 2: phase=ready remotePort=45979 startedByUs=false tunnel /health=200
  after disconnect -> web servers: 1, opencode servers: 1, RSS: 486 MB
connect 3: phase=ready remotePort=45979 startedByUs=false tunnel /health=200
  after disconnect -> web servers: 1, opencode servers: 1, RSS: 457 MB
```

That is the reporter's symptom, about 500 MB per connect on `main`, and flat on the branch. The remote CLI in this run is the published 1.24.1, which does not report `bindHost` yet, so the run also shows the desktop working against a CLI older than this change.

The password and version-replacement paths ran in a second live setup: the same manager driving the `openchamber serve` CLI from this checkout, with remote commands executed through `/bin/sh` in an isolated `HOME` and managed opencode start skipped.

```
password pw-one: port=30064 startedByUs=true -> servers alive: [30064*, 31325]
password pw-two: port=36159 startedByUs=true -> servers alive: [30064*, 31325, 36159*]
password pw-one: port=30064 startedByUs=false -> servers alive: [30064*, 31325, 36159*]
other app version: adopted=null -> servers alive: 2
```

`*` marks a password-protected server. A changed password gets its own server and never the old one, and the first password finds its server again. The last line is a passwordless manager with another `appVersion`: it stops the one server that fits it, asks for a fresh start, and leaves the two protected servers running.

Not verified: the packaged desktop app with its Remote Instances UI, a Windows desktop client (no ControlMaster there), and the network-publishing path over real ssh. Those paths build remote commands the same way the verified ones do.

## Visual evidence

No layout, component, or style changes. The one place a user can see this change is the text of the SSH connect log in Remote Instances, so the evidence is that log, taken from the two runs above.

`main`, connects 2 and 3. Every connect goes from detecting straight to starting:

```
[INFO] phase="server_detecting" detail=Detecting managed OpenChamber server
[INFO] phase="server_starting" detail=Starting managed OpenChamber server
[INFO] phase="server_detecting" detail=Detecting managed OpenChamber server
[INFO] phase="server_starting" detail=Starting managed OpenChamber server
```

This branch, connects 2 and 3. Detecting finds the running server and `server_starting` never appears:

```
[INFO] phase="server_detecting" detail=Detecting managed OpenChamber server
[INFO] phase="server_detecting" detail=Detecting managed OpenChamber server
```

New log lines a user can meet, all plain text in the same log: `Not reusing the server on remote port N: ...`, `Replacing the managed server on remote port N: ...`, and the WARN when the remote registry cannot be read. The memory numbers in the live run are the before and after measurement for the leak itself.

## Risks and failure behavior

- Replacing a daemon of another version interrupts agent work running on it. This happens once per app update, and it is the maintainer's choice over leaking a server per update.
- If the user started `openchamber serve` by hand in daemon mode with the same password, the desktop now attaches to it, and replaces it after an app update. Foreground servers are left alone.
- If `status --json` fails, the connect logs a WARN and starts a server as before. The leak returns for that host, the connection still works.
- If a stale server refuses to stop, the log says so. An unpinned instance starts its new server on another port. On a pinned `preferredPort` nothing else can start, so the connect falls back to the stale server instead of failing.
- A pinned `preferredPort` still reuses a server the registry does not know. What changed there is that a registered daemon of another version on that port gets replaced, and a server that rejects the password is no longer reused. On `main` that second case also failed, later, at the client token step.
- With `keepRunning` off, disconnecting stops the adopted daemon even if a second client (another laptop, or a second SSH instance to the same account) is still using it. On `main` each client had its own server, so this could not happen. It needs the non-default setting plus a second client, and the other choice, never stopping an adopted server, makes `keepRunning: false` stop working after the first dropped tunnel. Telling a desktop-started daemon from a hand-started one needs an owner marker in the CLI registry, which I left out of this fix.
- Each connect posts the instance password to `/auth/session` of every registered server on the account until one fits. A server with another password counts that as a failed login, and ten of those within the limiter's window lock its logins from loopback for 15 minutes. The connect log names the auth status when that happens.
- Changing the UI password leaves the server with the old password running. The app cannot prove that server is its own, so it does not stop it. That is one leftover per password change, cleared with `openchamber stop --port` on the remote.
- Two desktops on different app versions sharing one remote account will replace each other's server on connect. On `main` they already fought over the installed binary.
- On a host with wget but no curl the probe cannot authenticate, so a password-protected instance never adopts a server there and the old leak remains for that combination.
- A remote CLI older than this change reports no `bindHost`. Such a server is another version anyway, so it gets replaced.


yulia-ivashko:
/oc-review No code changed since the reviewed HEAD. Only the PR description changed: the Live run now covers a real ssh session against a Debian sshd container (main vs branch process and memory counts), and Visual evidence shows the connect log before and after. Please check only those two sections.

PR #3727, opened 2026-09-19
test(markdown): harden SVG sanitizer against event-handler XSS vectors

DeryFerd:
## Intent

Agent-generated markdown reaches the DOM through two layers: the marked renderer escapes all raw HTML, and a DOMPurify config (`USE_PROFILES { html, mathMl, svg }` plus SVG tag/attribute allowlists) sanitizes the HTML the pipeline generates, such as highlighter output, KaTeX math, and the disclosure markup the tokenizer constructs. DOMPurify's defaults strip event-handler attributes today, but that safety is implicit. Nothing in the repo pins it, and the config sits close to the edge since `ADD_ATTR` already includes `style` and the SVG attribute family, so a future config edit or a DOMPurify major upgrade could re-enable event handlers with nothing failing. This PR turns that boundary into a tested contract: a new vector suite runs the real DOMPurify pipeline over adversarial SVG and HTML inputs and asserts that event handlers, `script`/`style` elements, and `javascript:` URLs never survive, while legitimate SVG shapes still render. Alongside the tests, a three-line `uponSanitizeAttribute` hook rejects any attribute matching `/^on/i` explicitly, so the invariant no longer depends on library defaults. Rendered output is unchanged today; the hook only rejects attributes the allowlist already rejects.

The suite was also proven to bite. I ran it against the current config first (all green, documenting existing behavior), then temporarily added `onload`, `onclick`, `onbegin`, and `onerror` to `ADD_ATTR` to simulate the exact mistake a future edit could make. Two tests failed as expected. With the new hook in place, the same loosened config passes the full suite, which is the property this PR is after: the boundary holds even when someone loosens the config above it.

## Non-goals

This PR does not change which tags or attributes legitimate content may use. It does not touch the marked-layer escaping (`escapeRawMarkdownHtml`) or the anchor `href` and app-URL hooks that already handle `obsidian://`-style links and `noopener` attributes. It is not a fix for any known exploitable bypass; none was found while tracing this boundary, and I want to be clear that this is hardening plus regression coverage rather than a patch for an observed bug. Other call sites that render worker-highlighted code, such as permission cards and diff views, are out of scope here.

## Affected surfaces

`packages/ui` markdown rendering only: `markdownCore.ts` gains the event-handler hook and a test-only export (`__sanitizeForTests`, following the module's existing `__...ForTests` convention), and a new test file covers the vectors. No persisted data, API, or package contract changes.

|---|---|
| Web | Chat markdown rendering is unchanged; the sanitizer now also rejects event-handler attributes explicitly instead of relying on DOMPurify defaults. |
| Desktop (Electron) | Same shared UI, same unchanged rendered output. |
| VS Code | Same shared UI in the webview, same unchanged rendered output. |
| Hosted mobile | Same shared UI, same unchanged rendered output. |
| Capacitor mobile | Same shared UI, same unchanged rendered output. |

## Repository guidance

|---|---|---|
| AGENTS.md validation rules | Source change in `packages/ui` that adds an export, so focused tests, package type-check, oxlint on changed files, and `dead-code` all apply. | Ran each of these at package scope and report exact results below; the new export is consumed by the new test file and does not appear in the knip report. |
| `openchamber-change-discipline` skill | Governs any source change: smallest complete change, establish current behavior for behavior-preserving work, validate at the narrowest level covering the risk. | The diff is two files and 75 added lines; the vector suite pins current behavior first, then the hook is the only production change; validation stays inside the owning package. |
| Module contract comments (`markdownSecurity.ts`, sanitize section of `markdownCore.ts`) | The module documents its own invariant: raw markdown HTML stays escaped and active elements are "forbidden again at the final DOMPurify boundary", with the config explicitly framed as defense in depth for generated and highlighter HTML. The markdown folder has no separate DOCUMENTATION.md, so these comments are the nearest owning docs. | The hook and the vector suite extend exactly that documented boundary and the hook carries a comment stating why event handlers are never legitimate in generated markdown HTML. |

## Validation

The vector suite deliberately runs the real DOMPurify pipeline under happy-dom rather than the mocked DOMPurify that `markdownCore.test.ts` uses for its href-hook tests, so the sanitizer config itself is what gets exercised. One environment note: happy-dom's parser drops SVG children that follow a leading `script` or `style` element, so two vectors place the legitimate shape before the forbidden tag; real browsers parse either order correctly, and the test file comments explain this.

| Check | Result |
|---|---|
| `bun test src/components/chat/markdown/markdownSanitizeVectors.test.ts` (in `packages/ui`) | 6/6 pass against the current config |
| Mutation check: temporarily added `onload`, `onclick`, `onbegin`, `onerror` to `ADD_ATTR` | 2 tests failed as expected; with the new hook in place the same loosened config passes 6/6; config then restored |
| `bun test src/components/chat/markdown/markdownCore.test.ts src/components/chat/markdown/detachedMarkdownDomCache.test.ts` | 45/45 pass, existing behavior intact |
| `bun run type-check:ui` | pass (`tsc --noEmit`, exit 0) |
| `bunx oxlint` on both changed files | 9 findings, all pre-existing anti-slop backlog in `markdownCore.ts`; the count is identical on the base revision and none of the flagged lines are ones this PR adds. 0 findings in the new test file. |
| `bun run dead-code` | no entries for `__sanitizeForTests` or the new test file |
| Full `bun run test` and `bun run build` | not run locally; the change is confined to one module and its tests, and the package-scoped checks above cover that scope. CI will run the full suite. |

**Live run:** Not required here because no user-reachable behavior changes: the hook rejects only attributes the sanitizer's allowlist already rejects, so sanitized output is identical before and after this change, and the vector suite asserts exactly that on both sides. The closest thing to a runtime check is the suite itself, which drives the real DOMPurify pipeline rather than a mock. I did not run the built app for this change.

## Visual evidence

No visible change to show, and I want to give a concrete reason rather than leave this empty. The diff cannot affect rendered behavior: it adds tests, a test-only export, and a hook that rejects attributes the sanitizer already drops, so every rendered markdown message produces the same HTML as before. The vector suite verifies this directly by comparing sanitized output across the mutation runs described above. If a reviewer would still like side-by-side renders of a message containing inline SVG before and after, I am happy to capture them, but the output is unchanged by construction.

## Risks and failure behavior

The one real risk is over-blocking: if some future feature ever needed an event-handler attribute in generated markdown HTML, the `/^on/i` hook would reject it. Nothing in the current pipeline needs one (Shiki, KaTeX, and mermaid emit none), and the hook is a small, clearly commented block that is easy to relax with a test update if that day comes. Rollback is trivial: revert the commit and the suite still pins DOMPurify's default behavior on its own. Performance impact is one regex test per attribute during sanitization, a pass that already visits every attribute, with no caching, lifecycle, or persisted-state changes involved. Cross-runtime risk is low by construction since all five runtimes share this module and the rendered output is unchanged.


DeryFerd:
Greetings @btriapitsyn 

Could I get a review for this PR?

Thank you and apologize if I am tagging you

